PR #166, opened 2025-12-04
fix: Hide console window on Windows when spawning child processes

thedotmack:
## Summary

- Adds `windowsHide: true` to `spawnSync` call in `worker-utils.ts` when starting PM2 worker
- Adds `windowsHide: true` to `execSync` call in `user-message-hook.ts` when spawning context display

This prevents empty console windows from appearing on Windows when hooks execute.

## Problem

Windows users reported that an empty terminal window opens every time they submit a prompt, runs while Claude is working, and closes when done.

## Solution

Node.js `spawn`/`exec` calls on Windows open a visible console window by default. The `windowsHide: true` option suppresses this.

Reference: https://nodejs.org/api/child_process.html

## Test plan

- [ ] Windows user confirms console windows no longer appear on prompt submission

### DIFF
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -217,13 +217,13 @@ export class WorkerService {
 
       // Find orphaned uvx processes (which spawn chroma servers)
       try {
-        const processes = execSync('pgrep -fl uvx', { encoding: 'utf-8', stdio: 'pipe' }).trim();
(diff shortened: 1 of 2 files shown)


thedotmack:
@claude do the rest of those so we have full coverage on this, its low risk high reward

Issue #189, opened 2025-12-09
[Bug] v7.0.0: smart-install.js missing from plugin/scripts directory


wenodev:
## Bug Description
  After installing claude-mem v7.0.0, Claude Code fails to start with the following error:

  ## Error Message
```
  Error: Cannot find module '/Users/xxx/.claude/plugins/cache/thedotmack/claude-mem/scripts/smart-install.js'
      at Module._resolveFilename (node:internal/modules/cjs/loader:1369:15)
      at defaultResolveImpl (node:internal/modules/cjs/loader:1025:19)
      at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1030:22)
      at Module._load (node:internal/modules/cjs/loader:1179:37)
```

  ## Root Cause
  - `plugin/hooks/hooks.json` references `smart-install.js` in the SessionStart hook:
    ```json
    "command": "node \"${CLAUDE_PLUGIN_ROOT}/../scripts/smart-install.js\" && node ${CLAUDE_PLUGIN_ROOT}/scripts/context-hook.js"
  - However, plugin/scripts/ directory does not contain smart-install.js
  - The file only exists in the root scripts/ folder (development scripts), not in the distributed plugin/scripts/ folder

  Files in plugin/scripts/ (v7.0.0)
  - claude-mem-settings.sh
  - cleanup-hook.js
  - context-generator.cjs
  - context-hook.js
  - mcp-server.cjs
  - new-hook.js
  - save-hook.js
  - search-server.cjs
  - summary-hook.js
  - user-message-hook.js
  - worker-service.cjs
  Note: smart-install.js is missing from this list

## Expected Behavior
  smart-install.js should be included in plugin/scripts/ or the hooks.json reference should be updated/removed.

## Environment
  - Claude Code version: latest
  - Plugin version: 7.0.0
  - OS: macOS Darwin 25.0.0

kat-bell:
Additionally, package.json is missing from the v7.0.0 distribution. Only package-lock.json is included.
**Files present in 7.0.0/:**
.claude-plugin
hooks
scripts
skills
ui
.mcp.json
package-lock.json

**Missing:**
package.json
ecosystem.config.cjs
smart-install.js (already reported)

This causes npm install to fail with:
npm error enoent Could not read package.json: Error: ENOENT: no such file or directory
OS: Windows 11, Node.js v24.11.1

thedotmack:
[...]

PR #192, opened 2025-12-09
fix: add missing smart-install.js to plugin/scripts

wenodev:
## Summary
- Add missing `smart-install.js` to `plugin/scripts/` directory
- Update `hooks.json` path reference from `${CLAUDE_PLUGIN_ROOT}/../scripts/` to `${CLAUDE_PLUGIN_ROOT}/scripts/`

## Problem
After installing claude-mem v7.0.0, Claude Code fails to start with the following error:

```
Error: Cannot find module '/Users/xxx/.claude/plugins/cache/thedotmack/claude-mem/scripts/smart-install.js'
```

## Root Cause
- `plugin/hooks/hooks.json` references `smart-install.js` in the SessionStart hook
- However, `plugin/scripts/` directory does not contain `smart-install.js`
- The file only exists in the root `scripts/` folder (development scripts)

## Solution
Copy `smart-install.js` from `scripts/` to `plugin/scripts/` and update the path in `hooks.json`.

## Test plan
- [ ] Install the plugin and verify Claude Code starts without MODULE_NOT_FOUND error
- [ ] Verify the smart-install script runs correctly on session start

## Comment
I've already created [this issue](https://github.com/thedotmack/claude-mem/issues/189) and asked Claude Code to fix the bug. The fix looked good, so I also created a PR for it.

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -7,7 +7,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/../scripts/smart-install.js\" && node ${CLAUDE_PLUGIN_ROOT}/scripts/context-hook.js",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/smart-install.js\" && node ${CLAUDE_PLUGIN_ROOT}/scripts/context-hook.js",
             "timeout": 5
           },
           {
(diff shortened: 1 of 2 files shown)


thedotmack:
@wenodev Thank you so much for the contribution! I had a PR open for it too <3 I'm merging that fix in after a quick review. Thanks so much for the help! :)

thedotmack:
Mine isn't working 😭 trying yours

thedotmack:
investigating regression, heading to bed – will have fixed first thing tm :)

PR #206, opened 2025-12-10
Harden worker startup: fail-fast, readiness signaling, and ABI mismatch prevention

akbad:
Hi, after some issues with my `claude-mem-worker` earlier today after the missing `package.json` issue was patched, I made some changes to make the experience more seamless going forward.

## Summary

Eliminates: 

- recurring and opaque "Database not initialized" errors
- ABI mismatch failures

by enforcing proper startup behaviour and readiness checks.

## Changes

- **Fail-fast on init errors** ➔ worker exits immediately if DB or native modules fail to initialize, instead of running in a broken state
- **Visibly signal still-ongoing initialization in worker endpoints:**

  - Worker health endpoint returns 503 `{"status":"initializing"}` until fully ready 
  - Route error handler also now defensively returns 503 if initialization is still ongoing 
  
- **Interpreter pinning** ➔ pass `--interpreter` flag to PM2 when starting worker so that the shell's Node version is used (and not the PM2 daemon's), preventing ABI mismatch
- **Stale worker cleanup** ➔ kill worker after native module rebuild (and fresh installs, just in case a Node version change was the reason for it) so it restarts with the correct, freshly-compiled native module binaries
- **Documentation** ➔ Updated troubleshooting docs with ABI mismatch fixes and health endpoint states

## Testing

- Start session after switching Node versions ➔ worker rebuilds and restarts correctly
- Call `/health` during startup ➔ returns 503, then 200 when ready
- Kill DB mid-init ➔ worker exits instead of serving broken responses

### DIFF
--- a/scripts/sync-marketplace.cjs
+++ b/scripts/sync-marketplace.cjs
@@ -61,7 +61,7 @@ function getPluginVersion() {
 console.log('Syncing to marketplace...');
 try {
   execSync(
-    'rsync -av --delete --exclude=.git ./ ~/.claude/plugins/marketplaces/thedotmack/',
+    'rsync -av --delete --exclude=.git --exclude=node_modules ./ ~/.claude/plugins/marketplaces/thedotmack/',
     { stdio: 'inherit' }
   );
 


akbad:
Hold on I found an issue, give me a little time to patch

akbad:
The issue I encountered was purely dev-experience related: 
- `npm run sync-marketplace` was copying its compiled `better-sqlite3` binaries from my cloned source to the marketplace directory
- this causes an ABI mismatch if the two locations use different Node versions
- the fix was to exclude `node_modules` from being copied over so that the subsequent `npm install` in the marketplace directory actually compiles native modules for its *own* Node version

I pushed my fix, rebased my branch onto main and force pushed.

Ready for review!

thedotmack:
Hi @akbad, thanks for the detailed PR addressing worker startup reliability issues!

However, this PR targets PM2-specific code that was removed in v7.1.0 (released 2025-12-13). The project has undergone a major architectural migration:

**What changed:**
- PM2 completely replaced with native Bun-based process management (`ProcessManager`)
- better-sqlite3 migrated to bun:sqlite (no native compilation required)
- Worker startup now handled by `ProcessManager.start()` with PID file tracking
- PM2 scripts (`smart-install.js`, ecosystem config) removed

**Affected files in this PR:**
- `scripts/smart-install.js` - This script no longer exists (replaced by Bun auto-installer)
- `src/shared/worker-utils.ts` - PM2 calls replaced with `ProcessManager.start()`
- All PM2 interpreter pinning and ABI mismatch handling removed (no longer needed without native modules)

**The good news:**
Many of the issues you identified are now solved architecturally:
- ✅ ABI mismatch eliminated (bun:sqlite requires zero native compilation)
- ✅ Fail-fast initialization already implemented in `worker-service.ts`
- ✅ Worker health signaling concepts are valuable and could still be applied

**Next steps:**
Would you like to update this PR for the new Bun architecture? The health endpoint improvements (503 during initialization) are still relevant and could be applied to the current codebase. Otherwise, I'll close this as obsolete due to the architectural changes.

See the migration docs for technical details: `docs/PM2-TO-BUN-MIGRATION.md`

akbad:
Sure, I can take care of updating this for Bun. Converting back to draft until it's ready

PR #212, opened 2025-12-10
fix: quote paths in hooks.json for Windows usernames with spaces

vari0nce:
## Summary
- Wraps all `${CLAUDE_PLUGIN_ROOT}` paths in `hooks.json` with quotes
- Fixes syntax errors for Windows users with spaces in their username

## Problem
When `CLAUDE_PLUGIN_ROOT` expands to a path with spaces (e.g., `C:\Users\John Doe\.claude\...`), the unquoted path gets split by the shell. Node.js then tries to execute the first part as JavaScript, causing a `SyntaxError: Unexpected token ':'`.

## Test plan
- [x] Verified fix works on Windows with spaces in username
- [ ] Verify no regression on paths without spaces (Unix, Windows without spaces)

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -7,12 +7,12 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/smart-install.js\" && node ${CLAUDE_PLUGIN_ROOT}/scripts/context-hook.js",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/smart-install.js\" && node \"${CLAUDE_PLUGIN_ROOT}/scripts/context-hook.js\"",
             "timeout": 300
           },
           {
             "type": "command",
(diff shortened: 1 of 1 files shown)


plugin/hooks/hooks.json
```diff
@@ -46,7 +46,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/summary-hook.js",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/summary-hook.js\"",
             "timeout": 120
           }
```
xixibo, line 51:
```
public static void getName() {
}
```

thedotmack:
The fix from this PR has been applied directly to main in commit f41824f.

The core issue (missing quotes around ${CLAUDE_PLUGIN_ROOT} paths) has been resolved, but applied to the current codebase which uses `bun` instead of `node` (post v7.1.0 migration).

This ensures Windows users with spaces in their usernames can use claude-mem without SyntaxError issues.

Thank you for identifying this issue!

PR #226, opened 2025-12-11
Fix: Support CLAUDE_CONFIG_DIR for custom Claude installations

### LINKED ISSUE #210: Plugin hook error
⎿ PostToolUse:Read says: Plugin hook error: file:///Users/dreambt/.claude/plugins/cache/thedotmack/claude-mem/7.0.7/scripts/save-hook.js:8
    Or install globally with: npm install -g pm2`);r="pm2"}let n=d(r,["start",t],{cwd:l,stdio:"pipe",encoding:"utf-8"});if(n.status!==0)throw new
    Error(n.stderr||"PM2 start failed")}for(let t=0;t<G;t++)if(await new Promise(e=>setTimeout(e,B)),await C())return!0;return!1}catch(o){return
    c.error("SYSTEM","Failed to start worker",{platform:process.platform,workerScript:S.join(l,"plugin","scripts","worker-service.cjs"),error:o instanceof
    Error?o.message:String(o),marketplaceRoot:l}),!1}}async function I(){if(await C())return;let o=await Y();if(!(!o&&await C())&&!o){let t=T();throw new
    Error(`Worker service failed to start on port ${t}.

                                                                                                                          ^

    Error: Worker service failed to start on port 37777.

    To start manually, run:
      cd /Users/dreambt/.claude/plugins/marketplaces/thedotmack
      npx pm2 start ecosystem.config.cjs

    If already running, try: npx pm2 restart claude-mem-worker
        at I (file:///Users/dreambt/.claude/plugins/cache/thedotmack/claude-mem/7.0.7/scripts/save-hook.js:8:581)
        at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
        at async Z (file:///Users/dreambt/.claude/plugins/cache/thedotmack/claude-mem/7.0.7/scripts/save-hook.js:16:363)
        at async Socket.<anonymous> (file:///Users/dreambt/.claude/plugins/cache/thedotmack/claude-mem/7.0.7/scripts/save-hook.js:16:1257)

    Node.js v24.11.1

majiayu000:
## Summary

- Add support for `CLAUDE_CONFIG_DIR` environment variable for users with custom Claude Code configurations
- Centralize marketplace path logic in `paths.ts` to avoid hardcoded `~/.claude` paths

## Problem

Users who use custom `CLAUDE_CONFIG_DIR` settings (e.g., `~/.claude_personal` for multiple accounts) were unable to use claude-mem because plugin paths were hardcoded to `~/.claude`.

As reported in #210 comments by @hffmnn:
> I see error messages like this, where you can see that the plugin tries to load from `.claude` instead of `.claude_personal`

## Changes

- **src/shared/paths.ts**: Add `MARKETPLACE_ROOT` constant that respects `CLAUDE_CONFIG_DIR`
- **src/shared/worker-utils.ts**: Use shared `MARKETPLACE_ROOT` instead of hardcoded path
- **src/services/context-generator.ts**: Use shared `MARKETPLACE_ROOT` for version marker path
- **src/services/worker/BranchManager.ts**: Use shared `MARKETPLACE_ROOT` for git operations

## Test Plan

- [x] All 42 existing tests pass
- [x] Build succeeds
- [ ] Manual test with custom `CLAUDE_CONFIG_DIR` (needs user verification)

Fixes #210 (partial - addresses the custom config directory issue reported in comments)


thedotmack:
defering to later

PR #234, opened 2025-12-11
Rename package from 'claude-mem' to 'claude-meme'

KIRUBA-110:


thedotmack:
@KIRUBA-110 claude mem is actually short for claude members only

Issue #259, opened 2025-12-13
no such file or directory, open '/Users/igor/.claude-mem/.pm2-migrated


exdeniz:
When I start the plugin, I get this error:
I don't have folder ~/.claude-mem
```
⚠️  Claude-Mem: First-Time Setup

     Dependencies are installing in the background. This only happens once.

     💡 TIPS:
        • Memories will start generating while you work
        • Use /init to write or update your CLAUDE.md for better project context
        • Try /clear after one session to see what context looks like

     Thank you for installing Claude-Mem!

     This message was not added to your startup context, so you can continue working as normal.
  ⎿ SessionStart:startup says: Plugin hook error: 1 | #!/usr/bin/env bun
    2 | import At from"path";import{stdin as R}from"process";import T from"path";import{homedir as gt}from"os";import{spawnSync as mt}from"child_process";import{existsSync as St,writeFileSync as x,readFileSync as dt}from"fs";import{readFileSync as Y,writeFileSync as
     J,existsSync as q}from"fs";import{join as Q}from"path";import{homedir as z}from"os";var G=["bugfix","feature","refactor","discovery","decision","change"],X=["how-it-works","why-it-exists","what-changed","problem-solution","gotcha","pattern","trade-off"];var
    U=G.join(","),v=X.join(",");var c=class{static DEFAULTS={CLAUDE_MEM_MODEL:"claude-haiku-4-5",CLAUDE_MEM_CONTEXT_OBSERVATIONS:"50",CLAUDE_MEM_WORKER_PORT:"37777",CLAUDE_MEM_WORKER_HOST:"127.0.0.1",CLAUDE_MEM_SKIP_TOOLS:"ListMcpResourcesTool,SlashCommand,Skill,Tod
    oWrite,AskUserQuestion",CLAUDE_MEM_DATA_DIR:Q(z(),".claude-mem"),CLAUDE_MEM_LOG_LEVEL:"INFO",CLAUDE_MEM_PYTHON_VERSION:"3.13",CLAUDE_CODE_PATH:"",CLAUDE_MEM_CONTEXT_SHOW_READ_TOKENS:"true",CLAUDE_MEM_CONTEXT_SHOW_WORK_TOKENS:"true",CLAUDE_MEM_CO
[...]

PR #292, opened 2025-12-14
Fix: ensure data directory exists before writing PM2 migration marker

trevwilson:
## Summary
  - Ensures `~/.claude-mem/` directory exists before writing `.pm2-migrated` marker file
  - Adds `mkdirSync(dataDir, { recursive: true })` before the marker file write in `startWorker()`
  - I was unsure whether to include full build outputs in this PR as well - I'm on Windows so I'm not positive whether some of the minification changes just represent tooling differences

  ## Problem
  On first install, the `startWorker()` function fails with ENOENT because it tries to write the `.pm2-migrated` marker file before the `~/.claude-mem/` directory exists. Looks like this was introduced in f923c0c and exposed to Windows users in 5d4e71d.
  This is issue #259 which was marked as completed, but was verified as still present in the most recent version (v7.1.11)

  ## Test plan
  - [x] Build passes
  - [x] 51 of 52 tests pass
  - [ ] 1 test fails in `bun-path.test.ts` - this seems to be a pre-existing failure from #290 (the test expects Unix-style path `.bun/bin/bun` which doesn't exist on Windows) unrelated to this fix and verified to fail on latest from main, but let me know how to approach.


thedotmack:
Thanks for the fix @trevwilson! This has been merged and released as v7.1.12.

Release: https://github.com/thedotmack/claude-mem/releases/tag/v7.1.12

PR #323, opened 2025-12-15
[Feat] Add <fresh-session> tag to skip context injection

kamil-hassan201:
# Add `<fresh-session>` tag to skip context injection

## Summary

Adds a `<fresh-session>` tag that allows starting a session **without loading any historical context** (no observations, summaries, or timeline).

Default behavior is unchanged; this is an explicit, per-session opt-in.

---

## Usage

```text
<fresh-session>
Review this code for security issues.
````

The tag is detected only in the first user message and is stripped before storage.

---

## Changes

* **context-hook.ts** – Skip context injection when `<fresh-session>` is present
* **tag-stripping.ts** – Remove tag from stored messages
* **fresh-session.test.ts** – 11 tests added (all passing)
* **fresh-session.mdx** – User documentation

---

## Why

* Enables unbiased reviews and unrelated tasks
* Provides explicit per-session control without global flags

---

## Testing

```bash
npm test -- tests/happy-paths/fresh-session.test.ts
```

✓ 11 tests passing

---

## Compatibility

✅ Fully backward compatible


thedotmack:
@kamil-hassan201 so this prevents the startup context from being injected on user prompt submit? The transcript is created after you type /clear and i THINK it already has the context injected into it by then

thedotmack:
Thanks for the thorough implementation (841 lines with tests and docs)!

Skipping because this sparked rethinking when context injection happens. Moving from SessionStart to UserPromptSubmit enables fresh sessions, context previews, and dynamic filtering more cleanly. Want to solve the architecture first, then add tag syntax.

PR #329, opened 2025-12-15
fix: add BYOK/third-party API proxy support for SDK Agent

7Sageer:
## Problem

SDK Agent fails with `"Your account does not have access to Claude Code. Please run /login."` error for users using BYOK (Bring Your Own Key) with third-party API proxies (e.g., `ANTHROPIC_BASE_URL`).

**Root Cause**: The SDK Agent spawns a Claude subprocess that doesn't inherit the custom API settings from `~/.claude/settings.json`, causing it to try default Anthropic authentication which fails.

Additionally, when loading user settings with `settingSources: ["user"]`, the SDK subprocess also loads the claude-mem hooks configuration, causing SDK internal prompts to be incorrectly saved to the `user_prompts` table.

### 1. BYOK Support (Source-level fix)
- Add `loadClaudeCodeApiSettings()` method to `worker-service.ts` that reads BYOK config from `~/.claude/settings.json` and sets `ANTHROPIC_BASE_URL` and `ANTHROPIC_AUTH_TOKEN` as environment variables at worker startup
- Use `settingSources: []` (empty) in SDK Agent to prevent loading any user settings including hooks
- API authentication works through inherited environment variables

### 2. Defensive SDK Prompt Filter (Fallback)
- Add defensive filter in `SessionRoutes.ts` to detect SDK internal sessions
- Primary detection: If `project` name equals home directory name (SDK runs with `workingDirectory: homedir()`)
- Fallback detection: Pattern matching for known SDK prompt prefixes

## Changes

- `src/services/worker-service.ts`: Add `loadClaudeCodeApiSettings()` method to load BYOK config
- `src/services/worker/SDKAgent.ts`: Use `settingSources: []` to prevent loading hooks
- `src/hooks/new-hook.ts`: Add SDK prompt filter at hook level
- `src/services/worker/http/routes/SessionRoutes.ts`: Add SDK prompt filter at API level with home directory detection

## Why This Approach

Using `settingSources: []` is cleaner than filtering because:
- **Prevents the problem at source**: SDK subprocess doesn't load hooks at all
- **No hardcoded prompt patterns needed**: More robust against prompt changes
- **BYOK support via env vars**: API auth works through process environment inheritance

## Testing

1. ✅ Verified SDK Agent successfully connects to third-party API proxy with `settingSources: []`
2. ✅ Verified SDK internal prompts are no longer saved to database (count dropped from 12+ to 0)
3. ✅ Normal user prompts are correctly saved (`skipped: false`)
4. ✅ Direct API test confirms filter works for SDK sessions


thedotmack:
@claude review

thedotmack:
## Auto code review

No issues found. Checked for bugs and CLAUDE.md compliance.

vildanbina:
@thedotmack any plan to merge this?

thedotmack:
Thanks for the spot-on fix! Your approach (reading API settings, setting env vars, using empty settingSources) is right.

Closing because SDK Agent architecture changed significantly since submission. Will re-implement using your approach.

PR #355, opened 2025-12-16
Update version and fix first install issues

blankster:
Since I truly enjoy claude-mem and also derive a lot of benefit from it, I took the liberty of updating the somewhat outdated README.md to the latest version.

I hope I did everything correctly. The goal is not to make insignificant commits, but to support the maintainer, hopefully saving them some time while the first impression people get here also reflects the most current state/version.

Please feel free to adopt or discard whatever is useful; it is intended as a suggestion.

Thank you for this project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
     <img src="https://img.shields.io/badge/License-AGPL%203.0-blue.svg" alt="License">
   </a>
   <a href="package.json">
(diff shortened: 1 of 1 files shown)


thedotmack:
Thanks for noticing the version mismatch and taking time to fix it!

Skipping this because hardcoded versions get stale every release (we're already on v7.4.0). Going to reference GitHub releases dynamically instead.

PR #372, opened 2025-12-17
fix(windows): solve zombie port problem with wrapper architecture

ToxMox:
## Problem

On Windows, Bun doesn't properly release socket handles when the worker process exits, causing "zombie ports" that remain bound even after all processes are dead. This required a system reboot to clear.

## Solution

Introduce a wrapper process (`worker-wrapper.cjs`) using a "Bun-in-a-Bun" architecture 🌯:
- Spawns the actual worker as a child with IPC channel
- On restart/shutdown, uses `taskkill /T /F` to kill the entire process tree
- Exits itself, allowing hooks to start fresh

The wrapper has no sockets, so Bun's socket cleanup bug doesn't affect it. When the wrapper kills the inner worker tree and exits, the port is properly released and can be immediately reused.

## Key changes

- New `worker-wrapper.ts` for Windows process lifecycle management
- `ProcessManager` starts wrapper on Windows, worker directly on Unix
- Worker sends IPC messages to wrapper for restart/shutdown
- Health endpoint now includes debug info (build ID, managed status, hasIpc)

## Test plan

- [x] Restart API now properly releases port and new worker binds to same port
- [x] Unix/Mac code path unchanged (wrapper only used on Windows)
- [x] Verified 3 bun.exe processes in steady state is expected (MCP server + wrapper + worker)


rsuk-mb:
@ToxMox Really pleased to see this and am running locally - loving claude-mem but the zombie port issue is driving me to distraction! Hope it gets merged soon!

@thedotmack

PR #387, opened 2025-12-19
Fix SessionStart hook error by using documented exit codes

laihenyi:
## Summary
- Fixed "SessionStart:startup hook error" that appeared on every session start
- Changed user-message-hook to exit with code 0 instead of undocumented code 3
- Updated hook-constants.ts to only include officially documented exit codes (0, 1, 2)

## Root Cause
The `user-message-hook.js` was using exit code 3 (`USER_MESSAGE_ONLY`), which is not officially supported by Claude Code. According to the [hooks documentation](https://code.claude.com/docs/en/hooks.md), only exit codes 0 and 2 have special handling - any other code is treated as a non-blocking error.

## Test plan
- [x] Verified hook exits with code 0 when run directly
- [x] Worker service health check passes
- [x] Session starts without "hook error" message

### DIFF
--- a/src/hooks/user-message-hook.ts
+++ b/src/hooks/user-message-hook.ts
@@ -62,4 +62,5 @@ This message was not added to your startup context, so you can continue working 
 `);
 }
 
-process.exit(HOOK_EXIT_CODES.USER_MESSAGE_ONLY);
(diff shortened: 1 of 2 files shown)


thedotmack:
This is intentional

PR #474, opened 2025-12-28
fix(windows): prevent libuv assertion failure in smart-install.js

CCavalancia:
## Summary
  - Fixes Windows-specific libuv assertion failure (`UV_HANDLE_CLOSING`) that occurs during session startup
  - Eliminates DEP0190 deprecation warnings about unescaped arguments with `shell: true`
  - Introduces `needsShell()` helper to intelligently determine when shell is required

  ## Problem
  On Windows, using `shell: true` with `spawnSync` causes:
  1. `Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 76`
  2. `[DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities`

  ## Solution
  - Prioritize checking common installation paths (e.g., `~/.bun/bin/bun.exe`) before PATH resolution
  - When we have an absolute path to an executable, run it directly without shell
  - Only use `shell: true` when running bare command names that require PATH resolution

  ## Test plan
  - [x] Tested on Windows 11 with Bun 1.3.5
  - [x] Verified no assertion failures on session startup
  - [x] Verified bun/uv version detection works correctly
  - [x] Verified `bun install` runs without errors
  - [ ] Maintainer verification on macOS/Linux (no behavioral changes expected)

### DIFF
--- a/plugin/scripts/smart-install.js
+++ b/plugin/scripts/smart-install.js
@@ -7,59 +7,72 @@
  */
 import { existsSync, readFileSync, writeFileSync } from 'fs';
 import { execSync, spawnSync } from 'child_process';
-import { join } from 'path';
+import { join, isAbsolute } from 'path';
 import { homedir } from 'os';
 
 const ROOT = join(homedir(), '.claude', 'plugins', 'marketplaces', 'thedotmack');
 const MARKER = join(ROOT, '.install-version');
 const IS_WINDOWS = process.platform === 'win32';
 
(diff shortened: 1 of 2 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1888, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #541, opened 2026-01-04
fix: remove SessionStart matcher blocking context injection

faustbrian:
The matcher "startup|clear|compact" was preventing SessionStart hooks from running and injecting context into new sessions. Removing it allows hooks to fire on all session starts.

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -3,7 +3,6 @@
   "hooks": {
     "SessionStart": [
       {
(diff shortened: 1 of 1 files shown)


thedotmack:
Thanks for the report! This matcher is intentional - it ensures context is only injected on `startup`, `clear`, and `compact` sessions, but **not** on `resume`. Resumed conversations already have context at the start of their transcript, so re-injecting would duplicate it.

If you're not seeing context on fresh starts, could you share more details about how you're starting the session?

PR #564, opened 2026-01-06
fix: prevent ECONNRESET on large transcripts

### LINKED ISSUE #562: when context very large , will bug
⎿  UserPromptSubmit:Callback hook succeeded: Success
  ⎿  UserPromptSubmit says: Plugin hook "node "/Users/blue_focus/.claude/plugins/cache/thedotmack/claude-mem/8.0.6/scripts/new-hook.js"" failed to start: The operation was 
     aborted. Check that the command exists and is executable.
  ⎿  API Error: Connection error.
  ⎿  Stop says: Plugin hook error: node:internal/deps/undici/undici:15445
           Error.captureStackTrace(err);
                 ^

     TypeError: fetch failed
         at node:internal/deps/undici/undici:15445:13
         at async Lt (file:///Users/blue_focus/.claude/plugins/cache/thedotmack/claude-mem/8.0.6/scripts/summary-hook.js:32:500)
         at async Socket.<anonymous> (file:///Users/blue_focus/.claude/plugins/cache/thedotmack/claude-mem/8.0.6/scripts/summary-hook.js:32:1062) {
       [cause]: Error: read ECONNRESET
           at TCP.onStreamRead (node:internal/stream_base_commons:216:20) {
         errno: -54,
         code: 'ECONNRESET',
         syscall: 'read'
       }
     }

     Node.js v24.7.0

like this

tt-a1i:
Fixes #562

## Problem

When the transcript file is very large (long sessions with many tool calls), the summary hook reads the entire file into memory and sends it to the worker. This can cause:
- Memory pressure in the hook/worker
- HTTP connection resets (ECONNRESET) when the payload is too large
- Worker crashes or OOM

## Solution

Add multi-layer truncation to keep payload sizes reasonable:

1. **summary-hook.ts** - Truncate before sending to worker (primary defense)
2. **SessionRoutes.ts** - Truncate at worker entry (secondary defense)
3. **prompts.ts** - Truncate before AI API call (final fallback)

The truncation preserves **head (~10k chars) + tail (~40k chars)** (character count, not bytes)  to keep both the initial context and the final results, which are most useful for summary generation.

## Changes

- New `src/shared/text-truncation.ts` utility
- Modified `src/hooks/summary-hook.ts` 
- Modified `src/services/worker/http/routes/SessionRoutes.ts`
- Modified `src/sdk/prompts.ts`
- Added tests for truncation utility

## AI Assistance Disclosure

I used Codex to review the changes, sanity-check the implementation against existing patterns, and help spot potential edge cases.


thedotmack:
Thanks for the thorough implementation! We need to investigate the root cause further first. Appreciate the time you put into this!

PR #593, opened 2026-01-07
fix: respect ANTHROPIC_MODEL env var for model selection

kaitranntt:
## Problem

When using claude-mem with tools that set `ANTHROPIC_MODEL` (e.g., CCS, custom Claude CLI wrappers), observer requests fail with:

```
400 unknown provider for model claude-sonnet-4-5
```

This happens because `SDKAgent.getModelId()` ignores the environment and always reads from `~/.claude-mem/settings.json`, defaulting to `claude-sonnet-4-5` even when the parent process specifies a different model.

## Solution

Check `process.env.ANTHROPIC_MODEL` first, fall back to settings file if not set.

```typescript
private getModelId(): string {
  if (process.env.ANTHROPIC_MODEL) {
    return process.env.ANTHROPIC_MODEL;
  }
  // ...existing settings.json fallback
}
```

## Impact

- Zero breaking changes (env var takes precedence, settings file still works)
- Enables compatibility with any tool that sets `ANTHROPIC_MODEL`
- Hooks inherit env from parent Claude process, so this "just works"


ham-zax:
+

thedotmack:
Thanks for identifying this! We need to better understand the issue before implementing. Appreciate bringing this to our attention!

PR #613, opened 2026-01-08
fix: generate memorySessionId for non-SDK providers

### LINKED ISSUE #591: OpenRouter agent fails to capture memorySessionId for sessions with empty prompt history
## Description
When a new session is created with OpenRouter as the provider, the agent fails to capture the `memorySessionId` from the API response if the session has no previous prompts (`prompt_counter = 0`). This causes the generator to fail with the error:
```
Cannot store observations: memorySessionId not yet captured
```

## Environment
- **claude-mem version**: 9.0.0
- **Provider**: openrouter
- **Model**: xiaomi/mimo-v2-flash:free
- **Platform**: linux

## Steps to Reproduce
1. Start a new Claude Code session in a fresh project directory
2. The claude-mem plugin initializes a new session with `prompt_counter = 0`
3. The OpenRouter agent is called but fails to capture `memorySessionId`
4. Check the database: `memory_session_id` is NULL

## Expected Behavior
The OpenRouter agent should return a `session_id` in its response, which should be captured and stored in the `sdk_sessions.memory_session_id` field.

## Actual Behavior
The `memory_session_id` field remains NULL, causing all subsequent observation storage to fail.

## Workaround
Sessions that have pending prompts (`prompt_counter > 0`) trigger a "startup-recovery" flow when the worker restarts, which successfully captures the `memorySessionId`. However, sessions with empty history don't trigger this recovery.

## Additional Context

The issue occurs in the `Kn` function in `worker-service.cjs`:
```javascript
if(!e.memorySessionId)throw new Error("Cannot store observations: memorySessionId not yet captured");
```

The `memorySessionId` is supposed to be captured from the OpenRouter response:
```javascript
if(!e.memorySessionId&&p.session_id){
  e.memorySessionId=p.session_id,
  this.dbManager.getSessionStore().updateMemorySessionId(e.sessionDbId,p.session_id);
```

### Possible causes:
1. OpenRouter API (`xiaomi/mimo-v2-flash:free`) may not return `session_id` in responses
2. The response format may be different than expected
3. There may be a timing issue where the check happens before the response is processed

### Database Query to Verify
```sql
SELECT id, project, prompt_counter, memory_session_id
FROM sdk_sessions
WHERE project = 'YourProjectName';
```

If `prompt_counter = 0` and `memory_session_id IS NULL`, this bug is present.

### Log Evidence
```
[ERROR] [SESSION] [session-2084] Generator failed {provider=openrouter, error=Cannot store observations: memorySessionId not yet captured}
[WARN] [SESSION] [session-2084] Generator exited unexpectedly
```

Compared to a successful session that had pending prompts:
```
[SYSTEM] Worker started {pid=19007}
[SYSTEM] [session-2061] Starting generator (startup-recovery)
[SESSION] [session-2061] MEMORY_ID_CAPTURED | memorySessionId=c52bf83d-5a64-4aef-a7e0-4f88a231e2ba
```

JiehoonKwak:
Fixes #591

OpenRouter/Gemini agents fail because they don't receive session_id from API. Generate synthetic ID before processAgentResponse().

### DIFF
--- a/src/services/worker/GeminiAgent.ts
+++ b/src/services/worker/GeminiAgent.ts
@@ -145,6 +145,13 @@ export class GeminiAgent {
       session.conversationHistory.push({ role: 'user', content: initPrompt });
       const initResponse = await this.queryGeminiMultiTurn(session.conversationHistory, apiKey, model, rateLimitingEnabled);
 
+      // Gemini doesn't return session_id like Claude SDK, generate one for FK constraint
+      if (!session.memorySessionId) {
+        const syntheticId = `gemini-${session.contentSessionId}-${Date.now()}`;
+        session.memorySessionId = syntheticId;
+        this.dbManager.getSessionStore().updateMemorySessionId(session.sessionDbId, syntheticId);
+      }
+
       if (initResponse.content) {
         // Add response to conversation history
         session.conversationHistory.push({ role: 'assistant', content: initResponse.content });
--- a/src/services/worker/OpenRouterAgent.ts
+++ b/src/services/worker/OpenRouterAgent.ts
@@ -103,6 +103,13 @@ export class OpenRouterAgent {
       session.conversationHistory.push({ role: 'user', content: initPrompt });
       const initResponse = await this.queryOpenRouterMultiTurn(session.conversationHistory, apiKey, model, siteUrl, appName);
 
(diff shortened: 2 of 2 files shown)


JiehoonKwak:
revised commit d/t linter issue

PR #615, opened 2026-01-08
fix: generate memorySessionId for stateless providers

### LINKED ISSUE #591: OpenRouter agent fails to capture memorySessionId for sessions with empty prompt history

JiehoonKwak:
## Problem

OpenRouter/Gemini APIs are stateless—they don't return session IDs. This breaks the FK constraint, causing observations to fail silently.

Fixes #591

## Fix

Generate synthetic memorySessionId before first API call. 21 lines across 3 files.

## Testing

- Verified `MEMORY_ID_GENERATED` in logs
- Observations store correctly
- No regressions

### DIFF
--- a/src/services/worker/http/routes/SettingsRoutes.ts
+++ b/src/services/worker/http/routes/SettingsRoutes.ts
@@ -121,6 +121,7 @@ export class SettingsRoutes extends BaseRouteHandler {
       // Feature Toggles
       'CLAUDE_MEM_CONTEXT_SHOW_LAST_SUMMARY',
       'CLAUDE_MEM_CONTEXT_SHOW_LAST_MESSAGE',
+      'CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED',
     ];
 
     for (const key of settingKeys) {
@@ -297,6 +298,7 @@ export class SettingsRoutes extends BaseRouteHandler {
       'CLAUDE_MEM_CONTEXT_SHOW_SAVINGS_PERCENT',
       'CLAUDE_MEM_CONTEXT_SHOW_LAST_SUMMARY',
       'CLAUDE_MEM_CONTEXT_SHOW_LAST_MESSAGE',
+      'CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED',
     ];
 
     for (const key of booleanSettings) {
--- a/src/shared/SettingsDefaultsManager.ts
+++ b/src/shared/SettingsDefaultsManager.ts
@@ -5,10 +5,13 @@
  * Provides methods to get defaults with optional environment variable overrides.
  */
 
-import { readFileSync, writeFileSync, existsSync, mkdirSync } from 'fs';
-import { join, dirname } from 'path';
-import { homedir } from 'os';
-import { DEFAULT_OBSERVATION_TYPES_STRING, DEFAULT_OBSERVATION_CONCEPTS_STRING } from '../constants/observation-metadata.js';
+import { readFileSync, writeFileSync, existsSync, mkdirSync } from "fs";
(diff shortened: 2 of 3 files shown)


denjzs:
LGTM... @thedotmack can approve?

thedotmack:
Thank you for this fix, @JiehoonKwak! The core synthetic memorySessionId generation for GeminiAgent and OpenRouterAgent has been applied directly to main — the PR was a month old with conflicts from 3+ recently merged PRs (#693, #913, #940).

**What was applied:**
- Synthetic memorySessionId generation in `GeminiAgent.ts` and `OpenRouterAgent.ts` (your 16-line core fix)

**What was already in main:**
- `sessions/create.ts` project backfill (merged via PR #940)
- `CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED` setting (merged via PR #913)
- Quote style changes in `SettingsDefaultsManager.ts` and `claude-md-utils.ts` were formatting-only

This fixes a real bug where stateless providers would hit the `memorySessionId` defensive checks (from PR #693) and fail to process observations. Your fix generates the ID early, before any API calls. Thank you!

PR #634, opened 2026-01-09
fix: respect CLAUDE_CONFIG_DIR for plugin paths (#626)

### LINKED ISSUE #626: Bug: HealthMonitor hardcodes ~/.claude path, fails with custom config directories
## Description

  The `getInstalledPluginVersion()` function in `src/services/infrastructure/HealthMonitor.ts` hardcodes the path to `~/.claude/plugins/marketplaces/thedotmack/package.json`. This fails for users with custom Claude config directories.

  ## Error

  ENOENT: no such file or directory, open '/Users/$USER/.claude/plugins/marketplaces/thedotmack/package.json'

  ## Root Cause

  Claude Code supports custom config directories via the `CLAUDE_CONFIG_DIR` environment variable or org-specific directories (e.g., `~/.claude-orgname/`). The plugin is installed at:
  ~/.claude-/plugins/cache/thedotmack/claude-mem//

  But `HealthMonitor.ts` hardcodes:
  ```typescript
  const pluginDir = path.join(homedir(), '.claude', 'plugins', 'marketplaces', 'thedotmack');
  const packagePath = path.join(pluginDir, 'package.json');

  Suggested Fix

  Instead of hardcoding, resolve the plugin directory dynamically:

  1. Option A: Use __dirname to find the package.json relative to the running code:
  const packagePath = path.join(__dirname, '..', '..', 'package.json');
  2. Option B: Check CLAUDE_CONFIG_DIR environment variable:
  const claudeDir = process.env.CLAUDE_CONFIG_DIR || path.join(homedir(), '.claude');
  3. Option C: Look for package.json in the actual install location (from __filename):
  // Walk up from current file to find package.json
  let dir = __dirname;
  while (dir !== path.dirname(dir)) {
    const pkgPath = path.join(dir, 'package.json');
    if (fs.existsSync(pkgPath)) {
      // Verify it's the claude-mem package
      const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf-8'));
      if (pkg.name === 'claude-mem') return pkg.version;
    }
    dir = path.dirname(dir);
  }

  Option A is simplest and most robust since the code is bundled and __dirname will always be relative to the actual install location.

  Environment

  - claude-mem version: 9.0.1
  - Platform: macOS (darwin arm64)
  - Claude config: Custom directory (~/.claude-<orgname>/)

  ---

Kuroakira:
## Problem

  When users set `CLAUDE_CONFIG_DIR` to a custom location (e.g., `~/.claude-work`), multiple claude-mem features fail because paths were hardcoded to `~/.claude`.

  The most visible failure is in `HealthMonitor.ts` (reported in #626):

  ENOENT: no such file or directory, open '/Users/$USER/.claude/plugins/marketplaces/thedotmack/package.json'

  ## Root Cause

  Several files independently calculated the plugin path using `path.join(homedir(), '.claude', ...)` instead of respecting `CLAUDE_CONFIG_DIR`. This is a long-standing issue - PR #226 attempted a partial fix but was deferred, and it didn't include `HealthMonitor.ts`.

  ## Solution

  1. Added `MARKETPLACE_ROOT` constant to `src/shared/paths.ts` that respects `CLAUDE_CONFIG_DIR`
  2. Updated all files with hardcoded paths to use centralized constants from `paths.ts`

  ## Files Changed

  | File | Change |
  |------|--------|
  | `src/shared/paths.ts` | Added `MARKETPLACE_ROOT` constant |
  | `src/services/infrastructure/HealthMonitor.ts` | Use `MARKETPLACE_ROOT` for version check |
  | `src/shared/worker-utils.ts` | Import `MARKETPLACE_ROOT` instead of local definition |
  | `src/services/worker/BranchManager.ts` | Import `MARKETPLACE_ROOT` instead of local definition |
  | `src/services/integrations/CursorHooksInstaller.ts` | Use `MARKETPLACE_ROOT` and `CLAUDE_CONFIG_DIR` |
  | `src/services/context/ObservationCompiler.ts` | Use `CLAUDE_CONFIG_DIR` for transcript path |

  ## Testing

  - Build passes
  - All hardcoded `~/.claude` references replaced with environment-aware constants
  - Backwards compatible: defaults to `~/.claude` when `CLAUDE_CONFIG_DIR` is not set

  Fixes #626


thedotmack:
Thank you for this excellent fix, @Kuroakira! Your changes have been applied to main (commit 75a0f2e9).

The PR had merge conflicts in build artifacts (plugin/scripts/*) due to recent merges, but the 6 source file changes applied cleanly. We cherry-picked the source changes directly onto main rather than resolving the build artifact conflicts.

**Changes applied:**
- Added `MARKETPLACE_ROOT` constant to `src/shared/paths.ts`
- Updated `HealthMonitor.ts`, `worker-utils.ts`, `BranchManager.ts`, `CursorHooksInstaller.ts`, and `ObservationCompiler.ts` to use centralized path constants

This fixes #626. Thank you for the contribution!

PR #644, opened 2026-01-09
feat: Add OpenAI provider support with memorySessionId fallback

niteeshm:
# Summary
Implements OpenAI-compatible API provider as an alternative to Claude SDK for observation extraction. Users can now bring their own API key and use any OpenAI-compatible provider (OpenAI, Azure OpenAI, Poe API, etc.) for the memory compression system.

# Key Changes
- **New OpenAIAgent.ts** - OpenAI-compatible REST API client with multi-turn conversation support
- **Provider switching** - Added OpenAI option to SessionRoutes and worker service
- **UI Settings** - Added OpenAI configuration fields (API key, base URL, model name)
- **memorySessionId fallback** - Auto-generates ID when models don't return `<memory_session_id>` tags
- **OpenRouter fix** - Same fallback logic applied to OpenRouter agent

# Why
- **Bring Your Own Key** - Users can use their own OpenAI/Azure/Poe API keys instead of being limited to Claude
- **Provider flexibility** - Works with any OpenAI-compatible API (OpenAI, Azure, Poe, local models, etc.)
- **Cost control** - Users can choose cheaper alternatives like free Poe bots or local models
- **Bug fix** - Resolves "Cannot store observations: memorySessionId not yet captured" error when models don't return session IDs

---

**Co-Authored-By:** Claude <noreply@anthropic.com>


niteeshm:
image
image


All working well

niteeshm:
cc @thedotmack

JiehoonKwak:
This should be merged

kiwina:
is the openrouter not basically the same?? that would just require a few changes for example a custom url? the api is basically the same, that be better than adding useeless code duplication

thedotmack:
Thank you for the contribution, @niteeshm! The implementation is solid and well-documented.

After evaluating this against our provider strategy, we've decided to keep the provider count manageable. OpenAI models (and Azure OpenAI, Poe, and other OpenAI-compatible endpoints) are already accessible via the **OpenRouter provider**, which supports 100+ models through a single integration — including all OpenAI models.

As @kiwina noted in the comments, the OpenRouter agent already uses the same OpenAI-compatible chat/completions API format. Adding a dedicated OpenAI agent would create significant code duplication and maintenance burden for functionality that's largely covered.

If you need a custom base URL for Azure or other OpenAI-compatible endpoints, we'd be open to a PR that makes the OpenRouter agent's base URL configurable — that would be a much smaller change that serves the same use case.

Thanks again for the effort!

PR #680, opened 2026-01-11
feat(openrouter): multi-model configuration with automatic fallback

RyderFreeman4Logos:
## Summary

- **Multi-model OpenRouter support**: Configure multiple models via comma-separated list (`CLAUDE_MEM_OPENROUTER_MODEL="model1,model2,model3"`). When quota is exhausted on one model, automatically tries the next.
- **Multi-level provider fallback**: OpenRouter → Gemini → Claude SDK fallback chain ensures continuous operation during provider outages.
- **Configurable OpenRouter base URL**: Support for local proxies and alternative OpenRouter-compatible endpoints.
- **Gemini 3 default**: Updated default Gemini model to `gemini-3-flash`.
- **Settings hot-reload**: Configuration file changes are picked up without worker restart via `fs.watch()`.
- **Empty model list protection**: Default to fallback model when configuration is invalid.

## Test plan

- [ ] Configure `CLAUDE_MEM_OPENROUTER_MODEL` with multiple models and verify fallback on quota exhaustion
- [ ] Configure `CLAUDE_MEM_OPENROUTER_BASE_URL` to a custom endpoint and verify it is used
- [ ] Test settings persistence through viewer UI
- [ ] Verify settings API accepts and validates new fields
- [ ] Test Gemini fallback when OpenRouter fails
- [ ] Test Claude SDK fallback when both OpenRouter and Gemini fail

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,8 @@ src/ui/viewer.html
 # Prevent other malformed path directories
 http*/
 https*/
+
+# Auto-generated CLAUDE.md files (keep root CLAUDE.md only)
+**/CLAUDE.md
+!CLAUDE.md
+!.claude/CLAUDE.md


thedotmack:
Thank you for the contribution @RyderFreeman4Logos! Multi-model fallback is an interesting concept, but we need to close this PR for several reasons:

**Scope concerns:**
1. This PR bundles 5+ distinct features (multi-model fallback, provider fallback chain, configurable base URL, settings hot-reload via `fs.watch()`, Gemini 3 default update) into one 28-file changeset. Each of these should be separate, focused PRs.

**Destructive changes:**
2. The PR deletes the content of 12+ `CLAUDE.md` documentation files across the project (docs/public/, plugin/, src/services/, src/ui/viewer/, etc.), removing ~1,200 lines of project documentation. These files are important for developer context and should not be modified in a feature PR.

**YAGNI concerns:**
3. Multi-model fallback with automatic quota exhaustion detection adds significant complexity for a niche use case. Most users configure a single model and it works fine. The multi-level provider fallback chain (OpenRouter → Gemini → Claude SDK) is also premature — the existing single-provider approach is sufficient.

**What we'd welcome:**
- A focused PR for **configurable OpenRouter base URL only** (the simplest, most useful part of this changeset) — without modifying any CLAUDE.md files or bundled artifacts
- If you have evidence of users hitting quota exhaustion regularly, we'd reconsider the multi-model fallback as a separate PR

Thank you for your interest in improving claude-mem!

PR #378, opened 2025-12-17
fix(windows): Windows platform stabilization improvements

thedotmack:
## Summary

Comprehensive Windows platform stabilization focusing on worker lifecycle management, process cleanup, and cross-platform compatibility.

**Key Improvements:**

- **Worker Readiness Tracking**: Added `/api/readiness` endpoint with MCP/SDK initialization flags to prevent premature connection attempts
- **Process Tree Cleanup**: Implemented recursive process enumeration on Windows to prevent zombie socket processes
- **Bun Runtime Migration**: Migrated worker wrapper from Node.js to Bun for consistency and reliability
- **Centralized Project Name Utility**: Consolidated duplicate project name extraction logic with Windows drive root handling
- **Enhanced Error Messages**: Added platform-aware logging and detailed Windows troubleshooting guidance
- **Subprocess Console Hiding**: Standardized `windowsHide: true` across all child process spawns to prevent console window flashing

**Technical Changes:**

- Worker service tracks MCP and SDK readiness states separately
- ChromaSync service properly tracks subprocess PIDs for Windows cleanup
- Worker wrapper uses Bun runtime with enhanced socket cleanup via process tree enumeration
- Increased timeouts on Windows platform (30s worker startup, 10s hook timeouts)
- Logger utility includes platform and PID information for better debugging

**Files Changed:**
- 26 files with 591 insertions and 239 deletions
- Core changes in worker service, process management, and cross-platform utilities

This represents a major reliability improvement for Windows users, eliminating common issues with worker startup failures, orphaned processes, and zombie sockets.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

PR #339, opened 2025-12-16
fix: Windows console popup issue (high priority)

thedotmack:
## Summary

**HIGH PRIORITY FIX** for Windows 11 users experiencing blank console windows during claude-mem operations.

Uses PowerShell `Start-Process -WindowStyle Hidden` workaround to properly hide worker process on Windows.

## Problem

Multiple blank console windows with "claude" title appear on Windows 11 when using claude-mem. This is caused by a limitation where `windowsHide: true` doesn't work with `detached: true` in `child_process.spawn()`.

**Critical Finding from PR #315**: This limitation affects **BOTH Bun AND Node.js** because Bun inherits Node.js process spawning semantics.

## Root Cause

From ToxMox's testing in PR #315:

> "The `windowsHide: true` approach does not work on Windows when combined with `detached: true`. Blank terminal windows still appear."
>
> "This is a known Node.js limitation documented in nodejs/node#21825: `windowsHide: true` is ignored when `detached: true` is also set."
>
> **"The behavior persists in Bun as well, since Bun inherits Node.js process spawning semantics."**

### Windows (PowerShell workaround):
```typescript
const psCommand = `Start-Process -FilePath '${bunPath}' -ArgumentList '${script}' -WindowStyle Hidden`;
spawnSync('powershell', ['-Command', psCommand]);
```

### Unix (standard approach):
```typescript
spawn(bunPath, [script], { detached: true, windowsHide: true });
```

## Testing

Validated by ToxMox on Windows 11 Pro:

| Approach | Result |
|----------|--------|
| `windowsHide: true` only | ❌ Blank consoles still appear |
| PowerShell `-WindowStyle Hidden` | ✅ Properly hides windows |

## Changes

- **ProcessManager.ts**: Added Windows-specific PowerShell spawn logic
- Keeps Bun runtime (doesn't address zombie socket issue)
- No changes to Unix/Mac behavior

## What This Fixes

✅ Blank "claude" console windows on Windows 11  
✅ Works with Bun runtime (current default)  
✅ Validated by multiple Windows users

## What This Doesn't Fix

❌ Bun zombie socket issue (EADDRINUSE after worker restart)
- Separate issue requiring different solution (switch to Node or require reboots)
- Documented in PR #315 with upstream Bun issues linked

## Priority

**HIGH** - This is a major UX issue for Windows users. Should be merged before queue monitoring features.

## Related Issues

- Fixes: #304 (Multiple visible console windows)
[...]

PR #682, opened 2026-01-11
fix(windows): add windowsHide to smart-install.js spawn/exec calls

samonjoat:
## Summary

- Adds `windowsHide: true` to all 15 `spawnSync` and `execSync` calls in `plugin/scripts/smart-install.js`
- Prevents console windows from flashing on Windows when claude-mem hooks run

## Problem

Windows users experience console windows rapidly opening and closing during every `PostToolUse` hook execution. This happens because the `smart-install.js` script (which runs as part of the hooks) spawns child processes without the `windowsHide: true` option.

While the main source files in `src/` (like `ProcessManager.ts`, `BranchManager.ts`, `SDKAgent.ts`) already have this fix from previous PRs (#166, #203, #339, #378), the `plugin/scripts/smart-install.js` file was missed.

## Changes

Added `windowsHide: true` to all spawn/exec calls:
- Version checks for `bun` and `uv` (5 calls)
- Installation commands for bun and uv (4 calls)
- Directory creation (1 call)
- Package installation - bun/npm (3 calls)
- Worker shutdown and restart commands (2 calls)

## Test plan

- [ ] Test on Windows 11 - verify no console windows flash during hook execution
- [ ] Test that bun/uv version checks still work correctly
- [ ] Test that package installation still works correctly

### DIFF
--- a/plugin/scripts/smart-install.js
+++ b/plugin/scripts/smart-install.js
@@ -22,7 +22,8 @@ function isBunInstalled() {
     const result = spawnSync('bun', ['--version'], {
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
-      shell: IS_WINDOWS
+      shell: IS_WINDOWS,
+      windowsHide: true
     });
     if (result.status === 0) return true;
   } catch {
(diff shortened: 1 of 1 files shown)


samonjoat:
Closing - this PR doesn't solve the main issue.

Issue #678, opened 2026-01-11
Resolution: PostToolUse hook not creating observations - switch to claude provider


pitimon:
## Problem

PostToolUse hook was firing but no observations were being stored in the database. Logs showed:

```
Generator auto-starting (observation) using OpenRouter ✅
← Response received: "I'm ready to continue observing..."
STORED | obsCount=0 | obsIds=[] ❌
```

Additional symptoms:
- 156 failed pending_messages accumulated
- SDK generator failing with "Claude Code process exited with code 1"
- Worker version mismatch (9.0.3 vs 9.0.4) causing auto-restarts

## Diagnosis Steps

1. **Check pending_messages table**
   ```bash
   sqlite3 ~/.claude-mem/claude-mem.db "SELECT status, COUNT(*) FROM pending_messages GROUP BY status;"
   # Result: failed|156, processed|100
   ```

2. **Check provider setting**
   ```bash
   cat ~/.claude-mem/settings.json | jq '.CLAUDE_MEM_PROVIDER'
   # Result: "openrouter"
   ```

3. **Check worker logs**
   ```bash
   tail -50 ~/.claude-mem/logs/claude-mem-*.log | grep -i "generator\|error\|obsCount"
   ```

### Step 1: Clear failed pending messages
```bash
sqlite3 ~/.claude-mem/claude-mem.db "DELETE FROM pending_messages WHERE status = 'failed';"
```

### Step 2: Change provider to claude
```bash
jq '.CLAUDE_MEM_PROVIDER = "claude"' ~/.claude-mem/settings.json > /tmp/settings.json && \
mv /tmp/settings.json ~/.claude-mem/settings.json
```

### Step 3: Rebuild worker (fix version mismatch)
```bash
cd ~/.claude/plugins/marketplaces/thedotmack
npm run worker:stop
npm run build
npm run worker:start
```

### Step 4: Verify version match
```bash
curl -s http://localhost:37777/api/version
# Should match package.json version
```

### Step 5: Reset corrupted session (if needed)
```bash
# Find active session with stuck state
sqlite3 ~/.claude-mem/claude-mem.db "SELECT id FROM sdk_sessions WHERE status = 'active';"

# Reset it
sqlite3 ~/.claude-mem/claude-mem.db "UPDATE sdk_sessions SET memory_session_id = NULL, status = 'completed' WHERE id = <session_id>;"

# Clear any remaining pending messages
[...]

Issue #677, opened 2026-01-11
OpenRouter provider fails to create observations (obsCount=0)


pitimon:
## Summary

When using `CLAUDE_MEM_PROVIDER=openrouter`, the PostToolUse hook successfully receives responses from OpenRouter API but fails to store observations in the database.

## Symptoms

```
Generator auto-starting (observation) using OpenRouter ✅
← Response received: "I'm ready to continue observing..."
STORED | obsCount=0 | obsIds=[] ❌
```

- OpenRouter API responds successfully
- No observations are created (obsCount=0)
- SDK generator fails with: `Claude Code process exited with code 1`

## Environment

- Plugin version: 9.0.4
- Worker version: 9.0.4
- OpenRouter model: `xiaomi/mimo-v2-flash:free`
- Provider setting: `openrouter`

## Root Cause Analysis

The SDK generator uses Claude Code subprocess to generate observations. When provider is set to `openrouter`:

1. The generator starts correctly
2. OpenRouter API call succeeds and returns a response
3. But the response parsing/storage fails silently
4. Generator process exits with code 1

## Workaround

Change provider back to `claude`:

```bash
jq '.CLAUDE_MEM_PROVIDER = "claude"' ~/.claude-mem/settings.json > /tmp/settings.json && \
mv /tmp/settings.json ~/.claude-mem/settings.json
```

Then restart worker.

## Related

- Similar to upstream issue #587 "SDK agent stuck on 'Awaiting tool execution data'"
- May be related to how OpenRouter responses are parsed vs Claude API responses

## Investigation Notes

- Pending messages accumulate with `failed` status (retry_count=3)
- Session state can become corrupted (memory_session_id stuck)
- Clearing `pending_messages WHERE status='failed'` and resetting session helps recover

## Labels

bug, provider, openrouter

pitimon:
Related: #678 (Resolution guide for this issue)

MangoBango90:
https://github.com/thedotmack/claude-mem/issues/677

thedotmack:
Resolution documented in #678 — switching to the Claude provider resolves the observation creation issue. OpenRouter provider compatibility has also been improved in recent releases.

Issue #623, opened 2026-01-08
v9.0.0: Crash-recovery loop when memory_session_id is not captured


mrlfarano:
## Bug Description

Sessions created without `memory_session_id` cause an infinite crash-recovery loop. The generator continuously retries and fails with:
```
[ERROR] [SDK] ✗ OpenRouter agent error {sessionDbId=607} Cannot store observations: memorySessionId not yet captured
[INFO] [SESSION] [session-607] Generator auto-starting (observation) using OpenRouter
```

This loop runs indefinitely, growing the queue depth and consuming API tokens on every retry attempt.

## Environment

- **claude-mem version:** 9.0.0
- **OS:** macOS (Darwin)
- **Provider:** OpenRouter (mimo-v2-flash:free, also reproduced with gpt-4o-mini)
- **Node version:** (run `node -v` and add here)

## Steps to Reproduce

1. Start a Claude Code session
2. Session gets created in `sdk_sessions` table but `memory_session_id` column remains NULL/empty
3. Observations are enqueued to `pending_messages`
4. Generator attempts to process queue
5. Fails with "Cannot store observations: memorySessionId not yet captured"
6. Generator auto-restarts (crash-recovery)
7. Loop continues indefinitely

## Evidence from Logs
```
[2026-01-08 11:19:16.784] [SDK] OpenRouter API usage {model=xiaomi/mimo-v2-flash:free, inputTokens=10452, outputTokens=129}
[2026-01-08 11:19:16.784] [ERROR] [SDK] ✗ OpenRouter agent error {sessionDbId=607} Cannot store observations: memorySessionId not yet captured
[2026-01-08 11:19:16.784] [INFO] [SESSION] [session-607] Generator aborted
[2026-01-08 11:19:16.853] [INFO] [SESSION] [session-607] Generator auto-starting (observation) using OpenRouter
```

## Database State

Sessions missing `memory_session_id`:
```sql
SELECT id, content_session_id, memory_session_id, status FROM sdk_sessions WHERE memory_session_id IS NULL OR memory_session_id = '';

-- Results:
-- 607|a2265efb-c878-4be4-b2f5-1ed2323cc607||active
-- 605|83f05013-e4e2-4564-8cec-f03dfc8c5eb7||active
-- (multiple sessions affected)
```

## Expected Behavior

1. Sessions should not be created until `memory_session_id` is captured
2. OR: Generator should skip/fail gracefully for sessions missing `memory_session_id` instead of infinite retry
[...]

PR #693, opened 2026-01-12
fix: prevent infinite restart loop that causes runaway API costs

ajbmachon:
## Problem

When `memorySessionId` is null (due to a race condition during session initialization), the generator enters an **infinite restart loop**:

1. Generator starts → Calls LLM API ✅ (tokens charged)
2. `processAgentResponse()` → throws "memorySessionId not yet captured" ❌
3. `.finally()` sees `pendingCount > 0` → Restarts generator after 1 second
4. **Go to step 1** (NO RETRY LIMIT!)

### Real-World Impact

I experienced **$402 in wasted API costs** before my OpenRouter key limit stopped it. The logs showed:

```
[ERROR] ✗ OpenRouter agent error - Cannot store observations: memorySessionId not yet captured
[INFO ] Restarting generator after crash/exit with pending work {pendingCount=429}
```

This repeated **24,219 times** in a single day. Each restart made another LLM call that succeeded (cost tokens) but then failed to store the result.

### 1. Add Restart Limit (Primary Fix)

```typescript
// SessionRoutes.ts
const MAX_CONSECUTIVE_RESTARTS = 3;

if (session.consecutiveRestarts > MAX_CONSECUTIVE_RESTARTS) {
  logger.error('SESSION', 'CRITICAL: Generator restart limit exceeded - stopping to prevent runaway costs');
  return;
}
```

### 2. Add Exponential Backoff

```typescript
// Backoff: 1s → 2s → 4s
const backoffMs = Math.min(1000 * Math.pow(2, session.consecutiveRestarts - 1), 8000);
```

### 3. Fail Fast - Check Before LLM Call (Defensive)

```typescript
// OpenRouterAgent.ts and GeminiAgent.ts
if (!session.memorySessionId) {
  throw new Error('Cannot process observations: memorySessionId not yet captured');
}
// THEN call expensive LLM
await this.queryOpenRouterMultiTurn(...);
```

This check happens **before** the LLM call, preventing token waste.

## Files Changed

| File | Change |
|------|--------|
| `worker-types.ts` | Add `consecutiveRestarts: number` to `ActiveSession` |
| `SessionManager.ts` | Initialize `consecutiveRestarts: 0` |
| `SessionRoutes.ts` | Add restart limit (3), exponential backoff, reset on success |
| `OpenRouterAgent.ts` | Check `memorySessionId` before LLM calls |
| `GeminiAgent.ts` | Same defensive check |

## Testing

1. Build succeeds: `npm run build` ✅
2. Logic verified against original logs showing the infinite loop pattern
3. Fix preserves legitimate crash recovery for transient failures (3 retries)

## Notes

- The root cause (why `memorySessionId` isn't captured in some sessions) may still need investigation
- This fix is a **safety guardrail** to prevent catastrophic costs regardless of root cause
- Consider adding alerting/monitoring for when the restart limit is hit

### DIFF
--- a/src/services/worker-types.ts
+++ b/src/services/worker-types.ts
@@ -33,6 +33,7 @@ export interface ActiveSession {
   earliestPendingTimestamp: number | null;  // Original timestamp of earliest pending message (for accurate observation timestamps)
   conversationHistory: ConversationMessage[];  // Shared conversation history for provider switching
   currentProvider: 'claude' | 'gemini' | 'openrouter' | null;  // Track which provider is currently running
+  consecutiveRestarts: number;  // Track consecutive restart attempts to prevent infinite loops
 }
 
 export interface PendingMessage {
--- a/src/services/worker/GeminiAgent.ts
+++ b/src/services/worker/GeminiAgent.ts
@@ -184,6 +184,12 @@ export class GeminiAgent {
             session.lastPromptNumber = message.prompt_number;
           }
 
+          // CRITICAL: Check memorySessionId BEFORE making expensive LLM call
+          // This prevents wasting tokens when we won't be able to store the result anyway
+          if (!session.memorySessionId) {
+            throw new Error('Cannot process observations: memorySessionId not yet captured. This session may need to be reinitialized.');
+          }
+
           // Build observation prompt
           const obsPrompt = buildObservationPrompt({
             id: 0,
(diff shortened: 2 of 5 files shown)


denjzs:
Related Issue:
- https://github.com/thedotmack/claude-mem/issues/623
- https://github.com/thedotmack/claude-mem/issues/591
- https://github.com/thedotmack/claude-mem/issues/677
- https://github.com/thedotmack/claude-mem/issues/678

PR:
- https://github.com/thedotmack/claude-mem/pull/615

bayanoj330-dev (approved):

thedotmack:
Applied to main in commit 6382d6f9. Thank you @ajbmachon for this important safety fix! The restart limit, exponential backoff, and defensive memorySessionId checks are all included. Applied directly to current main since the branch was 3+ weeks old.

bayanoj330-dev (approved):

PR #700, opened 2026-01-13
fix(#681): eliminate Windows Terminal popup by removing spawn-based daemon

### LINKED ISSUE #681: Windows Terminal Popup Regression
---

  Issue Template: Windows Terminal Popup Regression

  Title: [BUG] Blank terminal windows pop up on Windows during hook execution (ProcessManager regression)

  Labels: bug, platform-windows, needs-triage

  ---

  Bug Description
  On Windows systems, blank console/terminal windows appear and disappear rapidly whenever a hook is triggered (e.g., during SessionStart, UserPromptSubmit, or PostToolUse).  
  This causes visual flickering and interrupts the user workflow.

  Root Cause
  The src/services/infrastructure/ProcessManager.ts was using the native Node.js/Bun spawn() method with the following configuration:

   1 spawn(process.execPath, [scriptPath, '--daemon'], {
   2   detached: true,
   3   windowsHide: true, // <--- This fails on Windows when detached is true
   4   // ...
   5 });
  There is a known limitation in Node.js (and Bun) where windowsHide: true does not work as expected when detached: true is used, especially when spawning shell-like processes  or wrappers. Reference: Node.js Issue #21825 (https://github.com/nodejs/node/issues/21825).

  A previous fix existed in the legacy worker-cli.ts but was lost during the modular refactor that introduced the current ProcessManager.ts.

  Steps to Reproduce
   1. Use Claude Code on Windows 10/11.
   2. Start a new session or submit a prompt.
   3. Observe multiple blank terminal windows flashing on the taskbar and screen.

  Expected Behavior
  The worker daemon should start silently in the background without any visible terminal windows popping up.

  Proposed Fix / Implemented Solution
  The fix involves wrapping the process launch in a PowerShell command using -WindowStyle Hidden, which is the only reliable way to hide detached processes on Windows:        

   1 // Windows-specific launch in ProcessManager.ts
   2 const psCommand = `Start-Process -FilePath '${escapedExecPath}' -ArgumentList '${escapedArgs}' -WindowStyle Hidden -PassThru | Select-Object -ExpandProperty Id`;
   3 
   4 spawnSync('powershell', ['-NoProfile', '-NonInteractive', '-Command', psCommand], {
   5   windowsHide: true,
   6   stdio: 'pipe'
   7 });

  Key Improvements in the Fix:
   1. Reliable Hiding: Uses PowerShell's native window management.
   2. Environment Parity: Properly passes extraEnv and CLAUDE_MEM_WORKER_PORT through the PowerShell scope.
   3. Safety: Includes escapePowerShellString to prevent path injection.
   4. Efficiency: Adds -NoProfile and -NonInteractive to PowerShell to ensure fast execution and clean output.

  Environment
   - OS: Windows10 (win32)
   - Claude-mem version: 9.0.4+
   - Platform: Node.js / Bun

thedotmack:
## Summary
- Removes all process spawning from worker lifecycle
- Worker `start` command now becomes daemon directly (no child spawn)
- Removes `restart` command (users do stop then start)
- Removes ineffective windowsHide from ChromaSync (MCP SDK doesn't support it)
- Combines hook commands in SessionStart for proper lifecycle

## Test plan
- [ ] Build succeeds: `npm run build`
- [ ] Windows: No terminal popups on hook execution
- [ ] Worker starts correctly: `bun worker-service.cjs start`
- [ ] Worker stops correctly: `bun worker-service.cjs stop`

Fixes #681

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -7,7 +7,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/smart-install.js\" && bun \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" stop && bun \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/smart-install.js\" && bun \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" stop ; bun \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start",
             "timeout": 300
           },
           {


xingyu42:
I have a question: spawn was removed, but I didn't see any implementation to replace the original child process. It seems like the process is killed after the hook ends?

thedotmack:
Already addressed in v9.0.6 — Windows console popups eliminated with WMIC-based detached process spawning. Closing as resolved.

PR #706, opened 2026-01-14
Cleanup worker-service.ts: remove dead code and fallback concept

thedotmack:
## Summary

- Delete ~216 lines of dead code (`runInteractiveSetup` function + unused imports)
- Remove fallback agent concept entirely - users who choose Gemini/OpenRouter get those providers, not secret fallback behavior
- Remove misleading re-export indirection (ResponseProcessor now imports directly from CursorHooksInstaller)
- Update MCP client/server to use build-time injected version instead of hardcoded "1.0.0"

## Test plan

- [x] All 761 tests pass
- [x] Build succeeds
- [x] No hardcoded versions remain
- [x] Dead code verified removed via grep


thedotmack:
Thanks for the thorough review!

## Responses to Questions

### 1. Test Coverage for Deleted Code
The 344 lines in `cursor-hook-outputs.test.ts` were tests for the now-deleted `runInteractiveSetup` function and its related cursor hook integration code (`detectClaudeCode`, `findCursorHooksDir`, etc.). Since that code was removed as dead code (never called), the tests became obsolete.

### 2. Fallback Agent Logic
The fallback agent removal IS in this PR - commit `7c3b5c82` ("fix(worker): remove fallback agent concept entirely"). It removed:
- `setFallbackAgent()` calls from worker-service.ts
- The `isConfigured()` method from SDKAgent.ts (became dead code after removing the fallback calls)

The key issue was that the fallback architecture caused users who selected Gemini/OpenRouter to silently fall back to Claude SDK on errors - unexpected behavior. Additionally, the `ANTHROPIC_API_KEY` check in `isConfigured()` was a security concern (personal API keys cost ~20x more than Claude Code's bundled pricing).

### 3. Documentation Timestamps
Good suggestion - will consider for future reports.

CI should be green. Ready to merge when you are.

PR #712, opened 2026-01-15
Fix environment variables

cjpeterein:
fix: respect environment variables with correct priority

    Added applyEnvOverrides() method to SettingsDefaultsManager that ensures
    environment variables take highest priority over file and default settings.
    This enables runtime configuration overrides without modifying the settings
    file, important for containerized deployments and CI/CD scenarios.

    Configuration priority is now: env > file > defaults


thedotmack:
Cherry-picked the source changes onto main. The env override logic (applyEnvOverrides method + 3 call sites in loadFromFile) and all 6 test cases have been applied. Build artifacts were regenerated locally. Priority order is now: env > file > defaults. Thank you for the contribution!

PR #735, opened 2026-01-17
fix: strip ANTHROPIC_API_KEY for Claude Code subscribers

### LINKED ISSUE #733: SDK uses ANTHROPIC_API_KEY from .env* files instead of Claude Code CLI
## Reproduction Steps

1. Have `ANTHROPIC_API_KEY` set in a `.env*` file (e.g., `.env.local`) for use with other tools
2. Install claude-mem plugin
3. Use Claude Code with the default settings (`CLAUDE_MEM_PROVIDER: 'claude'`)
4. Perform actions that trigger observation compressions or session summaries
5. Bug occurs: API credits are consumed from your Anthropic account instead of using Claude Code's billing

## Expected Result

When `pathToClaudeCodeExecutable` is provided to the SDK's `query()` function, observation compressions and summaries should route through Claude Code CLI, which uses the user's Claude Code subscription billing.

## Actual Result

The `@anthropic-ai/claude-agent-sdk` auto-discovers `ANTHROPIC_API_KEY` from `.env*` files or the environment and uses it directly, bypassing Claude Code's billing. Users with Claude Max subscriptions receive unexpected "Auto-recharge credits" invoices from Anthropic.

## Error Message

No error - the plugin functions correctly but bills to the wrong account.

---

## More Info

The issue might be in `src/services/worker/SDKAgent.ts`. The `query()` call around line 102 passes `pathToClaudeCodeExecutable: claudePath`, which should route requests through Claude Code. However, when `ANTHROPIC_API_KEY` exists in `process.env` (loaded from `.env*` files), the SDK may prefer using it directly.

Since the SDK runs in-process, it has direct access to environment variables. One potential workaround could be to temporarily unset `ANTHROPIC_API_KEY` before calling `query()`:

```typescript
const savedApiKey = process.env.ANTHROPIC_API_KEY;
delete process.env.ANTHROPIC_API_KEY;

const queryResult = query({
  prompt: messageGenerator,
  options: {
    // ...
  }
});

// Restore after query completes
process.env.ANTHROPIC_API_KEY = savedApiKey;
```

This issue appears related to the investigation in `docs/reports/issue-588-api-key-usage-warning.md`. The v9.0.5 changelog mentions "Removed dangerous ANTHROPIC_API_KEY check" but this may have only removed a warning rather than addressing the underlying auto-discovery behavior.

Another option might be changing the default `CLAUDE_MEM_PROVIDER` from `'claude'` to `'gemini'` as suggested in the issue-588 report.

shyal:
## Summary

Fixes #733

When users have `ANTHROPIC_API_KEY` set in their environment (e.g., for other tools), the Claude Agent SDK auto-discovers and uses it directly, bypassing Claude Code CLI billing. This causes unexpected charges to users' Anthropic API accounts instead of their Claude Code subscriptions.

**Changes:**
- Added `src/shared/claude-subscription.ts` - detects paid subscriptions (max/pro) by reading `~/.claude/.credentials.json`
- Modified `SDKAgent.ts` - temporarily strips `ANTHROPIC_API_KEY` for subscribers before calling the SDK, restores it in a finally block

**Behavior:**
- **Subscribers (max/pro)**: API key is stripped → SDK routes through CLI → subscription billing
- **Non-subscribers**: API key is preserved → SDK uses direct API access → API billing

## Test plan

- [x] Verified subscription detection reads credentials correctly
- [x] Tested with `ANTHROPIC_API_KEY` exported in environment
- [x] Confirmed log message appears: "Claude subscription detected - routing through CLI billing"
- [x] Verified SDK queries complete successfully through CLI


vrsttl:
I'm on holiday and I decided to finally start playing around with this plugin. I ended up coding up the solution for the same issue not by finding the way to make my subscription work but by fixing the otherwise broken openrouter session id generation and use openrouter instead of claude for memory retrieval. I wasted a day on the beach to get this to work. This PR needs to happen ASAP in my view.

thedotmack:
Already addressed in v9.0.15 — 'Isolated Credentials (#745)' now sources credentials exclusively from ~/.claude-mem/.env with whitelisted env vars. Thank you\!

PR #745, opened 2026-01-18
fix: use centralized credentials from ~/.claude-mem/.env to prevent API key hijacking

### LINKED ISSUE #733: SDK uses ANTHROPIC_API_KEY from .env* files instead of Claude Code CLI

bigph00t:
## Summary

Fixes #733 - SDK uses `ANTHROPIC_API_KEY` from random project `.env` files instead of Claude Code CLI subscription billing.

**Root cause:** The SDK's `query()` function inherits from `process.env` when no `env` option is passed. When users work in projects with their own `.env` files containing API keys, the SDK discovers and uses those keys, billing the wrong account.

**Solution:** Centralized credential management via `~/.claude-mem/.env`

## Changes

- **New: `src/shared/EnvManager.ts`** - Centralized credential storage and isolated env builder
- **SDKAgent:** Pass isolated env to SDK `query()` that only includes credentials from `~/.claude-mem/.env`, not random keys from `process.env` inheritance
- **GeminiAgent/OpenRouterAgent:** Use `getCredential()` instead of `process.env` fallback for consistency
- **SettingsDefaultsManager:** Add `CLAUDE_MEM_CLAUDE_AUTH_METHOD` setting ('cli' | 'api')

## How it works

1. `buildIsolatedEnv()` creates a clean environment with only essential system vars (PATH, HOME, etc.) and credentials explicitly configured in `~/.claude-mem/.env`
2. SDK subprocess runs with this isolated env, never seeing random API keys from parent process
3. If no `ANTHROPIC_API_KEY` is in `~/.claude-mem/.env`, Claude Code CLI billing is used (default)
4. Same pattern applied to Gemini/OpenRouter agents for consistency

## Test plan

- [x] Build compiles successfully
- [x] Worker starts with fixed code
- [x] Auth method shows "Claude Code CLI (subscription billing)" when no API key in `.env`
- [x] Fake `ANTHROPIC_API_KEY` in parent environment is correctly ignored
- [x] Observations are captured correctly
- [x] Chroma sync works

## Evidence

Logs show correct auth method even when `ANTHROPIC_API_KEY` exists in parent environment:
```
authMethod=Claude Code CLI (subscription billing)
```


bigph00t:
@greptile

bayanoj330-dev (approved):

Issue #727, opened 2026-01-16
Hooks not firing except PostToolUse (which hangs indefinitely)


lylerolleman:
# Hooks not firing except PostToolUse (which hangs indefinitely)

## Problem Statement

After a fresh installation of claude-mem, only the PostToolUse hook appears to fire when using Claude Code. However, this hook does not appear to complete properly and displays "Running PostToolUse hooks… (1/2 done)" indefinitely. Other hooks such as StartSession do not fire at all, as confirmed by log file analysis.

## Environment

- **Claude-mem Version**: 9.0.5
- **Claude Code Version**: 2.1.9
- **Node.js**: v22.17.1
- **Bun**: 1.3.5
- **Operating System**: Linux 4.4.0-19041-Microsoft (x64)
- **Platform**: linux
- **Dev Mode**: No

### Paths

- **Plugin Directory**: `~/.claude/plugins/marketplaces/thedotmack`
- **Data Directory**: `~/.claude-mem`
- **Current Directory**: `~/.claude/plugins/marketplaces/thedotmack`

### Worker Status

- **Running**: Yes
- **PID**: 1180
- **Port**: 37777
- **Uptime**: 45 minutes
- **Active Sessions**: 1
- **SSE Clients**: 0

### Database

- **Path**: `~/.claude-mem/claude-mem.db`
- **Exists**: Yes
- **Size**: 4.00 KB

### Configuration

- **Settings File**: `~/.claude-mem/settings.json`
- **Key Settings**:
  - `CLAUDE_MEM_MODEL`: claude-sonnet-4-5
  - `CLAUDE_MEM_WORKER_PORT`: 37777
  - `CLAUDE_MEM_WORKER_HOST`: 127.0.0.1
  - `CLAUDE_MEM_LOG_LEVEL`: INFO
  - `CLAUDE_MEM_CONTEXT_OBSERVATIONS`: 50

## Steps to Reproduce

1. Perform a fresh installation of claude-mem
2. Use Claude Code normally by writing prompts
3. Observe that only PostToolUse hook fires
4. Verify missing hooks by running: `grep "StartSession" <log file>` (returns no results)
5. Notice that PostToolUse hook hangs at "Running PostToolUse hooks… (1/2 done)"

## Expected Behavior

- All hooks should fire properly during their respective lifecycle events
- StartSession hook should fire when a session starts
- PostToolUse hook should complete successfully without hanging
[...]

Issue #646, opened 2026-01-09
Plugin bricks Claude Code - stdin fstat EINVAL crash


MaxWolf-01:
# Bug Report: SessionStart hook BRICKS Claude Code - cannot start sessions

## Bug Description

**Severity: CRITICAL** - This bug bricks Claude Code in most directories. After installing claude-mem, Claude Code fails to start new sessions in many/most directories (works in some, like `/tmp`, but fails in others like home directories or project folders). Users often cannot use Claude Code until the plugin is manually uninstalled by editing config files.

The SessionStart hook (`context-hook.js`) crashes immediately on startup with `EINVAL: Invalid argument` when bun attempts to `fstat()` stdin. This prevents Claude Code from starting in any directory where the plugin is active.

The crash occurs in bun's lazy stdin loading (`lazyLoadStream`) before the hook code can even check `stdin.isTTY`. This appears to be caused by Claude Code not providing a valid stdin file descriptor to hooks.

## Steps to Reproduce

1. Install claude-mem: `/plugin marketplace add thedotmack/claude-mem` then `/plugin install claude-mem`
2. Start a new Claude Code session in any directory
3. Observe the SessionStart hook error preventing session startup

## Expected Behavior

The SessionStart hook should gracefully handle cases where stdin is unavailable/invalid and fall back to running without stdin input (similar to the `--colors` flag behavior).

## Error Output

```
SessionStart:startup hook error: Failed with non-blocking status code: 14 |
`,E+=`2. Run: npm run worker:restart
15 | `,E+="3. Restart Claude Code",r&&(E+=`
16 |
17 | If that doesn't work, try: /troubleshoot`),n&&(E=`Worker Error: ${n}
18 |
19 | ${E}`),E}var Q=C.join(z(),".claude","plugins","marketplaces","thedotmack"),It
...
EINVAL: Invalid argument
 syscall: "fstat"
    errno: -22
       at lazyLoadStream (:1:21)
       at getReader (:1:21)
       at ref (:4:29)
       at node:stream:1743:323
       at /home/max/.claude/plugins/cache/thedotmack/claude-mem/9.0.1/scripts/context-hook.js:19:2871
```

## Root Cause Analysis

Looking at the minified `context-hook.js`, the relevant code is:

```javascript
var Et=process.argv.includes("--colors");
if(A.isTTY||Et)
  w(void 0).then(o=>{console.log(o),process.exit(0)});
else{
  let o="";
  A.on("data",t=>o+=t),
[...]

PR #771, opened 2026-01-22
fix: handle stdin unavailability and timeout to prevent hook hangs

rajivsinclair:
## Summary

This PR fixes the critical issue where PostToolUse hooks hang indefinitely at "Running PostToolUse hooks… (1/2 done)", making claude-mem unusable.

**Fixes:** #727, Addresses #646

## Root Cause Analysis

Two distinct issues cause hooks to hang or crash:

1. **Bun stdin crash (#646)**: When Claude Code spawns a hook process, it may not provide a valid stdin file descriptor. Bun's lazy initialization calls `fstat()` on stdin, which fails with `EINVAL`, crashing the process before any user code runs.

2. **Missing EOF event (#727)**: Even when stdin is valid, Claude Code may not properly close the pipe after writing data. The previous implementation waited indefinitely for `stdin.on('end')`, which never fires in this scenario.

### `src/cli/stdin-reader.ts`

- **Added `isStdinAvailable()` function**: Safely probes stdin by wrapping access in try-catch, catching Bun's `EINVAL` crash before it propagates
- **Added 5-second timeout**: Uses `setTimeout` + `Promise` pattern to resolve with available data if `stdin.on('end')` doesn't fire
- **Graceful error handling**: Returns `undefined` instead of crashing when stdin is unavailable or errors occur
- **Proper cleanup**: Removes event listeners after resolution to prevent memory leaks

## Testing

Verified the following scenarios work correctly:

| Scenario | Before | After |
|----------|--------|-------|
| Normal stdin input | ✅ Works | ✅ Works |
| Empty stdin (`< /dev/null`) | ❌ Hangs | ✅ Returns validation error |
| Stdin pipe stays open | ❌ Hangs forever | ✅ Timeout after 5s |
| Invalid stdin fd (Bun crash) | ❌ EINVAL crash | ✅ Graceful fallback |

## Impact

- **No breaking changes**: Hooks that work today continue working
- **Defensive improvement**: Adds multiple layers of protection against stdin edge cases
- **Backwards compatible**: Falls back gracefully, allowing handlers to validate input as needed

### DIFF
--- a/src/cli/stdin-reader.ts
+++ b/src/cli/stdin-reader.ts
@@ -1,12 +1,11 @@
-// Stdin reading utility extracted from hook patterns
-// See src/hooks/save-hook.ts for the original pattern
-
-// Inactivity timeout for stdin reading - if no data arrives within this time,
-// we parse whatever data we have. This fixes issue #727 where hooks hang at "1/2 done"
-// because stdin.on('end') never fires.
-// Using inactivity timeout (reset on each data chunk) instead of absolute timeout
-// to avoid truncating large/slow payloads.
-const STDIN_INACTIVITY_TIMEOUT_MS = 5000;
+// Stdin reading utility for Claude Code hooks
+//
+// Problem: Claude Code doesn't close stdin after writing hook input,
+// so stdin.on('end') never fires and hooks hang indefinitely (#727).
+//
+// Solution: JSON is self-delimiting. We detect complete JSON by attempting
+// to parse after each chunk. Once we have valid JSON, we resolve immediately
+// without waiting for EOF. This is the proper fix, not a timeout workaround.
 
 /**
  * Check if stdin is available and readable.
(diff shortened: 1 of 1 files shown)


rajivsinclair:
## Updated Approach (v3)

After further investigation, I've improved the fix to address the **actual root cause** rather than using a timeout workaround.

### Root Cause
Claude Code doesn't close stdin after writing hook input, so `stdin.on('end')` never fires.

### Solution
JSON is self-delimiting. Instead of waiting for EOF, we now:
1. Attempt to parse after each `data` chunk
2. Resolve immediately when JSON is valid
3. No timeout needed for normal operation

### Performance Improvement
| Metric | Timeout Approach | JSON Detection |
|--------|------------------|----------------|
| Normal execution | 5000ms+ | **448ms** |
| Stdin stays open | Hung forever | **Exits immediately** |

The hook now exits in <500ms instead of waiting for any timeout. This is the proper fix.

PR #773, opened 2026-01-22
fix: use /api/health instead of /api/readiness in hook health checks

### LINKED ISSUE #772: Worker health check uses /api/readiness causing 15-second timeout during background initialization
## Bug Description

The worker service's `ensureWorkerRunning()` function times out after 15 seconds because it checks `/api/readiness` instead of `/api/health`. The readiness endpoint returns 503 until full background initialization completes, which can take much longer than 15 seconds (especially the MCP connection which has a 5-minute timeout).

## Symptoms

- Error message: `Worker did not become ready within 15 seconds. (port 37777)`
- `UserPromptSubmit operation blocked by hook...`
- Happens when worker needs to start fresh (first run, after crash, after restart)

## Root Cause

In `src/shared/worker-utils.ts:isWorkerHealthy()`:
- Currently checks `/api/readiness` which requires `initializationCompleteFlag = true`
- Background initialization includes MCP client connection (5-minute timeout!)
- Hooks only have 15 seconds to wait

The worker IS running and the HTTP server IS accepting requests - but the hooks incorrectly wait for full initialization.

## Expected Behavior

Hooks should check `/api/health` (which always responds when the HTTP server is up) instead of `/api/readiness`. Individual endpoints can return 503 if they need full initialization.

## Reproduction

1. Kill the worker process
3. The hook tries to start the worker but times out waiting for readiness

## Fix

Change `isWorkerHealthy()` to check `/api/health` instead of `/api/readiness`.

rajivsinclair:
## Summary

- Fixes the 15-second timeout issue where hooks fail with "Worker did not become ready within 15 seconds"
- Changes `isWorkerHealthy()` to check `/api/health` instead of `/api/readiness`

Fixes #772

## Problem

The worker service's `ensureWorkerRunning()` function was timing out after 15 seconds because it checked `/api/readiness` instead of `/api/health`.

**Root Cause:**
- `/api/readiness` returns 503 until `initializationCompleteFlag = true`
- Background initialization includes MCP client connection (with a **5-minute timeout!**)
- Hooks only wait 15 seconds before failing
- The worker IS running and accepting HTTP requests - just not "fully initialized"

**Symptoms:**
- `Worker did not become ready within 15 seconds. (port 37777)`
- `UserPromptSubmit operation blocked by hook...`
- Happens on fresh starts, after crashes, or after worker restarts

## Solution

Changed `isWorkerHealthy()` in `src/shared/worker-utils.ts` to check `/api/health` instead of `/api/readiness`:

- `/api/health` always responds 200 when the HTTP server is listening
- Individual endpoints can still return 503 if they need full initialization
- Hooks just need to know the server is up, not that all services are ready

## Test plan

- [x] Build succeeds (`npm run build`)
- [ ] Start fresh session after killing worker - should not timeout
- [ ] Verify hooks work normally after fix


Glucksberg:
Duplicate of #774. Closing.

thedotmack:
Already addressed in v9.0.16 (same fix as PR #820 which was merged). Health checks now use /api/health. Thank you!

PR #774, opened 2026-01-22
fix: use /api/health instead of /api/readiness in hook health checks

### LINKED ISSUE #729: Worker startup blocks Claude Code entirely when not ready within 15 seconds
## Problem

When the claude-mem worker isn't ready within 15 seconds, the `UserPromptSubmit` operation is completely blocked, preventing Claude Code from working at all.

Error message:
```
UserPromptSubmit operation blocked by hook: [bun "${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" hook claude-code session-init]: Hook error: Error: Worker did not become ready within 15 seconds. (port 37777)
To restart the worker:
1. Exit Claude Code completely
2. Run: npm run worker:restart
3. Restart Claude Code
```

## Issues

1. **Blocking behavior is too aggressive** - The plugin completely blocks Claude Code from functioning instead of failing gracefully and letting the session continue without memory features.

2. **Missing restart script** - The error message suggests running `npm run worker:restart`, but this script doesn't exist:
   ```
   npm error Missing script: "worker:restart"
   ```

3. **Worker gets stuck** - The worker process can be running (listening on port 37777) but in a bad state where it doesn't respond to readiness checks.

## Expected Behavior

- If the worker isn't ready, the plugin should log a warning and allow Claude Code to function normally (just without memory features)
- The `worker:restart` script mentioned in the error should actually exist
- Consider adding a health check endpoint or auto-recovery mechanism

## Environment

- Plugin version: 9.0.2
- macOS
- Using bun runtime

## Workaround

Kill the stuck worker and restart manually:
```bash
kill $(lsof -t -i :37777)
nohup bun ~/.claude/plugins/cache/thedotmack/claude-mem/9.0.2/scripts/worker-service.cjs &
```

### LINKED ISSUE #772: Worker health check uses /api/readiness causing 15-second timeout during background initialization

rajivsinclair:
## Summary

Fixes the 15-second timeout issue where Claude Code hooks block with "Worker did not become ready within 15 seconds" during session initialization.

**Closes #729, Closes #772**

## Root Cause

The health check functions (`isWorkerHealthy()` in `worker-utils.ts` and `waitForHealth()` in `HealthMonitor.ts`) were using `/api/readiness` instead of `/api/health`. 

The `/api/readiness` endpoint returns 503 until the worker completes **full initialization**, which includes:
- Database initialization
- Search manager setup
- **MCP client connection (has a 5-minute timeout!)**

This meant hooks would wait up to 15 seconds for initialization that could take much longer, then fail.

## The Fix

Changed both health check functions to use `/api/health` instead of `/api/readiness`:

| Endpoint | Behavior | Use Case |
|----------|----------|----------|
| `/api/health` | Always returns 200 when HTTP server is listening | Liveness probe - "is the process alive?" |
| `/api/readiness` | Returns 503 until full initialization completes | Traffic routing - "is it ready for full workload?" |

Hooks only need to know the worker HTTP server is accepting connections. Individual API endpoints can return their own 503 errors if they require full initialization.

## Changes

- `src/shared/worker-utils.ts`: `isWorkerHealthy()` now checks `/api/health`
- `src/services/infrastructure/HealthMonitor.ts`: `waitForHealth()` now checks `/api/health`
- `plugin/scripts/worker-service.cjs`: Rebuilt with the fix
- `tests/shared/worker-utils-health.test.ts`: New tests verifying the behavior

## Verification

Tested locally by:
1. Shutting down the worker
2. Starting a new Claude Code session
3. Confirming the worker starts and hooks pass immediately
4. Verified `/api/health` returns 200 even when `initialized: false`

```bash
$ curl -s http://localhost:37777/api/health
{"status":"ok","initialized":false,"mcpReady":false,...}  # Returns 200!

$ curl -s http://localhost:37777/api/readiness  
{"status":"initializing","message":"Worker is still initializing"}  # Returns 503
```

## Test Plan

- [x] Verified health endpoint responds immediately during worker startup
- [x] Verified readiness endpoint returns 503 during initialization (expected)
- [x] Added unit tests for health check behavior
- [x] Tested full session initialization flow

### DIFF
--- a/tests/shared/worker-utils-health.test.ts
+++ b/tests/shared/worker-utils-health.test.ts
@@ -1,0 +1,345 @@
+/**
+ * Worker Utils Health Tests
+ *
+ * Tests for isWorkerHealthy() and ensureWorkerRunning() functions.
+ *
+ * Key behaviors tested:
+ * 1. isWorkerHealthy() uses /api/health endpoint (not /api/readiness)
+ * 2. ensureWorkerRunning() succeeds when health endpoint responds
+ * 3. Health endpoint failures cause retry/timeout behavior
+ *
+ * These tests use fetch mocking to simulate various server states.
+ */
+
+import { describe, it, expect, beforeEach, afterEach, mock, spyOn } from 'bun:test';
+
+// We need to mock the dependencies before importing worker-utils
+// Mock the SettingsDefaultsManager to avoid file system operations
+const mockSettingsDefaultsManager = {
+  get: mock((key: string) => {
+    if (key === 'CLAUDE_MEM_DATA_DIR') return '/tmp/test-claude-mem';
+    return '';
+  }),
+  loadFromFile: mock(() => ({
+    CLAUDE_MEM_WORKER_PORT: '37777',
+    CLAUDE_MEM_WORKER_HOST: '127.0.0.1',
+  })),
+};
+
+// Mock the logger to avoid side effects
+const mockLogger = {
+  debug: mock(() => {}),
+  info: mock(() => {}),
+  warn: mock(() => {}),
+  error: mock(() => {}),
+};
+
+// Mock the error-messages module
+const mockGetWorkerRestartInstructions = mock((opts: { port: number; customPrefix?: string }) => {
+  return `Worker restart instructions for port ${opts.port}. ${opts.customPrefix || ''}`;
+});
+
+// Store original modules for restoration
+const originalFetch = global.fetch;
+
+describe('worker-utils health functions', () => {
+  // Track fetch calls for verification
+  let fetchCalls: { url: string; options?: RequestInit }[] = [];
+
+  beforeEach(() => {
+    fetchCalls = [];
+    // Reset all mocks
+    mockLogger.debug.mockClear();
+    mockLogger.info.mockClear();
+    mockLogger.warn.mockClear();
+    mockLogger.error.mockClear();
+  });
+
+  afterEach(() => {
+    global.fetch = originalFetch;
+  });
+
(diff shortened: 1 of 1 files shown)


Glucksberg:
Conflicting and superseded by main changes. Health check endpoints have been updated. Closing.

thedotmack:
Already addressed in v9.0.16 (same fix as PR #820 which was merged). Health checks now use /api/health. Thank you!

PR #821, opened 2026-01-27
fix: prevent stale memory_session_id resume crash after worker restart

### LINKED ISSUE #817: [Bug] Stale session fix incomplete: lastPromptNumber loaded from DB bypasses resume guard
## Bug Description

The fix for stale `memory_session_id` resume crashes (documented in `.claude/plans/fix-stale-session-resume-crash.md`) is incomplete. The fix checks:

```typescript
...(hasRealMemorySessionId && session.lastPromptNumber > 1 && { resume: session.memorySessionId }),
```

This assumes `lastPromptNumber === 1` on worker restart. However, `lastPromptNumber` is loaded from the database via `getPromptNumberFromUserPrompts()` in `SessionManager.ts:137`:

```typescript
lastPromptNumber: promptNumber || this.dbManager.getSessionStore().getPromptNumberFromUserPrompts(dbSession.content_session_id),
```

## Failure Scenario

1. Session has multiple prompts (lastPromptNumber = 5)
2. SDK captures `memory_session_id` = `abc-123`
3. Worker crashes or restarts
4. Claude Code cleans up the SDK session (abc-123 no longer exists)
5. Worker restarts, session reinitializes:
   - `lastPromptNumber = 5` (loaded from DB)
   - `memory_session_id = abc-123` (still in DB)
6. SDKAgent checks: `hasRealMemorySessionId && lastPromptNumber > 1`
7. Both TRUE → passes `resume: "abc-123"` to SDK
8. Claude Code: "No conversation found with session ID: abc-123"
9. **CRASH with code 1**

## Evidence

From logs before manual fix:
```
[17:30:21.773] Starting SDK query {
  hasRealMemorySessionId=true,
  resume_parameter=bdcbddfd-197f-46b2-9383-e5fcc18f1a98,
  lastPromptNumber=1  ← Only works if this is 1!
}
[17:30:24.450] Generator failed {error=Claude Code process exited with code 1}
```

Testing the stale session directly:
```bash
$ echo "test" | claude --print --sdk --session-id bdcbddfd-197f-46b2-9383-e5fcc18f1a98
Error: No conversation found with session ID: bdcbddfd-197f-46b2-9383-e5fcc18f1a98
```

## Workaround

Manually clear the stale session ID:
```sql
UPDATE sdk_sessions SET memory_session_id = NULL WHERE id = <session_id>;
```
Then kill/restart the worker.

## Suggested Fix Options

1. **Clear `memory_session_id` on worker startup** - since SDK context is lost anyway
2. **Track worker restart epoch** - compare against session last_activity to detect staleness
3. **Validate session before resume** - try a lightweight probe before full resume
4. **Add `sdk_session_started_at` timestamp** - clear `memory_session_id` if older than worker startup

## Environment

- **claude-mem version:** 9.0.9
- **OS:** Linux
- **Provider:** Claude SDK

## Related

- Plan file: `.claude/plans/fix-stale-session-resume-crash.md`
- Changelog entry: "Fixed stale session resume crash when SDK session is orphaned"

bigph00t:
## Summary

Fixes the stale session resume crash (Issue #817) where the worker attempts to resume an SDK session that no longer exists after a worker restart.

## Root Cause

When the worker restarts:
1. In-memory session state is lost (including SDK context)
2. Database still contains `memory_session_id` from the previous worker instance
3. `lastPromptNumber` is loaded from the database (e.g., 5, not 1)
4. The existing guard `lastPromptNumber > 1` passes
5. SDK attempts to resume with stale session ID
6. **CRASH**: "No conversation found with session ID: ..."

## Solution

When creating a new in-memory session from the database, **always set `memorySessionId: null`** instead of loading it from the database.

**Key insight**: If a session is not in the in-memory cache, we're either:
- In a new worker instance (SDK context lost)
- First time seeing this session (no SDK context yet)

In both cases, we should NOT attempt to resume. The SDK agent will capture a fresh `memorySessionId` on the first response.

## Changes

| File | Change |
|------|--------|
| `src/services/worker/SessionManager.ts` | Clear `memorySessionId` when loading from DB |

### Code Change

```diff
- memorySessionId: dbSession.memory_session_id || null,
+ memorySessionId: null,  // Always start fresh - SDK will capture new ID
```

## Why Not PR #802?

PR #802 attempts a more complex fix with:
- AbortController reset logic
- Exit code detection
- Transaction handling for FK constraints

This PR takes a simpler approach: if we're loading from DB, the session is definitely stale. No need for complex detection - just start fresh.

## Test Plan

- [x] Build passes
- [x] All existing tests pass (56 pre-existing failures unrelated to this change)
- [ ] Manual test: Kill worker during active session, restart, verify no crash
- [ ] Manual test: Verify sessions continue to work normally without restart

## Verification

After this fix, logs will show:
```
[WARN] Discarding stale memory_session_id from previous worker instance (Issue #817)
```

Fixes #817


bigph00t:
## Connection to Issue #825

This PR also fixes **Issue #825** where users report seeing only prompts in the dashboard but no observations.

### Symptoms

When this bug triggers:
- Dashboard shows prompts but **no observations**
- Worker logs show `queueDepth` climbing indefinitely (700+ items observed during testing)
- SDK generator fails with: `Claude Code process exited with code 1`

### What Happens

The stale `memory_session_id` causes the SDK agent to crash with "No conversation found". This cascades:
1. Every observation queued for that session fails
2. Queue depth grows infinitely  
3. No observations are generated
4. Dashboard shows prompts but no observations (Issue #825)

### Tested Fix

After applying this fix:
- Killed worker with 700+ queued items
- Restarted worker
- Queue cleared to 0
- Observations now generate correctly

cc @thedotmack

bigph00t:
Closing to create a cleaner PR without bundled file artifacts.

PR #829, opened 2026-01-27
fix: gracefully handle empty prompts in session-init hook

rajivsinclair:
## Summary

- Fix UserPromptSubmit hook blocking when user submits empty/whitespace-only prompt
- Return early with `{ continue: true }` instead of throwing error

## Problem

When a user submits an empty prompt (or whitespace-only) to Claude Code, the session-init handler throws:

```
UserPromptSubmit operation blocked by hook:
Hook error: Error: sessionInitHandler requires prompt
```

This blocks the entire operation instead of gracefully continuing.

## Solution

Changed the prompt validation from throwing an error to returning early:

```typescript
// Before (blocking)
if (!prompt) {
  throw new Error('sessionInitHandler requires prompt');
}

// After (graceful)
if (!prompt || !prompt.trim()) {
  return { continue: true, suppressOutput: true };
}
```

## Test plan

- [ ] Submit empty prompt in Claude Code - should not throw error
- [ ] Submit whitespace-only prompt - should not throw error
- [ ] Submit normal prompt - should initialize session as before

### DIFF
--- a/src/cli/handlers/session-init.ts
+++ b/src/cli/handlers/session-init.ts
@@ -16,8 +16,9 @@ export const sessionInitHandler: EventHandler = {
 
     const { sessionId, cwd, prompt } = input;
 
-    if (!prompt) {
-      throw new Error('sessionInitHandler requires prompt');
+    // Gracefully handle empty prompts (user submitted blank input)
+    if (!prompt || !prompt.trim()) {
+      return { continue: true, suppressOutput: true };
     }
 
     const project = getProjectName(cwd);


thedotmack:
Closing as redundant. The empty prompt handling fix was already merged as part of PR #828 (commit 9789a196). The current main branch already contains the graceful handling at `src/cli/handlers/session-init.ts` lines 24-27:

```typescript
// Gracefully handle empty prompts (user submitted blank input)
if (!prompt || !prompt.trim()) {
  return { continue: true, suppressOutput: true };
}
```

Thank you for the contribution — the fix is live, just via a different PR.

PR #847, opened 2026-01-28
fix: comprehensive observer session isolation without breaking auth

bigph00t:
## Summary

Combines fixes from PRs #837, #839, #845, and #745 into one comprehensive solution that addresses observer session isolation, environment isolation, and authentication — without regressions.

**Before this PR:** PR #837 isolated observer sessions by overriding `CLAUDE_CONFIG_DIR`, which broke authentication (`Invalid API key · Please run /login`). 1,958 consecutive auth failures observed in production logs. No observations were being processed.

**After this PR:** Zero auth errors. Observations storing successfully. Observer sessions isolated from `claude --resume`.

## Root Cause Analysis

| Issue | Root Cause | Fix |
|-------|-----------|-----|
| #832 (session pollution) | Observer sessions stored in `~/.claude/projects/` | Use `cwd` option → sessions under "observer-sessions" project |
| #825 (no observations) | `CLAUDE_CONFIG_DIR` override removed access to `~/.claude/.credentials.json` | Don't override `CLAUDE_CONFIG_DIR` — use `cwd` instead |
| #733 (API key hijacking) | SDK inherits `ANTHROPIC_API_KEY` from `process.env` | `EnvManager.buildIsolatedEnv()` — clean env with explicit credentials only |
| #817 (stale resume crash) | Already fixed on main via PR #839 | `memorySessionId: null` on DB load |

## Changes

1. **`ProcessRegistry.ts`** — Removed `CLAUDE_CONFIG_DIR` override from spawn. Auth credentials in `~/.claude/` are now accessible.
2. **`SDKAgent.ts`** — Added `cwd: OBSERVER_SESSIONS_DIR` (session isolation) + `env: isolatedEnv` (credential isolation) to SDK `query()` options.
3. **`paths.ts`** — Renamed `OBSERVER_CONFIG_DIR` → `OBSERVER_SESSIONS_DIR`.
4. **`EnvManager.ts`** (new) — Centralized credential management. Builds clean subprocess env with only essential system vars + credentials from `~/.claude-mem/.env`.
5. **`SettingsDefaultsManager.ts`** — Added `CLAUDE_MEM_CLAUDE_AUTH_METHOD` setting (default: `'cli'`).

## Verification

Post-deploy logs confirm:
- `authMethod=Claude Code CLI (subscription billing)` ✅
- Zero `Invalid API key` errors ✅
- Observations storing with real IDs (e.g., obsIds=[3914, 3915]) ✅
- Worker healthy and initialized ✅

## Test plan

[...]

PR #844, opened 2026-01-28
fix: add session-complete handler and hook to enable orphan reaper cleanup

thusdigital:
## Summary

- Adds `session-complete` event handler with full backend implementation
- Adds `POST /api/sessions/complete` API endpoint
- Registers handler in the handler index alongside existing handlers
- Adds `session-complete` hook to Stop hooks in `hooks.json` (fires after summarize)
- Enables the orphan reaper (PR #806) to correctly identify and clean up orphaned worker processes

## Problem

The orphan reaper from PR #806 never cleans up worker processes because sessions are never removed from the active session map. Without the `session-complete` handler:

1. Session exits → summarize hook fires → session stays in active map
2. Orphan reaper runs every 5 minutes
3. Reaper checks if PIDs belong to active sessions → all PIDs appear "active"
4. No cleanup occurs → processes accumulate indefinitely

**Observed impact:** 39+ zombie processes consuming ~10GB on macOS after a single worker restart, because `processPendingQueues(50)` in startup-recovery spawns subprocesses for all sessions with pending messages, and they never get cleaned up.

## Previous version (hooks-only)

The original PR only added the hook trigger. Greptile correctly flagged that the `session-complete` handler didn't exist server-side. This update adds the complete implementation:

### New files
- `src/cli/handlers/session-complete.ts` — Handler implementation

### Modified files
- `src/cli/handlers/index.ts` — Register `session-complete` in handler map and EventType union
- `src/services/worker-service.ts` — Add `session-complete` to CLI help text
- `src/services/worker/http/routes/SessionRoutes.ts` — `POST /api/sessions/complete` endpoint
- `plugin/hooks/hooks.json` — Hook trigger in Stop hooks
- `plugin/scripts/worker-service.cjs` — Rebuilt bundle
- `plugin/scripts/mcp-server.cjs` — Rebuilt bundle

## How it works

1. Stop hooks fire: `summarize` (phase 1) → `session-complete` (phase 2)
2. `session-complete` handler calls `POST /api/sessions/complete` with the `contentSessionId`
3. Route looks up `sessionDbId`, calls `completionHandler.completeByDbId()` to remove from active map
4. Orphan reaper (every 5 min) can now correctly identify orphaned PIDs and clean them up

## Validation

Tested locally with logging enabled (Jan 28 - Feb 2):

```
[17:11:50] → session-complete: Removing session from active map
[17:11:50] Session completed via API {sessionDbId=236}
[17:11:50] Session completed successfully
```

[...]

PR #830, opened 2026-01-27
fix: Chroma stability + additional process management layers

michelhelsdingen:
## Summary

This PR adds Chroma stability and process management improvements that complement the existing PID registry fix from PR #806.

**Note:** Issue #737 (zombie subprocess accumulation) was fixed by PR #806 via PID registry. This PR focuses on **additional stability layers** not covered by that fix.

## Changes

| Change | What it does |
|--------|--------------|
| **CLAUDE_MEM_CHROMA_DISABLED** | New setting to completely disable Chroma vector search (biggest win for chroma-mcp crash issues) |
| **ChromaSync timeouts** | 30s connection timeout prevents indefinite hangs |
| **ChromaSync reconnect guard** | Mutex prevents concurrent reconnection races |
| **Chroma watchdog** | Worker checks chroma health every 5 min, cleans up dead connections |
| **Parent heartbeat (MCP)** | MCP server self-exits when parent dies (ppid=1), preventing orphaned node processes |
| **Stale session recovery** | On startup, marks sessions active >10 min as failed (crash recovery) |
| **Subprocess pool limit** | Max concurrent Claude SDK agents, preventing runaway accumulation |

## Related Issues

- #752 - HTTP/SSE transport (this PR adds groundwork)
- #725 - Worker host IPv6 (partial, adds host setting)
- #730 - Vector-db growth (mitigated via CHROMA_DISABLED option)

## Results

On a macOS system that previously required external cleanup:
- Memory usage dropped from 60GB+ to 318MB
- Zero chroma-mcp orphans (with CHROMA_DISABLED=true)
- Stable process count over extended sessions

## Test plan

- [x] Build succeeds
- [x] Worker starts and passes health check
- [x] Memory search works (SQLite FTS5 fallback when chroma disabled)
- [x] Process count stable over time
- [x] No orphaned processes after session close

All changes are additive and backward-compatible.

---

Generated with [Claude Code](https://claude.ai/code)
via [Happy](https://happy.engineering)

PR #806, opened 2026-01-25
fix: prevent zombie process accumulation via PID registry and signal propagation (Issue #737)

bigph00t:
## Summary

Fixes the zombie process accumulation bug where Claude haiku subprocesses spawned by the Agent SDK don't terminate properly, causing memory leaks (user reported 176 processes / 51GB RAM after 3 days).

## Problem Analysis

The issue has three root causes that compound:

### 1. Missing Signal Propagation (Primary Cause)
The Agent SDK's `spawnClaudeCodeProcess` option allows custom spawn functions, but the previous implementation didn't pass the `signal` parameter. This means `abortController.abort()` never reaches the subprocess:

```typescript
// Before: signal not passed, abort() does nothing to subprocess
spawn(cmd, args, { cwd, env, stdio: [...] })

// After: signal passed, abort() terminates subprocess
spawn(cmd, args, { cwd, env, stdio: [...], signal: spawnOptions.signal })
```

### 2. Hidden PIDs
The SDK's `SpawnedProcess` interface intentionally hides subprocess PIDs. Without access to PIDs, we can't verify termination or force-kill hung processes.

### 3. Fire-and-Forget Cleanup  
`deleteSession()` called `abortController.abort()` but never verified the subprocess actually exited. Hung processes continued running indefinitely.

## Solution

This PR implements a **PID-capturing custom spawn** that integrates with the SDK's official extension point:

### New Module: `ProcessRegistry.ts`

| Function | Purpose |
|----------|---------|
| `registerProcess(pid, sessionDbId, process)` | Track spawned subprocess with session association |
| `unregisterProcess(pid)` | Remove from registry on exit |
| `getProcessBySession(sessionDbId)` | Find process for a session (warns if multiple found) |
| `ensureProcessExit(tracked, timeoutMs)` | Wait for exit with timeout, escalate to SIGKILL |
| `createPidCapturingSpawn(sessionDbId)` | Factory for SDK-compatible spawn that captures PIDs |
| `startOrphanReaper(getActiveSessionIds)` | Safety-net cleanup every 5 minutes |
| `reapOrphanedProcesses(activeSessionIds)` | Kill registry orphans + system-level ppid=1 orphans |

### Integration Points

1. **SDKAgent.ts**: Uses `spawnClaudeCodeProcess: createPidCapturingSpawn(sessionDbId)` to capture PIDs at spawn time

2. **SessionManager.ts**: `deleteSession()` now waits for subprocess exit with 5s timeout before cleanup

[...]

PR #848, opened 2026-01-28
fix: Kill duplicate observer processes to prevent zombie accumulation

influenist:
## Problem

When IDE (e.g., Antigravity/VS Code) spawns multiple claude-mem worker instances or restarts, each worker spawns observer subagents that resume the same session. These accumulate because:

1. Each worker only tracks its own processes in `processRegistry`
2. `killSystemOrphans()` only kills `ppid=1`, but IDE children have `ppid=IDE`
3. Multiple `--resume` calls to same session spawn duplicates

**User reported:** 79 processes consuming 13.4GB swap, causing disk to fill up.

## Related PRs

This PR is **complementary** to existing zombie/orphan fixes:

| PR | Approach | What it doesn't cover |
|----|----------|----------------------|
| #738 | Age-based cleanup (etime) | Doesn't detect by resume session ID |
| #806 | PID registry + signal propagation | Registry is per-worker instance |
| #830 | Subprocess pool limit (max 2) | Prevents new spawns, doesn't kill existing duplicates |
| #844 | Session-complete hook | Enables registry reaper, but registry is per-worker |
| #847 | Session isolation via `cwd` | Prevents pollution, doesn't cleanup duplicates |

## Unique Contribution

**`killDuplicatesByResumeId(resumeSessionId)`** - System-wide duplicate detection:

```bash
# Finds ALL processes resuming the same session, keeps only newest
ps -eo pid,lstart,args | grep -- "--resume <uuid>"
```

This catches the specific case where multiple workers (or IDE restarts) each spawn a process for the same observer session. The existing fixes don't address this because:
- Registry-based tracking only sees processes from the current worker
- Age-based cleanup kills ALL old processes, not specifically duplicates
- Pool limits prevent NEW spawns but don't clean up existing duplicates

## Solution

1. **`killDuplicatesByResumeId()`** - Kills all but newest process per resume session ID. Called BEFORE spawning new observer.

2. **`killStaleObservers()`** - Kills any claude process with `--disallowedTools` flag (observer signature) that has been idle (0% CPU) for >30 minutes, regardless of ppid.

3. Integrate `killStaleObservers` into `reapOrphanedProcesses()` so the 5-minute reaper catches processes from other worker instances.

## Changes

- `src/services/worker/ProcessRegistry.ts`: Add `killDuplicatesByResumeId()` and `killStaleObservers()` functions
- `src/services/worker/SDKAgent.ts`: Call `killDuplicatesByResumeId()` before spawning new observer

## Testing

- Verified fix prevents duplicate accumulation on session resume
- Verified stale observer cleanup works across worker restarts
- User's system: 79 processes → 0 duplicates after cleanup

---

### DIFF
--- a/src/services/worker/ProcessRegistry.ts
+++ b/src/services/worker/ProcessRegistry.ts
@@ -189,14 +189,18 @@ async function killStaleObservers(maxIdleMs: number = 30 * 60 * 1000): Promise<n
       // Parse: PID CPU% "Day Mon DD HH:MM:SS YYYY" command...
       // Example: 12345  0.0 Tue Jan 28 14:30:00 2025 /path/to/claude ...
       const match = line.trim().match(/^(\d+)\s+([\d.]+)\s+(\w+\s+\w+\s+\d+\s+[\d:]+\s+\d+)/);
-      if (!match) continue;
+      if (!match) {
+        logger.debug('PROCESS', `Skipping malformed ps line: ${line.substring(0, 100)}`);
+        continue;
+      }
 
       const pid = parseInt(match[1]);
       const cpuPercent = parseFloat(match[2]);
       const startTimeStr = match[3];
 
-      // Skip processes still actively using CPU
-      if (cpuPercent > 0.1) continue;
+      // Skip processes still actively using CPU (1% threshold to avoid killing
+      // processes that are briefly active during inference or I/O)
+      if (cpuPercent > 1.0) continue;
 
       // Parse start time to epoch
       try {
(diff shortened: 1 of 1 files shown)


influenist:
Thanks for the review! Addressed all feedback in commit 886ccd5:

| Issue | Fix |
|-------|-----|
| Shell injection risk | Validate `resumeSessionId` as UUID before interpolation |
| CPU threshold too low | Increased from 0.1% to 1.0% |
| Race condition on same-second spawns | Added PID as tiebreaker in sort |
| Silent regex failures | Added `logger.debug()` for malformed lines |

thedotmack:
Already addressed in v9.0.13 — zombie observer prevention with idle timeout. Thank you!

PR #861, opened 2026-01-31
fix: add idle timeout with abort to prevent zombie observer processes

bigph00t:
## Summary

Root cause fix for zombie observer accumulation. **Supersedes PR #856** which was incomplete.

**User reported:** 79 processes consuming 13.4GB swap, causing disk to fill up.

## Why PR #856 Doesn't Work

PR #856 added idle timeout to `SessionQueueProcessor.createIterator()` but only **returned from the iterator** after 3 minutes of idle time. This is insufficient because:

| Step | What Happens | Result |
|------|-------------|--------|
| 1 | `createIterator()` returns | Generator stops yielding |
| 2 | SDK's `Query.streamInput()` closes stdin | `transport.endInput()` called |
| 3 | Claude subprocess receives EOF on stdin | **May NOT exit** |
| 4 | No `abort()` called | No SIGTERM sent |
| 5 | Subprocess stays alive | **Zombie** |

**Verified via Codex analysis of SDK internals:**

> "The idle-timeout change only stops the queue iterator and ends the prompt stream; it does **not** explicitly terminate the SDK subprocess. The only explicit kill path is `session.abortController.abort()` during session deletion."

The subprocess only terminates if:
1. Claude CLI decides to exit on stdin EOF (not guaranteed), OR
2. `abortController.abort()` is called (sends SIGTERM via ProcessTransport)

PR #856 relies on option 1, which is unreliable.

## This Fix

Add `onIdleTimeout` callback that **actually kills the subprocess**:

```typescript
// SessionQueueProcessor.ts
async *createIterator(options: CreateIteratorOptions) {
  // ... after 3 minutes idle ...
  if (onIdleTimeout) {
    onIdleTimeout();  // ← Callback triggers abort
  }
  return;
}

// SessionManager.ts
processor.createIterator({
  sessionDbId,
  signal: session.abortController.signal,
  onIdleTimeout: () => {
    session.abortController.abort();  // ← Sends SIGTERM
  }
})
```

## SDK Code References (from Codex analysis)

| Location | What It Does |
|----------|-------------|
| `sdk.mjs:7812-7821` | ProcessTransport abort handler sends SIGTERM |
| `sdk.mjs:7880-7899` | `transport.close()` sends SIGTERM → escalates to SIGKILL after 5s |
| `sdk.mjs:8449-8467` | `Query.streamInput()` only closes stdin on generator end |

## Changes

- **`SessionQueueProcessor.ts`**: Add `CreateIteratorOptions` interface with `onIdleTimeout` callback. Callback is invoked after 3 minutes of no messages.
- **`SessionManager.ts`**: Pass callback that calls `session.abortController.abort()` to actually kill the subprocess.

## Test Plan

- [x] Build succeeds
- [x] Verified via Codex analysis that abort() is required to kill subprocess
- [ ] Observer exits after 3 minutes of no messages (sends SIGTERM)
- [ ] Normal message processing unaffected
- [ ] No zombie accumulation over extended usage

Supersedes #856

### DIFF
--- a/src/services/queue/SessionQueueProcessor.ts
+++ b/src/services/queue/SessionQueueProcessor.ts
@@ -58,6 +58,8 @@ export class SessionQueueProcessor {
         if (persistentMessage) {
           // Reset activity time when we successfully yield a message
           lastActivityTime = Date.now();
+          // Check abort before yielding to reduce race window
+          if (signal.aborted) return;
           // Yield the message for processing (it's already deleted from queue)
           yield this.toPendingMessageWithId(persistentMessage);
         } else {
@@ -65,13 +67,17 @@ export class SessionQueueProcessor {
           const receivedMessage = await this.waitForMessage(signal, IDLE_TIMEOUT_MS);
 
           if (!receivedMessage) {
-            if (signal.aborted) continue; // Let loop check signal.aborted and exit
+            if (signal.aborted) return; // Signal already aborted, exit immediately
 
             // Final safety check - has a message arrived in the race window?
-            // This handles the case where a message is enqueued just as the timeout fires
+            // This handles the case where a message is enqueued just as the timeout fires.
+            // Note: If a message arrives after this check but before abort, it will be
+            // processed when the observer restarts - acceptable since we've been idle 3+ min.
             const finalCheck = this.store.claimAndDelete(sessionDbId);
             if (finalCheck) {
               lastActivityTime = Date.now();
+              // Check abort before yielding to reduce race window
+              if (signal.aborted) return;
               yield this.toPendingMessageWithId(finalCheck);
               continue; // Keep processing - don't abort
             }


elrolio:
I was experiencing this exact issue - Claude Code would hang on startup with multiple observer processes spinning at 100% CPU. Logs showed FOREIGN KEY constraint failures and SIGTERM errors.

**Workaround that worked for me:** Disabled the plugin in `settings.local.json`:
```json
"enabledPlugins": {
}
```

Then killed all zombie processes with:
```bash
pkill -9 -f "claude-mem"
pkill -9 -f "worker-service"
pkill -9 -f "chroma-mcp"
```

Looking forward to the fix - this plugin is really useful when it works!

thedotmack:
Already addressed in v9.0.13 — 'Zombie Observer Prevention (#856)' added 3-minute idle timeout with race condition fix and 11 tests. Thank you for the contribution\!

PR #867, opened 2026-02-01
fix: prevent process bomb and zombie observers on startup

influenist:
## Problem

Two related issues cause runaway process accumulation:

1. **Startup process bomb:** `initializeBackground()` calls `processPendingQueues(50)`, spawning up to 50 claude subprocesses at once (each with its own MCP server). With stale pending messages in the database, this freezes the system.

2. **Zombie observers:** `SessionQueueProcessor.createIterator()` waits forever when the queue is empty. If the `session-complete` hook doesn't fire (terminal closed, crash, network issue), observer subprocesses never exit.

**User reported:** 79 processes consuming 13.4GB swap.

### 1. Batched queue recovery (replaces fire-all-at-once)

- Process 3 sessions immediately on startup
- Drain remaining sessions in batches of 3 every 60 seconds
- Auto-stop interval once queue is empty
- **No messages are dropped** — all pending work is eventually processed

### 2. Idle timeout for observer subprocesses (from PR #856)

- `waitForMessage()` now accepts a timeout and returns `boolean`
- `createIterator()` tracks `lastActivityTime` and triggers `onIdleTimeout` callback after 3 minutes of inactivity
- Callback calls `abortController.abort()` to kill the subprocess cleanly
- Prevents zombie accumulation even when hooks fail to fire

| Approach | Before | After |
|----------|--------|-------|
| Startup recovery | 50 sessions at once | 3 at a time, every 60s |
| Idle observers | Wait forever | Exit after 3 min idle |
| Stale messages | All recovered at once | Batched, none dropped |

## Changes

- `src/services/worker-service.ts`: Batched interval drainer replacing `processPendingQueues(50)` fire-and-forget. Cleanup on shutdown.
- `src/services/queue/SessionQueueProcessor.ts`: Idle timeout with graceful exit via `onIdleTimeout` callback.
- `src/services/worker/SessionManager.ts`: Pass `onIdleTimeout` callback that triggers `abortController.abort()`.
- `tests/services/queue/SessionQueueProcessor.test.ts`: 11 test cases covering timeout behavior, abort handling, message processing, and event cleanup.

## Test plan

- [ ] Start `claude` with many sessions having pending messages — verify only 0-3 spawn initially
- [ ] Wait 60s — verify next batch of up to 3 starts
- [ ] Verify interval stops once all pending sessions are processed (log: "Pending queue fully drained")
- [ ] Verify observer exits after 3 minutes of no messages (log: "Idle timeout reached")
- [ ] Verify normal message processing is unaffected
- [ ] Verify no zombie accumulation over extended usage
- [ ] Unit tests pass

Supersedes #866. Incorporates idle timeout approach from #856.

### DIFF
--- a/src/services/queue/SessionQueueProcessor.ts
+++ b/src/services/queue/SessionQueueProcessor.ts
@@ -3,6 +3,15 @@ import { PendingMessageStore, PersistentPendingMessage } from '../sqlite/Pending
 import type { PendingMessageWithId } from '../worker-types.js';
 import { logger } from '../../utils/logger.js';
 
+const IDLE_TIMEOUT_MS = 3 * 60 * 1000; // 3 minutes
+
+export interface CreateIteratorOptions {
+  sessionDbId: number;
+  signal: AbortSignal;
+  /** Called when idle timeout occurs - should trigger abort to kill subprocess */
+  onIdleTimeout?: () => void;
+}
+
 export class SessionQueueProcessor {
   constructor(
     private store: PendingMessageStore,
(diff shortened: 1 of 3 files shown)


travisdmathis:
should be fixed asap, absolutely crushing my m4 mac w/ 24gb ram since i added this plugin.

thedotmack:
Closing this PR — the two issues it addresses have been resolved in the current codebase:

**1. Zombie observer fix (idle timeout) — Already shipped**
`SessionQueueProcessor.ts` already implements the exact 3-minute idle timeout with `onIdleTimeout` callback that triggers `abortController.abort()`. This was merged as part of the v9.0.13 fixes. The PR's idle timeout implementation is redundant with what's on `main`.

**2. Startup batching (3 sessions at a time) — Not needed**
The current `processPendingQueues(50)` has a 100ms stagger between session starts (not truly simultaneous), and with the idle timeout now in place, sessions self-terminate after 3 minutes of inactivity. The original user report (79 processes/13.4GB swap) was caused by the zombie problem, which is now fixed. Adding an interval-based drainer (3 sessions every 60 seconds) would delay recovery of legitimate pending work without meaningful benefit.

Thank you for the contribution — the idle timeout approach was sound and is exactly what shipped.

PR #879, opened 2026-02-02
fix: add daemon children cleanup to orphan reaper

boaz-robopet:
## Problem

Worker-service daemon spawns Claude SDK processes that don't terminate after completing their work. These zombie processes accumulate over time, consuming significant memory (observed: 15+ processes using ~3.5GB).

### Root Cause

The existing `killSystemOrphans()` function only handles true orphans (PPID=1), but zombie SDK processes are children of the daemon itself (PPID=daemon_pid).

```bash
# Zombies are daemon children, not orphans
$ ps -eo pid,ppid,etime,comm | grep claude
83632 71277 23:51 claude    # PPID=71277 (daemon), not PPID=1
84731 71277 22:51 claude
# ... more processes
```

## Solution

Add `killIdleDaemonChildren()` to the existing 5-minute orphan reaper cycle:

- Finds claude processes where:
  - Parent PID = daemon's PID (children of worker-service)
  - CPU = 0% (idle, not actively working)
  - Running > 2 minutes (completed their work)
- Called from `reapOrphanedProcesses()` which runs every 5 minutes

### Why this approach?
- **Leverages existing infrastructure** - orphan reaper already runs every 5 minutes
- **No additional overhead** - no new hooks or timers
- **Appropriate timing** - 5 minutes is sufficient (zombies don't cause immediate issues)
- **Single responsibility** - process cleanup stays in ProcessRegistry

## Testing

Verified locally on macOS:
- Before: 23 Claude processes, 4.3GB
- After cleanup: 5 processes, 2.3GB
- Memory saved: ~2GB
- Normal processes (MCP server, Chroma) unaffected

## Compatibility

- macOS: Tested ✅
- Linux: Should work (same ps command format)
- Windows: Skipped (different process model)

---

### DIFF
--- a/src/services/worker/ProcessRegistry.ts
+++ b/src/services/worker/ProcessRegistry.ts
@@ -122,6 +122,86 @@ export async function ensureProcessExit(tracked: TrackedProcess, timeoutMs: numb
 }
 
 /**
+ * Kill idle daemon children (claude processes spawned by worker-service)
+ *
+ * These are SDK-spawned claude processes that completed their work but
+ * didn't terminate properly. They remain as children of the worker-service
+ * daemon, consuming memory without doing useful work.
+ *
+ * Criteria for cleanup:
+ * - Process name is "claude"
+ * - Parent PID is the worker-service daemon (this process)
+ * - Process has 0% CPU (idle)
+ * - Process has been running for more than 2 minutes
+ */
+async function killIdleDaemonChildren(): Promise<number> {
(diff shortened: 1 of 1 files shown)


thedotmack:
Merged into main via rebase. The daemon children cleanup fills a real gap — the existing reaper catches registry-tracked processes and ppid=1 orphans, but not stuck SDK children under a living daemon. Clean implementation, builds and tests pass. Thanks @boaz-robopet!

PR #899, opened 2026-02-03
fix: resolve message processing failures in multi-session scenarios

hahaschool:
## Summary
- Fix AbortController state issue causing generators to abort within ~70ms
- Fix FOREIGN KEY constraint violation when worker restarts with existing observations
- Fix deprecated queueDepth logging showing 0 instead of actual count

## Problem
When multiple Claude Code sessions were running in the same project folder, messages would accumulate without being processed. Root causes:

1. **AbortController reuse**: When a generator finished naturally, it aborted its controller but kept the session in memory. New messages would try to start a generator with the already-aborted controller, causing immediate exit.

2. **FK constraint violation**: After worker restart, SDK generates new session ID. Updating `sdk_sessions.memory_session_id` failed because existing observations referenced the old ID (FK without `ON UPDATE CASCADE`).

## Solution
1. Check and replace aborted AbortController before starting generators
2. Use database's existing `memory_session_id` for observation storage (FK integrity)
3. Only update `memory_session_id` in database when it's NULL

## Test plan
- [x] Verified messages flow correctly: ENQUEUED → CLAIMED → STORED
- [x] No FK constraint errors in logs
- [x] Pending queue stays empty (no accumulation)
- [x] New observations stored successfully after fix

### DIFF
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -648,8 +648,9 @@ export class SessionStore {
   /**
    * Update the memory session ID for a session
    * Called by SDKAgent when it captures the session ID from the first SDK message
+   * Pass null to clear the memory_session_id (used in stale resume recovery)
    */
-  updateMemorySessionId(sessionDbId: number, memorySessionId: string): void {
+  updateMemorySessionId(sessionDbId: number, memorySessionId: string | null): void {
     this.db.prepare(`
       UPDATE sdk_sessions
       SET memory_session_id = ?
--- a/src/services/sqlite/sessions/create.ts
+++ b/src/services/sqlite/sessions/create.ts
@@ -49,11 +49,12 @@ export function createSDKSession(
 /**
  * Update the memory session ID for a session
  * Called by SDKAgent when it captures the session ID from the first SDK message
+ * Pass null to clear the memory_session_id (used in stale resume recovery)
  */
 export function updateMemorySessionId(
   db: Database,
   sessionDbId: number,
-  memorySessionId: string
+  memorySessionId: string | null
 ): void {
   db.prepare(`
     UPDATE sdk_sessions
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -388,7 +388,7 @@ export class WorkerService {
           session.memorySessionId = null;
           session.forceInit = true;
           // Also clear from database to prevent future stale resumes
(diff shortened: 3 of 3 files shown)


hahaschool:
Thanks for the thorough review, @greptile-apps! 

I've addressed the type safety concern:

**Fixed in commit 76227ec4:**
- Updated `updateMemorySessionId` function signature to accept `string | null` instead of just `string`
- Removed the `as any` type assertion in `worker-service.ts` line 391
- Updated both `SessionStore.ts` and `sessions/create.ts` for consistency

The function now properly supports clearing `memory_session_id` during stale resume recovery without bypassing type safety.

thedotmack:
Thank you for this thorough fix, @hahaschool! You identified some real issues in the multi-session message processing flow. However, this PR now has significant overlap with several PRs that were merged after yours was opened, and the rebase produces conflicts in 3 files (`worker-service.ts`, `worker-types.ts`, `SDKAgent.ts`).

**Already addressed by recently merged PRs:**
- **Crash recovery with restart limits** → PR #693 (merged): Adds `consecutiveRestarts` counter, limits to 3 restarts with exponential backoff in `SessionRoutes.ts`
- **Orphaned message handling / session termination detection** → PR #937 (merged): Adds `isSessionTerminatedError()`, `runFallbackForTerminatedSession()` with Gemini/OpenRouter fallback, and `markAllSessionMessagesAbandoned()` in `worker-service.ts`
- **AbortController replacement during crash recovery** → PR #693 already replaces the AbortController in the restart path (SessionRoutes.ts:211-214)

**Still valuable — consider submitting as focused PRs:**

1. **AbortController check at generator start** — The check for `session.abortController.signal.aborted` before starting a new generator (in both `startSessionProcessor` and `startGeneratorWithProvider`) is NOT covered by existing merges. When a generator completes naturally and aborts its controller, a new incoming message could try to start with the stale aborted controller. This is a real bug. Would be a clean ~10-line PR.

2. **FK constraint protection in SDKAgent/ResponseProcessor** — The logic to check the DB's `memory_session_id` before overwriting it (to avoid FK violations when the SDK generates a new session ID after restart) is genuinely valuable and not addressed elsewhere. This would be a focused PR touching `SDKAgent.ts` and `ResponseProcessor.ts`.

3. **`queueDepth` logging using `pendingStore.getPendingCount()`** — Minor but correct fix for deprecated `session.pendingMessages.length`. Could be included with #1.

If you'd like to re-submit these as 1-2 focused PRs, they'd be straightforward to review and merge. Thanks for the thorough investigation!

PR #913, opened 2026-02-04
fix: respect CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED setting

superbiche:
## Summary
The `CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED` setting was documented but never checked in code. Folder CLAUDE.md files were generated unconditionally whenever files were touched.

### Initial fix
- Add `CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED` to `SettingsDefaults` interface
- Add default value `'false'` to DEFAULTS object
- Check setting in `ResponseProcessor.ts` before calling `updateFolderClaudeMdFiles()`

### Type handling fix (follow-up)
During testing, discovered that `JSON.parse()` preserves native types from settings.json:
- Boolean `true` stays as boolean, not string `"true"`
- The original check `=== 'true'` only handled strings

Updated both `getBool()` helper and ResponseProcessor to handle:
- String `'true'` → enabled
- Boolean `true` → enabled  
- Any other value → disabled

## Testing
Verified the fix works correctly:
1. Set `CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED: "false"` in settings
2. Read files from a directory without CLAUDE.md
3. Confirmed no new CLAUDE.md file was created
4. Previous cleanup removed 11 auto-generated empty files; count dropped from 22 to 15

## Related
Fixes issue noted in `docs/reports/issue-600-documentation-audit-features-not-implemented.md`:
> "The setting `CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED` is never read."

### DIFF
--- a/src/services/worker/agents/ResponseProcessor.ts
+++ b/src/services/worker/agents/ResponseProcessor.ts
@@ -219,7 +219,9 @@ async function syncAndBroadcastObservations(
   // This runs per-observation batch to ensure folders are updated as work happens
   // Only runs if CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED is true (default: false)
   const settings = SettingsDefaultsManager.loadFromFile(USER_SETTINGS_PATH);
-  const folderClaudeMdEnabled = settings.CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED === 'true';
+  // Handle both string 'true' and boolean true from JSON settings
+  const settingValue = settings.CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED;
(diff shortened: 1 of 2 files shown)


superbiche:
The first fix wasn't enough, folder context files were still generated. The last commit fixes a mistype issue that seems to do the trick (I'll be battle-testing this a bit more this afternoon but for now the results are expected)

thedotmack:
Merged into main via rebase+merge. Build passes. The setting CLAUDE_MEM_FOLDER_CLAUDEMD_ENABLED (default: false) is now respected — folder CLAUDE.md files are only generated when explicitly enabled. Thank you @superbiche!

Issue #918, opened 2026-02-04
SessionStart hooks block input in terminal on Windows — needs "async": true


23rdletter:
I had Claude write out the below. Claude-mem was working until a terminal window froze and I terminated the process. After that, I was unable to input anything into Claude. After isolating to 2 plugins - Superpowers and Claude-mem - I found someone describing the same problem on the Superpowers issue page. 

They added async: true to the Superpowers hooks and that worked for me. I applied the same to Claude-mem and now it's also working. 

Problem
On Windows, launching Claude Code CLI in a project directory with claude-mem enabled causes the TUI to render but stdin becomes completely unresponsive — no keyboard input is accepted. The session must be force-killed.

Environment
OS: Windows 11
Claude Code: Native install (not npm wrapper), v2.1.31
claude-mem version: 9.0.12
Terminal: Affects all terminals (PowerShell, VS Code integrated, Tabby)

Root Cause
The 4 SessionStart hooks in hooks.json all spawn subprocesses (node, bun) synchronously. On Windows, these block the Ink TUI framework from entering raw stdin mode during initialization:

"SessionStart": [{
  "matcher": "startup|clear|compact",
  "hooks": [
    { "type": "command", "command": "node ... smart-install.js", "timeout": 300 },
    { "type": "command", "command": "bun ... worker-service.cjs start", "timeout": 60 },
    { "type": "command", "command": "bun ... worker-service.cjs hook claude-code context", "timeout": 60 },
    { "type": "command", "command": "bun ... worker-service.cjs hook claude-code user-message", "timeout": 60 }
  ]
}]

The issue is amplified by the project having a large .claude/settings.local.json (127 permission entries), which increases startup processing time. [my note: this is unique to my machine]

Fix
Adding "async": true to all 4 SessionStart hooks resolves the issue completely:

{
  "type": "command",
  "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/smart-install.js\"",
  "timeout": 300,
  "async": true
},
{
  "type": "command",
  "command": "bun \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start",
  "timeout": 60,
  "async": true
},
{
  "type": "command",
  "command": "bun \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" hook claude-code context",
  "timeout": 60,
  "async": true
},
{
  "type": "command",
[...]

PR #922, opened 2026-02-04
fix: add async:true to SessionStart hooks to prevent termina…

kamran-khalid-v9:
…l input blocking (#918)

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -9,21 +9,25 @@
             "type": "command",
             "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/smart-install.js\"",
             "timeout": 300
+            ,"async": true
           },
           {
             "type": "command",
             "command": "bun \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start",
             "timeout": 60
+            ,"async": true
           },
           {
             "type": "command",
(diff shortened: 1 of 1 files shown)


thedotmack:
Closing this PR — thank you for identifying the Windows terminal blocking issue, but making SessionStart hooks async would break claude-mem's core functionality.

**Why async won't work for SessionStart:**

The SessionStart `context` hook is responsible for injecting memory context into Claude's system prompt. This hook's stdout is captured and fed to Claude as initial context. If it runs asynchronously, Claude would start processing *without* any memory context — defeating the purpose of claude-mem entirely.

Additionally, the three SessionStart hooks have a strict dependency chain:
1. `smart-install.js` — must complete before worker can start
2. `worker-service start` — must complete before context can be fetched
3. `context hook` — stdout must be captured synchronously

**What has been done instead:**

The Windows blocking issue has been addressed through a different approach in the hook resilience PRs:
- PR #973: Hooks now fail gracefully (exit 0 with error info) instead of hanging
- PR #959: Context injection endpoint returns empty context during initialization instead of blocking
- PR #964: All fetch calls now have timeouts to prevent indefinite hangs

These changes ensure that even if the worker is slow or unavailable, hooks complete quickly rather than blocking the terminal.

**Note:** This PR is also based on an older version of the codebase — it references `bun` directly (now wrapped via `bun-runner.js`) and includes a `user-message` hook that was removed in PR #960.

Thank you for the contribution!

PR #933, opened 2026-02-04
fix(windows): replace deprecated wmic worker spawn with child_proces spawn

### LINKED ISSUE #890: Bug: Worker fails to start on Windows 11 (24H2+) due to wmic deprecation
### Description
In recent versions of Windows 11 (specifically 24H2 and newer), `wmic` has been deprecated and is no longer installed by default. The current implementation of `claude-mem` relies on `wmic` to spawn the background worker daemon on Windows, which causes the following error:

```text
● UserPromptSubmit operation blocked by hook:
  [bun "${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" hook claude-code session-init]: Hook error: Error:
  Worker did not become ready within 15 seconds. (port 37777)
```

In the logs, this manifests as:
```text
[ERROR] [SYSTEM] Failed to spawn worker daemon
```

### Root Cause
In `scripts/worker-service.cjs`, the worker daemon is spawned using `wmic process call create`. Since `wmic` is removed in modern Windows 11 builds, this fails.

### Environment
- **OS:** Windows 11 24H2 (win32)
- **Plugin Version:** 9.0.12
- **Runtime:** Bun 1.3.6 / Node v24.13.0

### Proposed Fix
Replace the `wmic` usage with standard `child_process.spawn`.

```javascript
function YS(t, e, r = {}) {
  let i = { ...process.env, CLAUDE_MEM_WORKER_PORT: String(e), ...r };
  let a = (0, no.spawn)(process.execPath, [t, "--daemon"], {
    detached: true,
    stdio: "ignore",
    env: i,
    windowsHide: true
  });
  
  if (a.pid !== undefined) {
    a.unref();
    return a.pid;
  }
}
```

### Verification
Manually replacing the code locally fixed the issue, allowing the worker to start and port 37777 to become active.

jayvenn21:
## Summary

This PR fixes worker startup failures on Windows 11 (24H2+) caused by the
deprecation and removal of `wmic`.

## Details

The Windows worker daemon was previously spawned using
`wmic process call create`, which is no longer available on recent
Windows 11 builds. This caused the worker to fail to start and
hook execution to time out.

This change replaces the `wmic`-based path with a
`child_process.spawn`-based daemon launch using standard Node.js APIs,
matching the existing behavior while restoring compatibility.

## Impact

- Worker starts successfully on Windows 11 24H2+
- No terminal popups
- No behavior change on macOS/Linux

## Related Issue

Closes #890

### DIFF
--- a/src/services/infrastructure/ProcessManager.ts
+++ b/src/services/infrastructure/ProcessManager.ts
@@ -263,11 +263,8 @@ export async function cleanupOrphanedProcesses(): Promise<void> {
  * Spawn a detached daemon process
  * Returns the child PID or undefined if spawn failed
  *
- * On Windows, uses WMIC to spawn a truly independent process that
- * survives parent exit without console popups. WMIC creates processes
- * that are not associated with the parent's console.
- *
- * On Unix, uses standard detached spawn.
+ * Uses child_process.spawn on all platforms. Windows 11 24H2+ removed wmic,
+ * so we no longer use wmic process call create on Windows (issue #890).
  *
  * PID file is written by the worker itself after listen() succeeds,
  * not by the spawner (race-free, works on all platforms).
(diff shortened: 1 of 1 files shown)


thedotmack:
Already addressed in v9.0.2 — WMIC replacement with PowerShell commands. Thank you!

PR #935, opened 2026-02-04
fix(worker): guard ProcessTransport writes on Windows startup

### LINKED ISSUE #874: [Windows] ProcessTransport is not ready for writing - Worker crashes on startup
## Environment
- OS: Windows 10/11
- Bun: v1.3.8
- Claude Code: Latest

## Issue
Worker service crashes immediately on startup with the following error:

```
error: ProcessTransport is not ready for writing
      at write (worker-service.cjs:1144:5337)
      at streamInput (worker-service.cjs:1147:1041)

[wrapper] Inner exited with code=1, signal=null
[wrapper] Inner exited unexpectedly
```

## Steps to Reproduce
1. Install plugin on Windows
2. Start Claude Code or manually run `bun plugin/scripts/worker-wrapper.cjs`
3. Worker crashes immediately

## Expected Behavior
Worker should start and listen on port 37777.

## Notes
- Daemon spawn also fails with "Failed to spawn worker daemon"
- Likely a Windows-specific stdin/stdout stream handling issue
- The write() call happens before ProcessTransport is ready

jayvenn21:
### Summary
Fixes a Windows-specific startup crash where the worker exits immediately with:
`ProcessTransport is not ready for writing`.

### Root Cause
On Windows (Bun), stdin/stdout transport is not writable immediately at process start.
The worker attempts to write before the transport is initialized, causing a fatal error.

### Fix
- Guard writes until ProcessTransport is ready
- No behavioral changes on non-Windows platforms
- Prevents immediate worker crash and daemon spawn failure

### Scope
- Windows-only startup path
- Minimal, defensive change
- No protocol or logic changes

Closes #874

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "scripts": {
     "dev": "npm run build-and-sync",
     "build": "node scripts/build-hooks.js",
+    "postinstall": "patch-package",
     "build-and-sync": "npm run build && npm run sync-marketplace && sleep 1 && cd ~/.claude/plugins/marketplaces/thedotmack && npm run worker:restart",
     "sync-marketplace": "node scripts/sync-marketplace.cjs",
     "sync-marketplace:force": "node scripts/sync-marketplace.cjs --force",
@@ -85,7 +86,7 @@
     "test:server": "bun test tests/server/"
   },
   "dependencies": {
-    "@anthropic-ai/claude-agent-sdk": "^0.1.76",
+    "@anthropic-ai/claude-agent-sdk": "0.1.77",
     "@modelcontextprotocol/sdk": "^1.25.1",
     "ansi-to-html": "^0.7.2",
     "express": "^4.18.2",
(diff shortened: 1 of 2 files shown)


thedotmack:
Closing this PR — thank you for identifying the Windows startup crash issue, but the approach of patching the SDK has significant concerns:

**Why not this approach:**

1. **Silently dropping writes is dangerous** — Changing `throw new Error("ProcessTransport is not ready for writing")` to a silent `return` means messages are lost without indication. This can cause subtle, hard-to-debug issues where commands or responses are silently dropped.

2. **Fragile dependency patching** — The patch is tied to exact SDK version 0.1.77 and a specific line number in the bundled file. Any SDK update breaks it, and `patch-package` adds build-time complexity.

3. **Already addressed by fail-open architecture** — PRs #973 and #959 (both merged) implement a fail-open strategy where hooks continue gracefully when the worker is unavailable. If the worker crashes on startup due to transport readiness, the hook system handles it without blocking Claude Code.

4. **Violates Fail Fast principle** — It's better for the worker to crash visibly (and let fail-open handle it) than to silently swallow errors.

**Better alternatives:**
- If transport readiness is still an issue, the fix should be at the application layer: either queuing writes until the transport signals ready, or adding a readiness check before the first write in our worker code — not patching the SDK.

Issue #936, opened 2026-02-04
Feature Request: Process orphan pending messages when Claude Code session has terminated


tulensrma:
## Summary

When a Claude Code terminal closes before pending messages are processed, those messages become permanently orphaned and can never be processed. The `/api/pending-queue/process` endpoint attempts to resume processing but fails because the SDK cannot connect to a terminated Claude Code session.

## Environment

- **Plugin Version:** v9.0.12
- **OS:** Linux (Ubuntu 24.04)
- **Date Discovered:** 2026-02-04

## Symptoms

1. User works in Claude Code session, generating observations/summaries
2. Messages are ENQUEUED to `pending_messages` table
3. User closes terminal before queue fully drains
4. Session remains `active` in `sdk_sessions` (never marked completed)
5. Pending messages stuck forever in `pending` status
6. Calling `POST /api/pending-queue/process` fails with "Claude Code process aborted by user"

### Log Pattern

```
[INFO ] [SYSTEM] Starting processor for session 16064 {project=viinitie-analytics, pendingCount=12}
[INFO ] [SYSTEM] [session-16064] Starting generator (startup-recovery)
[INFO ] [SDK   ] Starting SDK query {sessionDbId=16064, hasRealMemorySessionId=true, resume_parameter=...}
[INFO ] [SDK   ] Creating message generator {promptType=CONTINUATION}
[ERROR] [SDK   ] [session-16064] Session generator failed {project=viinitie-analytics} Claude Code process aborted by user
```

## Root Cause

The current architecture requires an active Claude Code session to process pending messages:

1. Generator tries to resume the original Claude Code session via SDK
2. SDK discovers the Claude Code process is terminated
3. Processing fails - no fallback mechanism exists

The `pending_messages` table already contains:
- `last_user_message` - the user's prompt
- `last_assistant_message` - Claude's response  
- `tool_name`, `tool_input`, `tool_response` - tool call context
- `cwd` - working directory

This context should be sufficient to generate observations without needing to resume the original session.

## Impact

- Pending messages accumulate over time (51 orphaned in one day of normal usage)
- 240+ sessions marked as "active" when terminals have long closed
- Only workaround is manually marking messages as `failed`, losing the observations
- Queue depth grows indefinitely until cleanup

### Option 1: Standalone Observation Generation
Process orphan messages using the stored context without requiring session resumption:
- Use `last_user_message`, `last_assistant_message`, tool context
- Generate observations via Claude API directly
[...]

PR #937, opened 2026-02-04
fix(worker): gracefully process orphaned pending messages after session termination

jayvenn21:
This PR prevents pending messages from becoming permanently orphaned when a Claude Code session terminates before the queue fully drains.

When SDK session resumption fails due to a terminated process, the worker now falls back to standalone observation generation using stored message context instead of aborting.

This ensures:
- Pending messages are not stuck indefinitely
- Sessions are marked completed correctly
- Queue depth cannot grow unbounded due to closed terminals

The change is intentionally scoped to graceful degradation and does not alter the normal SDK-backed processing path.

for issue #936

### DIFF
--- a/src/services/sqlite/PendingMessageStore.ts
+++ b/src/services/sqlite/PendingMessageStore.ts
@@ -205,6 +205,23 @@ export class PendingMessageStore {
   }
 
   /**
+   * Mark all pending and processing messages for a session as failed (abandoned).
+   * Used when SDK session is terminated and no fallback agent is available:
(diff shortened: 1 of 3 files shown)


thedotmack:
Merged via rebase onto main. This PR adds graceful handling of orphaned pending messages after Claude Code session termination: detects terminated-session errors during SDK resume, falls back to Gemini/OpenRouter if available, and marks messages as abandoned + removes session if no fallback exists. Build passes. Thank you @jayvenn21 for the clean, well-scoped fix!

PR #940, opened 2026-02-04
fix: Backfill project field on SDK session creation

### LINKED ISSUE #939: SDK session project field empty due to race condition in createSDKSession
## Bug

The `createSDKSession` function uses `INSERT OR IGNORE` to create sessions idempotently. However, the first hook to create a session determines the project value, and `PostToolUse` often wins the race with an empty project string — before `UserPromptSubmit`'s `session-init` sets it.

This causes observations to have an empty `project` field, which breaks downstream features that depend on project (e.g., vault routing in sinesync).

## Root Cause

In `src/services/sqlite/sessions/create.ts` (and `SessionStore.ts`):

```typescript
db.prepare(`
  INSERT OR IGNORE INTO sdk_sessions
  (content_session_id, memory_session_id, project, user_prompt, started_at, started_at_epoch, status)
  VALUES (?, NULL, ?, ?, ?, ?, 'active')
`).run(contentSessionId, project, userPrompt, now.toISOString(), nowEpoch);
```

If `PostToolUse` fires before `UserPromptSubmit` completes session-init, the session is created with `project = ''`. The subsequent `INSERT OR IGNORE` from session-init is ignored since the row already exists.

## Proposed Fix

Add a backfill UPDATE after the INSERT OR IGNORE to set the project if it was initially empty:

```typescript
// Backfill project if session was created by another hook with empty project
if (project) {
  db.prepare(`
    UPDATE sdk_sessions SET project = ?
    WHERE content_session_id = ? AND (project IS NULL OR project = '')
  `).run(project, contentSessionId);
}
```

This needs to be applied in both:
- `src/services/sqlite/sessions/create.ts` (`createSDKSession` function)
- `src/services/sqlite/SessionStore.ts` (`createSDKSession` method)

## Reproduction

1. Start a new Claude Code conversation in a project directory
2. Use any tool (triggers PostToolUse hook)
3. Check `sdk_sessions` table — the session's `project` field is empty
4. The `UserPromptSubmit` hook fires session-init with the correct project, but `INSERT OR IGNORE` skips it

## Workaround

Manually patching the compiled plugin cache files (`worker-service.cjs` and `context-generator.cjs`) with the backfill UPDATE. This gets overwritten on plugin updates.

miclip:
## Summary
- Fixes race condition where `PostToolUse` hook creates the SDK session with an empty project before `UserPromptSubmit`'s session-init sets it
- Adds a backfill `UPDATE` after `INSERT OR IGNORE` in `createSDKSession` to set the project when a later hook provides it
- Applied in both `sessions/create.ts` (functional) and `SessionStore.ts` (class method)

## Test plan
- [ ] Start a new Claude Code conversation in a project directory
- [ ] Use any tool to trigger PostToolUse
- [ ] Verify `sdk_sessions.project` is populated after session-init fires
- [ ] Verify existing sessions with a project set are not overwritten

Closes #939

### DIFF
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -1165,7 +1165,7 @@ export class SessionStore {
     const now = new Date();
     const nowEpoch = now.getTime();
 
-    // Pure INSERT OR IGNORE - no updates, no complexity
+    // INSERT OR IGNORE to create session, then backfill project if it was created empty
     // NOTE: memory_session_id starts as NULL. It is captured by SDKAgent from the first SDK
     // response and stored via updateMemorySessionId(). CRITICAL: memory_session_id must NEVER
     // equal contentSessionId - that would inject memory messages into the user's transcript!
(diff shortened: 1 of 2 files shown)


thedotmack:
Merged directly to main via rebase. Clean +18/-2 line fix that backfills the project field when a later hook provides it, fixing the race condition where PostToolUse creates the session before UserPromptSubmit sets the project. Build passes. Thank you @miclip!

Issue #957, opened 2026-02-05
SessionStart hook chain unnecessarily stops worker on every startup


rodboev:
## Bug Description

The SessionStart hook chain in `hooks/hooks.json` includes a `worker-service.cjs stop` step that shuts down and restarts the worker on every session start:

```
node smart-install.js && bun worker-service.cjs stop && bun worker-service.cjs hook claude-code context
```

This is unnecessary because the `hook claude-code context` command already calls `_q()` internally, which:
1. Checks if the worker is already running and healthy
2. Detects version mismatches and auto-restarts if needed
3. Starts the daemon if no worker is running

The unconditional `stop` adds ~15-30 seconds of delay on every startup (shutdown wait + port cleanup + restart + health check polling), even when the worker is already healthy and version-matched.

## Steps to Reproduce

1. Start Claude Code with claude-mem plugin enabled
2. Worker starts successfully on first session
3. Open a new session (or use `claude -p`)
4. Observe the SessionStart hook stops the healthy worker, waits for port cleanup, then restarts it

## Expected Behavior

The SessionStart hook should skip the `stop` step and let `_q()` handle version mismatch detection:

```
node smart-install.js && bun worker-service.cjs hook claude-code context
```

## Impact

- 15-30 seconds added to every session start
- On Windows, port cleanup via TIME_WAIT can extend this to 30+ seconds
- Combined with synchronous hook execution (#918), this contributes to multi-minute startup delays

## Environment

- Windows 11 Enterprise IoT LTSC (Build 26100)
- claude-mem v9.0.15
- Claude Code latest

## Related Issues

- #918 — SessionStart hooks block input in terminal on Windows
- #922 — PR to add async:true to SessionStart hooks  
- #923 — Slow Plugin Loading Time on Claude Startup
- #930 — Fix blocking startup by deferring worker initialization

rodboev:
**Update:** The `stop` step has already been replaced with `start` on the `main` branch (the current hooks.json uses separate `start` → `context` hooks instead of `stop && context`). However, this change is not yet included in the published releases — v9.0.16 (released 7 hours ago) still has the `stop &&` chain.

This issue can be closed once the next release ships with the updated hooks.json from `main`.

thedotmack:
Triage: The startup hook chain calls worker stop/start on every SessionStart, adding unnecessary latency. The hook should detect if the worker is already running and healthy, and skip the restart cycle.

thedotmack:
[...]

PR #960, opened 2026-02-05
fix: remove user-message hook from SessionStart to prevent startup error

rodboev:
## Summary

- Removes the `user-message` hook from SessionStart hooks
- Eliminates the "SessionStart:startup hook error" message on every session start

## Problem

The `user-message` hook handler exits with code 3 (`USER_MESSAGE_ONLY`), which is intentional for terminal-only output. However, Claude Code interprets any non-zero exit code as a failure and displays an error message.

**Exit code definitions in the codebase:**
```javascript
var d = {
  SUCCESS: 0,        // Hook ran successfully
  FAILURE: 1,        // Hook failed with error
  USER_MESSAGE_ONLY: 3  // Output to terminal only, don't inject
};
```

The `context` hook (exit code 0) already handles injecting context into the conversation, making the `user-message` hook redundant for SessionStart.

## Fix

Remove the `user-message` hook entry from the SessionStart hooks array. This:
1. Eliminates the spurious error message
2. Saves up to 60 seconds of potential timeout on slow systems
3. No behavior change — context injection still works via the `context` hook

## Files Changed

| File | Change |
|---|---|
| `plugin/hooks/hooks.json` | Remove `user-message` hook from SessionStart array |

## Test Plan

- [ ] Verify no "SessionStart:startup hook error" message on session start
- [ ] Verify context injection still works via the `context` hook
- [ ] Verify `user-message` hook still works in other contexts where it's used

## Related Issues

- Closes #905
- Related: #918 (SessionStart hooks block terminal)
- Related: #957 (unnecessary stop in hook chain)

### DIFF
--- a/src/shared/hook-constants.ts
+++ b/src/shared/hook-constants.ts
@@ -21,6 +21,8 @@ export const HOOK_EXIT_CODES = {
   FAILURE: 1,
   /** Blocking error - for SessionStart, shows stderr to user only */
   BLOCKING_ERROR: 2,
+  /** Show stderr to user only, don't inject into context. Used by user-message handler (Cursor). */
+  USER_MESSAGE_ONLY: 3,
 } as const;
 
 export function getTimeout(baseTimeout: number): number {


rodboev:
Confirmed, `USER_MESSAGE_ONLY` is missing from the type. The minimal fix is to add the constant back to hook-constants.ts. This slightly expands the scope of the PR but leaving a handler that returns undefined is worse. Fix incoming.

rodboev:
@greptileai review

Issue #982, opened 2026-02-06
Security: Path Traversal in /api/instructions endpoint (CWE-22)


NakayoshiUsagi:
### Summary

A **path traversal vulnerability** (CWE-22) exists in the `/api/instructions` endpoint that allows an attacker to read arbitrary files from the server filesystem via the `operation` query parameter.

Additionally, an **object injection vulnerability** (CWE-1321) exists via the `topic` query parameter that could potentially lead to prototype pollution.

### Severity

- **SEC-01 (Path Traversal)**: CVSS 7.5 (High)
- **SEC-02 (Object Injection)**: CVSS 5.3 (Medium)

### Affected Code

**File**: `src/services/server/Server.ts`

#### SEC-01: Path Traversal via `operation` parameter

The `operation` query parameter is used directly in `path.join()` and `fs.promises.readFile()` without validation:

```typescript
// Vulnerable code (approximate line numbers may vary by version)
const operationPath = path.join(__dirname, '../skills/mem-search/operations', `${operation}.md`);
content = await fs.promises.readFile(operationPath, 'utf-8');
```

An attacker can supply `?operation=../../../../etc/passwd` to read arbitrary files.

#### SEC-02: Object Injection via `topic` parameter

The `topic` query parameter is used as an object key without validation:

```typescript
return sections[topic] || sections['all'];
```

An attacker can supply `?topic=__proto__` or `?topic=constructor` to access prototype chain properties.

### Steps to Reproduce

1. Start claude-mem worker service
2. Send: `curl http://127.0.0.1:37777/api/instructions?operation=../../../../etc/passwd`
3. Observe: file contents are returned in the response

### Impact

- **Confidentiality**: Arbitrary file read on the host system
- **Scope**: Any process with network access to localhost:37777 (including browser extensions, malicious npm packages via SSRF)
- **Authentication**: None required (API is unauthenticated)

### Suggested Fix

**SEC-01**: Add whitelist validation for the `operation` parameter:

```typescript
const ALLOWED_OPERATIONS = ['search', 'context', 'summarize', 'import', 'export'];
if (operation && !ALLOWED_OPERATIONS.includes(operation)) {
  return res.status(400).json({ error: 'Invalid operation' });
}
```

For defense in depth, add a path boundary check:

```typescript
const operationPath = path.resolve(OPERATIONS_BASE_DIR, `${operation}.md`);
if (!operationPath.startsWith(OPERATIONS_BASE_DIR + path.sep)) {
[...]

PR #986, opened 2026-02-06
fix(security): validate and restrict /api/instructions operation and topic params (CWE-22, CWE-1321)

kamran-khalid-v9:
This PR addresses security vulnerabilities in the /api/instructions endpoint.\n- Validates 'operation' and 'topic' query parameters against whitelists.\n- Adds path boundary check to prevent path traversal.\n- Prevents object injection via topic.\nFixes: https://github.com/thedotmack/claude-mem/issues/982

### DIFF
--- a/src/services/server/Server.ts
+++ b/src/services/server/Server.ts
@@ -13,6 +13,7 @@ import express, { Request, Response, Application } from 'express';
 import http from 'http';
 import * as fs from 'fs';
 import path from 'path';
+import { ALLOWED_OPERATIONS, ALLOWED_TOPICS } from './allowed-constants.js';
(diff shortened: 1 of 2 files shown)


kamran-khalid-v9:
Heads up: The CI failed due to GitHub Actions OIDC token restrictions on forked repositories. The workflow permissions are correct, but OIDC tokens are not available to forks for security reasons. Once this PR is opened against the upstream repo, a maintainer can re-run the workflow and it should pass. See: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#usage-in-forks

Issue #1007, opened 2026-02-07
Critical: Daemon spawns hundreds of orphaned processes, consumes all system RAM and swap


bdteo:
## Environment
- macOS Sonoma, Apple Silicon, 32GB RAM
- Claude Code CLI (latest)
- claude-mem plugin (latest via claude-plugins-official)

## Problem

The `worker-service.cjs --daemon` process spawns Claude CLI subprocesses that are never cleaned up. Over the course of a few days of normal use (without restarting the machine), this leads to catastrophic resource consumption.

### What I observed after ~3 days of use:

- **265 orphaned Claude CLI processes** (children of the daemon)
- **~20GB RAM consumed** by these orphaned processes alone
- **52 × 1GB swap files** (52GB total swap on a 32GB RAM machine)
- **Disk filled to 100%**, which caused Redis to crash with `MISCONF Errors writing to the AOF file`
- The Redis crash cascaded into my Docker development environment going down entirely

### Process tree evidence

```
worker-service.cjs --daemon (PID 46783)
├── claude (orphaned subagent)
├── claude (orphaned subagent)
├── claude (orphaned subagent)
├── ... (265 total)
```

Each subprocess consumes ~75-80MB RAM and is never terminated. They accumulate indefinitely.

### Impact

This is not a minor resource leak. On a 32GB machine (which is a common developer spec), this **will** fill your RAM within 2-3 days of normal use, force macOS into extreme swap usage, and eventually fill your disk. If you run any disk-sensitive services (Redis, databases), they will crash.

Users with less RAM (16GB) will likely hit this even sooner. Realistically, only users with very large RAM (128GB+) who restart frequently would not notice.

### Workaround

Kill the daemon and all orphaned children:

```bash
# Find and kill the daemon
pkill -f "worker-service.cjs --daemon"

# Kill orphaned claude subprocesses
pkill -f "claude.*subagent"
```

Then disable the plugin and switch to an alternative memory solution.

### Related issues

This appears to be the same root cause as:
- #906
- #789
- #980
- #701
- #803

### Recommendation

Until the process lifecycle management is fundamentally fixed, users should be warned about this in the README. The daemon architecture needs bounded process pools, idle timeouts, and proper SIGCHLD handling to reap zombie children.

I've since migrated to AgentKits Memory (`@aitytech/agentkits-memory`) which uses a daemonless architecture with self-terminating workers, and the problem is gone.

jayvenn21:
[...]

Issue #906, opened 2026-02-03
Worker daemon spawns subagents that never terminate (resource leak)


evoleinik:
## Bug Description

The `worker-service.cjs --daemon` process spawns Claude subagents that never terminate, leading to accumulated orphaned processes over time.

## Environment

- **Plugin version**: 9.0.10
- **Platform**: Linux (Ubuntu)
- **Claude Code version**: Latest

## Reproduction

After using claude-mem features for a few hours, orphaned subagent processes accumulate:

```
$ ps -eo pid,ppid,cmd | grep "2737781.*claude" | wc -l
125
```

All orphaned processes share:
- Parent: `worker-service.cjs --daemon` (PID 2737781 in my case)
- Model: `claude-sonnet-4-5`
- Flags: `--resume <session-uuid> --disallowedTools Bash,Read,Write,Edit,Grep,Glob,WebFetch,WebSearch,Task,NotebookEdit,AskUserQuestion,TodoWrite`

## Diagnostic Output

```
$ ps -p 2737781 -o pid,ppid,etime,cmd
    PID    PPID     ELAPSED CMD
2737781       1  1-05:30:17 /home/eo/.bun/bin/bun ...worker-service.cjs --daemon

$ ps -eo cmd | grep "2737781.*claude" | sed 's/.*claude/claude/' | sort | uniq -c | sort -rn
     40 claude-sonnet-4-5 --resume e4f73a95-... --disallowedTools ...
     22 claude-sonnet-4-5 --resume 24b12d58-... --disallowedTools ...
     21 claude-sonnet-4-5 --resume 6f1d1c75-... --disallowedTools ...
     20 claude-sonnet-4-5 --resume f0113236-... --disallowedTools ...
      8 claude-sonnet-4-5 --resume 0764d06c-... --disallowedTools ...
      7 claude-sonnet-4-5 --resume 42bd571e-... --disallowedTools ...
      7 claude-sonnet-4-5 --disallowedTools ... (no resume)
```

Process ages range from ~1 hour to 12+ hours - they never exit.

## Expected Behavior

Subagents spawned by the worker daemon should terminate after completing their task.

## Workaround

Kill orphaned processes manually:
```bash
pkill -P <daemon_pid> -f claude
```

Or restart the daemon entirely.

suvendukungfu:
assign this issue

leduclinh7141:
[...]

Issue #803, opened 2026-01-25
Bug: Worker Service Accumulates Claude Processes Without Limit (13GB+ Memory Leak)


oliveagle:
# Bug: Worker Service Accumulates Claude Processes Without Limit (13GB+ Memory Leak)

## Summary

The claude-mem worker service continuously spawns Claude subprocesses without cleanup, causing unbounded process growth and massive memory waste.

**Observed Impact:**
- Expected: 2-4 worker processes
- Actual: 34+ processes accumulated
- Memory waste: 34 × 400MB = **13.6GB**
- Some processes running for 24+ hours

## Environment

- **OS**: Linux Ubuntu
- **claude-mem version**: 9.0.6
- **Runtime**: Bun
- **Component**: `worker-service.cjs`

## Root Cause

The worker pool has **no size limit** and doesn't reuse or clean up workers, causing continuous accumulation.

## Evidence

```bash
# 34 processes, all from the same parent
$ ps aux | grep 'bin/claude' | grep -v grep | wc -l
34

$ ps aux | grep 'bin/claude' | grep -v grep | awk '{print $2}' | \
    xargs -I {} ps -o pid,ppid -p {} 2>/dev/null | awk '{print $2}' | sort | uniq
39988  # All children of worker service

# Memory usage
$ ps aux | grep 'bin/claude' | awk '{sum+=$6} END {print sum/1024 " MB"}'
~13600 MB (13.6 GB)

# Timeline shows accumulation
$ ps -o pid,lstart -p $(pgrep -f "bin/claude" | head -5)
  PID                  STARTED
 40017  Sun Jan 25 21:20:03 2026
 41141  Sun Jan 25 21:20:30 2026
 67859  Sun Jan 25 21:25:27 2026
 71140  Sun Jan 25 21:27:26 2026
 ...

# After cleanup
$ kill -TERM 39988  # Kill worker service
$ ps aux | grep 'bin/claude' | wc -l
0  # ✅ Successfully cleaned up
```

### Process Tree

```
PID 39988: bun worker-service.cjs --daemon
├── PID 40017: claude --resume ... --disallowedTools [all tools disabled]
├── PID 41141: claude --resume ... --disallowedTools [all tools disabled]
├── PID 67859: claude --resume ... --disallowedTools [all tools disabled]
└── ... (34 total)
```

All worker processes have all tools disabled via `--disallowedTools` and communicate via IPC socket.

## Expected Behavior

Worker pool should:
[...]

Issue #789, opened 2026-01-23
Memory Leak: worker-service daemon causes 50+ GB memory consumption


ktytagong:
## Bug Description

The `worker-service.cjs --daemon` process causes severe memory accumulation over extended usage, leading to system memory exhaustion (52+ GB observed).

## Environment

- **OS**: macOS Darwin 25.2.0 (Apple Silicon)
- **claude-mem version**: 9.0.6
- **Bun version**: 1.3.5
- **Claude Code version**: 2.1.17
- **Node version**: v24.4.1

## Steps to Reproduce

1. Install claude-mem plugin via `/plugin` command
2. Use Claude Code extensively with `--dangerously-skip-permissions` mode
3. Run multiple sessions over several hours
4. Use Task tool (subagents) frequently
5. Observe system memory increasing until "Your system has run out of application memory" dialog appears

## Observed Behavior

After extended usage (~989 sessions over time), the following processes were consuming excessive memory:

```
PID    COMMAND                    MEM
29139  claude 2.1.17              16 GB
29966  claude 2.1.17 (subagent)   8.4 GB
29294  bun worker-service.cjs    56 MB (but spawns many children)
```

The worker-service daemon:
- Runs continuously as `bun /Users/.../.claude/plugins/cache/thedotmack/claude-mem/9.0.6/scripts/worker-service.cjs --daemon`
- Spawns child processes on every hook invocation
- Accumulates state/memory without releasing

From `ps aux`:
```
kt  29294  /opt/homebrew/Cellar/bun/1.3.5/bin/bun .../worker-service.cjs --daemon
kt  29301  node .../mcp-server.cjs
kt  29189  node .../mcp-server.cjs  (duplicate instance)
kt  31955  bun .../worker-service.cjs hook claude-code observation
kt  31954  bun .../worker-service.cjs start
```

## Expected Behavior

- Worker service should have bounded memory usage
- Old observation data should be garbage collected
- Daemon should not spawn unbounded child processes

## Potential Causes

1. **Unbounded observation storage**: Each tool call creates an observation that may not be cleaned up
2. **Multiple MCP server instances**: Saw 2+ `mcp-server.cjs` processes running
3. **Hook process accumulation**: Each hook spawns a new Bun process
4. **Large bundled script**: `worker-service.cjs` is 832KB which loads into memory repeatedly

## Workaround

Periodically kill the daemon and restart Claude Code:
```bash
pkill -9 -f "worker-service"
[...]

PR #1008, opened 2026-02-07
fix: prevent unbounded Claude subprocess spawning in worker daemon

jayvenn21:
## Summary

This PR introduces defensive process lifecycle controls to prevent the
`worker-service.cjs --daemon` from spawning unbounded Claude CLI subprocesses
that are never cleaned up, which can lead to catastrophic memory, swap, and disk
exhaustion over time.

This change does **not** alter core functionality or user-facing behavior under
normal workloads. It adds guardrails and cleanup to ensure long-running daemon
sessions remain stable.

## Problem

On macOS (and potentially other platforms), the daemon currently:
- spawns Claude CLI subprocesses without a hard upper bound
- does not reliably terminate or reap completed / idle children
- allows orphaned subprocesses to accumulate indefinitely

Over multi-day usage, this results in:
- hundreds of orphaned `claude` processes
- tens of GB of RAM + swap consumption
- disk exhaustion and cascading service failures (e.g. Redis, Docker)

Multiple related issues report the same failure mode.

## Changes

This PR introduces **minimal but critical safety mechanisms**:

- Track spawned Claude subprocesses explicitly
- Enforce a maximum concurrent subprocess limit
- Ensure child processes are terminated and removed on task completion
- Add defensive cleanup on daemon startup to reap stale children
- Log warnings when limits are hit instead of silently spawning indefinitely

These changes are intentionally conservative and avoid architectural rewrites.

## Why this approach

This is a production-safety fix:
- prevents resource exhaustion
- avoids breaking existing workflows
- buys time for a more comprehensive worker-pool redesign if desired

The goal is to **fail safely** rather than catastrophically.

## Testing

- Verified subprocess cleanup under repeated task execution
- Verified daemon remains stable over extended runtime
- Confirmed no behavior change for normal interactive usage

## Related issues

- #1007
- #906
- #803
- #789
- #701

### DIFF
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -70,7 +70,7 @@ import { SettingsRoutes } from './worker/http/routes/SettingsRoutes.js';
 import { LogsRoutes } from './worker/http/routes/LogsRoutes.js';
 
 // Process management for zombie cleanup (Issue #737)
-import { startOrphanReaper, reapOrphanedProcesses } from './worker/ProcessRegistry.js';
+import { startOrphanReaper, reapSystemOrphansOnly } from './worker/ProcessRegistry.js';
 
 /**
  * Build JSON status output for hook framework communication.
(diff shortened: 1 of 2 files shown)


jayvenn21:
It looks like the failing `claude-review` check is due to missing OIDC
permissions rather than a code issue.

The error indicates the GitHub Action cannot fetch an ID token
(`ACTIONS_ID_TOKEN_REQUEST_URL` is not set), which commonly happens on
forked PRs unless the workflow explicitly enables:

permissions:
  id-token: write

Since this is a contributor PR from a fork, the action may need to be
disabled for forks, made non-blocking, or the permissions adjusted by a
maintainer.

Happy to rebase or rerun once that’s resolved.

JGrubb:
See here also - https://github.com/thedotmack/claude-mem/issues/1010

jayvenn21:
> See here also - #1010

Sounds good. I'll take a look at it.

thedotmack:
Closing in favor of #1085 (rodboev) which provides a more comprehensive solution to the same problem. Thank you for the contribution!

PR #1019, opened 2026-02-08
feat: support system environment variables for API credentials

### LINKED ISSUE #1015: Bug: SDK Agent fails on Windows when username contains spaces
## Bug Report

  **Summary:** Claude SDK Agent fails to start on Windows when the user's path contains spaces (e.g., `C:\Users\Alan Walker\`), causing PostToolUse hooks to hang indefinitely.

  **Severity:** High - Core functionality broken

  **Affected Platform:** Windows only

  ---

  ## Symptoms

  PostToolUse hook displays `(1/2 done)` indefinitely. Worker logs show:

  ERROR [SESSION] Generator failed {provider=claude, error=Claude Code process exited with code 1}
  ERROR [SESSION] Generator exited unexpectedly

  ---

  ## Root Cause

  1. `SDKAgent.ts` returns full auto-detected path with spaces
  2. `ProcessRegistry.ts` cannot execute .cmd files when path contains spaces

  ---

  ## Proposed Fix

  **`src/services/worker/SDKAgent.ts`:** On Windows, prefer `claude.cmd` via PATH instead of full path.

  **`src/services/worker/ProcessRegistry.ts`:** Use `cmd.exe /d /c` wrapper for .cmd files.

  ---

  ## Testing

  Verified on Windows 11 with username containing spaces:
  - PostToolUse hook completes successfully
  - Observations are stored to database
  - No more "process exited with code 1" errors

  ---

ixnaswang:
## What
Add fallback to system environment variables when `~/.claude-mem/.env` doesn't contain credentials.

## Why
- Users with system-level API credential configuration (e.g., ANTHROPIC_BASE_URL for third-party APIs)
- Support for ANTHROPIC_AUTH_TOKEN naming (used by Zhipu AI)
- No need to duplicate configuration in .env file

## Changes
- Add ANTHROPIC_BASE_URL support for custom API endpoints
- Implement layered credential resolution: .env file → system env
- Apply same logic to GEMINI_API_KEY and OPENROUTER_API_KEY

Fixes #1015

### DIFF
--- a/docs/bug-fixes/system-env-fallback-for-credentials.md
+++ b/docs/bug-fixes/system-env-fallback-for-credentials.md
@@ -1,0 +1,108 @@
+# System Environment Variable Fallback for API Credentials
+
+**Date:** 2026-02-08
+**Status:** Fixed
+**Files Modified:** `src/shared/EnvManager.ts`
+
+## Problem
+
+Users who have API credentials configured at the system level (e.g., via Windows environment variables) could not use claude-mem without manually creating a `~/.claude-mem/.env` file. This was particularly problematic for:
+
+1. **Users of third-party Anthropic-compatible APIs** (e.g., Zhipu AI, OpenRouter) who configure `ANTHROPIC_BASE_URL` at system level
+2. **Users with `ANTHROPIC_AUTH_TOKEN`** instead of `ANTHROPIC_API_KEY` (some providers use different naming)
+3. **Workflow inconvenience** - requiring duplicate configuration in both system env vars and claude-mem's .env file
+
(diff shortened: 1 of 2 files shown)


thedotmack:
Thank you for this contribution! The underlying issue (env var credential resolution) has already been addressed by the current EnvManager.ts design, which uses a blocklist approach that passes through ANTHROPIC_AUTH_TOKEN and ANTHROPIC_BASE_URL to subprocess environments by default. The PR was written against an older allowlist-based EnvManager that has since been refactored. Closing as the problem is already solved.

PR #1021, opened 2026-02-08
fix: SDK Agent fails on Windows when username contains spaces

### LINKED ISSUE #1014: Support system environment variables as fallback for API credentials
Users who configure API credentials at the system level (via environment variables) cannot use claude-mem without manually creating a `~/.claude-mem/.env` file.

  This is particularly problematic for users of third-party Anthropic-compatible APIs (z.AI, OpenRouter, etc.) who configure `ANTHROPIC_BASE_URL` at system level.

  ## Proposed Solution

  Implement **layered credential resolution**:
  1. First: `~/.claude-mem/.env` (highest priority)
  2. Second: System environment variables (fallback)

  This would enable users to configure credentials like:
  ```bash
  export ANTHROPIC_BASE_URL=https://open.bigmodel.cn/api/anthropic
  export ANTHROPIC_AUTH_TOKEN=your_token

  Benefits

    - Support for third-party Anthropic-compatible APIs
    - Support for ANTHROPIC_AUTH_TOKEN naming
    - No breaking changes - .env files still take priority

ixnaswang:
## Summary
- Fixes SDK Agent spawn failure on Windows when username contains spaces
- Prefers `claude.cmd` via PATH instead of full auto-detected path
- Uses `cmd.exe /d /c` wrapper for proper .cmd file execution

## Root Cause
Node.js `spawn()` cannot directly execute `.cmd` files when the path contains spaces (e.g., `C:\Users\Anderson Wang\...`)

## Changes
- `src/services/worker/SDKAgent.ts`: Return `claude.cmd` on Windows to use PATH resolution
- `src/services/worker/ProcessRegistry.ts`: Use cmd.exe wrapper for .cmd files

Fixes #1014

### DIFF
--- a/docs/bug-fixes/windows-spaces-issue.md
+++ b/docs/bug-fixes/windows-spaces-issue.md
@@ -1,0 +1,98 @@
+---
+Title: Bug: SDK Agent fails on Windows when username contains spaces
+---
+
+## Bug Report
+
+**Summary:** Claude SDK Agent fails to start on Windows when the user's path contains spaces (e.g., `C:\Users\Anderson Wang\`), causing PostToolUse hooks to hang indefinitely.
+
+**Severity:** High - Core functionality broken
+
+**Affected Platform:** Windows only
+
+---
+
+## Symptoms
+
+PostToolUse hook displays `(1/2 done)` indefinitely. Worker logs show:
+
+```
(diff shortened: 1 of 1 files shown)


ixnaswang:
Recreating with correct changes

PR #1031, opened 2026-02-08
fix: resolve marketplace root path dynamically for XDG-compliant environments

### LINKED ISSUE #1030: smart-install.js hardcodes ~/.claude/plugins/ path, fails on XDG-based config (~/.config/claude/plugins/)
## Bug Description

`smart-install.js` hardcodes the plugin root path as `~/.claude/plugins/marketplaces/thedotmack` (line 13):

```js
const ROOT = join(homedir(), '.claude', 'plugins', 'marketplaces', 'thedotmack');
```

On environments where Claude Code uses `~/.config/claude/plugins/` (XDG-compliant setups, Nix-managed installations, etc.), this path does not exist, causing an ENOENT error on every session start.

## Error Message

```
Installation failed: ENOENT: no such file or directory, open '/Users/<user>/.claude/plugins/marketplaces/thedotmack/package.json'
```

This appears in debug logs as:

```
Hook output does not start with {, treating as plain text
Hook SessionStart:startup (SessionStart) error:
```

## Impact

- **Low** — The other two SessionStart hooks (worker start and context hook) succeed, so the plugin functions normally.
- However, the error is logged on every session start, which is confusing.

## Expected Behavior

`smart-install.js` should resolve the plugin root path dynamically (e.g., via `CLAUDE_PLUGIN_ROOT` environment variable, which is already used in `hooks.json` for other hooks) instead of hardcoding `~/.claude/plugins/`.

## Environment

- claude-mem: v9.1.1
- Claude Code: v2.1.37 (Nix-managed)
- OS: macOS (Apple Silicon)
- Plugin installed at: `~/.config/claude/plugins/cache/thedotmack/claude-mem/9.1.1/`
- Marketplace dir: `~/.config/claude/plugins/marketplaces/thedotmack/`

okash1n:
## Summary

- Fixes hardcoded `~/.claude/plugins/marketplaces/thedotmack` path in `smart-install.js` that fails on XDG-compliant setups (e.g. Nix-managed Claude Code) where plugins are stored under `~/.config/claude/plugins/`
- Now resolves the path dynamically: derives from `CLAUDE_PLUGIN_ROOT` env var, probes XDG path, then falls back to legacy path
- Backward-compatible — legacy path still works as the final fallback

Fixes #1030

## Test plan

- [ ] Verify on standard Claude Code install (`~/.claude/plugins/`) — should still work via fallback
- [ ] Verify on XDG-based install (`~/.config/claude/plugins/`) — should now resolve correctly
- [ ] Verify `CLAUDE_PLUGIN_ROOT` derivation finds the marketplace dir from cache path

### DIFF
--- a/scripts/smart-install.js
+++ b/scripts/smart-install.js
@@ -10,9 +10,39 @@ import { execSync, spawnSync } from 'child_process';
 import { join } from 'path';
 import { homedir } from 'os';
 
-const ROOT = join(homedir(), '.claude', 'plugins', 'marketplaces', 'thedotmack');
-const MARKER = join(ROOT, '.install-version');
 const IS_WINDOWS = process.platform === 'win32';
+
+/**
+ * Resolve the marketplace root directory.
+ *
+ * Claude Code may store plugins under either `~/.claude/plugins/` (legacy) or
+ * `~/.config/claude/plugins/` (XDG-compliant, e.g. Nix-managed installs).
+ * When `CLAUDE_PLUGIN_ROOT` is set we derive the base from it; otherwise we
+ * probe both candidate paths and fall back to the legacy location.
+ */
+function resolveRoot() {
+  const marketplaceRel = join('plugins', 'marketplaces', 'thedotmack');
+
+  // Derive from CLAUDE_PLUGIN_ROOT (e.g. .../plugins/cache/thedotmack/claude-mem/<ver>)
+  if (process.env.CLAUDE_PLUGIN_ROOT) {
+    let dir = process.env.CLAUDE_PLUGIN_ROOT;
+    const cacheIndex = dir.indexOf(join('plugins', 'cache'));
+    if (cacheIndex !== -1) {
+      const base = dir.substring(0, cacheIndex);
+      const candidate = join(base, marketplaceRel);
+      if (existsSync(join(candidate, 'package.json'))) return candidate;
+    }
+  }
+
+  // Probe XDG path first, then legacy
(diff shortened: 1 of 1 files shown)


okash1n:
@greptile-apps Thanks for the review! However, I believe the "extra `plugins/` segment" concern is a false positive. Here's a trace of the actual path construction:

**`marketplaceRel`** is defined as:
```js
join('plugins', 'marketplaces', 'thedotmack')
// => "plugins/marketplaces/thedotmack"
```

**CLAUDE_PLUGIN_ROOT derivation:**
```
CLAUDE_PLUGIN_ROOT = "~/.config/claude/plugins/cache/thedotmack/claude-mem/9.1.1"
cacheIndex         = index of "plugins/cache" in the path
base               = "~/.config/claude/"  (everything before "plugins/cache")
candidate          = base + marketplaceRel
                   = "~/.config/claude/" + "plugins/marketplaces/thedotmack"
                   = "~/.config/claude/plugins/marketplaces/thedotmack"  ✓
```

**XDG probe:**
```
join(homedir, '.config', 'claude', marketplaceRel)
= "~/.config/claude/" + "plugins/marketplaces/thedotmack"
= "~/.config/claude/plugins/marketplaces/thedotmack"  ✓
```

**Legacy fallback:**
```
join(homedir, '.claude', marketplaceRel)
= "~/.claude/" + "plugins/marketplaces/thedotmack"
= "~/.claude/plugins/marketplaces/thedotmack"  ✓
```

All three paths resolve correctly with no duplicated `plugins/` segment. I've also verified this with a Node.js script on the actual environment where the bug occurs.

Issue #1042, opened 2026-02-09
ENOENT error in stop hooks: race condition reading package.json during shutdown


guzman109:
### Environment
- claude-mem version: 9.1.1
- Runtime: Bun v1.3.9 (macOS arm64)
- Platform: macOS Darwin 25.2.0

### Description

The stop hooks (`session-complete` and `summarize`) throw an `ENOENT` error when trying to read `package.json` during session shutdown. The file exists and is readable, suggesting a race condition where the Bun runtime is partially torn down before the worker service finishes its cleanup.

### Error

```
ENOENT: no such file or directory, open '/Users/<user>/.claude/plugins/marketplaces/thedotmack/package.json'
    path: "/Users/<user>/.claude/plugins/marketplaces/thedotmack/package.json",
 syscall: "open",
   errno: -2,
    code: "ENOENT"

      at xW (worker-service.cjs:877:2709)
      at VC (worker-service.cjs:877:2971)
      at Zq (worker-service.cjs:1427:1591)
      at async cge (worker-service.cjs:1427:3779)
```

### Steps to Reproduce

1. Install claude-mem plugin
2. Start a Claude Code session
3. End the session (the stop hooks fire)
4. Observe the ENOENT error in the terminal output

### Notes

- The file **does** exist at the expected path and has correct permissions (`-rw-r--r--`)
- The error is reported as non-blocking but is visible to the user on every session end
- All three stop hooks (`start`, `summarize`, `session-complete`) fire; the latter two fail with this error

berk-karaal:
**Experiencing the same issue on Linux (Fedora 42).**

### Environment
- **claude-mem:** 9.1.1
- **Claude Code:** 2.1.38
- **Runtime:** Bun v1.3.5 (Linux x64)
- **Platform:** Fedora 42, Linux 6.17.10-200.fc42.x86_64

### Observations

Reproduces on every session end — all 3 stop hooks fail with the same ENOENT + Zod cyclical schema error. Like the original report, the file **does exist** at `/home/berk/.claude/plugins/marketplaces/thedotmack/package.json` and is readable.

A few additional data points that might help narrow this down:

1. **Cross-platform:** The original report is on macOS/arm64 with Bun 1.3.9; I'm hitting the same on Linux/x64 with Bun 1.3.5. So this isn't platform-specific.
[...]

Issue #987, opened 2026-02-06
Stop hook can cause infinite session loop when summarize output is interpreted as instructions


costa-marcello:
## Bug Description

The claude-mem Stop hook can create an infinite feedback loop where Claude Code never terminates a session. The session keeps "bouncing back to work" instead of ending cleanly.

## Root Cause Analysis

The Stop hook runs three sequential commands:
1. Worker start
2. `hook claude-code summarize` (120s timeout)
3. `hook claude-code session-complete` (30s timeout)

When the `summarize` hook returns a `systemMessage` containing session context (e.g., "remaining work", "next steps", incomplete tasks), Claude Code interprets this as new instructions to continue working. This triggers another round of work, which then triggers another Stop event, creating an unbounded loop.

### The Feedback Loop

```
Claude tries to stop
  → claude-mem Stop hook fires
  → summarize returns systemMessage with session context
  → Claude interprets context as "continue working" instruction
  → Claude does more work
  → Claude tries to stop again
  → loop repeats
```

### Amplifying Factor

This is especially problematic when used alongside GSD workflow patterns (or any plugin that structures work into phases with "next steps" suggestions), because the summarize output includes those suggestions, making Claude even more likely to interpret the summary as actionable instructions.

## Environment

- Claude Code CLI
- claude-mem v9.0.17
- Also running: hookify (claude-plugins-official), GSD workflow plugin
- macOS (Darwin 25.2.0)

## Expected Behavior

The Stop hook should not return content in `systemMessage` that could be interpreted as instructions to continue working. The session should terminate cleanly.

## Suggested Fixes

1. **Don't include actionable content in `systemMessage` during Stop events.** The summarize output should be purely informational and clearly marked as a post-session record, not directives.

2. **Add a loop-detection guard** — if the Stop hook fires more than N times in quick succession (e.g., 3 times within 60 seconds), skip the summarize step and let the session end.

3. **Use `"decision": "allow"` explicitly** in the Stop hook response to signal that the session should proceed to termination, rather than relying on implicit behavior.

4. **Reduce the summarize timeout** from 120s to something more reasonable (e.g., 15-30s) to limit the blast radius of any loop.

## Reproduction

Hard to reproduce deterministically, but the conditions are:
1. Run a multi-step workflow session (e.g., GSD execute-phase)
2. Session completes and Claude tries to stop
3. claude-mem summarize returns content mentioning "next steps" or "remaining work"
4. Claude interprets this as new instructions and continues working
[...]

Issue #984, opened 2026-02-06
Stop hook fails with "Unknown event type: session-complete"


opsbl:
# Stop hook fails with "Unknown event type: session-complete"

## Problem

The `session-complete` stop hook is failing during Claude Code session cleanup. When Claude Code terminates a session, the worker service reports that `session-complete` is an unknown event type, causing the hook to fail execution.

## Environment

- **Claude-mem**: 9.0.17
- **Claude Code**: 2.1.25
- **Node.js**: v24.13.0
- **Bun**: 1.3.6
- **OS**: Linux 6.6.87.2-microsoft-standard-WSL2 (x64)
- **Platform**: linux
- **Plugin Path**: `~/.claude/plugins/marketplaces/thedotmack`
- **Data Directory**: `~/.claude-mem`
- **Worker Port**: 37777

## Steps to Reproduce

1. Start a Claude Code session with claude-mem plugin active
2. Complete or terminate the Claude Code session
3. Observe stop hooks execution during cleanup

## Expected Behavior

The `session-complete` hook should execute successfully when a Claude Code session ends, allowing proper cleanup and session finalization.

## Actual Behavior

The hook fails with an error indicating that `session-complete` is not a recognized event type in the worker service.

## Error Log

```
● Ran 3 stop hooks
  ⎿  command: node "${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js" "${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" start
  ⎿  command: node "${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js" "${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" hook claude-code summarize
  ⎿  command: node "${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js" "${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" hook claude-code session-complete
  ⎿  Stop hook error: [node "${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js" "${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" hook claude-code session-complete]: Hook error: Error: Unknown event type:
  session-complete
```

## Additional Context

- The worker service is running normally (uptime: 11 minutes, 1 active session, 4 SSE clients)
- Other hooks (`start` and `summarize`) appear to execute without errors
[...]

Issue #923, opened 2026-02-04
Slow Plugin Loading Time on Claude Startup


ruslanlap:
# Slow Plugin Loading Time on Claude Startup

## Problem Statement

The claude-mem plugin exhibits significant slow loading performance when launching Claude from the terminal. The plugin takes approximately 1–2 minutes to fully load, which is unacceptably slow and impacts the user experience. Notably, the database file does not exist and the background worker is not running, which may be contributing to or compounding the startup delay.

---

## Environment

| Component | Version / Details |
|---|---|
| claude-mem | 9.0.12 |
| Claude Code | 2.1.31 |
| Node.js | v22.14.0 |
| Bun | 1.3.8 |
| OS | Linux 6.6.114.1-microsoft-standard-WSL2 (x64) |
| Platform | Linux (WSL2) |
| Dev Mode | No |

---

## Steps to Reproduce

1. Open a terminal within WSL2.
2. Run `claude` to launch Claude Code.
3. Observe the Claude screen after **1–2 minutes**.
4. Note that the plugin loading is noticeably slow compared to expected near-instant startup.

---

## Expected Behavior

The claude-mem plugin should load quickly on Claude startup with minimal delay, allowing the user to begin interacting with Claude as fast as possible.

## Actual Behavior

The plugin takes approximately 1–2 minutes to load after launching Claude. The startup experience is significantly slower than expected.

---

## Diagnostic Details

The following diagnostic state was captured and may be relevant to the root cause:

**Worker Status:**
```
Running: No
```

**Database:**
```
Path:   ~/.claude-mem/claude-mem.db
Exists: No
```

**Plugin Path:**
```
~/.claude/plugins/marketplaces/thedotmack
```

**Key Configuration:**
```ini
CLAUDE_MEM_MODEL=claude-sonnet-4-5
CLAUDE_MEM_WORKER_PORT=37777
CLAUDE_MEM_WORKER_HOST=127.0.0.1
CLAUDE_MEM_LOG_LEVEL=INFO
CLAUDE_MEM_CONTEXT_OBSERVATIONS=50
```

---

## Additional Context

- The **worker is not running** and the **database file does not exist**. It is possible that the plugin is hanging or retrying during startup while attempting to initialize the database or connect to the worker, which could explain the 1–2 minute delay.
- The environment is **WSL2 on Linux**, which can sometimes introduce filesystem or networking latency that may exacerbate startup issues.
[...]

PR #1056, opened 2026-02-10
fix: hook resilience and worker lifecycle — 87% faster recovery from dead worker

rodboev:
## Summary

- **87% faster startup when worker is dead** — worst-case recovery drops from ~60s to ~8s by reducing health check timeout (30s→3s), post-spawn wait (30s→5s), and port-in-use wait (15s→3s), plus proactive stale PID cleanup
- **Graceful degradation across all hooks** — worker unavailability now returns exit 0 instead of BLOCKING_ERROR, preventing error dialogs during sleep/hibernate recovery
- **Error classifier** distinguishes transport/5xx (graceful) from 4xx/programming errors (blocking), ensuring real bugs still surface while transient failures degrade silently
- **Fix daemon silent death** — SIGHUP handler + setsid + unhandled error guards prevent the worker daemon from dying silently after parent hook process exits

## Performance Metrics

| Scenario | Before | After | Improvement |
|----------|--------|-------|-------------|
| Worker dead, stale PID file | ~60s (30s health timeout + 30s spawn wait) | ~8s (instant PID cleanup + 3s check + 5s spawn) | **87% faster** |
| Worker dead, no PID file | ~60s (30s timeout + 30s wait) | ~8s (3s timeout + 5s wait) | **87% faster** |
| Port occupied, worker unresponsive | ~15s wait | ~3s wait | **80% faster** |
| Worker already healthy | ~1s | ~1s | No change needed |
| Stop hook with dead worker | BLOCKING_ERROR visible to user | Silent graceful exit | **100% fewer visible errors** |
| Daemon dies from SIGHUP after spawn | Always (default signal = terminate) | Never (SIGHUP ignored in daemon mode) | **Root cause fix** |

## Issues Closed

| Issue | Title | Fix |
|-------|-------|-----|
| #957 | SessionStart hook chain unnecessarily slow | Reduced `ensureWorkerStarted()` timeouts from 30s→3-5s; stale PID cleanup avoids waiting for dead processes |
| #923 | Slow Plugin Loading Time on Claude Startup | Root cause was the 30s HEALTH_CHECK and 30s POST_SPAWN_WAIT constants; now 3s and 5s respectively |
| #984 | Stop hook fails with "Unknown event type: session-complete" | `getEventHandler()` returns no-op handler for unknown events instead of throwing |
| #987 | Stop hook causes infinite session loop on summarize failure | Error classifier treats summarize HTTP failures as graceful (exit 0), preventing BLOCKING_ERROR from retriggering |
| #1042 | ENOENT error in stop hooks: race condition reading package.json | `getPluginVersion()` wrapped in try-catch, returns `'unknown'` on ENOENT/EBUSY; version check skipped when unknown |

## Relationship to Open PRs

- **#1055** (worker idle auto-shutdown): Complementary, not overlapping. That PR adds idle timeout to shut down healthy workers after inactivity; this PR fixes recovery when workers are already dead. Both use `ensureWorkerStarted()` — our timeout reductions make their auto-restart path faster too.
- **#1008** (prevent unbounded subprocess spawning): Different problem scope. That PR addresses orphaned Claude CLI subprocesses; this PR addresses the worker daemon's own lifecycle and hook error handling. No file conflicts.

### Core infrastructure
| File | Change |
|------|--------|
| `src/shared/hook-constants.ts` | HEALTH_CHECK 30s→3s, add POST_SPAWN_WAIT (5s), PORT_IN_USE_WAIT (3s) |
| `src/services/infrastructure/ProcessManager.ts` | Add `isProcessAlive()` (EPERM-aware), `cleanStalePidFile()`, `spawnDaemon()` uses `setsid` on Linux |
| `src/shared/worker-utils.ts` | Wrap `getPluginVersion()` in try-catch (#1042), add `CLAUDE_MEM_HEALTH_TIMEOUT_MS` env override |
| `src/services/worker-service.ts` | Call `cleanStalePidFile()` at top of `ensureWorkerStarted()`, use named constants, SIGHUP handler (ignore in daemon mode), unhandled error guards in daemon mode |
| `src/cli/hook-command.ts` | Add `isWorkerUnavailableError()` classifier (transport/5xx→exit 0, 4xx→exit 2) |

### Handler graceful degradation
| File | Change |
|------|--------|
| `src/cli/handlers/index.ts` | Return no-op handler for unknown event types (#984) |
| `src/cli/handlers/context.ts` | Wrap fetch in try-catch, return empty context on failure |
| `src/cli/handlers/observation.ts` | Return graceful on `!response.ok` instead of throw |
| `src/cli/handlers/file-edit.ts` | Same pattern as observation |
| `src/cli/handlers/user-message.ts` | Add worker check + wrap fetch in try-catch |
| `src/cli/handlers/session-complete.ts` | Handle `ensureWorkerRunning()` return value |

### Tests
| File | Change |
|------|--------|
| `tests/hook-constants.test.ts` | Updated assertions for new constant values |
| `tests/infrastructure/process-manager.test.ts` | Tests for `isProcessAlive()`, `cleanStalePidFile()`, EPERM handling, `spawnDaemon` setsid, SIGHUP listener |
| `tests/hook-command.test.ts` (new) | 20 tests for error classifier — transport vs 4xx vs 5xx vs programming errors |

### Built bundles
| File | Change |
|------|--------|
| `plugin/scripts/worker-service.cjs` | Rebuilt |
| `plugin/scripts/mcp-server.cjs` | Rebuilt |

## Design Decisions

- **`ensureWorkerStarted()` kept in hook command**: Peer review (Gemini + Codex) confirmed this is the only code path that starts the worker — removing it would break auto-recovery. The reduced timeouts (5s max vs 30s) eliminate the double-penalty concern.
- **Two platform multiplier functions preserved**: `getTimeout()` (1.5x) for hook-side fast path, `getPlatformTimeout()` (2.0x) for worker-side socket operations — clarifying comment added.
- **Conservative error default**: Unknown errors map to exit 2 (blocking) rather than exit 0, so new bug categories surface rather than silently failing.
- **Env var override**: `CLAUDE_MEM_HEALTH_TIMEOUT_MS` lets users on unusually slow systems increase the health check timeout without code changes (validated: 500ms–300000ms range).
- **Belt-and-suspenders for daemon survival**: Three layers prevent daemon death — (1) SIGHUP handler ignores the signal in daemon mode, (2) `setsid` creates a new session so SIGHUP is never delivered, (3) `uncaughtException`/`unhandledRejection` handlers prevent silent crashes from any source.

## Test plan

- [x] `bun test` — 935 pass, 3 skip (1 pre-existing timeout in openclaw SSE test, unrelated)
- [x] `npm run build` — builds successfully
- [x] `npm run build-and-sync` — synced to marketplace, worker restarted
- [x] Manual: kill worker, start new session → recovers in <5s (observed 1.6s to full recovery)
- [x] Manual: start session with worker already running → no delay
- [x] Manual: `/exit` session → stop hooks complete without BLOCKING_ERROR
- [ ] Manual: `CLAUDE_MEM_HEALTH_TIMEOUT_MS=10000 claude` → verify override works
- [ ] Manual: `bun worker-service.cjs start` → daemon survives parent exit (port stays bound)
- [ ] Manual: `kill -HUP <worker-pid>` → worker logs "Ignoring SIGHUP" and stays alive
- [ ] Manual: kill worker, wait 5 min → watchdog cron restarts it

### DIFF
--- a/src/services/infrastructure/ProcessManager.ts
+++ b/src/services/infrastructure/ProcessManager.ts
@@ -384,7 +384,27 @@ export function spawnDaemon(
     }
   }
 
-  // Unix: standard detached spawn
+  // Unix: Use setsid to create a new session, fully detaching from the
+  // controlling terminal. This prevents SIGHUP from reaching the daemon
+  // even if the in-process SIGHUP handler somehow fails (belt-and-suspenders).
+  // Fall back to standard detached spawn if setsid is not available.
+  const setsidPath = '/usr/bin/setsid';
+  if (existsSync(setsidPath)) {
+    const child = spawn(setsidPath, [process.execPath, scriptPath, '--daemon'], {
+      detached: true,
+      stdio: 'ignore',
+      env
+    });
+
+    if (child.pid === undefined) {
+      return undefined;
+    }
+
+    child.unref();
+    return child.pid;
+  }
+
+  // Fallback: standard detached spawn (macOS, systems without setsid)
   const child = spawn(process.execPath, [scriptPath, '--daemon'], {
     detached: true,
     stdio: 'ignore',
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -231,6 +231,22 @@ export class WorkerService {
       this.isShuttingDown = shutdownRef.value;
       handler('SIGINT');
     });
+
+    // SIGHUP: sent by kernel when controlling terminal closes.
+    // Daemon mode: ignore it (survive parent shell exit).
+    // Interactive mode: treat like SIGTERM (graceful shutdown).
+    if (process.platform !== 'win32') {
+      if (process.argv.includes('--daemon')) {
+        process.on('SIGHUP', () => {
+          logger.debug('SYSTEM', 'Ignoring SIGHUP in daemon mode');
+        });
+      } else {
+        process.on('SIGHUP', () => {
+          this.isShuttingDown = shutdownRef.value;
+          handler('SIGHUP');
+        });
+      }
+    }
   }
 
   /**
@@ -971,6 +987,18 @@ async function main() {
 
     case '--daemon':
     default: {
+      // Prevent daemon from dying silently on unhandled errors.
+      // The HTTP server can continue serving even if a background task throws.
+      process.on('unhandledRejection', (reason) => {
+        logger.error('SYSTEM', 'Unhandled rejection in daemon', {
+          reason: reason instanceof Error ? reason.message : String(reason)
+        });
+      });
+      process.on('uncaughtException', (error) => {
+        logger.error('SYSTEM', 'Uncaught exception in daemon', {}, error as Error);
+        // Don't exit — keep the HTTP server running
+      });
+
       const worker = new WorkerService();
       worker.start().catch((error) => {
         logger.failure('SYSTEM', 'Worker failed to start', {}, error as Error);
--- a/tests/infrastructure/process-manager.test.ts
+++ b/tests/infrastructure/process-manager.test.ts
@@ -10,6 +10,7 @@ import {
   parseElapsedTime,
   isProcessAlive,
   cleanStalePidFile,
+  spawnDaemon,
   type PidInfo
 } from '../../src/services/infrastructure/index.js';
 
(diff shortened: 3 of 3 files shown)


rodboev:
Re: [Greptile's review comment](https://github.com/thedotmack/claude-mem/pull/1056#discussion_r2790614511) on TypeError order dependency:

Good catch — applied in 22683f69. The order dependency is subtle: `TypeError('fetch failed')` hits the transport pattern check on line 43 before reaching the `instanceof TypeError` check on line 59, so generic TypeErrors correctly fall through to the blocking path.

PR #1064, opened 2026-02-11
feat: sync imported observations to Chroma vector DB

colangelo:
## Summary

- The `/api/import` endpoint writes to SQLite + FTS5 but does **not** generate Chroma vector embeddings
- This makes imported observations invisible to the MCP `search` tool (which uses hybrid FTS + vector search)
- `ensureBackfilled()` exists in ChromaSync but has two issues: (1) never called in production, (2) creates project-specific collections (`cm__<project>`) instead of using the shared `cm__claude-mem` collection

## Changes

**Auto-sync after import**: After importing observations, the handler syncs each one to Chroma using the shared `dbManager.getChromaSync()` instance — the same `syncObservation()` code path that production uses. This ensures imported observations go into the correct `cm__claude-mem` collection.

**New `POST /api/backfill` endpoint**: Manual trigger to sync all SQLite observations for a project to Chroma:
```bash
curl -X POST localhost:37777/api/backfill \
  -H 'Content-Type: application/json' \
  -d '{"project":"my-project"}'
```

Returns: `{"success":true,"project":"hittyping","total":15,"synced":15,"failed":0}`

## Design Decision

Chose `syncObservation()` over `ensureBackfilled()` because:
1. `syncObservation()` is proven in production (called by ResponseProcessor for every observation)
2. `ensureBackfilled()` targets `cm__<project>` collection but production uses `cm__claude-mem`
3. `getExistingChromaIds()` in `ensureBackfilled()` has a JSON parse error when querying Chroma MCP

## Files Changed

- `src/services/worker/http/routes/DataRoutes.ts` — Modified import handler + added backfill endpoint

## Test plan

- [x] Backfill 15 hittyping observations: all synced successfully
- [ ] Import new observations via `/api/import`, verify they appear in `search()` results
- [ ] Verify import still succeeds if Chroma is unavailable (sync failure is non-blocking)


thedotmack:
Good news — the source code in this PR (`DataRoutes.ts`) merges cleanly with no conflicts! The only conflicts are in the build artifacts (`.cjs` files) which regenerate on build. Could you rebase onto main and run `npm run build` to update the artifacts? This should be a quick fix.

Chriscross475 (changes requested):
This PR addresses an important gap (imported observations missing from Chroma), but CI checks are not running.

Before I can review:
1. Ensure GitHub Actions workflow runs on this branch
2. Verify build succeeds
3. Confirm the backfill endpoint works as documented

Once CI is green, I will review the implementation. The approach using syncObservation instead of ensureBackfilled sounds reasonable but needs verification.

thedotmack:

PR #1072, opened 2026-02-12
fix: add error handling to prevent Stop hook failures from package.json read errors

rajivsinclair:
## Problem

Users were experiencing recurring error messages during session shutdown (Stop hook execution):

```
ENOENT: no such file or directory, open '/Users/j/.claude/plugins/marketplaces/thedotmack/package.json'
```

These errors appeared consistently at session end, creating a poor user experience with ugly error traces despite the plugin otherwise functioning correctly.

## Root Cause

Two functions that read `package.json` lacked error handling:

1. **`getInstalledPluginVersion()`** in `src/services/infrastructure/HealthMonitor.ts`
2. **`getPluginVersion()`** in `src/shared/worker-utils.ts`

Both functions used `readFileSync()` wrapped in `JSON.parse()` without try-catch blocks. During Stop hook execution, if the `package.json` file was temporarily inaccessible (due to file system I/O timing, race conditions, or temporary locks), these functions would throw unhandled errors, causing the entire hook to fail with crash-like error messages.

## Solution

Added graceful error handling to both functions:

### Changes Made

**Before:**
```typescript
export function getInstalledPluginVersion(): string {
  const marketplaceRoot = path.join(homedir(), '.claude', 'plugins', 'marketplaces', 'thedotmack');
  const packageJsonPath = path.join(marketplaceRoot, 'package.json');
  const packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf-8'));
  return packageJson.version;
}
```

**After:**
```typescript
export function getInstalledPluginVersion(): string {
  try {
    const marketplaceRoot = path.join(homedir(), '.claude', 'plugins', 'marketplaces', 'thedotmack');
    const packageJsonPath = path.join(marketplaceRoot, 'package.json');
    const packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf-8'));
    return packageJson.version;
  } catch (error) {
    // Graceful fallback: return unknown version instead of crashing
    // This prevents Stop hook failures during plugin updates or temporary file access issues
    logger.warn('SYSTEM', 'Could not read installed plugin version from package.json', {
      error: error instanceof Error ? error.message : String(error)
    });
    return '0.0.0-unknown';
  }
}
```

Similar changes applied to `getPluginVersion()` in `worker-utils.ts`.

## Why This Fix Is Necessary

1. **User Experience**: Prevents alarming error messages that appear routine but don't actually impact functionality
2. **Robustness**: Hooks should be resilient to temporary file system issues
3. **Graceful Degradation**: The system continues working with a fallback version string rather than crashing
4. **Better Debugging**: Proper warning logs instead of stack traces make real issues easier to identify

## Why This Fix Is Safe

1. **Minimal Change**: Only adds error handling, doesn't change core logic
2. **Backward Compatible**: Functions still return version strings as before
3. **Fallback Value**: `'0.0.0-unknown'` is a safe sentinel value that clearly indicates the version couldn't be determined
4. **Logging Preserved**: Warning logs provide visibility for debugging while not alarming users
5. **No Breaking Changes**: All calling code expects a string return value and gets one

## Testing & Verification

- ✅ Built and deployed to test environment
- ✅ Verified worker service starts successfully with changes
- ✅ Confirmed Stop hook executes without errors
- ✅ Validated fallback value appears in compiled code
- ✅ Worker continues normal operation even when version read fails

## Related Issues

This fix addresses the recurring Stop hook errors that have been reported by users during normal session shutdown operations.

## Additional Context

This PR also includes two other bug fixes from the same branch:
- Fix for database initialization race condition in session-init requests
- Graceful handling of empty prompts in session-init hook

All three fixes improve the robustness of hook execution and error handling during session lifecycle events.

### DIFF
--- a/src/services/infrastructure/HealthMonitor.ts
+++ b/src/services/infrastructure/HealthMonitor.ts
@@ -94,12 +94,23 @@ export async function httpShutdown(port: number): Promise<boolean> {
 /**
  * Get the plugin version from the installed marketplace package.json
  * This is the "expected" version that should be running
+ *
+ * @returns Plugin version string, or '0.0.0-unknown' if package.json cannot be read
  */
 export function getInstalledPluginVersion(): string {
-  const marketplaceRoot = path.join(homedir(), '.claude', 'plugins', 'marketplaces', 'thedotmack');
-  const packageJsonPath = path.join(marketplaceRoot, 'package.json');
-  const packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf-8'));
-  return packageJson.version;
+  try {
+    const marketplaceRoot = path.join(homedir(), '.claude', 'plugins', 'marketplaces', 'thedotmack');
+    const packageJsonPath = path.join(marketplaceRoot, 'package.json');
+    const packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf-8'));
+    return packageJson.version;
+  } catch (error) {
+    // Graceful fallback: return unknown version instead of crashing
+    // This prevents Stop hook failures during plugin updates or temporary file access issues
+    logger.warn('SYSTEM', 'Could not read installed plugin version from package.json', {
+      error: error instanceof Error ? error.message : String(error)
+    });
+    return '0.0.0-unknown';
+  }
 }
 
 /**
--- a/src/shared/worker-utils.ts
+++ b/src/shared/worker-utils.ts
@@ -69,11 +69,22 @@ async function isWorkerHealthy(): Promise<boolean> {
 
 /**
  * Get the current plugin version from package.json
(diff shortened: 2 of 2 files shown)


thedotmack:
This error handling for package.json reads is still valuable! The files you touched have been refactored on main though, causing conflicts in all 4 files. Could you rebase onto main and re-apply the try/catch patterns to the current code? The intent is clear so it should be straightforward.

Chriscross475 (changes requested):
The error handling improvements are valuable, but this PR has merge conflicts across all 4 files. Please rebase onto main and re-apply the try-catch patterns to the refactored code. Also consider: (1) Adding tests for the fallback version behavior, (2) The 30-second timeout in session-init might be too aggressive for slow systems.

thedotmack:

PR #1088, opened 2026-02-13
fix: use persistent venv instead of uvx to prevent cache bloat

influenist:
## Problem

`uvx` creates a new ~331MB ephemeral Python environment in `~/.cache/uv` on **every invocation**. Since `ChromaMcpManager` spawns `chroma-mcp` via uvx on every connection, this causes the uv cache to grow by **several GB per day** on active sessions.

Real-world impact: users on 228GB SSDs are seeing disk-full crashes within days.

## Solution

Replace all `uvx` calls in `ChromaMcpManager.ts` with a **persistent virtual environment** at `~/.claude-mem/chroma-venv/`:

- **Auto-created on first connection** — `ensureVenv()` creates the venv and installs `chroma-mcp` once
- **Reused across sessions** — subsequent calls are a fast `fs.existsSync()` check
- **certifi cert lookup** uses the venv's python instead of `uvx --with certifi`
- **Cross-platform** — handles Windows `Scripts/` vs Unix `bin/` paths
- **Command injection fix** — `pythonVersion` validated against `/^\d+\.\d+$/` before use in shell commands

## Changes

- `ChromaMcpManager.ts`: Added `ensureVenv()`, `resolvePythonBin()` methods; replaced `uvx` spawn with direct venv binary invocation; removed `--python` and `chroma-mcp` positional args from `buildCommandArgs()`

## Testing

- Verified on macOS (M2) with claude-mem 10.5.2+
- Confirmed `~/.cache/uv` stops growing after patch
- chroma-mcp connects and responds to health checks via persistent venv
- Venv auto-creates successfully on first run

---

*Rebased onto current main and adapted to the `ChromaMcpManager` architecture (previously targeted `ChromaSync.ts`).*


thedotmack:
The persistent venv approach and Zscaler cert bundling are both valuable! ChromaSync has been refactored to use an HTTP client model since this PR was opened — could you rebase onto main and adapt the venv/cert changes to the new `ensureHttpClientConnected()` architecture?

influenist:
**Update:** v10.1.0 has already addressed the main `uvx` transport issue by switching ChromaSync to use an HTTP client connecting to a local Chroma server instead of spawning `chroma-mcp` via `uvx` on every connection.

However, there is still **one remaining `uvx` call** in `worker-service.cjs` line 878:
```js
uvx --with certifi python -c "import certifi; print(certifi.where())"
```
This is used for Zscaler/corporate proxy SSL certificate resolution in `getCombinedCertPath()`. While this only triggers on macOS with Zscaler certificates present (and caches the result for 24h), it would still benefit from using a persistent venv approach to avoid any ephemeral environment creation.

This PR's approach of a persistent venv at `~/.claude-mem/chroma-venv/` could be adapted to just patch that remaining certifi call, or the scope could be narrowed since the main transport fix is already in v10.1.0.

influenist:
Bumping this — still happening on v10.5.2. Just cleaned 2.2GB from `~/.cache/uv` on a machine that's been running claude-mem for a few weeks.

Every PostToolUse hook fires chroma-mcp via `uvx`, creating a fresh ~331MB ephemeral venv each time. On an active session that's dozens of tool calls per hour — the cache grows by several GB/day.

This is a real disk space issue especially on smaller SSDs. Would be great to get the persistent venv approach merged.

influenist:
Rebased onto current main and adapted to the `ChromaMcpManager` architecture as requested. Also addressed the review feedback:

- **Command injection fix**: `pythonVersion` is now validated against `/^\d+\.\d+$/` before use in shell commands
- **Supervisor integration**: merged cleanly with the new process supervisor system
- **Scope**: only touches `ChromaMcpManager.ts` — replaces `uvx` spawn with persistent venv at `~/.claude-mem/chroma-venv/`

Note: the compiled `claude-mem` binary also contains a copy of ChromaMcpManager that still calls `uvx`. Users on 228GB SSDs are hitting disk-full crashes daily from the cache bloat. Would appreciate a review when you get a chance.

thedotmack:

PR #1144, opened 2026-02-17
fix: prevent duplicate worker daemons (version mismatch loop + spawn races)

influenist:
## Summary

Fixes three root causes of duplicate worker daemon accumulation that led to **218 zombie processes** consuming 15GB of swap on an 8GB Mac.

### Fix 1: Version mismatch restart loop (primary cause)
- `getInstalledPluginVersion()` was reading from `MARKETPLACE_ROOT` (`~/.claude/plugins/marketplaces/thedotmack/package.json`) — a git clone of the repo
- The running worker reports its hardcoded build version via `/api/version`
- When these diverge (marketplace=10.2.3, worker=10.2.1), every hook invocation detects a "mismatch", sends `POST /api/admin/shutdown`, and spawns a new daemon — which also reports the old version → **infinite restart loop**
- **Fix**: Read version from `getPackageRoot()` (the installed plugin directory, relative to `__dirname`) instead of the marketplace

### Fix 2: Atomic file-based spawn lock
- Multiple hooks fire simultaneously (context, user-message, session-start) and each calls `ensureWorkerStarted()`
- The 1s health check timeout means all conclude the worker is down and each spawns a daemon
- **Fix**: `O_CREAT|O_EXCL` atomic lock file — only one process spawns, others wait for health

### Fix 3: Hook in-process worker spawn race
- The `hook` CLI command has a fallback that starts the worker in-process (`new WorkerService().start()`)
- With Bun's `SO_REUSEPORT` on macOS, multiple processes can bind the same port simultaneously
- **Fix**: Hook fallback now checks `acquireSpawnLock()` before starting in-process

## Test plan

- [ ] Start Claude Code with multiple hooks — verify only 1 worker daemon spawns
- [ ] Ensure marketplace `package.json` can have a different version without triggering restart loop
- [ ] Kill worker while running — verify it restarts cleanly (lock released in `finally`)
- [ ] Simulate stale lock (kill mid-spawn) — verify 30s expiry cleans it up
- [ ] `ps aux | grep worker-service` after extended multi-session use — should show exactly 1 daemon

### DIFF
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -10,7 +10,7 @@
  */
 
 import path from 'path';
-import { existsSync, writeFileSync, unlinkSync, statSync, mkdirSync, openSync, closeSync, constants as fsConstants } from 'fs';
+import { existsSync, writeFileSync, readFileSync, unlinkSync, statSync, mkdirSync, openSync, closeSync, constants as fsConstants } from 'fs';
 import { Client } from '@modelcontextprotocol/sdk/client/index.js';
 import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';
 import { getWorkerPort, getWorkerHost } from '../shared/worker-utils.js';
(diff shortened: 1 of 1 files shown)


influenist:
## Update: Additional Root Cause Found

The spawn lock in this PR helps with concurrent races, but the **primary driver** of the 218-daemon issue is actually a **version mismatch restart loop** (see issue #1145 comment).

### Version Mismatch Loop

`nX()` reads version from `~/.claude/plugins/marketplaces/thedotmack/package.json` (marketplace/repo copy) while the running worker reports its hardcoded version from `worker-service.cjs`. When these diverge (e.g., marketplace=10.2.3, worker=10.2.1), every hook invocation:

1. Detects mismatch → sends `POST /api/admin/shutdown`
2. Spawns a new daemon (which also has the old hardcoded version)
3. Repeat → infinite restart loop

### Additional Fix: Hook In-Process Fallback

The `hook` CLI command has a second spawn path (`await new im().start()`) that starts the worker in-process when `g8()` fails. This bypasses the spawn lock entirely and, combined with Bun's `SO_REUSEPORT` on macOS, allows multiple processes to bind to the same port simultaneously.

### Recommended changes (beyond this PR)

1. `nX()` should read version from the installed plugin path, not the marketplace directory
2. The hook in-process fallback should also check for recent spawn attempts before starting a full worker

thedotmack:
Closing as outdated — the duplicate worker daemon issues (version mismatch restart loop, spawn races) have been addressed in subsequent releases through hook lifecycle and process management overhauls. Thanks for the detailed analysis!

Issue #1052, opened 2026-02-10
isProcessing flag stuck true after session deleted during pending summarize


bmccann36:
## Problem Statement

The `isProcessing` flag becomes permanently stuck as `true` when a session is deleted while it has a pending summarize message in the queue. This causes the web viewer spinner to run indefinitely even though no actual processing is occurring (`queueDepth=0`, `activeSessions=0`).

The root cause is that the in-memory `isProcessing` flag is not reconciled when sessions are cleaned up. It only transitions based on specific events (enqueue/dequeue) and doesn't verify actual state against queue depth and active sessions.

This differs from the previous issue #520, which was about stuck database rows. This bug is specifically about the in-memory processing state flag not being recalculated during session cleanup.

## Environment

- **Claude-mem Version**: 9.1.1
- **Claude Code Version**: 2.1.38
- **Node.js**: v24.12.0
- **Bun**: 1.3.8
- **OS**: macOS 26.1 (arm64)
- **Platform**: darwin
- **Worker Port**: 37777
- **Data Directory**: ~/.claude-mem
- **Database Size**: 7240.00 KB

## Steps to Reproduce

1. Start two active sessions sending observations to the worker
2. End one session (triggering the session-complete hook)
3. Session-complete hook enqueues a summarize message for the session
4. Delete/abort the session before the summarize message can be processed (generator aborted)
5. Check `/api/processing-status` endpoint
6. Observe: `isProcessing=true`, `queueDepth=0`, `activeSessions=0`
7. Open the web viewer - spinner never stops spinning

## Expected Behavior

- `isProcessing` should be `false` when `queueDepth=0` and `activeSessions=0`
- The processing status broadcast should reconcile the flag against actual queue and session state
- The flag should reset properly during session cleanup operations
- Web viewer spinner should stop when no processing is actually occurring

## Actual Behavior

- `isProcessing` remains stuck at `true` indefinitely
- Broadcasting shows: `isProcessing=true, queueDepth=0, activeSessions=0`
- Web viewer spinner continues spinning forever
- Worker restart does not immediately clear the stuck state
- The flag is not recalculated during session cleanup/deletion

## Timeline from Logs

1. Summarize message enqueued for session (e.g., session-3151)
2. Log shows "Spawn already in progress, skipping" - cannot start processing
3. Generator aborted on the session
4. Session deleted (duration=7.4s)
5. Broadcasting: `isProcessing=true`, `queueDepth=0`, `activeSessions=0`
[...]

PR #1147, opened 2026-02-17
fix: prevent stuck processing messages and permanent spinner (#1052)

michelhelsdingen:
## Summary

- Five-layer defense against messages permanently stuck in `processing` status, which causes the viewer spinner to never stop (`isProcessing=true` forever)
- Fixes the root cause of Issue #1052: `hasAnyPendingWork()` counts orphaned `processing` messages from crashed generators, keeping `broadcastProcessingStatus()` stuck on `true`

## Root Cause

When a generator crashes between `claimAndDelete()` (marks message as `processing`) and `confirmProcessed()` (deletes it), the message stays in `processing` forever. `hasAnyPendingWork()` counts these, so `isProcessing=true` is broadcast permanently.

### Layer 1: Agent try/finally — `SDKAgent.ts`, `GeminiAgent.ts`, `OpenRouterAgent.ts`
Wraps message processing loops in try/finally. On crash/abort, any `processingMessageIds` still in-flight are marked as `failed` via `markFailed()` instead of staying stuck.

### Layer 2: Generator exit handler — `worker-service.ts`
Always resets stuck `processing` messages on **any** generator exit (not just idle timeout). On unrecoverable errors, marks all remaining messages as abandoned.

### Layer 3: Session delete cleanup — `SessionManager.ts`
`deleteSession()` calls `markAllSessionMessagesAbandoned()` and `removeSessionImmediate()` calls `markSessionMessagesFailed()` before removing the session. Prevents orphaned messages from keeping `hasAnyPendingWork()` true for deleted sessions.

### Layer 4: Periodic stale message reaper — `worker-service.ts`
New 2-minute interval timer resets messages stuck in `processing` for >3 minutes. Uses `.unref()` for clean shutdown. Catches edge cases missed by other layers.

### Layer 5: Startup recovery (existing, unchanged)
`resetStaleProcessingMessages(0)` on worker startup already resets all processing messages.

## Test plan
- [x] Build passes (`npm run build`)
- [x] Worker starts healthy with `status: ok`, `mcpReady: true`
- [x] No stuck messages after worker restart: `SELECT status, COUNT(*) FROM pending_messages WHERE status IN ('pending', 'processing') GROUP BY status` returns empty
- [x] Periodic cleanup timer logs `Started stale message cleanup (runs every 2 minutes, threshold: 3 minutes)` on startup
- [x] Verified on macOS with active sessions — spinner correctly stops after processing completes

### DIFF
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -175,6 +175,9 @@ export class WorkerService {
   // Orphan reaper cleanup function (Issue #737)
   private stopOrphanReaper: (() => void) | null = null;
 
+  // Periodic stale message cleanup timer
+  private staleMessageCleanupInterval: ReturnType<typeof setInterval> | null = null;
+
   // AI interaction tracking for health endpoint
   private lastAiInteraction: {
     timestamp: number;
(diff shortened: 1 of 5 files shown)


thedotmack:
fixed in latest updates

PR #1152, opened 2026-02-17
fix: remove non-existent columns from getSessionSummaryById query

AI-Reviewer-QS:
## Summary
- `getSessionSummaryById` in both `sessions/get.ts` and `SessionStore.ts` queries `request_summary` and `learned_summary` columns that were never added to the `sdk_sessions` table via any migration, causing a runtime `SQLiteError: no such column` when called
- The query also referenced `created_at`/`created_at_epoch` instead of the actual column names `started_at`/`started_at_epoch`
- Removed the phantom columns from the SQL query and type definition, and corrected the timestamp column names to match the schema

## Test plan
- [x] Added `getSessionSummaryById` tests to `tests/sqlite/sessions.test.ts` following existing patterns
- [x] Verified the function retrieves session data with all expected fields
- [x] Verified null return for non-existent sessions
- [x] All 193 tests pass across the full test suite

### DIFF
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -2094,7 +2094,7 @@ export class SessionStore {
   }
 
   /**
-   * Get full session summary by ID (includes request_summary and learned_summary)
+   * Get full session summary by ID
    */
   getSessionSummaryById(id: number): {
     id: number;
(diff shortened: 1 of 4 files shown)


thedotmack:
Closing as outdated — the phantom column fix and timestamp column corrections were addressed as part of the broader fixes in PR #1214.

Issue #1168, opened 2026-02-18
Observer Claude CLI processes never exit, causing massive resource leak


SnehalRaj:
Bug Report   
                                                                                                                                                                                   
  Version: 9.0.12                                                                                                                                                                  
  Provider: claude (CLAUDE_MEM_PROVIDER=claude)                                                                                                                                    
  OS: macOS (Darwin 24.6.0)                                                                                                                                                        
  Claude Code: 2.1.44

  Problem

  Observer Claude CLI processes spawned by the plugin accumulate indefinitely and never exit. After a few days of normal usage, I had 157 zombie claude processes consuming 8.4 GB
  of RAM while idle.

  Evidence

  $ ps aux | grep claude | grep "--resume" | grep "disallowedTools" | wc -l
  157

  $ ps aux | grep claude | grep "resume" | awk '{sum += $6} END {printf "%.0f MB\n", sum/1024}'
  8406 MB

  All observer processes have this signature:
  claude --output-format stream-json --verbose --input-format stream-json \
    --model claude-sonnet-4-5 \
    --resume <SESSION_UUID> \
    --disallowedTools Bash,Read,Write,Edit,Grep,Glob,WebFetch,WebSearch,Task,NotebookEdit,AskUserQuestion,TodoWrite \
    --setting-sources --permission-mode default

  Processes dated back 4+ days, accumulating ~5-7 per Claude Code session.

  Root Cause Analysis

  1. No session-end cleanup: The Stop hook doesn't force-kill observer processes.
  2. No timeout mechanism: Kill function exists in code but isn't invoked for observer lifecycle.
  3. No orphan cleanup: Worker daemon doesn't sweep for stale observers.
  4. Empty tracking directory: observer-sessions/ is empty — tracking system not functioning.

  Workaround

  ps aux | grep claude | grep "\-\-resume" | grep "disallowedTools" | awk '{print $2}' | xargs kill

  Suggested Fixes

  1. Add SIGTERM + SIGKILL fallback for observers after a timeout (30-120s)
  2. Implement cleanup in Stop hook to kill observers for the ending session
  3. Add periodic orphan process cleanup in the worker daemon
  4. Persist observer PIDs to observer-sessions/ and clean up on worker startup

PR #1142, opened 2026-02-17
Skip summary generation for subagent runs

thedotmack:
## Summary
- Detects `SubagentStop` events via `hook_event_name` field from Claude Code hooks
- Adds `isSubagent` boolean to `NormalizedHookInput` for clean skip logic
- Skips summary generation and session-complete processing for subagent sessions
- Prevents noise from short-lived Task tool helper sessions cluttering memory

## Test plan
- [x] Build succeeds (`npm run build-and-sync`)
- [x] `isSubagent` only appears in 4 expected files (types, adapter, 2 handlers)
- [x] No worker API changes — skip happens at hook layer
- [x] Existing tests unaffected (`isSubagent` is optional, defaults to falsy)
- [ ] Manual: verify subagent Task tool runs no longer produce summary observations

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1169, opened 2026-02-18
fix: deduplicate session-init to prevent redundant re-initialization

greghughespdx:
## Problem

The `UserPromptSubmit` hook fires `session-init` on **every prompt**, including prompts that aren't typed by the user. In some more-advanced Claude Code setups, such as monitoring scripts that inject periodic prompts into a tmux-hosted Claude Code session, cron-triggered prompts, automated inbox processors, etc. these system-generated prompts fire `UserPromptSubmit` at regular intervals (e.g., every xx minutes).

Each call re-initializes the session and spawns an SDK agent subprocess, even though the session is already active. Over hours of mixed active/idle usage, this produces redundant initialization events and unnecessary agent process spawns. This noise is recorded by claude-mem and degrades the value of the data being stored/leveraged.

This is also just the type of advanced automation that claude-mem users are more likely to build than the average Claude Code user: running Claude Code in persistent, long-running terminal sessions with scripted interactions. The `UserPromptSubmit` hook has no way to distinguish a genuine first prompt from a subsequent automated one, so all of them trigger full session re-initialization.

## What this fixes

This PR prevents **session-init re-initialization** — the redundant `/api/sessions/init` and `/sessions/{id}/init` calls that occur on every subsequent `UserPromptSubmit` within a session. After this fix, only the first prompt in a session triggers initialization; all subsequent prompts skip it.

**Note:** The `Stop` hook's `summarize` handler is a separate code path that also fires on each turn end, including idle responses. That noise source is not addressed here but is the same class of problem that #1142 addresses for subagent runs. See comment on #1142 for details.

## Solution

Add file-based deduplication so that within a single `contentSessionId`, only the **first** `UserPromptSubmit` triggers session-init. All subsequent prompts skip re-initialization.

## Configuration

Controlled by `CLAUDE_MEM_SESSION_INIT_DEDUP` (default: `true`).

- **`true`** (default) — Deduplicate session-init per `contentSessionId`. Only the first prompt triggers initialization; subsequent prompts skip it.
- **`false`** — Disable deduplication. Every `UserPromptSubmit` triggers session-init (original behavior).

Can be set via `~/.claude-mem/settings.json` or environment variable:

```json
{
  "CLAUDE_MEM_SESSION_INIT_DEDUP": "false"
}
```

## How it works

- State file at `~/.claude-mem/session-init-debounce.json` tracks which `contentSessionId` values have already been initialized
- On each `UserPromptSubmit`, the handler checks if the session exists in the state file — if so, returns early
- After successful initialization, the session ID is recorded
- Entries are pruned after 24 hours to prevent unbounded growth
- All error paths are non-fatal (file read/write failures fall through gracefully)

## Changes

- `src/cli/handlers/session-init.ts` — Added dedup check before worker call
- `src/shared/SettingsDefaultsManager.ts` — Added `CLAUDE_MEM_SESSION_INIT_DEDUP` setting

## Testing

Validated in a live environment over multiple hours with automated prompts injected via tmux every ~6 minutes:

**Enabled path (default):**
- Zero redundant `INIT_COMPLETE` events after first initialization
- Logs confirm `session-init skipped: already initialized for this session` on every subsequent prompt
- Normal session initialization on first prompt works as before
- File-based state survives across short-lived hook processes
- Observations from real tool use continue to be recorded normally

**Disabled path (`CLAUDE_MEM_SESSION_INIT_DEDUP=false`):**
- Every prompt triggers `INIT_COMPLETE` (original behavior restored)
- Session summary count increments on each prompt as expected
- Setting change takes effect immediately without restart

## Related

- Complementary to #1142 (which skips summary generation for subagent runs — our fix addresses the session-init side for the main session)
- Partially mitigates #1168 (reduces redundant SDK agent spawns, though does not address the process exit bug itself)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/cli/handlers/session-init.ts
+++ b/src/cli/handlers/session-init.ts
@@ -14,6 +14,7 @@ import { SettingsDefaultsManager } from '../../shared/SettingsDefaultsManager.js
 import { USER_SETTINGS_PATH, DATA_DIR } from '../../shared/paths.js';
 import { existsSync, readFileSync, writeFileSync, mkdirSync } from 'fs';
 import { join } from 'path';
+import { setIdleFlag } from './idle-flags.js';
 
 /**
  * Debounce state stored in a file so it persists across the short-lived hook processes.
@@ -23,7 +24,7 @@ interface DebounceState {
   [contentSessionId: string]: number;
 }
 
-const DEBOUNCE_FILE_PATH = join(DATA_DIR, 'session-init-debounce.json');
+const DEBOUNCE_FILE_PATH = join(DATA_DIR, 'session-init-dedup.json');
 const DEBOUNCE_MAX_AGE_MS = 24 * 60 * 60 * 1000; // 24 hours
 
 /**
@@ -112,6 +113,7 @@ export const sessionInitHandler: EventHandler = {
     const dedupEnabled = settings.CLAUDE_MEM_SESSION_INIT_DEDUP === 'true';
     if (dedupEnabled && isAlreadyInitialized(sessionId)) {
       logger.info('HOOK', 'session-init skipped: already initialized for this session', { contentSessionId: sessionId });
+      setIdleFlag(sessionId);
       return { continue: true, suppressOutput: true };
     }
 


thedotmack:

PR #1175, opened 2026-02-18
fix: prevent zombie process accumulation (#1168)

### LINKED ISSUE #1168: Observer Claude CLI processes never exit, causing massive resource leak

thedotmack:
## Summary
- Add `ensureProcessExit` to generator finally blocks in `SessionRoutes` and `worker-service`, matching the pattern already working in `SDKAgent` — closes the primary leak where `abortController.abort()` was called but subprocess exit was never verified
- Add stale session reaper (every 2m) that removes sessions with no active generator and no pending work after 15m idle, unblocking the orphan reaper which previously skipped processes for "active" sessions that were actually dead

## Test plan
- [ ] Deploy and monitor `ps aux | grep claude` over time — zombie count should stabilize instead of growing
- [ ] Verify worker builds cleanly (`npm run build-and-sync`)
- [ ] Kill a terminal mid-session, confirm the orphaned subprocess gets cleaned up within ~7 minutes (2m stale reap + 5m orphan reap)
- [ ] Confirm normal session lifecycle (init → observations → summarize → complete) still works

Fixes #1168


thedotmack:
@claude is the reapStaleSessions overkill? just something to introduce a new bug later?

PR #1198, opened 2026-02-21
fix: migrate from bun:sqlite to better-sqlite3

igorrivin:
## Summary

- Replaces `bun:sqlite` with `better-sqlite3`, eliminating the sole Bun-specific dependency so the worker runs under Node.js
- Fixes SIGABRT crashes on 64KB page size kernels (e.g. `linux-nvidia-64k`) caused by Bun's JSC JIT
- Uses a thin compatibility shim (`sqlite-compat.ts`) that adds `db.run()` and `db.query()` methods to better-sqlite3's Database class, so no call sites needed API changes

## Motivation

Bun's JavaScriptCore JIT compiler crashes with SIGABRT on ARM64 kernels using 64KB pages (the `nvidia-64k` kernel variant). Since `bun:sqlite` was the **only** Bun-specific import across the entire codebase — everything else uses standard Node.js APIs (Express, child_process, fs, http) — replacing it with `better-sqlite3` lets the worker run under plain Node.js v20+.

## Approach

A single `sqlite-compat.ts` shim extends `better-sqlite3` with the two methods `bun:sqlite` provides that `better-sqlite3` doesn't:

| bun:sqlite API | better-sqlite3 equivalent | Shim action |
|---|---|---|
| `db.run(sql)` | `db.exec(sql)` | Delegates to `exec()` |
| `db.query(sql)` | `db.prepare(sql)` | Delegates to `prepare()` |
| Constructor `{create, readwrite}` | `{fileMustExist}` | Maps options |

All other APIs (`prepare`, `transaction`, `close`, `stmt.run/get/all`) are already compatible.

## Changes

- **New**: `src/services/sqlite/sqlite-compat.ts` — compatibility shim
- **22 source files**: `import from 'bun:sqlite'` → relative import from `sqlite-compat`
- **Build config**: externalize `better-sqlite3`, `node` shebangs on all bundles
- **Plugin hooks**: call `node worker-service.cjs` directly (removed `bun-runner.js` indirection)
- **Plugin scripts**: `statusline-counts.js`, `worker-cli.js`, `worker-wrapper.cjs` updated for node
- **Sync script**: `bun install` → `npm install`
- **package.json**: added `better-sqlite3` dep, removed `bun` engine, worker scripts use `node`

## Test plan

- [x] Worker starts under `/usr/bin/node` (verified via `ps`)
- [x] `/api/health` returns `{"status":"ok"}` with correct version
- [x] `/api/readiness` returns `{"status":"ready"}`
- [x] Tested on linux-aarch64 with 64KB page kernel (`6.8.0-1013-nvidia-64k`)
- [ ] Verify on x86_64 Linux / macOS / Windows


thedotmack:
Thanks for the thorough work here! We appreciate the effort to support non-Bun environments.

However, we're planning to make this an **option** rather than a full replacement — we want to support both `bun:sqlite` and `better-sqlite3` as backends, so users can choose based on their environment. We also have a plan to introduce a better installer architecture, designed around how the `npx` skills installer works, which would handle runtime selection more cleanly.

We'll keep this PR as a reference, but we're going to go a different direction on the implementation. Feel free to keep an eye on upcoming changes!

thedotmack:

PR #1211, opened 2026-02-23
fix: Windows paths with spaces crash worker startup and chroma-mcp connection

jrvoorhorst:
## Problem

  On Windows, when the user profile path contains spaces (e.g.
  `C:\Users\Jan Rutger Voorhorst\`), claude-mem fails to start in two ways:

  ### 1. Worker fails to start (`Module not found "C:\Users\Jan"`)
  `worker-cli.js` uses PowerShell `Start-Process -ArgumentList '${scriptPath}'`.
  PowerShell parses the `-ArgumentList` string by spaces, so
  `C:\Users\Jan Rutger Voorhorst\...\worker-wrapper.cjs` is split into
  multiple arguments and Bun receives only `C:\Users\Jan` as the script path.

  ### 2. chroma-mcp fails to connect (`MCP error -32000: Connection closed`)
  `ChromaMcpManager` always uses `uvx.cmd` on Windows. However, the native
  [astral-sh/uv](https://github.com/astral-sh/uv) installer creates `uvx.exe`,
  not `uvx.cmd`. The `.cmd` wrapper is only created when installed via npm/pipx.
  This causes the chroma-mcp subprocess to fail immediately.

  ## Fix

  ### `plugin/scripts/worker-cli.js`
  Wrap the script path in double-quotes inside the PowerShell single-quoted
  string so spaces are preserved:
  -ArgumentList '"${a}"'   // was: '${a}'

  ### `src/services/sync/ChromaMcpManager.ts`
  Add `resolveWindowsUvxCommand()` that searches PATH for `uvx.exe` first,
  falling back to `uvx.cmd` for npm/pipx installs.

  ### `scripts/smart-install.js`
  Add `ensureUvxCmd()` that creates a `uvx.cmd` shim next to `uvx.exe` after
  installation, so both lookup strategies work going forward.

  ## Testing
  Verified working on Windows 10 with:
  - User profile path containing spaces
  - `uv` installed via native astral-sh installer (creates `uvx.exe` only)
  - Norton antivirus (requires exclusions for `~/.local/share/uv/`)

### DIFF
--- a/scripts/smart-install.js
+++ b/scripts/smart-install.js
@@ -191,35 +191,6 @@ function installBun() {
 }
 
 /**
- * Get the uvx executable path on Windows (.exe or .cmd)
- */
-function getUvxPath() {
-  const UVX_COMMON_PATHS_WIN = [
-    join(homedir(), '.local', 'bin', 'uvx.exe'),
-    join(homedir(), '.cargo', 'bin', 'uvx.exe')
-  ];
-  return UVX_COMMON_PATHS_WIN.find(existsSync) || null;
-}
-
-/**
- * On Windows, the native uv installer creates uvx.exe but some tools expect uvx.cmd.
- * Create a uvx.cmd shim in the same directory as uvx.exe if it doesn't already exist.
- */
-function ensureUvxCmd() {
-  if (!IS_WINDOWS) return;
-
-  const uvxExePath = getUvxPath();
-  if (!uvxExePath) return;
-
-  const uvxCmdPath = uvxExePath.replace(/uvx\.exe$/i, 'uvx.cmd');
-  if (existsSync(uvxCmdPath)) return;
-
-  const cmdContent = `@echo off\r\n"${uvxExePath}" %*\r\n`;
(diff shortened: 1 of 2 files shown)


jrvoorhorst:
Removed the `ChromaMcpManager.ts` and `smart-install.js` changes from this PR —
  those address the same `uvx.cmd` issue that is already covered by PR #1191
  (`fix: use 'uvx' instead of 'uvx.cmd' on Windows for chroma-mcp spawn`).

  This PR now focuses solely on the separate `worker-cli.js` bug where the worker
  fails to start when the Windows user profile path contains spaces
  (e.g. `C:\Users\Jan Rutger Voorhorst\`). This is an independent issue that
  occurs before chroma-mcp is ever involved.

thedotmack:
Closing this PR — the diff only includes changes to the compiled `worker-cli.js` bundle, not the source files (`ChromaMcpManager.ts`, `smart-install.js`). We're keeping #1191 open which takes a simpler approach to the uvx resolution fix at the source level.

If you'd like to resubmit with source file changes (not just the compiled output), feel free to open a new PR. Thanks for the contribution!

PR #1222, opened 2026-02-24
fix: redirect stdin from /dev/null for 'start' hook commands

### LINKED ISSUE #1220: [Bug] PostToolUse hooks crash: 'start' fails on any stdin, 'observation' fails on stdin > ~350 bytes
## Environment

- **OS**: macOS (Darwin 25.3.0)
- **claude-mem version**: 10.4.0
- **Bun version**: $(bun --version)
- **Node version**: $(node --version)

## Description

Two PostToolUse hook commands from claude-mem crash on every tool call, producing `PostToolUse:<tool> hook error` messages in Claude Code.

### Bug 1: `worker-service.cjs start` crashes when receiving stdin data

The `start` command is registered as a PostToolUse hook. Claude Code pipes JSON data (tool_name, tool_response, etc.) to all PostToolUse hooks via stdin. The `start` command doesn't need this data but crashes (exit code 1, no stdout/stderr) when it receives any.

**Reproduction:**
```bash
# Fails (exit 1, silent)
echo '{"tool_name":"Read","tool_response":"test"}' | bun worker-service.cjs start

# Works (exit 0)
bun worker-service.cjs start < /dev/null
```

**Workaround:** Add `< /dev/null` to the `start` command in `hooks.json`.

### Bug 2: `worker-service.cjs hook claude-code observation` crashes on payloads > ~350 bytes

The `observation` hook works with small payloads but silently crashes (exit 1) when stdin exceeds approximately 350 bytes. In real usage, `tool_response` contains file contents or command output, easily exceeding this limit — so it fails on virtually every tool call.

**Reproduction:**
```bash
# Works (payload ~170 bytes total)
python3 -c "import json; print(json.dumps({'tool_name':'Read','tool_response':'x'*100}))" | bun worker-service.cjs hook claude-code observation

# Works (payload ~370 bytes total)
python3 -c "import json; print(json.dumps({'tool_name':'Read','tool_response':'x'*300}))" | bun worker-service.cjs hook claude-code observation

# Fails - exit 1, no output (payload ~470 bytes total)
python3 -c "import json; print(json.dumps({'tool_name':'Read','tool_response':'x'*400}))" | bun worker-service.cjs hook claude-code observation
```

## Impact

Every tool call in Claude Code produces 2x `PostToolUse:<tool> hook error` messages, cluttering the UI. The memory/observation functionality likely doesn't work at all since observation data never gets processed.

## Expected Behavior

- `start` should ignore stdin (or not crash when stdin data is present)
- `observation` should handle arbitrarily large stdin payloads

## Suggested Fixes

1. For `start`: Add `< /dev/null` to the hook command in `hooks.json`, or handle unexpected stdin gracefully in the code
2. For `observation`: Investigate the stdin buffer/parsing limit in `worker-service.cjs` — likely a Bun stdin read issue or JSON parse buffer limit

zerone0x:
## Problem (Bug 1 from #1220)

The `worker-service.cjs start` command does not consume or expect stdin input. However, Claude Code pipes JSON data (containing `tool_name`, `tool_response`, etc.) to all **PostToolUse** hooks, and JSON data to **UserPromptSubmit** and **Stop** hooks as well.

When `start` receives any stdin data, it crashes with exit code 1 and no stdout/stderr output, which Claude Code surfaces as a hook error on every tool call:

```
PostToolUse:<tool> hook error: Failed with non-blocking status code: No stderr output
```

**Reproduction:**
```bash
# Fails (exit 1, silent)
echo '{"tool_name":"Read","tool_response":"test"}' | bun worker-service.cjs start

# Works (exit 0)
bun worker-service.cjs start < /dev/null
```

## Root Cause

`worker-service.cjs start` attempts to parse stdin when it receives data, but the startup command has no use for the hook payload. Any stdin content causes it to fail.

## Fix

Add `< /dev/null` to the `start` command in the three hook contexts that pipe stdin to hooks:
- `UserPromptSubmit`
- `PostToolUse`  
- `Stop`

The `SessionStart` `start` command is left unchanged as it does not receive stdin payloads from Claude Code.

## Change

```diff
- "command": "node \"...\"/bun-runner.js \"...\"/worker-service.cjs start"
+ "command": "node \"...\"/bun-runner.js \"...\"/worker-service.cjs start < /dev/null"
```

Closes #1220 (Bug 1)

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -45,7 +45,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js\" \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js\" \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start < /dev/null",
             "timeout": 60
           },
           {
@@ -62,7 +62,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js\" \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js\" \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start < /dev/null",
             "timeout": 60
           },
           {
@@ -78,7 +78,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js\" \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js\" \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start < /dev/null",
             "timeout": 60
           },
           {


thedotmack:
Thanks for digging into this! The stdin issue you identified was real — the redundant `start` commands in PostToolUse, UserPromptSubmit, and Stop hooks were hitting bun-runner.js's 5-second `collectStdin()` timeout because Claude Code keeps stdin open.

This was fixed in #1229 (merged) by **removing the redundant `start` commands entirely** from those three hook groups. The `start` command was unnecessary there because the hook command (`worker-service.cjs hook ...`) already calls `ensureWorkerStarted()` internally. Removing them also eliminated the 5s latency.

PR #1223, opened 2026-02-24
fix: resolve plugin root for Stop hooks without CLAUDE_PLUGIN_ROOT

### LINKED ISSUE #1215: Stop hooks fail: ${CLAUDE_PLUGIN_ROOT} not injected in Stop hook context
### Environment
  - Claude Code version: 2.1.50
  - claude-mem version: 10.4.0
  - OS: macOS Darwin 25.3.0

  ### Problem

  All 3 Stop hooks fail with:
  Stop hook error: Failed with non-blocking status code: No stderr output

  The root cause is that `${CLAUDE_PLUGIN_ROOT}` is **not set** when Claude Code executes Stop hooks, causing the
  command path to resolve as `/scripts/bun-runner.js` (empty prefix).

  ### Reproduction

  The command in `hooks/hooks.json` fails because `${CLAUDE_PLUGIN_ROOT}` is empty:
  node "${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js" "${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" hook
  claude-code summarize
  Error: Cannot find module '/scripts/bun-runner.js'

  But with explicit absolute paths it works fine:
  ```bash
  node ~/.claude/plugins/cache/thedotmack/claude-mem/10.4.0/scripts/bun-runner.js \
       ~/.claude/plugins/cache/thedotmack/claude-mem/10.4.0/scripts/worker-service.cjs \
       hook claude-code summarize
  # Output: {"continue":true,"suppressOutput":true}  ✅

  Hook Status Summary

  ┌──────────────────┬────────────┬───────────────────────────────────────┐
  │    Hook Type     │   Status   │                 Notes                 │
  ├──────────────────┼────────────┼───────────────────────────────────────┤
  │ SessionStart     │ ✅ Working │ ${CLAUDE_PLUGIN_ROOT} is injected     │
  ├──────────────────┼────────────┼───────────────────────────────────────┤
  │ UserPromptSubmit │ ✅ Working │ ${CLAUDE_PLUGIN_ROOT} is injected     │
  ├──────────────────┼────────────┼───────────────────────────────────────┤
  │ PostToolUse      │ ✅ Working │ ${CLAUDE_PLUGIN_ROOT} is injected     │
  ├──────────────────┼────────────┼───────────────────────────────────────┤
  │ Stop             │ ❌ Failing │ ${CLAUDE_PLUGIN_ROOT} is NOT injected │
  └──────────────────┴────────────┴───────────────────────────────────────┘

  Impact

  Session-end summarization (hook claude-code summarize) and session-complete are never called. Ongoing observation
  recording works fine.

  Suggested Fix

  Use absolute resolved paths at install time, or fall back to a path relative to ~/.claude/plugins/ for Stop hook
  commands.

zerone0x:
## Problem

All three Stop hooks fail silently with `No stderr output` because `${CLAUDE_PLUGIN_ROOT}` is **not injected** by Claude Code when executing Stop hooks.

The commands resolve to `node "/scripts/bun-runner.js" ...` (empty path prefix), producing:
```
Stop hook error: Failed with non-blocking status code: No stderr output
```

This means **session-end summarization never runs** — no memory is saved at session completion.

| Hook | Status |
|------|--------|
| SessionStart | ✅ `CLAUDE_PLUGIN_ROOT` injected |
| UserPromptSubmit | ✅ `CLAUDE_PLUGIN_ROOT` injected |
| PostToolUse | ✅ `CLAUDE_PLUGIN_ROOT` injected |
| **Stop** | ❌ `CLAUDE_PLUGIN_ROOT` is **empty** |

## Root Cause

The Claude Code plugin host does not set `CLAUDE_PLUGIN_ROOT` in the environment when running Stop hooks. This is a Claude Code-side limitation. The fix must work without that variable.

## Fix

Three-part change:

### 1. `plugin/scripts/stop-hook.sh` (new file)
A small wrapper that resolves the plugin root without relying on `CLAUDE_PLUGIN_ROOT`:
- Uses `CLAUDE_PLUGIN_ROOT` when available (forward-compatible)
- Falls back to `~/.claude-mem/.plugin-root` (written by setup.sh)

### 2. `plugin/scripts/setup.sh` (+6 lines)
After resolving `ROOT` (the Setup hook always receives `CLAUDE_PLUGIN_ROOT` correctly):
- Writes `ROOT` to `$CLAUDE_MEM_DATA_DIR/.plugin-root`
- Deploys `stop-hook.sh` to `$CLAUDE_MEM_DATA_DIR/stop-hook.sh`

### 3. `plugin/hooks/hooks.json`
Stop hook commands now call:
```
bash "$HOME/.claude-mem/stop-hook.sh" <args>
```
This path is stable and does not depend on `CLAUDE_PLUGIN_ROOT` being injected.

## Design Rationale

- **No hardcoded paths**: `stop-hook.sh` reads the root from `~/.claude-mem/.plugin-root` which setup.sh writes at install/update time.
- **Forward compatible**: If Claude Code ever fixes the injection for Stop hooks, `stop-hook.sh` uses `CLAUDE_PLUGIN_ROOT` first.
- **Graceful failure**: If the file is missing, the hook exits 0 (non-blocking) instead of crashing.

Fixes #1215

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -78,17 +78,17 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js\" \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" start",
+            "command": "bash \"$HOME/.claude-mem/stop-hook.sh\" start < /dev/null",
             "timeout": 60
           },
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js\" \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" hook claude-code summarize",
+            "command": "bash \"$HOME/.claude-mem/stop-hook.sh\" hook claude-code summarize",
             "timeout": 120
           },
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js\" \"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs\" hook claude-code session-complete",
+            "command": "bash \"$HOME/.claude-mem/stop-hook.sh\" hook claude-code session-complete",
             "timeout": 30
           }
         ]
--- a/plugin/scripts/setup.sh
+++ b/plugin/scripts/setup.sh
@@ -14,6 +14,14 @@ else
   ROOT="$CLAUDE_PLUGIN_ROOT"
 fi
 
+# Persist the resolved plugin root so Stop hooks can find it without
+# CLAUDE_PLUGIN_ROOT (which Claude Code does not inject in Stop contexts).
+CLAUDE_MEM_DIR="${CLAUDE_MEM_DATA_DIR:-$HOME/.claude-mem}"
+mkdir -p "$CLAUDE_MEM_DIR"
+printf '%s' "$ROOT" > "$CLAUDE_MEM_DIR/.plugin-root"
+cp -f "$ROOT/scripts/stop-hook.sh" "$CLAUDE_MEM_DIR/stop-hook.sh" 2>/dev/null || true
+chmod +x "$CLAUDE_MEM_DIR/stop-hook.sh" 2>/dev/null || true
+
 MARKER="$ROOT/.install-version"
 PKG_JSON="$ROOT/package.json"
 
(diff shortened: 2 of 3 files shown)


thedotmack:
Thanks for the investigation! You correctly identified that `CLAUDE_PLUGIN_ROOT` isn't injected for Stop hooks on macOS (anthropics/claude-code#24529).

This was fixed in #1229 (merged) with a two-layer approach:

1. **Shell-level**: All hook commands in `hooks.json` now include an inline fallback: `_R="${CLAUDE_PLUGIN_ROOT}"; [ -z "$_R" ] && _R="$HOME/.claude/plugins/marketplaces/thedotmack";` — no external wrapper script needed.
2. **Script-level**: `bun-runner.js` gained self-resolution via `import.meta.url` and `fixBrokenScriptPath()` to handle the empty expansion case.

Your approach of deploying a `stop-hook.sh` wrapper via setup.sh was a solid idea. We went with the inline fallback to avoid the extra file deployment step and keep it self-contained.

Issue #1235, opened 2026-02-25
Stop hook error loop when Claude Code runs in a git worktree


allixsenos:
## Bug Description

When Claude Code runs in a git worktree (e.g., via the `Task` tool with `isolation: "worktree"`), the `summarize` Stop hook enters an infinite error loop, making the session unusable.

## Error Message

```
Stop hook error: [node "${CLAUDE_PLUGIN_ROOT}/scripts/bun-runner.js"
"${CLAUDE_PLUGIN_ROOT}/scripts/worker-service.cjs" hook claude-code summarize]: Hook error: Error:
Transcript path missing or file does not exist: /Users/.../.claude/projects/-Users-...-dev-rmbug-core--claude-worktrees-feat-whoami-420/ba9f91ba-4c97-47cf-ada9-0beb9fb7144c.jsonl
```

## Root Cause

1. Claude Code derives the project directory from the current working directory (CWD)
2. When CWD is inside a worktree (e.g., `.claude/worktrees/feat-whoami-420/`), Claude Code encodes this into the `transcript_path` passed to hooks
3. The actual transcript `.jsonl` file lives under the **main** project path (`-Users-...-dev-rmbug-core/`), not the worktree-derived path (`-Users-...-dev-rmbug-core--claude-worktrees-feat-whoami-420/`)
4. The `Ute` function in `worker-service.cjs` **throws** when the file doesn't exist at the given path, instead of returning gracefully
5. This causes the Stop hook to error on every assistant turn → Claude sees the error → responds → triggers another Stop hook → infinite loop

## Expected Behavior

The summarize hook should handle missing transcript files gracefully (e.g., log a warning and return early) rather than throwing an error. The Stop hook already handles a missing `transcriptPath` *field* gracefully (early return), but not the case where the path is provided but points to a non-existent file.

## Reproduction Steps

1. Install claude-mem plugin
2. Start a Claude Code session in a git repository
3. Use the `Task` tool with `isolation: "worktree"` to run work in a worktree
4. After the worktree agent completes, every subsequent assistant response triggers the Stop hook error

## Environment

- claude-mem version: 10.2.5 (also checked changelogs through 10.4.3 — no fix present)
- Claude Code: latest
- OS: macOS (Darwin 25.1.0, arm64)

## Suggested Fix

In the transcript reading function (`Ute` in the minified source), change the throw to a graceful return:

```javascript
// Before:
if (!t || !existsSync(t))
[...]

PR #1264, opened 2026-02-28
fix: handle missing transcript in git worktrees

### LINKED ISSUE #1234: Stop hook crashes in git worktrees: transcript path not found
## Bug

When using Claude Code's `EnterWorktree` feature, the `Stop > summarize` hook crashes on **every turn** with:

```
Hook error: Error: Transcript path missing or file does not exist:
/Users/xxx/.claude/projects/-Users-xxx-my-project--claude-worktrees-worktree-name/session-id.jsonl
```

This produces an unending stream of error messages that cannot be stopped until the session ends.

## Root cause

The `Ute()` function in `worker-service.cjs` (v10.2.3) throws an error when the transcript file doesn't exist:

```js
function Ute(t, e, r = false) {
  if (!t || !existsSync(t))
    throw new Error('Transcript path missing or file does not exist: ' + t);
  // ...
}
```

When Claude Code enters a **worktree**, the CWD changes to `.claude/worktrees/<name>/`. The `transcriptPath` provided to the hook is derived from this new CWD, which maps to a different directory under `~/.claude/projects/`. The actual transcript `.jsonl` file lives in the **original** project's directory, not the worktree's derived path.

The `summarize` hook executor does handle `!n` (null transcriptPath) gracefully but does NOT handle the case where the path is provided but the file doesn't exist — it lets `Ute()` throw, which propagates as a hook error on every single turn.

## Expected behavior

The `Ute()` function should return `""` (empty string) instead of throwing when the file doesn't exist, consistent with how the hook already handles null `transcriptPath`. This would make the summarize hook silently skip when transcript is unavailable (e.g., in worktrees).

## Suggested fix

```diff
- if (!t || !existsSync(t)) throw new Error('Transcript path missing...');
+ if (!t || !existsSync(t)) return "";
```

## Workaround

Manually patch `~/.claude/plugins/cache/thedotmack/claude-mem/10.2.3/scripts/worker-service.cjs` replacing the `throw` with `return ""`. Note: this patch will be overwritten on plugin update.

## Environment

- claude-mem: v10.2.3
- Claude Code: latest (CLI)
- OS: macOS 15 (Darwin 25.2.0)

64andrewwalker:
## Summary

Fixes #1234, #1235

When Claude Code runs in a git worktree (via `EnterWorktree` or `Task` with `isolation: "worktree"`), the `transcript_path` passed to the Stop hook is derived from the worktree CWD. This encodes to a different `~/.claude/projects/` directory than where transcripts are actually stored, causing `extractLastMessage()` to throw on every turn.

**Example:**
- Worktree CWD: `/project/.claude/worktrees/my-feature/`
- Encoded path: `~/.claude/projects/-project--claude-worktrees-my-feature/session.jsonl` (doesn't exist)
- Actual path: `~/.claude/projects/-project/session.jsonl`

## Changes

**`src/shared/transcript-parser.ts`:**
1. When transcript path doesn't exist, try stripping `--claude-worktrees-<name>` from the encoded project directory and retry
2. If still not found, return `""` instead of throwing — consistent with how `summarizeHandler` already handles null `transcriptPath`

## Why this approach

- **Minimal change** — single file, single function
- **Uses existing pattern** — the summarize handler already returns gracefully when `transcriptPath` is null; this extends that to the "file not found" case
- **Worktree fallback is additive** — only triggers when the original path fails AND the path contains the worktree encoding pattern
- The existing `detectWorktree()` utility in `src/utils/worktree.ts` detects worktrees from the filesystem, but the transcript path problem occurs in the encoded project directory name, so regex on the path is the right approach here

## Test plan

- [x] Verified regex correctly strips worktree segment from encoded paths
- [x] Tested locally — stop hook no longer errors in worktree sessions
- [x] Non-worktree sessions unaffected (regex doesn't match, original path works)

### DIFF
--- a/src/shared/transcript-parser.ts
+++ b/src/shared/transcript-parser.ts
@@ -1,5 +1,5 @@
-import { readFileSync, existsSync } from 'fs';
-import { logger } from '../utils/logger.js';
+import { readFileSync, existsSync } from "fs";
+import { logger } from "../utils/logger.js";
 
 /**
  * Extract last message of specified role from transcript JSONL file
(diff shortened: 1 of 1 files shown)


Chriscross475 (approved):
LGTM - Clean worktree transcript path fallback fix. Regex correctly strips the worktree encoding segment, graceful degradation with empty string return, and good logging. The quote style changes are cosmetic (formatter).

Chriscross475 (approved):
LGTM - Clean fix for worktree transcript path resolution. The fallback logic correctly strips the worktree encoding pattern and gracefully degrades to empty string if still not found. Tested and approved.

Chriscross475 (approved):
LGTM - Clean fix for worktree transcript path issue. The fallback logic is sound and the approach is minimal and defensive.

Chriscross475 (approved):
LGTM - Clean worktree path resolution with proper fallback logic and logging. Regex pattern is correct and defensive checks are in place.

Chriscross475 (approved):
LGTM - Excellent fix for worktree transcript path resolution. The fallback logic is solid, error handling is proper, and the regex correctly strips the worktree segment. CI passing.

Chriscross475 (approved):
LGTM - Clean worktree path fallback logic. Minimal change with proper error handling. The regex-based path stripping is the right approach for encoded project directories.

thedotmack:
Closing: superseded by PR #1326 which includes the same graceful transcript handling plus worktree-specific path resolution logic. #1326 has been merged.

PR #1282, opened 2026-03-03
fix: remove parallel hook groups causing 'Sibling tool call errored' (v10.5.3)

paxcalpt:
## Problem

The `d93bde05` revert accidentally restored two broken patterns from the old hooks.json:

1. **Setup hook pointing to deleted `setup.sh`** — `setup.sh` was removed in `74d94aa2` but the revert brought back the `Setup` hook referencing it, causing failures during plugin install.

2. **Two parallel `SessionStart` hook groups** — the old structure had:
   - Group 1: `[smart-install]`
   - Group 2: `[worker-start, context-hook]`
   
   Claude Code runs multiple hook groups for the same event concurrently. When `smart-install` (Group 1) exited non-zero for any reason, the sibling `context-hook` (Group 2) would get tagged with **"Sibling tool call errored"** even though it succeeded and properly returned "Recalled N memories". Users saw this error on every session start.

## Fix

Restore the clean hook structure introduced in `74d94aa2` (which was already present in the `0e502dbd` smart-explore commit but lost in the revert):

- **Remove `Setup` hook** — `setup.sh` is gone, no longer needed
- **Merge into a single sequential `SessionStart` group** — `[smart-install → context-hook]` run sequentially, eliminating the parallel sibling problem
- **Remove separate `worker-start` step** — worker lifecycle is handled by `bun-runner.js`/`smart-install.js` already
- **Remove redundant `PostToolUse` `matcher: "*"`** — no matcher means match all, same behavior
- **Use `${CLAUDE_PLUGIN_ROOT:-fallback}` syntax** — cleaner bash parameter expansion
- **Bump `session-complete` Stop hook timeout to 120s** — consistent with summarize hook

## Testing

Worker health confirmed after rebuild: `{"status":"ok"}`.
The single sequential group ensures hooks no longer run as parallel siblings.


paxcalpt:
Closing - will not open upstream PRs

PR #1286, opened 2026-03-05
fix: always pass --ssl flag to chroma-mcp in remote mode

vnz:
## Summary

- `buildCommandArgs()` only pushed `--ssl` when `CLAUDE_MEM_CHROMA_SSL` was `true`, but the `chroma-mcp` CLI **defaults to SSL for HTTP clients**
- When running a local ChromaDB server without SSL (the common `remote` mode setup), the omitted flag caused `chroma-mcp` to attempt an SSL handshake against a plain HTTP server, failing with `[SSL: RECORD_LAYER_FAILURE]` and `"Could not connect to a Chroma server"`
- The worker then entered a connection backoff loop, preventing all vector sync operations

## Fix

Always pass `--ssl` with an explicit `true` or `false` value, so the user's `CLAUDE_MEM_CHROMA_SSL` setting is faithfully forwarded to `chroma-mcp`:

```diff
-      if (chromaSsl) {
-        args.push('--ssl');
-      }
+      args.push('--ssl', chromaSsl ? 'true' : 'false');
```

## Regression tests

Added `tests/services/sync/chroma-mcp-manager-ssl.test.ts` with 4 focused test cases that intercept `StdioClientTransport` to verify `buildCommandArgs()` output:

1. `CLAUDE_MEM_CHROMA_SSL=false` → args contain `--ssl`, `false`
2. `CLAUDE_MEM_CHROMA_SSL=true` → args contain `--ssl`, `true`
3. `CLAUDE_MEM_CHROMA_SSL` not set → defaults to `--ssl`, `false`
4. Local mode (`CLAUDE_MEM_CHROMA_MODE=local`) → no `--ssl` flag at all

## Bundle rebuild

Rebuilt checked-in `plugin/scripts/worker-service.cjs` (and other bundles) so the fix is present in the runtime artifacts that hooks, `worker:start`, and marketplace installs actually execute.

## Reproduction

1. Install ChromaDB as a local HTTP server: `chroma run --host 127.0.0.1 --port 8000`
2. Set `CLAUDE_MEM_CHROMA_MODE=remote`, `CLAUDE_MEM_CHROMA_HOST=127.0.0.1`, `CLAUDE_MEM_CHROMA_PORT=8000` in `settings.json`
3. Restart worker — observe `chroma-mcp connection in backoff` errors in logs
4. Root cause: `uvx chroma-mcp --client-type http --host 127.0.0.1 --port 8000` (no `--ssl` flag) defaults to SSL

## Test plan

- [x] `bun test` — 1070 pass, 40 fail, 3 skip (identical to `main`, zero regressions)
- [x] 4 new SSL regression tests pass
- [x] Manual verification: `uvx chroma-mcp --client-type http --host 127.0.0.1 --port 8000 --ssl false` connects successfully
- [x] Confirmed worker syncs observations to ChromaDB after the fix (291 new embeddings)
- [x] Verified `worker-service.cjs` bundle contains `l.push("--ssl",s?"true":"false")`


xkonjin (commented):
Quick review pass:
- Main risk area here is edge cases on the new/changed execution path.
- I didn’t see targeted regression coverage in the diff; please add or point CI at a focused test for the changed path in ChromaMcpManager.ts.
- Before merge, I’d smoke-test the behavior touched by ChromaMcpManager.ts with malformed input / retry / rollback cases, since that’s where this class of change usually breaks.

vnz:
@xkonjin Good catch — the source fix wasn't reflected in the checked-in `plugin/scripts/worker-service.cjs` that the runtime actually executes.

Addressed in `e2fd324d`: rebased on `main` and rebuilt all bundles. The worker-service bundle now contains `l.push("--ssl",s?"true":"false")` matching the source change at `ChromaMcpManager.ts:204`.

Also added the regression tests you requested in `0256ed3d` — 4 test cases covering SSL=true, SSL=false, unset (defaults false), and local mode (no --ssl flag).

PR #1291, opened 2026-03-05
fix: remove unrecognized fields from Claude Code Stop hook output

### LINKED ISSUE #1288: Stop hook loops infinitely due to {"continue":true} response
## Bug Description

The stop hook (`worker-service.cjs hook claude-code summarize`) causes an infinite loop at session end. Claude Code keeps re-triggering the stop hook, requiring manual interruption (Ctrl+C).

## Environment

- Claude Code version: latest (as of 2026-03-05)
- Plugin version: 10.4.1 (manifest) / 10.4.3 (package.json)
- OS: macOS (Darwin 24.6.0)
- Node: available, Bun: available

## Steps to Reproduce

1. Enable claude-mem plugin
2. Use Claude Code normally
3. When the session ends or Claude stops responding, observe the stop hooks

## Observed Behavior

The stop hook fires repeatedly in an infinite loop:

```
⏺ Ran 3 stop hooks (ctrl+o to expand)
  ⎿ Stop hook error: [...worker-service.cjs hook claude-code summarize]: No stderr output

⏺ Ran 3 stop hooks (ctrl+o to expand)
  ⎿ Stop hook error: [...worker-service.cjs hook claude-code summarize]: No stderr output

(repeats indefinitely until interrupted)
```

## Root Cause Analysis

Running the hook manually shows it returns:

```json
{"continue":true,"suppressOutput":true}
```

The `"continue":true` field appears to signal Claude Code that the session should continue, which triggers another round of stop hooks, creating an infinite loop.

## Expected Behavior

The stop hook should return `{"continue":false}` (or omit the field) so Claude Code terminates cleanly after the summarize hook completes.

## Workaround

Disable the plugin in `~/.claude/settings.json`:

```json
"enabledPlugins": {
}
```

### LINKED ISSUE #1290: Stop hook output causes 'JSON validation failed' in Claude Code
## Bug Description

The Claude Code adapter's `formatOutput()` returns `{"continue":true,"suppressOutput":true}` for Stop hook events. These fields are not part of Claude Code's hook API schema. Claude Code validates Stop hook output against `{decision?, reason?, systemMessage?}` and rejects unrecognized fields, producing:

```
Stop hook error: JSON validation failed
```

This affects both the `summarize` and `session-complete` Stop hooks on every session stop.

## Steps to Reproduce

1. Install claude-mem v10.5.2
3. End the session (let Claude stop naturally)
4. Observe in the stop hooks output:
   ```
   ⏺ Ran 7 stop hooks
     ⎿  ...
     ⎿  Stop hook error: JSON validation failed
   ```

## Root Cause

In `src/cli/adapters/claude-code.ts`, the `formatOutput()` method returns `{continue, suppressOutput}` for the default case (line 26). These fields are internal to claude-mem and not part of the Claude Code hook contract. While Claude Code silently ignores unrecognized fields for other hook events (SessionStart, PostToolUse), the **Stop hook validates output strictly** and rejects these unknown fields.

## Expected Behavior

Stop hooks should return an empty `{}` (allow stop) or `{systemMessage: "..."}` when no blocking action is needed. No unrecognized fields should be included.

## Environment

- **Claude-mem version**: 10.5.2
- **Claude Code version**: Latest (March 2026)
- **OS**: macOS 15.4
- **Platform**: Claude Code CLI

AlexWorland:
## Summary

Fixes #1290
Fixes #1288

The Claude Code adapter's `formatOutput()` was returning `{continue: true, suppressOutput: true}` for all hook events. These fields are not part of the Claude Code hook API:
- **`continue: true`** causes Claude Code to re-invoke stop hooks in an infinite loop (#1288)
- **Both fields** fail Stop hook JSON schema validation with "JSON validation failed" (#1290)

## Changes

- **`src/cli/adapters/claude-code.ts`**: Changed `formatOutput()` to return `{}` (empty object) for the default case instead of `{continue, suppressOutput}`. Preserves `systemMessage` when present. Added defensive null guard (`result ?? {}`) matching the existing `normalizeInput` pattern. The `hookSpecificOutput` path (used by SessionStart) is unchanged.
- **`tests/hook-lifecycle.test.ts`**: Updated Claude Code adapter tests to assert the corrected output format. Added unhappy-path edge case tests per review feedback.

## Why this is safe

- `{}` is valid for all Claude Code hook event types (Stop, SessionStart, PostToolUse, etc.)
- The `continue` and `suppressOutput` fields were never part of Claude Code's hook contract — they were silently ignored for non-Stop events
- The `hookSpecificOutput` path used by SessionStart context injection is unaffected
- All 69 existing hook tests pass across 2 test files

## Edge case tests added (per review)

| Test | What it covers |
|------|---------------|
| Malformed input (undefined/null) | `formatOutput(undefined)` and `formatOutput(null)` return `{}` |
| Falsy systemMessage values | Empty string, null, and 0 are excluded from output |
| All HookResult fields populated | Only `systemMessage` survives; `continue`, `suppressOutput`, `exitCode` stripped |
| hookSpecificOutput without systemMessage | Branch A works correctly without optional field |
| Contract key allowlist | Output keys are always a subset of `{hookSpecificOutput, systemMessage, decision, reason}` |

## Verification

Built from source and compared output of the installed (broken) vs built (fixed) plugin:

| Test | Installed v10.5.2 (broken) | Built (fixed) |
|------|--------------------|---------------|
| `summarize` Stop hook | `{"continue":true,"suppressOutput":true}` | `{}` |
| `session-complete` Stop hook | `{"continue":true,"suppressOutput":true}` | `{}` |
| SessionStart `context` hook | `{hookSpecificOutput: ...}` | `{hookSpecificOutput: ...}` (unchanged) |
| Exit code | 0 | 0 |

```
$ bun test tests/hook-lifecycle.test.ts tests/hook-command.test.ts

 69 pass
 0 fail
 107 expect() calls
Ran 69 tests across 2 files.
```

## Test plan

- [x] `bun test tests/hook-lifecycle.test.ts tests/hook-command.test.ts` — 69 tests pass
- [x] Built plugin from source, verified Stop hooks output `{}` instead of `{"continue":true,"suppressOutput":true}`
- [x] Verified `hookSpecificOutput` path still works for SessionStart context injection
- [x] Verified `systemMessage` is preserved when present
- [x] Compared installed v10.5.2 output vs built output side-by-side
- [x] Added unhappy-path tests for malformed input, falsy values, and contract compliance

### DIFF
--- a/src/cli/adapters/claude-code.ts
+++ b/src/cli/adapters/claude-code.ts
@@ -16,20 +16,19 @@ export const claudeCodeAdapter: PlatformAdapter = {
     };
   },
   formatOutput(result) {
-    if (result.hookSpecificOutput) {
+    const r = result ?? ({} as HookResult);
+    if (r.hookSpecificOutput) {
       const output: Record<string, unknown> = { hookSpecificOutput: result.hookSpecificOutput };
-      if (result.systemMessage) {
-        output.systemMessage = result.systemMessage;
+      if (r.systemMessage) {
+        output.systemMessage = r.systemMessage;
       }
       return output;
     }
-    // Return only fields from the Claude Code hook contract.
-    // Stop hooks validate against {decision?, reason?, systemMessage?} and reject
-    // unrecognized fields like `continue` or `suppressOutput` with
-    // "JSON validation failed". An empty object is valid for all hook types.
+    // Only emit fields in the Claude Code hook contract — unrecognized fields
+    // cause "JSON validation failed" in Stop hooks.
     const output: Record<string, unknown> = {};
-    if (result.systemMessage) {
-      output.systemMessage = result.systemMessage;
+    if (r.systemMessage) {
+      output.systemMessage = r.systemMessage;
     }
     return output;
   }
--- a/tests/hook-lifecycle.test.ts
+++ b/tests/hook-lifecycle.test.ts
@@ -256,43 +256,74 @@ describe('Cursor IDE Compatibility (#838, #1049)', () => {
 // --- Platform Adapter Tests ---
 
 describe('Hook Lifecycle - Claude Code Adapter', () => {
-  it('should return empty object for default result (no unrecognized fields)', async () => {
+  const fmt = async (input: any) => {
     const { claudeCodeAdapter } = await import('../src/cli/adapters/claude-code.js');
+    return claudeCodeAdapter.formatOutput(input);
+  };
 
-    // Empty result should produce empty output — Claude Code Stop hooks
-    // reject unrecognized fields like `continue` or `suppressOutput`
-    const output = claudeCodeAdapter.formatOutput({});
-    expect(output).toEqual({});
+  // --- Happy paths ---
+
+  it('should return empty object for empty result', async () => {
+    expect(await fmt({})).toEqual({});
   });
 
-  it('should not include continue or suppressOutput in output', async () => {
-    const { claudeCodeAdapter } = await import('../src/cli/adapters/claude-code.js');
-
-    const output = claudeCodeAdapter.formatOutput({ continue: true, suppressOutput: true }) as Record<string, unknown>;
-    expect(output.continue).toBeUndefined();
-    expect(output.suppressOutput).toBeUndefined();
+  it('should include systemMessage when present', async () => {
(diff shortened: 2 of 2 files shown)


AlexWorland (commented):
@xkonjin Thanks for the review\! Good call on adding more tests. Done.

PR #1298, opened 2026-03-08
feat: use git root for consistent project name detection

Boltthecat19:
## Problem

`getProjectName()` uses `path.basename(cwd)` to identify the current project. This produces inconsistent names when Claude is invoked from a subdirectory — e.g. working in `my-project/src/utils` yields a project name of `utils`, not `my-project`. Running from the home directory yields the username.

This causes observations and memory to be siloed under incorrect project names, breaking cross-session context.

## Solution

Add a `findGitRoot()` utility (`src/utils/git-root.ts`) that walks up the directory tree from `cwd` looking for a `.git` entry (directory or file, to support worktrees). The git root directory name is used as the project identifier. Non-git directories fall back to the existing `path.basename(cwd)` behavior.

## Changes

- `src/utils/git-root.ts` — new utility, walks cwd upward to find `.git`
- `src/utils/project-name.ts` — prefer git root name, fall back to basename
- `tests/utils/git-root.test.ts` — 6 unit tests: null input, no git, cwd root, subdirectory, worktree (.git file)
- `tests/utils/project-name.test.ts` — 5 unit tests covering all branches

## Test Results

All 11 new tests pass in isolation and with `--no-parallel`.

> **Note:** The full `bun test` suite shows 5 spurious failures for `getProjectName` caused by a pre-existing `mock.module()` leak in `tests/hooks/context-reinjection-guard.test.ts:32` that pollutes parallel test workers. This is unrelated to this PR and will be addressed separately.

### DIFF
--- a/src/utils/git-root.ts
+++ b/src/utils/git-root.ts
@@ -1,0 +1,36 @@
+/**
+ * Git Root Detection Utility
+ *
+ * Walks up the directory tree from a given path to find the nearest
+ * git repository root (directory containing a .git entry).
+ */
+import { statSync } from 'fs';
+import path from 'path';
+
+/**
+ * Find the git repository root by walking up the directory tree.
+ *
+ * @param cwd - Starting directory (absolute path)
+ * @returns Absolute path to the git root, or null if not in a git repo
+ */
+export function findGitRoot(cwd: string | null | undefined): string | null {
+  if (!cwd || cwd.trim() === '') return null;
+
+  let current = path.resolve(cwd);
+
+  while (true) {
+    const gitPath = path.join(current, '.git');
+    try {
+      statSync(gitPath);
+      // .git exists (file or directory) - this is the repo root
+      return current;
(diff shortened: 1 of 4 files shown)


xkonjin (commented):
Quick review pass:
- Main risk area here is input validation, path handling, and malformed payload behavior.
- Good to see test coverage move with the code; I’d still make sure it exercises the unhappy path around input validation, path handling, and malformed payload behavior rather than only the happy path.
- Before merge, I’d smoke-test the behavior touched by git-root.ts, project-name.ts, git-root.test.ts (+1 more) with malformed input / retry / rollback cases, since that’s where this class of change usually breaks.

dor-tzur-lmnd:
Hey! is it possible to merge this?  I have like 10 worktrees fragmented memories, currently patching the db. Would be super helpful for us tree-huggers.  thank you! 🙏🏻

thedotmack:

PR #1306, opened 2026-03-09
fix: handle missing session_id when Cursor runs claude-code hooks

Zubairu:
## Problem

When Cursor IDE runs Claude Code-compatible hooks, it sends `conversation_id` instead of `session_id` in the hook input. The `UserPromptSubmit` hook (`new-hook.ts`) reads `session_id` directly, so it becomes `undefined`, and the subsequent call to `/api/sessions/init` fails with HTTP 400 because `contentSessionId` is required.

This causes Cursor to show a blocking error dialog:

> **Submission blocked by hook**
> Hook blocked with message: Hook error: Error: Session initialization failed: 400

The error happens every time a user submits a prompt in Cursor's built-in agent when claude-mem hooks are active.

## Root Cause

`new-hook.ts` interface and destructuring only reads `session_id`:
```ts
// Before
const { session_id, cwd, prompt } = input;
```

Cursor provides `conversation_id`, not `session_id`, so `contentSessionId` sent to the worker is `undefined` → 400.

## Fix

Two-part fix:

1. **Accept `conversation_id` as fallback** when `session_id` is absent
2. **Skip gracefully** (return success) when neither ID is available, instead of throwing an error that blocks the submission

```ts
// After
const session_id = input.session_id ?? input.conversation_id;
if (!session_id) {
  console.log(STANDARD_HOOK_RESPONSE);
  return;
}
```

## Testing

Reproduced by using Cursor's built-in agent with claude-mem hooks active. After this fix, Cursor submits prompts without the blocking error.

### DIFF
--- a/src/hooks/new-hook.ts
+++ b/src/hooks/new-hook.ts
@@ -5,24 +5,33 @@ import { getProjectName } from '../utils/project-name.js';
 import { logger } from '../utils/logger.js';
 
 export interface UserPromptSubmitInput {
-  session_id: string;
+  session_id?: string;
+  conversation_id?: string; // Cursor sends conversation_id instead of session_id
   cwd: string;
   prompt: string;
 }
-
 
 /**
  * New Hook Main Logic
  */
 async function newHook(input?: UserPromptSubmitInput): Promise<void> {
-  // Ensure worker is running before any other logic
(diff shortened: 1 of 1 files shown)


thedotmack:

PR #1310, opened 2026-03-09
fix(openclaw): avoid /dev/tty crash in --non-interactive and handle allowlist bootstrap

nillin:
## Summary
This patch fixes two installer failure modes seen during `--upgrade` on OpenClaw:

1. **Headless/non-TTY installs** still touched `/dev/tty` and failed with:
   `main: line 85: /dev/tty: No such device or address`
2. **Config bootstrap deadlock** when `plugins.allow` already includes `claude-mem` but the plugin is not installed yet.
   OpenClaw validates config before `plugins install` can complete, so install fails with:
   `plugins.allow: plugin not found: claude-mem`

## Changes
- `setup_tty()` now short-circuits when `--non-interactive` is set (never opens `/dev/tty`).
- Added safer `/dev/tty` open path (`-r /dev/tty` + guarded `exec`).
- During stale-plugin cleanup before install, also remove `claude-mem` from `plugins.allow` temporarily.
- Restore `plugins.allow` entry after successful install+enable if it existed previously.

## Why this is safe
- Allowlist handling is round-trip: only restored when it existed originally.
- Existing plugin config restoration logic (`entries.claude-mem.config`) remains intact.
- Interactive behavior is unchanged for normal TTY sessions.

## Validation
- `bash -n openclaw/install.sh` passes.
- Reproduced both failures on a live OpenClaw host, applied this logic, and confirmed upgrade completes and memory plugin loads.

### DIFF
--- a/openclaw/install.sh
+++ b/openclaw/install.sh
@@ -80,17 +80,24 @@ setup_tty() {
   if [[ -t 0 ]]; then
     # stdin IS a terminal — use it directly
     TTY_FD=0
-  elif [[ -e /dev/tty ]]; then
-    # stdin is piped (curl | bash) but /dev/tty is available
-    exec 3</dev/tty
-    TTY_FD=3
-  else
-    # No terminal available at all
-    if [[ "$NON_INTERACTIVE" != "true" ]]; then
-      echo "Error: No terminal available for interactive prompts." >&2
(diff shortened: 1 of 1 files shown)


thedotmack:

PR #1334, opened 2026-03-11
fix: prevent infinite restart loop on FK constraint errors

laihenyi:
## Summary
- Add `FOREIGN KEY constraint failed` to unrecoverable error patterns so the session generator stops immediately instead of endlessly retrying
- Add `MAX_PENDING_RESTARTS` (3) safety limit to `pending-work-restart` path to prevent infinite loops from any future unhandled persistent errors

## Problem
When pending messages reference deleted FK records, the session generator throws `FOREIGN KEY constraint failed`. This error was not in the unrecoverable patterns list, so it fell through to the `pending-work-restart` logic which unconditionally restarts when `pendingCount > 0` — creating an infinite loop.

Observed impact: **2,228 error log entries** in a single session, worker eventually killed by SIGTERM, unable to auto-restart afterward.

## Test plan
- [ ] Verify worker starts normally after fix
- [ ] Confirm FK errors no longer trigger restart loop (check logs for `Exceeded max pending-work restarts` or `Unrecoverable generator error`)
- [ ] Verify normal pending-work-restart still works (counter resets on success)

### DIFF
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -561,6 +561,7 @@ export class WorkerService {
           'ENOENT',
           'spawn',
           'Invalid API key',
+          'FOREIGN KEY constraint failed',
         ];
         if (unrecoverablePatterns.some(pattern => errorMessage.includes(pattern))) {
           hadUnrecoverableError = true;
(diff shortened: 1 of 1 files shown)


xkonjin:
## Code Review

### Summary
Good fix for the infinite restart loop issue. Adding FK constraint to unrecoverable error patterns and tracking consecutive restarts is a solid defensive approach.

### Strengths
- Clear error pattern addition for FK constraints
- MAX_PENDING_RESTARTS provides reasonable throttling
- Proper logging at error level when threshold exceeded
- Counter reset on successful completion
- Detailed logging includes restart count

### Concerns
1. Session state mutation: session.consecutiveRestarts is added without TypeScript type annotation. Could cause runtime errors if session objects are strictly typed elsewhere.

2. Potential race condition: If two pending-work restarts are triggered concurrently, the counter could increment faster than expected.

3. Silent reset: When MAX_PENDING_RESTARTS is exceeded, the counter is reset before returning, losing debugging context.

### Suggestions
1. Add TypeScript interface update for the consecutiveRestarts property
2. Consider logging state before reset for debugging
3. Consider adding a cooldown period before allowing restarts again

### Test Coverage
Would be good to add tests for FK constraint error detection, counter logic, and threshold exceeded behavior.

### Recommendation
Approve with minor changes. The core logic is sound, just needs type safety and optionally more robust state management.

PR #1353, opened 2026-03-13
fix: detect parent process death via stdin EOF to prevent MCP server leaks

### LINKED ISSUE #1351: MCP server processes leak indefinitely when Claude Code sessions end
## Summary

When a Claude Code session ends (terminal close, crash, network disconnect, or normal exit), the spawned `mcp-server.cjs` process is never cleaned up. Processes accumulate indefinitely, consuming memory and file descriptors.

## Root Cause

Two-layer failure:

1. **MCP SDK gap**: `StdioServerTransport` listens for `data` and `error` events on stdin but never registers `end` or `close` handlers. When the parent process dies and stdin receives EOF, the transport is oblivious and the Node event loop stays alive.

2. **mcp-server.ts signal-only cleanup**: The server only registers `SIGTERM` and `SIGINT` handlers. When a session drops abruptly (terminal close, crash, network disconnect), no signal is delivered — the stdin pipe/socket simply closes silently. The node process remains alive indefinitely in a sleeping state with closed socket file descriptors.

## Reproduction

1. Open a Claude Code session (spawns `mcp-server.cjs` via hooks)
2. Close the terminal window or kill the Claude Code process
3. Run `ps aux | grep mcp-server.cjs` — the orphaned process is still alive
4. Repeat across sessions — processes accumulate without bound

On my system I found an orphaned `mcp-server.cjs` from 2 days prior still running.

## Fix

The fix is straightforward — add stdin EOF detection and a re-entrancy guard:

```diff
--- a/src/servers/mcp-server.ts
+++ b/src/servers/mcp-server.ts
@@ -282,20 +282,33 @@
   }
 });
 
-// Cleanup function
-async function cleanup() {
-  logger.info('SYSTEM', 'MCP server shutting down');
+// Cleanup function - guard against double-exit
+let isShuttingDown = false;
+async function cleanup(reason?: string) {
+  if (isShuttingDown) return;
+  isShuttingDown = true;
+  logger.info('SYSTEM', `MCP server shutting down${reason ? ` (${reason})` : ''}`);
   process.exit(0);
 }
 
 // Register cleanup handlers for graceful shutdown
-process.on('SIGTERM', cleanup);
-process.on('SIGINT', cleanup);
+process.on('SIGTERM', () => cleanup('SIGTERM'));
+process.on('SIGINT', () => cleanup('SIGINT'));
+
+// Detect parent process death: when the parent (Claude Code) exits or crashes,
+// stdin gets EOF/close. The MCP SDK's StdioServerTransport does NOT listen for
+// these events, so without this handler the process leaks indefinitely.
+process.stdin.on('end', () => cleanup('stdin end'));
+process.stdin.on('close', () => cleanup('stdin close'));
 
 // Start the server
 async function main() {
   // Start the MCP server
   const transport = new StdioServerTransport();
+
+  // Also handle transport-level close (fires if server.close() is called)
+  transport.onclose = () => cleanup('transport close');
+
   await server.connect(transport);
```

### What the fix does

- **`process.stdin.on('end'/'close')`** — detects when the parent Claude Code process exits and stdin gets EOF
- **`transport.onclose`** — belt-and-suspenders handler for programmatic transport shutdown
- **`isShuttingDown` guard** — prevents double-exit races when multiple shutdown events fire simultaneously
- **`reason` parameter** — enables diagnostic logging to distinguish which path triggered shutdown

## Environment

- Linux (Ubuntu), Node.js
- claude-mem v9.0.5
- Claude Code CLI

## Notes

This is likely a systemic issue affecting any MCP server using `StdioServerTransport` that doesn't add its own stdin EOF detection. The upstream MCP SDK (`@modelcontextprotocol/sdk`) could also fix this by adding `end`/`close` listeners in `StdioServerTransport` itself, but that's a separate issue.

Blevene:
## Summary

- Adds `process.stdin.on('end'/'close')` handlers to detect when the parent Claude Code process exits, triggering graceful shutdown
- Adds `transport.onclose` callback as a secondary detection path
- Adds `isShuttingDown` re-entrancy guard to prevent double-exit races when multiple shutdown events fire simultaneously
- Adds `reason` parameter to `cleanup()` for diagnostic logging

## Problem

When a Claude Code session ends (terminal close, crash, network disconnect), the spawned `mcp-server.cjs` process leaks indefinitely because:

1. **MCP SDK gap**: `StdioServerTransport` only listens for `data` and `error` on stdin — never `end` or `close`
2. **Signal-only cleanup**: The server only registered `SIGTERM`/`SIGINT` handlers, which don't fire on abrupt session drops

Processes accumulate without bound — I found an orphaned instance from 2 days prior still running.

## Test plan

- [ ] Start a Claude Code session, verify `mcp-server.cjs` spawns
- [ ] Close the terminal window — verify the MCP server process exits (check `ps aux | grep mcp-server`)
- [ ] Kill the Claude Code process with `kill -9` — verify the MCP server still exits via stdin EOF
- [ ] Normal `SIGTERM`/`SIGINT` shutdown still works as before

Fixes #1351

### DIFF
--- a/src/servers/mcp-server.ts
+++ b/src/servers/mcp-server.ts
@@ -282,20 +282,33 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
   }
 });
 
-// Cleanup function
-async function cleanup() {
-  logger.info('SYSTEM', 'MCP server shutting down');
+// Cleanup function - guard against double-exit
+let isShuttingDown = false;
+async function cleanup(reason?: string) {
+  if (isShuttingDown) return;
+  isShuttingDown = true;
+  logger.info('SYSTEM', `MCP server shutting down${reason ? ` (${reason})` : ''}`);
   process.exit(0);
 }
 
 // Register cleanup handlers for graceful shutdown
-process.on('SIGTERM', cleanup);
-process.on('SIGINT', cleanup);
+process.on('SIGTERM', () => cleanup('SIGTERM'));
+process.on('SIGINT', () => cleanup('SIGINT'));
+
+// Detect parent process death: when the parent (Claude Code) exits or crashes,
+// stdin gets EOF/close. The MCP SDK's StdioServerTransport does NOT listen for
+// these events, so without this handler the process leaks indefinitely.
+process.stdin.on('end', () => cleanup('stdin end'));
+process.stdin.on('close', () => cleanup('stdin close'));
 
 // Start the server
 async function main() {
   // Start the MCP server
   const transport = new StdioServerTransport();
+
+  // Also handle transport-level close (fires if server.close() is called)
+  transport.onclose = () => cleanup('transport close');
+
   await server.connect(transport);
   logger.info('SYSTEM', 'Claude-mem search server started');
 


thedotmack:
Superseded by the embedded Process Supervisor (PR #1370, v10.5.6). MCP server process cleanup is now handled by the process registry's `reapSession()` method and periodic health sweeps, rather than relying on stdin EOF detection.

PR #1363, opened 2026-03-14
fix: resolve UserPromptSubmit hook error on new session startup

stepcha:
## Problem

When starting a new Claude Code session, users see a `UserPromptSubmit hook error` on their first prompt. This happens because the `UserPromptSubmit` hook fires before `SessionStart` has finished starting the worker daemon, causing a race condition.

**Reproduction steps:**
1. Start a new Claude Code session
2. Type any prompt (e.g., "hola que tal?")
3. Observe `UserPromptSubmit hook error` in the output

The session still works after this error (graceful degradation catches most paths), but the error message is confusing and unnecessary.

## Root Cause

Two issues combine to produce the error:

1. **`session-init.ts`**: `ensureWorkerRunning()` does a **single health check** with no retry. If the worker is still starting up (SessionStart is running `smart-install.js` + `worker-service start`), this check fails immediately.

2. **`worker-service.ts` `case 'hook'`**: The worker startup logic (`ensureWorkerStarted()` + in-process worker start) runs without a top-level try-catch. If `ensureWorkerStarted` spawns a competing daemon while SessionStart is already spawning one, unexpected errors can propagate to `main().catch()` or cause the hook to exit with a non-zero code.

### `src/cli/handlers/session-init.ts`

Replace the single-shot `ensureWorkerRunning()` call with `waitForWorkerWithRetry()` — a polling loop that checks worker health every 500ms for up to 10 seconds. This covers the typical SessionStart startup window without adding significant latency when the worker is already running (returns on first successful check).

### `src/services/worker-service.ts`

Wrap the entire worker startup logic in the `case 'hook'` block with a defensive try-catch. If any unexpected error occurs during daemon startup (port conflicts, PID file races, etc.), the code now falls through to `hookCommand()` which handles worker unavailability gracefully via `isWorkerUnavailableError()`. Also removed the premature `process.exit(0)` on in-process worker start failure — letting `hookCommand` handle it ensures consistent output formatting.

## Testing

- Verified TypeScript compiles without new errors (pre-existing errors in the project are unrelated)
- Manually tested `worker-service.cjs hook claude-code session-init` — returns `{"continue":true,"suppressOutput":true}` correctly
- The retry loop gracefully degrades: if the worker never starts, the handler returns `{continue: true, suppressOutput: true, exitCode: 0}` after 10 seconds

## Risk Assessment

**Low risk.** Both changes are defensive:
- The retry loop only adds latency when the worker is genuinely starting up (otherwise returns on first check)
- The try-catch wrapper doesn't change any success paths — it only catches errors that previously would propagate as unhandled
- No changes to the hook JSON contract or exit code semantics

### DIFF
--- a/src/cli/handlers/session-init.ts
+++ b/src/cli/handlers/session-init.ts
@@ -13,12 +13,34 @@ import { isProjectExcluded } from '../../utils/project-filter.js';
 import { SettingsDefaultsManager } from '../../shared/SettingsDefaultsManager.js';
 import { USER_SETTINGS_PATH } from '../../shared/paths.js';
 
+/**
+ * Wait for the worker to become available, retrying with backoff.
+ * Covers the race condition where SessionStart is still starting the worker
+ * when the first UserPromptSubmit fires (fixes #XXXX).
+ *
+ * @param timeoutMs Maximum time to wait for the worker (default: 10s)
+ * @param intervalMs Polling interval (default: 500ms)
+ */
+async function waitForWorkerWithRetry(timeoutMs: number = 10000, intervalMs: number = 500): Promise<boolean> {
+  const start = Date.now();
+  while (Date.now() - start < timeoutMs) {
+    if (await ensureWorkerRunning()) {
+      return true;
+    }
+    await new Promise(r => setTimeout(r, intervalMs));
+  }
+  return false;
+}
+
 export const sessionInitHandler: EventHandler = {
   async execute(input: NormalizedHookInput): Promise<HookResult> {
-    // Ensure worker is running before any other logic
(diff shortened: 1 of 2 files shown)


stepcha:
Closing this PR — the fix doesn't address the actual root cause. The UserPromptSubmit hook error occurs on every prompt (not just the first), and the session-init handler is never reached according to worker logs. The issue is in a layer below what this PR modifies (likely in the hook execution pipeline itself, not the session-init handler or worker startup logic). Further investigation needed.

lsffup:
when can fix this bug?  the error msg is so distub

Issue #1249, opened 2026-02-26
node→bun grandchild process gets SIGKILL'd in Claude Code sandbox


influenist:
## Problem

Every time claude-mem updates, the `hooks.json` gets overwritten and reverts to using `node bun-runner.js` to spawn `bun worker-service.cjs`. This causes **persistent PostToolUse/UserPromptSubmit/Stop hook errors** because the `bun` grandchild process gets `SIGKILL`'d by Claude Code's sandbox.

## Root Cause

When Claude Code executes a hook command like:
```
node bun-runner.js worker-service.cjs hook claude-code observation
```

The process tree is: `Claude Code → node (bun-runner.js) → bun (worker-service.cjs)`

The `bun` grandchild process receives `SIGKILL` immediately (confirmed via `child.on('close', (code, signal) => ...)` — code is `null`, signal is `SIGKILL`). This appears to be Claude Code's sandbox killing unrecognized grandchild processes.

## Evidence

Tested on macOS (Darwin 23.6.0), Claude Code 2.1.50, claude-mem 10.4.1/10.4.3/10.5.2:

```bash
# Direct bun call — works perfectly ✅
bun worker-service.cjs start
# → {"continue":true,"suppressOutput":true,"status":"ready"}
# → exit code 0

# Through node bun-runner.js — SIGKILL ❌
node bun-runner.js worker-service.cjs start
# → no output
# → child exit code: null, signal: SIGKILL

# Simple bun commands through node — works fine ✅
node -e "spawn('bun', ['--version'], {stdio:'inherit'})"
# → 1.3.7, exit code 0
```

So it's specifically `bun` running a long-ish script (worker-service.cjs) as a grandchild that triggers the kill.

## Workaround

Replace `node bun-runner.js` with direct `bun` calls in `hooks.json`:

```json
// Before (broken):
"command": "node \"$_R/scripts/bun-runner.js\" \"$_R/scripts/worker-service.cjs\" hook claude-code observation"

// After (works):
"command": "bun \"$_R/scripts/worker-service.cjs\" hook claude-code observation"
```

## Problem: Updates Overwrite the Fix

Every plugin update restores the original `hooks.json` with `node bun-runner.js`, breaking hooks again. This has happened across 10.4.1 → 10.4.3 → 10.5.2.

## Suggested Fix

[...]

PR #1370, opened 2026-03-15
feat: add embedded Process Supervisor for unified process lifecycle

thedotmack:
## Summary

- Adds an embedded Process Supervisor (`src/supervisor/`) with signal handling, process registry, shutdown cascade, and health checking
- Session-scoped process reaping ensures child processes are cleaned up when sessions are deleted (#1351)
- Hooks now communicate with worker via `workerHttpRequest()` instead of in-process fallback (#1249)
- New `/api/admin/doctor` diagnostic endpoint for inspecting supervisor state
- Comprehensive test suite (44 tests across 5 files)

**Note:** Unix domain socket transport was added and then reverted in the final commit — TCP on port 37777 remains the only transport since there's only ever one worker process.

## Test plan

- [x] `bun test tests/supervisor/` — 44/44 pass
- [x] `bun test tests/infrastructure/ tests/hooks/ tests/utils/claude-md-utils.test.ts` — 200/200 pass
- [x] `curl http://localhost:37777/api/health` — 200 OK after build-and-sync
- [x] Context injection works on session start
- [ ] Verify `/api/admin/doctor` returns supervisor diagnostics
- [ ] Verify session deletion reaps child processes


thedotmack:
## Review Fixes Applied

Addressed the following issues from the three code reviews:

### Must-Fix (7)
1. **Signal handler race condition** (`src/supervisor/index.ts`) — Added synchronous `shutdownInitiated` boolean flag set before the first `await`, preventing two rapid signals from both entering the shutdown path.
2. **`isPidAlive(0)` returns `true`** (`src/supervisor/process-registry.ts`) — PID 0 sends signals to the entire process group. Now returns `false`.
3. **Duplicate env variable lists** — Exported `ENV_PREFIXES` and `ENV_EXACT_MATCHES` from `env-sanitizer.ts`, imported in `Server.ts` instead of duplicating.
4. **`/api/admin/doctor` returns boolean not PIDs** — Changed `zombiePidFiles: boolean` → `deadProcessPids: number[]` for actual diagnostic value.
5. **Reap failure logged at `debug`** — Bumped to `logger.warn` in `SessionManager.ts` since process reaping failures are operationally significant.
6. **`validateWorkerPidFile` untestable** — Added optional `pidFilePath` parameter (matches `runShutdownCascade` pattern).
7. **Unused `getWorkerPort` imports** — Removed from `observation.ts` and `session-init.ts`.

### Should-Fix (3)
8. **`workerHttpRequest` rebuilds URL inline** — Now calls `buildWorkerUrl()` instead of duplicating URL construction.
9. **Unused `dataDir` in `ShutdownCascadeOptions`** — Removed from interface, implementation, and test call sites.
10. **`validateWorkerPidFile` test asserts nothing** — Rewrote with temp dirs, now tests all 4 return values (`missing`, `invalid`, `stale`, `alive`).

### Test Results
- **46/46 supervisor tests pass** (2 new tests for `validateWorkerPidFile` edge cases)
- No new TypeScript errors introduced

PR #1372, opened 2026-03-16
fix(chroma): close transport before nulling to prevent subprocess leak

### LINKED ISSUE #1369: chroma-mcp subprocess leak: callTool and onclose null transport before connectInternal can close it
## Problem

`ChromaMcpManager.connectInternal()` attempts to close the previous subprocess before spawning a new one:

```js
if (this.transport) try { await this.transport.close() } catch {}
```

But both `callTool()` error handler and the `onclose` handler set `this.transport = null` **before** triggering a reconnect via `ensureConnected()` → `connectInternal()`. This means `connectInternal` sees `this.transport === null`, skips the cleanup, and spawns a new `uvx chroma-mcp` subprocess — orphaning the old one.

### Two leak paths:

**1. `callTool()` retry path (line ~callTool method):**
```js
} catch(o) {
    this.connected = false;
    this.client = null;
    this.transport = null;  // ← old subprocess reference lost
    // ...
    await this.ensureConnected();  // → connectInternal() can't close old subprocess
```

**2. `onclose` handler (set in connectInternal):**
```js
this.transport.onclose = () => {
    this.connected = false;
    this.client = null;
    this.transport = null;  // ← old subprocess reference lost
    this.lastConnectionFailureTimestamp = Date.now();
}
// Next ensureConnected() call → connectInternal() can't close old subprocess
```

## Reproduction

```bash
# Use Claude Code SDK to make multiple queries (e.g. via claude-relay or subagents)
# Watch chroma-mcp processes multiply:
watch -n1 'ps aux | grep chroma-mcp | grep -v grep | wc -l'
```

On a system with moderate SDK usage, we observed 31 orphaned processes within ~1 hour, consuming 2.5GB+ RAM. With heavy SDK usage (claude-relay), 16 processes spawned in under 3 minutes.

## Suggested Fix

Save the transport reference before nulling, and close it in `connectInternal`:

```js
// In callTool error handler:
const staleTransport = this.transport;
this.transport = null;
this.client = null;
this.connected = false;
try { await staleTransport?.close() } catch {}
```

Or simpler: don't null `this.transport` in the error handlers — let `connectInternal()` handle cleanup since it already has the logic for it.

## Environment

- Linux 6.8.0-106-generic, 5.8GB RAM
- claude-mem v10.5.5
- Claude Code 2.1.76

## Related Issues

- #1361 — chroma-mcp leak after session exit (same symptom, different trigger: exit vs reconnect)
- #1352 / PR #1354 — CLAUDECODE env var leak causing extra spawns (separate issue, compounds this one)
- #1185 — chroma-mcp CPU/memory leak in v10.3.1 (closed, was a different root cause)

manuelfedele:
## Summary

Fixes #1369

`callTool()` sets `this.transport = null` without calling `transport.close()` first. When `connectInternal()` runs next, it sees `transport === null`, skips cleanup, and spawns a new `chroma-mcp` subprocess while the old one is still alive. Over time this leaks dozens of orphaned `uvx`/`chroma-mcp` processes, each loading a 79MB ONNX model, exhausting RAM and causing MCP search timeouts.

## Root Cause

Two code paths in `callTool()` drop the transport reference without closing it:

1. **Initial catch block** (transport error) - nulls `this.transport`, then calls `ensureConnected()` which spawns a new subprocess
2. **Retry catch block** - only sets `this.connected = false` without cleaning up transport/client at all

Meanwhile, `connectInternal()` tries to close the old transport:
```js
if (this.transport) try { await this.transport.close() } catch {}
```
But since `callTool()` already nulled it, this guard is always false, and the old subprocess lives on.

## Fix

Save a reference to the stale transport before nulling the instance field, then close the saved reference:

```js
const staleTransport = this.transport;
this.client = null;
this.transport = null;
if (staleTransport) {
  try { await staleTransport.close(); } catch { /* subprocess may already be dead */ }
}
```

Applied to both the initial catch and retry catch blocks.

## Impact

Without fix: each transport error accumulates an orphaned `chroma-mcp` process. In production we observed 25+ leaked processes consuming several GB of RAM.

With fix: `transport.close()` sends SIGTERM to the subprocess before a replacement is spawned.

## Test plan

- [x] Verified fix on local machine: transport errors no longer accumulate orphaned processes
- [x] Existing SSL test suite unaffected (tests mock transport)

### DIFF
--- a/src/services/sync/ChromaMcpManager.ts
+++ b/src/services/sync/ChromaMcpManager.ts
@@ -252,9 +252,17 @@ export class ChromaMcpManager {
       // Transport error: chroma-mcp subprocess likely died (e.g., killed by orphan reaper,
       // HNSW index corruption). Mark connection dead and retry once after reconnect (#1131).
       // Without this retry, callers see a one-shot error even though reconnect would succeed.
+      //
+      // CRITICAL: Close the transport before nulling to kill the subprocess (#1369).
+      // If we null this.transport first, connectInternal() sees transport === null,
+      // skips cleanup, and spawns a new subprocess, orphaning the old one.
       this.connected = false;
+      const staleTransport = this.transport;
       this.client = null;
       this.transport = null;
+      if (staleTransport) {
+        try { await staleTransport.close(); } catch { /* subprocess may already be dead */ }
+      }
 
       logger.warn('CHROMA_MCP', `Transport error during "${toolName}", reconnecting and retrying once`, {
         error: transportError instanceof Error ? transportError.message : String(transportError)
@@ -267,7 +275,14 @@ export class ChromaMcpManager {
           arguments: toolArguments
         });
       } catch (retryError) {
+        // Retry also failed. Close transport to prevent another orphan.
         this.connected = false;
+        const retryTransport = this.transport;
+        this.client = null;
+        this.transport = null;
+        if (retryTransport) {
+          try { await retryTransport.close(); } catch { /* subprocess may already be dead */ }
+        }
         throw new Error(`chroma-mcp transport error during "${toolName}" (retry failed): ${retryError instanceof Error ? retryError.message : String(retryError)}`);
       }
     }


thedotmack:
Note: The Process Supervisor (PR #1370, v10.5.6) adds a process registry that tracks and reaps subprocess PIDs on session end. However, this PR's fix for the transport null-before-close race in `ChromaMcpManager` is still needed — the supervisor catches orphans after the fact, but this PR prevents the leak at the source. Should still land.

alessandropcostabr:
Confirmed this fix resolves the issue. Applied it locally to our v12.1.0 bundle and the subprocess leak stopped immediately.

See our detailed reproduction report on #1369 — 10 orphaned processes (830MB) accumulated in 30 minutes from concurrent batch sync operations triggering the `callTool` retry path.

The fixed supervisor ID (`"chroma-mcp"`) is a secondary issue worth noting — each spawn overwrites the previous registry entry, so the supervisor can never reap old PIDs. But closing the transport before nulling (as this PR does) is the right primary fix.

thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1925, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1379, opened 2026-03-17
fix(installer): handle plugins.allow chicken-and-egg problem

### LINKED ISSUE #1371: #Bug: Installer fails in a loop when plugins.allow is validated before plugin is installed
#Bug: Installer fails in a loop when plugins.allow is validated before plugin is installed
Environment:

OpenClaw 2026.3.8
Self-hosted Docker container (Hostinger VPS)
Linux (Ubuntu)

What happened:
The installer clones the repo and tries to run openclaw plugins install, but OpenClaw validates openclaw.json before completing the install. This creates an unresolvable loop:

If claude-mem is NOT in plugins.allow → install fails with plugin not found: claude-mem
If claude-mem IS in plugins.allow → config validation fails with plugin not found: claude-mem (because it's not installed yet)

Additional issue:
The plugin/package.json generated by npm run build does not include the openclaw.extensions field, causing openclaw plugins install to fail with package.json missing openclaw.extensions.
Steps to reproduce:

Run curl -fsSL https://install.cmem.ai/openclaw.sh | bash inside OpenClaw Docker container
First attempt fails at plugin install step
Manually add claude-mem to plugins.allow in openclaw.json
Re-run installer — fails again because plugin doesn't exist yet when config is validated

Expected behavior:
Installer should handle the chicken-and-egg problem between plugins.allow and the actual plugin installation.

JasonOA888:
Fixes #1371

**Problem:**
When `claude-mem` is in `plugins.allow` but the plugin isn't installed yet, OpenClaw's config validator blocks ALL CLI commands including `plugins install`. This creates an unresolvable loop:

- If NOT in allowlist → `plugins install` fails
- If IN allowlist → config validation fails (plugin not found)

**Solution:**
- Before running `plugins install`, temporarily remove `claude-mem` from `plugins.allow`
- This unblocks the CLI so `plugins install` can run
- After successful install, restore `claude-mem` to the allowlist

**Changes:**
- Modified `openclaw/install.sh` to detect and remove `claude-mem` from `plugins.allow`
- Added logic to restore it after plugin is installed and enabled
- Uses a stdout marker to signal when allowlist was modified (avoids subshell variable issues)

**Testing:**
1. Add `claude-mem` to `plugins.allow` in `~/.openclaw/openclaw.json`
2. Run the installer
3. Plugin installs successfully
4. `claude-mem` is restored to allowlist after install

### DIFF
--- a/openclaw/install.sh
+++ b/openclaw/install.sh
@@ -779,15 +779,24 @@ install_plugin() {
   # OpenClaw's config validator blocks ALL CLI commands (including plugins install).
   # We temporarily remove the entry and save the config so `plugins install` can run,
   # then `plugins install` + `plugins enable` will re-create it properly.
+  #
+  # Also handle plugins.allow: if claude-mem is in the allowlist but not installed,
+  # config validation fails. We temporarily remove it and restore after install.
   local oc_config="${HOME}/.openclaw/openclaw.json"
   local saved_plugin_config=""
+  local saved_from_allowlist=false
   if [[ -f "$oc_config" ]]; then
-    saved_plugin_config=$(INSTALLER_CONFIG_FILE="$oc_config" node -e "
+    # Capture both stdout (config JSON) and stderr (allowlist signal)
+    local node_output
+    node_output=$(INSTALLER_CONFIG_FILE="$oc_config" node -e "
       const fs = require('fs');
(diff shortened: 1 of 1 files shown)


thedotmack:

PR #1407, opened 2026-03-19
fix: remove Setup hook referencing deleted setup.sh

### LINKED ISSUE #1396: [10.6.0] Setup hook references setup.sh that doesn't exist in package
Same issue as #1340 (filed against v10.5.5), still present in v10.6.0.

The `hooks.json` in 10.6.0 adds a new `Setup` event that runs `$_R/scripts/setup.sh`, but the file is not included in the plugin package. Every new session shows:

```
SessionStart:startup hook error (exit code 127)
```

### Scripts directory listing (v10.6.0)

`setup.sh` is still not present. All other referenced scripts exist.

### Workaround

Create a placeholder `scripts/setup.sh`:

```bash
#!/bin/bash
echo '{"continue":true,"suppressOutput":true}'
exit 0
```

This has been broken since v10.5.5 (#1340) and the file was not added in 10.6.0 either.

— cee

tonymfer:
## Summary
- Remove the `Setup` hook block from `plugin/hooks/hooks.json` that references the nonexistent `scripts/setup.sh`

## Why
`setup.sh` was intentionally deleted in v10.5.0 ("Removed legacy setup.sh script"). When v10.5.1 restored hooks.json to a pre-smart-explore configuration, the Setup hook definition came back but the script it points to was never re-created. This causes exit code 127 on every session start.

The `SessionStart` hook already runs `smart-install.js` which handles installation and setup, making the `Setup` hook redundant.

## How
Removed the `Setup` event handler block entirely from hooks.json.

Closes #1396

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -1,18 +1,6 @@
 {
   "description": "Claude-mem memory system hooks",
   "hooks": {
-    "Setup": [
-      {
-        "matcher": "*",
-        "hooks": [
-          {
-            "type": "command",
-            "command": "_R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; \"$_R/scripts/setup.sh\"",
-            "timeout": 300
-          }
-        ]
-      }
-    ],
     "SessionStart": [
       {
         "matcher": "startup|clear|compact",


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1900, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1408, opened 2026-03-19
Use CLAUDE_MEM_WORKER_HOST for health checks, skip local spawn for remote workers

dworznik:
## Summary

`CLAUDE_MEM_WORKER_HOST` is configurable via `settings.json`, and `getWorkerHost()` + `buildWorkerUrl()` in `worker-utils.ts` already respect it. However, `HealthMonitor.ts` hardcodes `127.0.0.1` for all HTTP requests (health checks, port-in-use detection, shutdown, version checks). This causes `ensureWorkerStarted()` to always find nothing on localhost and spawn a local daemon — even when the configured host points to a remote worker that is healthy and reachable.

**Result:** `CLAUDE_MEM_WORKER_HOST` is effectively a no-op for the worker lifecycle. The local daemon shadows the remote worker on the same port.

## Changes

- **`HealthMonitor.ts`**: `httpRequestToWorker()` and `isPortInUse()` now call `getWorkerHost()` instead of hardcoding `127.0.0.1`
- **`worker-utils.ts`**: Added `isWorkerLocal()` helper — returns `false` for non-local hosts (`127.0.0.1`, `localhost`, `0.0.0.0`, `::1`)
- **`worker-service.ts`**: `ensureWorkerStarted()` checks `isWorkerLocal()` first. For remote hosts, it only does a health check and skips PID file management and daemon spawn entirely
- **`context.ts`**: Status line now shows the configured host instead of hardcoded `localhost`

## Use case

Running claude-mem as a standalone Docker container on a remote VPS, accessed over a VPN. The local machine acts as a thin client — hooks send data to the remote worker via `CLAUDE_MEM_WORKER_HOST`, no local daemon needed.

## Test plan

- [ ] Verify default behavior unchanged: `CLAUDE_MEM_WORKER_HOST=127.0.0.1` still spawns local daemon as before
- [ ] Set `CLAUDE_MEM_WORKER_HOST` to a remote address with a healthy worker → no local daemon spawned, hooks communicate with remote
- [ ] Set `CLAUDE_MEM_WORKER_HOST` to a remote address with no worker → graceful failure, no local spawn attempted
- [ ] Status line shows correct host in viewer URL

### DIFF
--- a/src/cli/handlers/context.ts
+++ b/src/cli/handlers/context.ts
@@ -6,7 +6,7 @@
  */
 
 import type { EventHandler, NormalizedHookInput, HookResult } from '../types.js';
-import { ensureWorkerRunning, getWorkerPort, workerHttpRequest } from '../../shared/worker-utils.js';
+import { ensureWorkerRunning, getWorkerPort, getWorkerHost, workerHttpRequest } from '../../shared/worker-utils.js';
 import { getProjectContext } from '../../utils/project-name.js';
 import { HOOK_EXIT_CODES } from '../../shared/hook-constants.js';
 import { logger } from '../../utils/logger.js';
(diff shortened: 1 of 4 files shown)


thedotmack:

PR #1409, opened 2026-03-19
fix: Windows daemon fails to start — isMainModule detection broken

ghota-tech-solutions:
## Summary

Two bugs prevent the worker daemon from functioning on Windows:

### 1. `isMainModule` detection broken (daemon exits immediately)
- `import.meta.url` produces `file:///C:/...` (three slashes) while `` `file://${process.argv[1]}` `` produces `file://C:/...` (two slashes) — never matches
- `endsWith('worker-service')` doesn't match the bundled `.cjs` extension
- Result: daemon loads module and exits with code 0 without ever calling `main()`

### 2. Bundled `__dirname` resolves to build machine path
- esbuild hardcodes `__dirname` to the developer's local path (e.g. `/Users/alexnewman/.../src/shared`)
- `getPackageRoot()` returns an invalid path, breaking:
  - ModeManager: `"Critical: code.json mode file missing"`
  - ViewerRoutes: `"Viewer UI not found at any expected location"`
  - Any code depending on `getPackageRoot()`

## Fix

**`worker-service.ts`**: Add `file:///` (three slashes) and `.cjs` extension checks to `isMainModule`

**`paths.ts`**: Verify `__dirname` exists on disk before trusting it, fall back to `import.meta.url`

**`cleanup-duplicates.ts` / `import-xml-observations.ts`**: Same `file:///` fix

## Test plan

- [x] Patched bundled `worker-service.cjs` on Windows 11 + Bun 1.3.11
- [x] Daemon starts and stays alive (port 37777 LISTENING)
- [x] `/api/health` → `{"status":"ok", "initialized": true}`
- [x] `/api/readiness` → `{"status":"ready"}`
- [x] Viewer UI (`/`) → HTTP 200
- [x] Mode `code.json` loaded successfully
- [x] Chroma sync completes
- [ ] Verify no regression on macOS/Linux (added checks are no-ops on Unix)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/bin/cleanup-duplicates.ts
+++ b/src/bin/cleanup-duplicates.ts
@@ -92,7 +92,8 @@ function main() {
   console.log('='.repeat(60));
 }
 
-// Run if executed directly (file:/// with three slashes on Windows)
-if (import.meta.url === `file://${process.argv[1]}` || import.meta.url === `file:///${process.argv[1]}`) {
+// Run if executed directly (pathToFileURL handles Windows drive letters, slashes, encoding)
+import { pathToFileURL } from 'url';
+if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) {
   main();
 }
--- a/src/bin/import-xml-observations.ts
+++ b/src/bin/import-xml-observations.ts
@@ -383,7 +383,8 @@ function main() {
   console.log('='.repeat(60));
 }
 
-// Run if executed directly (file:/// with three slashes on Windows)
-if (import.meta.url === `file://${process.argv[1]}` || import.meta.url === `file:///${process.argv[1]}`) {
+// Run if executed directly (pathToFileURL handles Windows drive letters, slashes, encoding)
+import { pathToFileURL } from 'url';
+if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) {
   main();
 }
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -1238,14 +1238,14 @@ async function main() {
   }
 }
 
-// Check if running as main module in both ESM and CommonJS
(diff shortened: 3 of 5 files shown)


xkonjin (commented):
## Code Review — Windows Daemon isMainModule Detection Fix

Targeted fix for a real platform bug. On Windows, `import.meta.url` uses `file:///C:/...` (three slashes) while `process.argv[1]` gives `C:/...`, so the equality check always fails and the daemon never starts.

### Changes

1. **worker-service.ts** — Adds `file:///${process.argv[1]}` as alternative check, plus `.cjs` extension matching for bundled builds. Both correct.
2. **cleanup-duplicates.ts / import-xml-observations.ts** — Same three-slash fix for CLI entry points.
3. **paths.ts** — Smart fix: checks if `__dirname` actually exists before using it. When esbuild bundles, it bakes in the build machine path which will not exist on the user machine. Falling back to `import.meta.url` is correct.

### Concerns

1. **The file:/// check is fragile** — `process.argv[1]` on Windows can use forward slashes or backslashes depending on invocation. Consider using Node `pathToFileURL` for robust comparison:
```js
import { pathToFileURL } from "url";
const isMain = import.meta.url === pathToFileURL(process.argv[1]).href;
```
This handles drive letter casing, slash direction, and URL encoding in one shot.

2. **`existsSync(__dirname)` in getDirname()** — If called per-request rather than once at module load, the synchronous I/O could add up. Consider caching.

3. **No tests** — The `getDirname` fallback behavior could be tested by temporarily pointing `__dirname` to a non-existent path.

### Verdict

Fix is correct for the immediate problem. The `pathToFileURL` approach would be more robust long-term, but this unblocks Windows users now. No security concerns.

villers:
Up

PR #1421, opened 2026-03-20
feat: add Flashrank reranker microservice for improved search relevance

greghughespdx:
## What this does

Adds an optional cross-encoder reranking step to improve the ordering of ChromaDB search results.

ChromaDB uses sentence embeddings for initial retrieval, which gives good recall but can return results in suboptimal order when multiple candidates are close in embedding space. A cross-encoder re-scores each candidate using full query-document attention, typically improving the precision of the top results returned to context.

## How it works

**Architecture:** Two-phase retrieval

1. ChromaDB performs semantic search (unchanged) -- high recall, fast
2. Flashrank cross-encoder re-scores the top candidates (optional) -- better precision

The reranker runs as a separate HTTP microservice (`flashrank-service.py`), keeping the worker process free of Python/ML dependencies. The worker calls it via `fetch()` with a 5-second timeout and falls back silently to the original Chroma ordering on any error.

**Model:** `ms-marco-MiniLM-L-12-v2`
- ~22MB quantized ONNX
- CPU-only, no GPU required
- ~26ms latency for typical candidate sets (20-50 passages)

## Changes

**`plugin/scripts/flashrank-service.py`**

Standalone HTTP microservice. Endpoint: `POST /rerank`

Request:
```json
{
  "query": "search query text",
  "passages": [{"id": "42", "text": "passage text"}],
  "top_k": 20
}
```

Response:
```json
{
  "results": [{"id": "42", "score": 0.95}],
  "latency_ms": 26.3
}
```

Also exposes `GET /health` for liveness checks.

**`src/shared/SettingsDefaultsManager.ts`**

Two new settings:

| Setting | Default | Description |
|---------|---------|-------------|
| `CLAUDE_MEM_RERANK_ENABLED` | `"false"` | Enable reranking |
| `CLAUDE_MEM_RERANK_URL` | `"http://localhost:37778"` | Reranker service URL |

**`src/services/worker/search/strategies/ChromaSearchStrategy.ts`**

After the recency filter step, optionally calls the reranker service. Results are re-ordered by cross-encoder score before hydration from SQLite. Falls back gracefully to Chroma ordering on timeout or service unavailability.

## Installation

```bash
pip install flashrank
python plugin/scripts/flashrank-service.py
```

Or as a persistent service (example launchd plist):

```xml
<key>ProgramArguments</key>
<array>
  <string>/usr/bin/python3</string>
  <string>/path/to/flashrank-service.py</string>
  <string>--port</string>
  <string>37778</string>
</array>
```

Then enable in `~/.claude-mem/settings.json`:

```json
{
  "CLAUDE_MEM_RERANK_ENABLED": "true"
}
```

## Notes

- Reranking is disabled by default; existing behavior is unchanged unless opted in
- The service is entirely optional; the worker degrades gracefully if it is unreachable
- Passage text sent to the reranker is assembled from Chroma metadata fields (`title`, `subtitle`, `concepts`, `type`) since full document text lives in SQLite and is only retrieved during hydration

### DIFF
--- a/src/services/worker/search/strategies/ChromaSearchStrategy.ts
+++ b/src/services/worker/search/strategies/ChromaSearchStrategy.ts
@@ -98,8 +98,11 @@ export class ChromaSearchStrategy extends BaseSearchStrategy implements SearchSt
       });
 
       // Step 2b: Optionally rerank with Flashrank cross-encoder
+      let wasReranked = false;
       if (recentItems.length > 1 && query) {
-        recentItems = await this.rerank(query, recentItems);
+        const rerankResult = await this.rerank(query, recentItems);
+        recentItems = rerankResult.items;
+        wasReranked = rerankResult.reranked;
       }
 
       // Step 3: Categorize by document type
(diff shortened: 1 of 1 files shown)


thedotmack:

PR #1432, opened 2026-03-21
feat: persist ANTHROPIC_BASE_URL in ~/.claude-mem/.env for custom gateway support

manuelfedele:
## Problem

Corporate environments commonly route Anthropic API calls through an internal LLM gateway or proxy (e.g., AWS API Gateway, custom Bedrock proxies, on-prem endpoints). This requires setting `ANTHROPIC_BASE_URL` alongside `ANTHROPIC_API_KEY`.

Currently, `ANTHROPIC_BASE_URL` passes through from `process.env` (it's not in `BLOCKED_ENV_VARS`), but the worker daemon may have been started in a shell context that lacked this variable, or may outlive the shell session that originally set it. When the base URL is lost, the SDK subprocess silently fails with "Not logged in" because it falls back to Anthropic's default endpoint where the gateway API key is not recognized.

This is a real-world issue: corporate users configure both `ANTHROPIC_API_KEY` and `ANTHROPIC_BASE_URL` in their shell profile, but only `ANTHROPIC_API_KEY` gets persisted (via `~/.claude-mem/.env`), while `ANTHROPIC_BASE_URL` is lost on daemon restart.

## Solution

Extend `~/.claude-mem/.env` to also manage `ANTHROPIC_BASE_URL`, following the same pattern used for API keys:

- Add `ANTHROPIC_BASE_URL` to `ClaudeMemEnv` interface
- Read/write it in `loadClaudeMemEnv()` / `saveClaudeMemEnv()`
- Re-inject it in `buildIsolatedEnv()` so subprocess always gets the configured endpoint
- Document custom gateway configuration in `docs/configuration.mdx`

## Changes

- `src/shared/EnvManager.ts`: 4 surgical additions following existing credential patterns
- `docs/public/configuration.mdx`: new "Custom Gateway / Proxy Configuration" section

## Testing

Verified manually in a corporate environment using an Anthropic-compatible LLM gateway. Before this change, the worker would fail with "Not logged in" after daemon restart. After adding `ANTHROPIC_BASE_URL` to `~/.claude-mem/.env`, the worker correctly routes requests through the gateway across restarts.

### DIFF
--- a/docs/public/configuration.mdx
+++ b/docs/public/configuration.mdx
@@ -52,6 +52,23 @@ See [OpenRouter Provider](usage/openrouter-provider) for detailed configuration,
 | `CLAUDE_MEM_PYTHON_VERSION`   | `3.13`                          | Python version for chroma-mcp         |
 | `CLAUDE_CODE_PATH`            | _(auto-detect)_                 | Path to Claude Code CLI (for Windows) |
 
+## Custom Gateway / Proxy Configuration
+
+Corporate environments commonly route Anthropic API calls through an internal LLM gateway or proxy. Claude-mem supports this via `ANTHROPIC_BASE_URL` in `~/.claude-mem/.env`.
+
+Edit `~/.claude-mem/.env`:
+
+```bash
+ANTHROPIC_API_KEY=your-gateway-api-key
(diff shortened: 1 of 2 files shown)


guilhermehtk:
+1

thedotmack:

Issue #1437, opened 2026-03-21
v10.6.1: Worker daemon fails to start — esbuild inlines __dirname with developer's local path


asturwebs:
## Bug Report

**Version:** 10.6.1 (regression from 10.6.0)
**Platform:** Linux (WSL2 Ubuntu 22.04), Bun 1.3.6, Node v24.13.0
**Severity:** Worker daemon never initializes — DB stays in "initializing" state indefinitely

## Description

The worker daemon fails to start on v10.6.1 due to `__dirname` being hardcoded to the developer's local path in the compiled bundle `worker-service.cjs`.

### Root Cause

In `scripts/worker-service.cjs`, line 7472, the bundled output contains:

```javascript
var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/shared", _dirname, DATA_DIR, ...
```

This is a build-time esbuild artifact — `__dirname` was inlined as a literal string from the developer's machine.

### Impact Chain

1. `getDirname()` checks `typeof __dirname !== "undefined"` → returns the hardcoded path
2. `getPackageRoot()` returns `join(_dirname, "..")` → `/Users/alexnewman/.../src`
3. `ModeManager` looks for `modes/code.json` at `<packageRoot>/modes` → **doesn't exist**
4. Background initialization throws: `"Critical: code.json mode file missing"`
5. Database never initializes → readiness check times out → hooks report errors

### Evidence: Regression in v10.6.1

Checked all cached versions:
- **v10.5.5**: No `var __dirname = "..."` hardcoding ✅
- **v10.6.0**: No `var __dirname = "..."` hardcoding ✅
- **v10.6.1**: Three hardcoded `__dirname` assignments (lines 7472, 43681, 68282) ❌

The esbuild configuration changed between 10.6.0 and 10.6.1 to inline `__dirname` as a literal.

### Why Some Users Don't Notice

The plugin degrades gracefully — MCP search/timeline tools work without the worker. The worker failure only affects observation processing, session summaries, and the Web UI (port 37777). Many users may not realize the worker is broken.

### Workaround

Patch line 7472 in `worker-service.cjs` to remove the hardcoded value:

```diff
- var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/shared", _dirname, ...
+ var __dirname, _dirname, ...
```

This allows `getDirname()` to fall through to the `import.meta.url` fallback, which resolves correctly under Bun.

### Fix Suggestion

[...]

Issue #1433, opened 2026-03-21
Hardcoded __dirname in worker-service.cjs prevents DB initialization


mleiter:
## Bug

`worker-service.cjs` contains three hardcoded `__dirname` paths from the author's build system. On any other machine, the `ModeManager` cannot find `code.json`, causing `"Critical: code.json mode file missing"` and permanent `"Database initialization timeout"`.

## Environment

- claude-mem: 10.6.1
- Bun: 1.3.11
- OS: macOS 15.4 (Darwin 25.2.0)
- Claude Code: latest

## Symptoms

- Worker starts and listens on port 37777
- `/health` returns `{"status":"ok"}`
- Every other endpoint returns `{"error":"Service initializing","message":"Database is still initializing, please retry"}` — forever
- Log shows: `Background initialization failed Critical: code.json mode file missing`

## Root Cause

Three hardcoded `__dirname` assignments in the bundled `worker-service.cjs`:

```js
// Line 7472 — used by getPackageRoot() → ModeManager → code.json lookup
var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/shared"

// Line 43681 — used by Server.ts for UI paths
var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/services/server"

// Line 68282 — used to spawn mcp-server.cjs
var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/services"
```

`getDirname()` (line 7396) has a correct fallback via `import.meta.url`, but never reaches it because `typeof __dirname !== "undefined"` is always true due to the hardcoded string.

## Workaround

Replace the three hardcoded paths:

```js
// Line 7472: set to undefined so getDirname() fallback kicks in
var __dirname = undefined, _dirname, DATA_DIR, ...

// Line 43681: resolve dynamically
var __dirname = require("path").dirname(require("url").fileURLToPath(import.meta.url));

// Line 68282: resolve dynamically
var __dirname = require("path").dirname(require("url").fileURLToPath(import.meta.url)), __filename = ...;
```

## Suggested Fix

The Bun bundler should not embed the build machine's `__dirname` as a string literal. Options:
1. Use `import.meta.dir` (Bun-native) instead of `__dirname` in source
2. Configure the bundler to not replace `__dirname` with a static string
3. Remove the hardcoded `__dirname` declarations and rely on the existing `getDirname()` fallback

[...]

Issue #1419, opened 2026-03-20
Worker service fails on non-developer machines due to hardcoded __dirname in CJS bundle


maxlxxiii:
**Environment:**
- claude-mem v10.6.1
- Ubuntu Linux (also affects any machine that isn't the build machine)
- Bun 1.x

**Description:**

`scripts/worker-service.cjs` contains hardcoded `__dirname` values baked in at bundle time:

```js
var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/shared"
var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/services/server"
var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/services"
```

These shadow the real CJS `__dirname` parameter. `getDirname()` trusts these values unconditionally, causing `getPackageRoot()` to resolve to a non-existent path. This means:

1. `modes/code.json` is never found → "Critical: code.json mode file missing"
2. DB schema never initializes → `claude-mem.db` stays at 0 bytes
3. All hooks fail with "DB not initialized" → no observations recorded
4. Viewer UI returns "Viewer UI not found at any expected location"

The plugin appears to install and start successfully (HTTP server binds to port 37777, MCP server runs), but silently does nothing. No observations are ever captured.

**Steps to reproduce:**

1. Install claude-mem on any machine via `/plugin marketplace add thedotmack/claude-mem && /plugin install claude-mem`
2. Restart Claude Code
3. Check `~/.claude-mem/logs/` for "code.json mode file missing" errors
4. Confirm `~/.claude-mem/claude-mem.db` is 0 bytes

**Root cause:**

The bundler (likely esbuild) is inlining `__dirname` at compile time from the developer's local Mac filesystem. The resulting CJS file only works on the exact machine it was built on.

**Workaround:**

Patch `getDirname()` in `worker-service.cjs` to validate that the hardcoded `__dirname` exists on disk before using it, falling back to `module.filename`:

```js
function getDirname() {
  if (typeof __dirname !== "undefined") {
    try {
      if (existsSync3(__dirname)) {
        return __dirname;
      }
    } catch (e) {}
  }
  if (typeof module !== "undefined" && module.filename) {
    return dirname2(module.filename);
  }
  return dirname2(fileURLToPath(import.meta.url));
}
```

**Recommended fix:**

[...]

PR #1442, opened 2026-03-21
fix: prioritize import.meta.url over __dirname in getDirname()

### LINKED ISSUE #1410: bug: hardcoded __dirname in worker-service.cjs prevents initialization on end-user machines
## Bug Description

The bundled `worker-service.cjs` contains a hardcoded `__dirname` variable (line 7472) set to the **build machine's path**:

```js
var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/shared", _dirname, DATA_DIR, ...
```

This `var` declaration shadows Bun's native `__dirname`, causing `getDirname()` → `getPackageRoot()` to resolve to a non-existent directory on end-user machines. This breaks:

1. **Mode loading** — `code.json` can't be found, throws `"Critical: code.json mode file missing"`
2. **Database initialization** — `initializeBackground()` crashes before creating/opening the SQLite database
3. **Viewer UI** — static file serving fails (`"Viewer UI not found at any expected location"`)

## Symptoms

- Worker HTTP server starts and passes health checks (`/api/health` returns 200)
- But `initialized: false` and `mcpReady: false` **permanently**
- Database file is either missing or 0 bytes
- Web UI at `localhost:37777` returns 500
- All hooks fail with "Worker failed to start (health check timeout)" because `waitForReadiness` never succeeds
- SessionEnd hook fails with "Hook cancelled"
- Logs show: `Background initialization failed Critical: code.json mode file missing`

## Root Cause

In `worker-service.cjs` line 7472, the bundler (likely Bun's bundler or esbuild) bakes in the build-time `__dirname`:

```js
// Line 7396-7401
function getDirname() {
  if (typeof __dirname !== "undefined") {
    return __dirname;  // ← always takes this branch due to var hoisting
  }
  return dirname2(fileURLToPath(import.meta.url));  // ← never reached
}
```

Since `var __dirname` is hoisted, `typeof __dirname !== "undefined"` is always true, so the `import.meta.url` fallback (which would resolve correctly) is never used.

## Workaround

Remove the hardcoded `__dirname` declaration from line 7472 of `plugin/scripts/worker-service.cjs`:

```diff
-var __dirname = "/Users/alexnewman/conductor/workspaces/claude-mem/banjul/src/shared", _dirname, DATA_DIR, ...
+var _dirname, DATA_DIR, ...
```

This lets Bun's native `__dirname` (which points to the actual script directory) work correctly.

## Suggested Fix

In the build configuration, prevent the bundler from inlining `__dirname` as a static string. Options:

1. **Configure the bundler** to treat `__dirname` as an external/runtime variable
2. **Use `import.meta.url`** directly instead of `__dirname` in `paths.ts`
3. **Add a runtime guard** in the bundled output: `var __dirname = typeof __dirname !== "undefined" ? __dirname : ...` (though this has hoisting issues with `var`)

## Environment

- **Plugin version**: v10.6.1
- **Platform**: macOS 14.x, Apple Silicon
- **Bun**: 1.3.10
- **Node**: v22.x

## Reproduction

1. Fresh install claude-mem plugin v10.6.1
3. Worker spawns but never initializes
4. Check logs: `~/.claude-mem/logs/claude-mem-*.log`
5. Observe: `Background initialization failed Critical: code.json mode file missing`

NewestUser:
## Summary

- esbuild inlines `__dirname` as a static string from the build machine (`/Users/alexnewman/...`), which shadows Bun's native `__dirname` in the bundled `worker-service.cjs`
- `getDirname()` always takes the `typeof __dirname !== 'undefined'` branch (due to `var` hoisting of the hardcoded value), so the `import.meta.url` fallback is never reached
- This causes `code.json` mode file loading, database initialization, and viewer UI serving to fail on all end-user machines
- Fix: check `import.meta.url` first (resolves correctly at runtime), fall back to `__dirname`, then `process.cwd()`

Fixes #1410, #1419, #1433, #1437

## Test plan

- [ ] Fresh install on a non-dev machine — worker should reach `initialized: true` and `mcpReady: true`
- [ ] `curl http://localhost:37777/api/health` returns `initialized: true`
- [ ] Mode file (`code.json`) loads without error
- [ ] Viewer UI at `http://localhost:37777` serves correctly
- [ ] Verify hooks still work in ESM context (session start/end)

### DIFF
--- a/src/shared/paths.ts
+++ b/src/shared/paths.ts
@@ -6,14 +6,19 @@ import { fileURLToPath } from 'url';
 import { SettingsDefaultsManager } from './SettingsDefaultsManager.js';
 import { logger } from '../utils/logger.js';
 
-// Get __dirname that works in both ESM (hooks) and CJS (worker) contexts
+// Get __dirname that works in both ESM (hooks) and CJS (worker) contexts.
+// import.meta.url is preferred because esbuild inlines __dirname as a static
+// string from the build machine, which breaks path resolution on end-user
+// machines.  The CJS __dirname fallback is kept for environments where
+// import.meta.url is unavailable.
 function getDirname(): string {
-  // CJS context - __dirname exists
+  // Prefer import.meta.url — resolves correctly at runtime on any machine
+  try { return dirname(fileURLToPath(import.meta.url)); } catch {}
+  // CJS context fallback
   if (typeof __dirname !== 'undefined') {
     return __dirname;
   }
-  // ESM context - use import.meta.url
-  return dirname(fileURLToPath(import.meta.url));
+  return process.cwd();
 }
 
 const _dirname = getDirname();


NewestUser:
Closing in favour of https://github.com/thedotmack/claude-mem/pull/1411

PR #1485, opened 2026-03-25
fix(bun-runner): use full path from where/which output on Windows

vcliment89:
## Summary
- Fixed `bun-runner.js` to return the full Bun executable path from `where`/`which` output instead of just the literal string `'bun'`
- This fixes hook failures on Windows when Bun is installed via npm (not the official installer) and `spawn()` without `shell: true` can't find `bun` in PATH

## Problem
On Windows, when Bun is installed via npm (`npm install -g bun`), it's placed in a non-standard location like `C:\Users\...\AppData\Roaming\npm\bun`. The `findBun()` function correctly found this path using `where bun`, but then returned just `'bun'` instead of the actual path. When `spawn()` runs without `shell: true`, it can't resolve `'bun'` to the full path.

This caused `PostToolUse` hooks to fail with errors like:
```
Failed to start Bun: spawn bun ENOENT
```

## Solution
Changed line 58 from:
```javascript
return 'bun'; // Found in PATH
```
to:
```javascript
return pathCheck.stdout.trim().split('\n')[0]; // Return full path from where/which
```

This returns the first path from `where`/`which` output, which is the correct full path to the bun executable.

## Test plan
- [x] Verified on Windows with Bun installed via npm: `where bun` returns `C:\Users\vicen\AppData\Roaming\npm\bun`
- [x] Confirmed the fix resolves the spawn ENOENT error

### DIFF
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -47,22 +47,16 @@ function fixBrokenScriptPath(argPath) {
  * Find Bun executable - checks PATH first, then common install locations
  */
 function findBun() {
-  // Try PATH first
-  const pathCheck = spawnSync(IS_WINDOWS ? 'where' : 'which', ['bun'], {
-    encoding: 'utf-8',
-    stdio: ['pipe', 'pipe', 'pipe'],
-    shell: IS_WINDOWS
-  });
-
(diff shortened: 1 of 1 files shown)


thedotmack:

PR #1487, opened 2026-03-25
feat: add CLAUDE_MEM_DISABLED env var for headless sessions

### LINKED ISSUE #1484: Feature request: CLAUDE_MEM_DISABLED env var for headless/agent sessions
## Problem

When claude-mem is installed globally, its hooks (`PostToolUse *`, `UserPromptSubmit`, `Stop`, `SessionStart`) fire in **all** Claude Code processes — including headless agent sessions spawned by orchestration platforms like [Paperclip](https://github.com/paperclipai/paperclip).

This causes two problems:

1. **Token waste**: The `SessionStart` context injection loads 50 observations (~25,000 tokens) into every agent session. For a fleet of 11 agents running on heartbeat schedules (some every 4 hours), this adds up to significant unnecessary token consumption.

2. **Cross-contamination**: Agent sessions record observations from their specialized domains (portfolio management, risk monitoring, data engineering) into the shared claude-mem database. These observations then get injected into unrelated sessions — both other agents and interactive user sessions — polluting context with irrelevant information.

### Why existing mechanisms don't work

- **`CLAUDE_CONFIG_DIR` override**: Claude Code stores OAuth tokens in the macOS Keychain keyed to the config directory path. Changing `CLAUDE_CONFIG_DIR` for agent sessions breaks authentication ("Not logged in").
- **`--bare` mode**: Disables keychain auth entirely, requiring `ANTHROPIC_API_KEY`. Not viable for OAuth-only setups.
- **Project-level settings**: Affects interactive sessions in the same working directory.

### Current disable check

`bun-runner.js` already has an early-exit check (line 84-98):

```javascript
function isPluginDisabledInClaudeSettings() {
  const configDir = process.env.CLAUDE_CONFIG_DIR || join(homedir(), '.claude');
  const settingsPath = join(configDir, 'settings.json');
  if (!existsSync(settingsPath)) return false;
  const settings = JSON.parse(readFileSync(settingsPath, 'utf-8'));
  return settings?.enabledPlugins?.['claude-mem@thedotmack'] === false;
}
```

This only works via `CLAUDE_CONFIG_DIR`, which breaks auth as described above.

## Proposed solution

Add a single env var check before the existing disable check in `bun-runner.js`:

```javascript
if (process.env.CLAUDE_MEM_DISABLED === '1') {
  process.exit(0);
}
```

This allows orchestration platforms to disable claude-mem per-session via environment variable injection (e.g., Paperclip's `adapterConfig.env`) without affecting:
- Global plugin state
- Authentication
- Other plugins or hooks
- Interactive user sessions

## Environment

- claude-mem v10.6.2
- Claude Code (latest)
- macOS (Apple Silicon)
- Paperclip with `claude_local` adapter (11 agents on heartbeat schedules)

Thank you for the excellent plugin — this small addition would make it fully compatible with multi-agent setups.

bluzername:
## What this does

Adding support for `CLAUDE_MEM_DISABLED=1` environment variable so you can turn off claude-mem in specific sessions without touching global settings or breaking auth.

I was running into this problem when using multi-agent setup with Paperclip - all my agent sessions were getting context injected (wasting tokens) and recording observations from different domains into same database (making context messy for unrelated sessions).

## Changes

- `plugin/scripts/bun-runner.js` - env var check before anything else runs, so it exits fast with code 0
- `src/shared/plugin-state.ts` - exported `isDisabledByEnvVar()` helper so other entry points can also use this check
- `tests/infrastructure/plugin-disabled-check.test.ts` - tests covering the new function (true only for exact "1" value)
- `README.md` - documented the env var under Configuration section

## Why not use existing disable methods

Like the issue author mentioned:
- `CLAUDE_CONFIG_DIR` override breaks OAuth because macOS Keychain is tied to the config dir path
- `--bare` mode needs `ANTHROPIC_API_KEY`, doesnt work for OAuth-only setups
- Project-level settings affect interactive sessions too

The env var approach is cleanest - just set it per-process and everything else stays untouched.

## How to use

```bash
CLAUDE_MEM_DISABLED=1 claude --print "do something"
```

Or in orchestration config (like Paperclip's adapterConfig.env).

## Testing

Added 5 test cases to existing test file - checks that only exact value "1" triggers disable, other values like "0", "true", empty string are all treated as not disabled.

Closes #1484

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,21 @@ Settings are managed in `~/.claude-mem/settings.json` (auto-created with default
 
 See the **[Configuration Guide](https://docs.claude-mem.ai/configuration)** for all available settings and examples.
 
+### Disabling for Headless/Agent Sessions
+
+If you run Claude Code in headless or multi-agent setups (e.g. Paperclip, CI pipelines), you can disable claude-mem per-session with an environment variable:
+
+```bash
+CLAUDE_MEM_DISABLED=1 claude --print "your prompt here"
+```
+
+This is useful when:
+- Agent sessions should not inject or record observations
+- You want to avoid token waste from context injection in automated pipelines
+- Cross-contamination between agent and user sessions is a concern
+
+Unlike `CLAUDE_CONFIG_DIR` overrides, this does not break OAuth authentication.
+
 ---
 
 ## Development
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -79,6 +79,13 @@ function findBun() {
   return null;
 }
 
+// Early exit if disabled via environment variable (#1484).
+// Allows orchestration platforms (e.g. Paperclip) to disable claude-mem
+// per-session without affecting auth or global plugin state.
+if (process.env.CLAUDE_MEM_DISABLED === '1') {
+  process.exit(0);
+}
+
 // Early exit if plugin is disabled in Claude Code settings (#781).
 // Sync read + JSON parse — fastest possible check before spawning Bun.
 function isPluginDisabledInClaudeSettings() {
--- a/src/shared/plugin-state.ts
+++ b/src/shared/plugin-state.ts
@@ -10,6 +10,15 @@ import { homedir } from 'os';
 const PLUGIN_SETTINGS_KEY = 'claude-mem@thedotmack';
 
 /**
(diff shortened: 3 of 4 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The feature ask is now tracked in #1976, which is the single canonical issue for prioritization. Thanks for the suggestion — your context is captured there.

Issue #1490, opened 2026-03-25
bug: Worker daemon repeatedly killed by version mismatch shutdown, startup fratricide, and session SIGTERM — no recovery


rymalia:
### Summary

The worker daemon dies repeatedly from three independent application-level kill mechanisms and has no auto-restart capability, leaving it permanently dead until the next successful cold start. On macOS ARM64 with the marketplace-distributed plugin (v10.6.1), the worker rarely survives more than a few hours.

This is not a single bug — it's three separate code paths that each independently kill the worker, compounded by a cold start timeout that prevents recovery. This issue documents all three mechanisms with log evidence and reproduction steps, and proposes targeted fixes for the two most impactful ones.

### Environment

- **claude-mem**: v10.6.1 (marketplace distribution)
- **Platform**: macOS 26.3.x (ARM64, Apple Silicon)
- **Bun**: 1.2.x
- **ChromaDB**: enabled (local mode)

### Kill Mechanism 1: Version Mismatch HTTP Shutdown (most frequent)

**Root cause**: `BUILT_IN_VERSION` falls back to `"development"` because the marketplace build pipeline does not replace `__DEFAULT_PACKAGE_VERSION__`:

```js
// Server.ts:26 — in the marketplace bundle, __DEFAULT_PACKAGE_VERSION__ is never substituted
const BUILT_IN_VERSION = typeof __DEFAULT_PACKAGE_VERSION__ !== 'undefined'
  ? __DEFAULT_PACKAGE_VERSION__
  : 'development';
```

When a hook calls `ensureWorkerStarted()` and the PID file is absent or stale (but the worker is responding on the port), `checkVersionMatch()` compares `"10.6.1"` (from `package.json`) against `"development"` (from the worker's `/api/version` endpoint). The mismatch triggers `httpShutdown()` — sending `POST /api/admin/shutdown` to the running worker. The worker shuts down gracefully. The hook then tries to respawn, but often fails the `POST_SPAWN_WAIT` timeout (5 seconds, too short for macOS cold starts).

**This is 100% reproducible** — the version will always mismatch because the build constant is never baked into the marketplace bundle.

**Log evidence**:
```
[23:21:49.401] Worker PID file points to a live process, skipping duplicate spawn
[23:21:49.411] Worker became healthy while waiting on live PID
[23:21:49.420] Worker available {}
[23:21:49.510] Shutdown initiated          ← HTTP shutdown, no SIGTERM logged
[...]

Issue #1435, opened 2026-03-21
Binary embeds stale worker version (10.3.1) mismatching plugin   version (10.6.1), causes infinite restart loop


merlinhong:
```markdown
  ## Bug Description

  The `claude-mem` binary embeds an older worker version (10.3.1), while
  the plugin distributed via the marketplace is 10.6.1. The version
  check fails on every startup, triggering an infinite restart loop that
  makes the service permanently unavailable and wastes system resources.

  ## Environment

  - OS: macOS Tahoe 26.3
  - Plugin version: 10.6.1
  - Binary embedded worker version: 10.3.1

  ## Symptoms

  - Worker process restarts continuously in a tight loop
  - CPU usage spikes from repeated process spawning
  - `/api/health` never stabilizes
  - MCP connection cannot be established

  ## Root Cause

  Two separate issues compound each other:

  **1. Binary not updated alongside plugin release**

  The `claude-mem` binary ships with an embedded worker at version 10.3.1,
  but the plugin scripts in the marketplace are 10.6.1. The version
  mismatch check triggers a restart, which never resolves because the
  binary always re-embeds the old worker.

  **2. Version fallback defaults to placeholder strings**

  ```js
  // Line 43682
  const version = pkg.version || "development"

  // Line 68316
  const version = pkg.version || "0.0.0-dev"

  When pkg.version is unavailable (e.g. bundled without package.json),
  the fallback is a non-semver placeholder that fails version comparison
  and triggers another restart cycle.

  Fix

  1. Release process: Ensure the binary is rebuilt and re-published
  in sync with every plugin version bump. The embedded worker version
  in the binary must match the plugin version exactly.
  2. Version fallback: Inject the actual version string at build time
  rather than using placeholder strings:
  const version = pkg.version || "10.6.1"  // injected at build time
  2. Or better, fail fast with a clear error rather than falling back to
  an invalid semver string.

  Workaround (for affected users)

  Use bun to run the marketplace worker directly instead of the binary:

  ~/.bun/bin/bun \
    ~/.claude/plugins/cache/thedotmack/claude-mem/10.6.1/scripts/worker-service.
  cjs \
    daemon &

  This bypasses the binary entirely and uses the correct version.

  Impact

  Severity: High — Renders the service completely non-functional after
  plugin updates. The restart loop is not self-healing and requires manual
  intervention. Users have no indication of the root cause from logs alone.

thedotmack:
[...]

Issue #1426, opened 2026-03-21
aggressiveStartupCleanup SIGKILLs hook process on cold start, causing SessionStart hook error and empty context


sangfansh:
## Summary

On cold start (after reboot), the worker daemon's `aggressiveStartupCleanup()` kills the hook process that spawned it, causing:
- "SessionStart:startup hook error" shown to user
- Empty context output (no claude-mem banner)
- Everything works on second launch because daemon is already running

## Root Cause

`aggressiveStartupCleanup()` kills all processes whose command line matches `worker-service.cjs` (defined in the `QS` array), excluding only its own PID. However, the hook process (`bun worker-service.cjs hook claude-code context`) also matches this pattern. The daemon SIGKILL's its own parent — the hook that's waiting for readiness.

Introduced in commit `d333c7d` ("Expand startup orphan cleanup to target mcp-server and worker-service processes"), present since v9.1.0. Was masked by the in-process worker fallback until v10.5.6 removed it (PR #1370).

## Fix

Add `process.ppid` to the exclusion check in `aggressiveStartupCleanup()`:

```diff
- if (!Number.isInteger(u) || u <= 0 || u === e) continue;
+ if (!Number.isInteger(u) || u <= 0 || u === e || u === process.ppid) continue;
```

## Additional Issues Found

During investigation, several other cold start issues were identified:

### 1. Stale `Setup` hook references deleted `setup.sh`
`hooks.json` still has a `Setup` hook calling `setup.sh`, which was deleted in v10.5.0 (commit `0e502db`, "refactor: simplify hooks configuration and remove setup script"). Causes a hook error on every startup.

**Fix:** Remove the `Setup` entry from `hooks.json`.

### 2. Redundant `start` hook in SessionStart
The `start` command and `hook claude-code context` command both call `ensureWorkerStarted()`. Since hooks run in parallel, both try to spawn the daemon simultaneously, causing "Is port 37777 in use?" errors.

**Fix:** Remove the explicit `start` hook from SessionStart. The `context` hook already handles daemon startup internally.

### 3. `collectStdin` 5-second timeout wastes startup time
`bun-runner.js` waits up to 5 seconds for stdin data before launching bun. SessionStart hooks don't receive stdin, so this is pure wasted time. 500ms is sufficient — if Claude Code pipes data, it arrives instantly.

**Fix:** Change `setTimeout` in `collectStdin()` from `5000` to `500`.

### 4. `hook` command doesn't wait for DB readiness
[...]

Issue #1423, opened 2026-03-20
Worker daemon cold start exceeds POST_SPAWN_WAIT timeout (5s) on macOS ARM64


kishorebudha:
### Description
Worker daemon cold start exceeds `POST_SPAWN_WAIT` timeout (5 seconds) on macOS ARM64. The daemon spawns successfully via `spawnDaemon()` but takes ~6+ seconds to become healthy, causing every `SessionStart` hook to fail with `"Failed to start worker"`.

### Environment
- **claude-mem**: v10.6.1
- **Claude Code**: 2.1.80
- **Platform**: macOS 26.3.1 (arm64)
- **Bun**: 1.3.11 (at ~/.bun/bin/bun, not in PATH)
- **Node**: v25.8.1
- **ChromaDB**: enabled (v1.5.5, local mode, port 8000)

### Steps to Reproduce
1. Ensure no worker is running (`lsof -i :37777` shows nothing)
2. Start a new Claude Code session
3. `SessionStart` hook fires, runs `worker-service.cjs start`
4. `ensureWorkerStarted()` calls `spawnDaemon()` which spawns `~/.bun/bin/bun worker-service.cjs --daemon`
5. `waitForHealth(port, 5000)` polls for health
6. Daemon becomes healthy at ~6s, but timeout (5s) already expired
7. PID file is removed, hook returns `{status: "error", message: "Failed to start worker"}`

### Evidence from logs
```
[2026-03-20 17:36:37.343] [INFO ] [SYSTEM] Starting worker daemon
[2026-03-20 17:36:42.411] [ERROR] [SYSTEM] Worker failed to start (health check timeout)
```
Consistently 5.0-5.1 seconds between spawn and timeout — exactly matching `POST_SPAWN_WAIT: 5000`.

### Expected Behavior
Worker should start within the timeout. On macOS ARM64, 5 seconds is not enough for Bun cold start + ChromaDB initialization. Suggest increasing `POST_SPAWN_WAIT` to 10-15 seconds, or making it configurable.

### Workaround
Pre-start the worker daemon via a macOS LaunchAgent so it is already running when Claude Code launches:
```xml
<!-- ~/Library/LaunchAgents/com.claude-mem.worker.plist -->
<plist version="1.0">
<dict>
    <key>Label</key><string>com.claude-mem.worker</string>
    <key>ProgramArguments</key>
    <array>
        <string>~/.bun/bin/bun</string>
[...]

PR #1491, opened 2026-03-25
fix: prevent worker daemon from being killed by its own hooks (#1490)

rymalia:
Three independent fixes for worker daemon instability:

1. Remove version mismatch auto-restart from ensureWorkerStarted() (#1435). The marketplace bundle ships with __DEFAULT_PACKAGE_VERSION__ unbaked, causing BUILT_IN_VERSION to fall back to "development". This creates a 100% reproducible mismatch on every hook call, killing a healthy worker and often failing to restart. Same pattern across these issues (8+ releases): 
- #566
- #665
- #667
- #669
- #689
- #1124
- #1145

2. Add process.ppid and PID-file PID to aggressiveStartupCleanup() exclusions (#1426). Without this, a newly spawned daemon SIGKILLs the hook process that spawned it and any already-running worker the PID file points to.

3. Increase POST_SPAWN_WAIT from 5s to 15s (#1423). The 5s timeout was sized for Linux (<1s startup) but macOS ARM64 cold starts take 6-8s with Chroma enabled.


rymalia:
Addressed CodeRabbit review feedback in 25c1aa1f:

1. **Fixed POST_SPAWN_WAIT test assertion** (`tests/hook-constants.test.ts`) — updated from 5000 to 15000 to match the constant change
2. **Removed redundant `readPidFile()` from `aggressiveStartupCleanup()`** (`ProcessManager.ts`) — `start()` writes the new PID before `initializeBackground()` calls cleanup, so `readPidFile()` always returns `process.pid` (already protected). Added comment explaining why. The pre-existing worker is protected by `ensureWorkerStarted()` returning early when a healthy worker is detected — we never reach `aggressiveStartupCleanup()` in that case.
3. **Added `waitForReadiness()` to the reused-worker path** (`worker-service.ts`) — mirrors the fresh-spawn path so concurrent hooks don't race past a cold-starting worker's initialization guard

rymalia:
addressed CodeRabbit review feedback in fa959ba

rymalia:
@coderabbitai resolve

PR #1494, opened 2026-03-26
fix: use null-byte delimiter in observation content hash to prevent collisions

3em0:
## Summary
- Observation dedup hash concatenated 3 fields (`memorySessionId`, `title`, `narrative`) without delimiters, allowing different tuples to produce identical hashes — a memory poisoning vector
- Now joins with `\x00` so field boundaries are unambiguous
- Added regression test covering 4 collision scenarios that previously all hashed identically

## Example collision (before fix)

| memorySessionId | title | narrative | concat result |
|---|---|---|---|
| `session-abc` | `debug log` | `""` | `session-abcdebug log` |
| `session-ab` | `cdebug log` | `""` | `session-abcdebug log` |
| `session-` | `abcdebug log` | `""` | `session-abcdebug log` |
| `""` | `session-abcdebug log` | `""` | `session-abcdebug log` |

## Test plan
- [x] All 13 existing data-integrity tests pass
- [x] New collision regression test verifies 4 ambiguous tuples produce 4 distinct hashes

### DIFF
--- a/src/services/sqlite/observations/store.ts
+++ b/src/services/sqlite/observations/store.ts
@@ -22,7 +22,7 @@ export function computeObservationContentHash(
   narrative: string | null
 ): string {
   return createHash('sha256')
-    .update((memorySessionId || '') + (title || '') + (narrative || ''))
+    .update([memorySessionId || '', title || '', narrative || ''].join('\x00'))
     .digest('hex')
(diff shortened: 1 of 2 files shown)


3em0:
@thoraxe @rodboev @delorenj hello?

PR #1525, opened 2026-03-30
fix: expand tilde (~) to home directory in project resolver

### LINKED ISSUE #1478: Project resolver doesn't expand ~ to home directory path
## Summary

When a terminal reports `cwd` as `~` (unexpanded tilde) instead of the full path (e.g., `/Users/logan`), the project name resolver doesn't recognize it and falls back to `unknown-project`. The full path `/Users/logan` resolves correctly.

## Root cause

The resolver receives the literal string `~` and doesn't expand it to the user's home directory before attempting project resolution. Since `~` doesn't match any known path patterns, it falls through to the `unknown-project` fallback.

## Reproduction

1. Open a terminal where `cwd` is reported as `~` (not `/Users/<username>`)
3. Check the pending queue:
   ```
   curl -s http://localhost:37777/api/pending-queue | jq '.queue.messages[].project'
   # Returns: "unknown-project"
   ```

## Expected behavior

`~` should be expanded to the home directory path (e.g., `/Users/logan`) before project resolution, so it resolves the same way as the full path.

## Suggested fix

Expand `~` to `$HOME` (or equivalent) at the start of the project name resolver, before any other logic runs. This is a single-point fix that avoids special-casing and lets the existing resolution logic handle it normally.

## Environment

- claude-mem: 10.6.2
- OS: macOS (Darwin 25.3.0, Apple Silicon)

## Related Issues

- #1256 — basename project detection causes fragmentation in monorepos
- #1458 — Same folder name shares memories due to basename-only scoping

SeongjunKim36:
## Summary
- Expand literal `~` and `~/...` to the actual home directory path in `getProjectName()` before resolution
- Prevents fallback to `unknown-project` when terminals report cwd as unexpanded tilde

## Details
Some terminals report `cwd` as the literal string `~` instead of the full path (e.g., `/Users/logan`). The project name resolver didn't handle this, causing it to fall through to `unknown-project`.

The fix adds a simple tilde expansion using `os.homedir()` at the start of `getProjectName()`, before any other resolution logic runs.

Fixes #1478

## Test plan
- [ ] Verify `getProjectName('~')` returns the home directory basename (e.g., `logan`)
- [ ] Verify `getProjectName('~/projects/my-app')` returns `my-app`
- [ ] Verify existing paths without `~` are unaffected

### DIFF
--- a/src/utils/project-name.ts
+++ b/src/utils/project-name.ts
@@ -1,4 +1,5 @@
 import path from 'path';
+import { homedir } from 'os';
 import { logger } from './logger.js';
 import { detectWorktree } from './worktree.js';
 
@@ -15,8 +16,13 @@ export function getProjectName(cwd: string | null | undefined): string {
     return 'unknown-project';
   }
 
+  // Expand ~ to home directory (some terminals report cwd as literal ~)
+  const resolvedCwd = (cwd === '~' || cwd.startsWith('~/'))
+    ? cwd.replace('~', homedir())
+    : cwd;
+
   // Extract basename (handles trailing slashes automatically)
-  const basename = path.basename(cwd);
+  const basename = path.basename(resolvedCwd);
 
   // Edge case: Drive roots on Windows (C:\, J:\) or Unix root (/)
   // path.basename('C:\') returns '' (empty string)


thedotmack:

PR #1553, opened 2026-04-01
fix: persist session completion to database in completeByDbId (#1532)

### LINKED ISSUE #1532: `completeByDbId` does not update `sdk_sessions` status/completed_at in database
## Summary

When a session is completed via `POST /api/sessions/complete`, the `sdk_sessions` table is never updated — `status` stays `'active'` and `completed_at` remains `NULL`. This causes stale sessions to accumulate and eventually exhaust the worker's agent pool slots.

## Root Cause

`handleCompleteByClaudeId` in `SessionRoutes` calls `completionHandler.completeByDbId(sessionDbId)`, which only:
1. Calls `sessionManager.deleteSession()` — removes the session from the in-memory map
2. Calls `eventBroadcaster.broadcastSessionCompleted()` — broadcasts an event

Neither step updates the database. The `SessionStore` has no `markSessionCompleted()` method that would execute:
```sql
UPDATE sdk_sessions SET status = 'completed', completed_at = datetime('now'), completed_at_epoch = ? WHERE id = ?
```

## Impact

- Sessions accumulate as `status='active'` indefinitely
- Worker's agent pool slots are consumed by ghost sessions
- Eventually causes `Timed out waiting for agent pool slot after 60000ms` errors (63 occurrences in a single day in our case)
- New observation/summarize processing gets blocked, causing pending queue backlog

## Reproduction

1. Start a session via OpenClaw or any integration that calls `/api/sessions/complete`
2. End the session (triggers `POST /api/sessions/complete`)
3. Query: `SELECT id, status, completed_at FROM sdk_sessions ORDER BY id DESC LIMIT 5;`
4. Observe: `status='active'`, `completed_at=NULL` despite session being "completed"

## Expected Behavior

`completeByDbId` should update the database record:
- Set `status = 'completed'`
- Set `completed_at` and `completed_at_epoch` to current timestamp

## Environment
- claude-mem version: 10.6.3
- Worker: running on port 37777
- Platform: WSL2 (Linux)

ousamabenyounes:
## Summary

- `completeByDbId` only cleaned up in-memory state — `sdk_sessions` rows kept `status='active'` and `completed_at=NULL` forever
- Ghost sessions accumulated and exhausted the agent pool, causing `Timed out waiting for agent pool slot after 60000ms`
- Fix: add `SessionStore.markSessionCompleted()` and call it at the start of `completeByDbId`

## Root cause

```ts
// Before — no DB write, sessions accumulate as ghost 'active' rows
async completeByDbId(sessionDbId: number): Promise<void> {
  await this.sessionManager.deleteSession(sessionDbId);
  this.eventBroadcaster.broadcastSessionCompleted(sessionDbId);
}
```

```ts
// After — DB marked before in-memory cleanup
async completeByDbId(sessionDbId: number): Promise<void> {
  this.sessionStore.markSessionCompleted(sessionDbId);  // ← new
  await this.sessionManager.deleteSession(sessionDbId);
  this.eventBroadcaster.broadcastSessionCompleted(sessionDbId);
}
```

## Changes

| File | Change |
|------|--------|
| `src/services/sqlite/SessionStore.ts` | Add `markSessionCompleted()` — sets `status='completed'`, `completed_at`, `completed_at_epoch` |
| `src/services/worker/session/SessionCompletionHandler.ts` | Inject `SessionStore`, call `markSessionCompleted` before delete |
| `src/services/worker/http/routes/SessionRoutes.ts` | Pass `dbManager.getSessionStore()` to `SessionCompletionHandler` constructor |
| `tests/services/sqlite/session-store-mark-completed.test.ts` | 4 new tests (real SQLite `:memory:`, 0 mocks) |

## Test plan

- [x] New tests: 4 pass, 0 fail (real SQLite, no mocks)
- [x] Full suite: 1110 pass, 0 fail (+4 new)
- [x] Verified: `SELECT status, completed_at FROM sdk_sessions` shows `completed` after `completeByDbId`

Closes #1532

### DIFF
--- a/src/services/worker/http/routes/SessionRoutes.ts
+++ b/src/services/worker/http/routes/SessionRoutes.ts
@@ -41,7 +41,7 @@ export class SessionRoutes extends BaseRouteHandler {
     this.completionHandler = new SessionCompletionHandler(
       sessionManager,
       eventBroadcaster,
-      dbManager.getSessionStore()
+      dbManager
     );
   }
 
--- a/src/services/worker/session/SessionCompletionHandler.ts
+++ b/src/services/worker/session/SessionCompletionHandler.ts
@@ -5,21 +5,30 @@
  * Used by DELETE /api/sessions/:id and POST /api/sessions/:id/complete endpoints.
  *
  * Completion flow:
- * 1. Delete session from SessionManager (aborts SDK agent, cleans up in-memory state)
- * 2. Broadcast session completed event (updates UI spinner)
+ * 1. Persist completion to database (fix for #1532)
+ * 2. Delete session from SessionManager (aborts SDK agent, cleans up in-memory state)
+ * 3. Broadcast session completed event (updates UI spinner)
+ *
+ * Note: DatabaseManager is injected instead of SessionStore to avoid calling
+ * getSessionStore() during route construction, before DB initialization (#1553).
+ * The store is resolved lazily at call time via the private getter.
  */
 
 import { SessionManager } from '../SessionManager.js';
 import { SessionEventBroadcaster } from '../events/SessionEventBroadcaster.js';
-import { SessionStore } from '../../sqlite/SessionStore.js';
+import { DatabaseManager } from '../DatabaseManager.js';
 import { logger } from '../../../utils/logger.js';
 
 export class SessionCompletionHandler {
   constructor(
     private sessionManager: SessionManager,
     private eventBroadcaster: SessionEventBroadcaster,
-    private sessionStore: SessionStore
+    private dbManager: DatabaseManager
   ) {}
+
+  private get sessionStore() {
+    return this.dbManager.getSessionStore();
+  }
 
   /**
    * Complete session by database ID
--- a/tests/services/worker/session-completion-handler-lazy-init.test.ts
+++ b/tests/services/worker/session-completion-handler-lazy-init.test.ts
@@ -1,0 +1,35 @@
+/**
+ * Tests for SessionCompletionHandler lazy DB initialization (fix for #1553)
+ *
+ * Regression: SessionCompletionHandler previously called dbManager.getSessionStore()
+ * in the constructor, which threw "Database not initialized" on Linux because
+ * SessionRoutes is constructed before dbManager.initialize() runs in start().
+ *
+ * Mock Justification: MINIMAL — only DatabaseManager is mocked, with a controlled
+ * flag to simulate uninitialized vs initialized state. All other behavior is tested
+ * in session-store-mark-completed.test.ts using real SQLite.
+ */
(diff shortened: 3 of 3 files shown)


alessandropcostabr:
## Bug: `getSessionStore()` called before DB initialization on Linux

### Environment
- **OS:** Ubuntu 24.04 (Linux x86_64)
- **Bun:** 1.3.11
- **claude-mem:** 10.7.0+ (build including this PR)

### Problem

This PR injects `SessionStore` into `SessionCompletionHandler` via the constructor:

```ts
// SessionRoutes constructor
this.completionHandler = new SessionCompletionHandler(
  sessionManager,
  eventBroadcaster,
  dbManager.getSessionStore()  // ← fails here
);
```

The issue is that `SessionRoutes` is instantiated during `WorkerService.registerRoutes()`, which runs **before** `dbManager.initialize()` is called. At that point, `sessionStore` is still `null`, so `getSessionStore()` throws:

```
Error: Database not initialized
    at getSessionStore (worker-service.cjs:1188)
    at new SessionRoutes (worker-service.cjs:1491)
    at registerRoutes (worker-service.cjs:1507)
    at new WorkerService (worker-service.cjs:1507)
```

The worker crashes silently (`process.exit(0)`) and retries in a loop. On our setup this caused **~48h of lost observations** — the worker kept failing and no memory was being recorded.

### Why it works on macOS but not Linux

The previous version (without this PR) passes only 2 args to `SessionCompletionHandler` and doesn't touch the DB during route setup. The build artifact shipped in the plugin marketplace includes this change, even though the PR hasn't been merged to main.

### Suggested fix

Lazy initialization — pass `dbManager` instead of `dbManager.getSessionStore()`, and resolve the store inside the methods that need it:

```ts
// Option A: lazy getter
export class SessionCompletionHandler {
  constructor(
    private sessionManager: SessionManager,
    private eventBroadcaster: SessionEventBroadcaster,
    private dbManager: DatabaseManager
  ) {}

  private get sessionStore() {
    return this.dbManager.getSessionStore();
  }
}
```

Or move `registerRoutes()` to after `dbManager.initialize()` in the `start()` method.

### Reproduction

```bash
# On Linux with Bun 1.3.11
cd ~/.claude/plugins/marketplaces/thedotmack/plugin/scripts
CLAUDE_MEM_WORKER_PORT=37777 bun -e "
  const mod = require('./worker-service.cjs');
  try { new mod.WorkerService(); } catch(e) { console.error(e.message); }
"
# Output: Database not initialized
```

### Workaround

Reverting `worker-service.cjs` to the cached version (pre-PR build from Apr 3) restores normal operation.

---

cc @thedotmack — this is likely why the build in the marketplace plugin is broken on Linux. Happy to submit a fix PR if helpful.

thedotmack:

Issue #1505, opened 2026-03-27
SessionStart hooks fail on cold start: worker-start exits 137, context hook races worker


nimesh-kumar-sh:
## Summary

Two of the three `SessionStart` hooks produce non-zero exit codes on cold start (worker not running), causing Claude Code to display `SessionStart:startup hook error` twice on every fresh session.

## Root Cause

Since Claude Code runs **all matching hooks in parallel**, the three SessionStart hooks race each other:

1. **`smart-install`** — always succeeds (exit 0)
2. **`worker-start`** (`bun-runner.js worker-service.cjs start`) — the `start` subcommand forks a daemon, then sends **SIGKILL to its own process group**. This kills both the bun parent AND the node wrapper (`bun-runner.js`) before any `close` event handler can fire. Result: **exit 137, no stdout output**.
3. **`context`** (`bun-runner.js worker-service.cjs hook claude-code context`) — runs simultaneously with `worker-start`, tries to reach `localhost:37777` before the worker is listening. Result: **connection refused, non-zero exit**.

Despite the errors, the worker daemon **does start successfully** — the exit code is a false positive from the parent process being killed.

## Reproduction

```bash
# Kill the worker
kill -9 $(lsof -t -i :37777)
sleep 3

# Run worker-start — exits 137 despite starting the daemon
_R="$HOME/.claude/plugins/cache/thedotmack/claude-mem/10.6.2"
CLAUDE_PLUGIN_ROOT="$_R" node "$_R/scripts/bun-runner.js" "$_R/scripts/worker-service.cjs" start
echo "EXIT: $?"  # prints 137

# Worker IS running:
lsof -i :37777  # shows bun listening
```

### Option A: Fix `worker-service.cjs start` exit behavior
The `start` command should not SIGKILL its own process group. After forking the daemon, the parent should exit cleanly with code 0. This is the proper fix.

### Option B: Fix the hooks.json commands (workaround)
Wrap the `worker-start` command with `;` continuation and a health-check wait loop so the shell continues after the SIGKILL and returns 0:

```json
{
  "type": "command",
  "command": "... node \"$_R/scripts/bun-runner.js\" \"$_R/scripts/worker-service.cjs\" start; for i in 1 2 3 4 5 6 7 8; do curl -sf http://localhost:37777/health >/dev/null 2>&1 && break; sleep 1; done; echo '{\"continue\":true,\"suppressOutput\":true}'",
  "timeout": 60
}
```

[...]

Issue #1245, opened 2026-02-26
[Bug] worker-service.cjs start subcommand causes SIGKILL under systemd (fork-then-exit)


KJT125:
## Environment
- OS: WSL2 Ubuntu on Windows 11
- Runtime: Bun 1.x
- claude-mem: v10.4.1 (bundled CJS, via OpenClaw plugin install)
- Supervisor: systemd user service (Type=simple)

## Problem

Running `bun worker-service.cjs start --daemon` under systemd results in consistent SIGKILL (signal 9) exactly 2 seconds after start. 100% reproducible across multiple days.

## Root Cause

The `start` subcommand triggers fork-then-exit / self-daemonize. Under `systemd Type=simple`, systemd tracks the original PID. When the parent exits after forking, systemd sends SIGKILL to the entire cgroup.

**Evidence:**
- `bun worker-service.cjs` → stable 10+ min ✅
- `bun worker-service.cjs start --daemon` → dead in 2s ❌
- Same binary, same env, same systemd config.

## Workaround

Run foreground under systemd (no `start`, no `--daemon`):

```ini
[Service]
Type=simple
ExecStart=/path/to/bun worker-service.cjs
Restart=always
```

## Suggestion

1. Provide `--foreground` or `run` subcommand for supervisor-friendly usage
2. Document systemd unit file examples
3. If daemon mode needed, support `Type=forking` + `PIDFile=`

thedotmack:
The Process Supervisor (PR #1370, shipped in v10.5.6) restructures process spawning with `detached: true` and proper signal handling (SIGTERM/SIGINT with graceful shutdown cascade). The `shutdownInitiated` flag prevents duplicate cascades. This may help with the systemd SIGKILL issue since the supervisor now handles its own lifecycle rather than relying on the fork-then-exit pattern. If you're still hitting this on v10.5.6, please reopen.

thedotmack:
Closed by @thedotmack. Please reopen if still relevant. Thank you for your contribution!

PR #1567, opened 2026-04-02
feat: drain orphaned pending messages on SIGTERM session completion

alessandropcostabr:
## Summary

Drains orphaned pending messages when a session completes via the API (Stop hook → `completeByDbId`), preventing message accumulation in dead sessions.

## Problem

When the Stop hook fires, `completeByDbId()` calls `deleteSession()` which aborts the SDK agent via SIGTERM. The generator exits before processing its queue, leaving messages in `pending` status forever — no future generator will pick them up because the session is already `completed`.

**Observed in production:** 15 orphaned `summarize` messages across completed sessions, with ages between 3 hours and 3 days. These messages were never processed and accumulated indefinitely.

```
[SESSION] Generator auto-starting (init)  {queueDepth=0}
[QUEUE  ] ENQUEUED | messageId=10289 | type=summarize | depth=1
[SESSION] Generator aborted                     ← SIGTERM
[SESSION] Session deleted {duration=28.9s}
→ msg 10289 remains pending forever
```

## Solution

After `deleteSession()`, call `markAllSessionMessagesAbandoned(sessionDbId)` to drain any remaining pending messages. This method already exists in `PendingMessageStore` and is used by `worker-service.ts` in `terminateSession()` — we're reusing it in the completion handler path where it was missing.

```typescript
await this.sessionManager.deleteSession(sessionDbId);

// Drain orphaned pending messages left by SIGTERM
try {
  const pendingStore = this.sessionManager.getPendingMessageStore();
  const drainedCount = pendingStore.markAllSessionMessagesAbandoned(sessionDbId);
  if (drainedCount > 0) {
    logger.warn('SESSION', `Drained ${drainedCount} orphaned pending messages`);
  }
} catch (e) {
  logger.debug('SESSION', 'Failed to drain pending queue');
}

this.eventBroadcaster.broadcastSessionCompleted(sessionDbId);
```

**Why `completeByDbId` is the right place:** It's the entry point for ALL intentional session termination via the API (DELETE and POST /complete endpoints). `deleteSession()` itself is also called by `reapStaleSessions`, which already ensures zero pending before calling.

## Related Issues

- #1490 — Worker daemon killed by SIGTERM, no recovery
- #1505 — SessionStart hooks fail on cold start
- #1245 — worker-service.cjs start causes SIGKILL under systemd

## Production Data

| Metric | Before | After |
|--------|--------|-------|
| Orphaned pending messages | 15 (ages 3h-3d) | 0 |
| Days of operation | — | 23 days |
| Observations processed | — | 3,400+ |
| Sessions completed cleanly | — | 79 |

## Testing

- [x] Build compiles (worker-service.cjs)
- [x] Fix verified in compiled output
- [x] 23 days production usage, zero orphaned messages after fix
- [x] No regression — reuses existing `markAllSessionMessagesAbandoned` method
- [x] Graceful degradation — try-catch ensures completion never fails from drain

### DIFF
--- a/src/services/worker/session/SessionCompletionHandler.ts
+++ b/src/services/worker/session/SessionCompletionHandler.ts
@@ -31,6 +31,9 @@ export class SessionCompletionHandler {
     // When deleteSession() aborts the generator, pending messages in the queue
     // are never processed. Without drain, they stay in 'pending' status forever
     // since no future generator will pick them up for a completed session.
+    // Note: this is best-effort — if a generator outlives the 30s SIGTERM timeout
+    // (SessionManager.deleteSession), it may enqueue messages after this drain.
+    // In practice this race is rare (zero orphans over 23 days, 3400+ observations).
     try {
       const pendingStore = this.sessionManager.getPendingMessageStore();
       const drainedCount = pendingStore.markAllSessionMessagesAbandoned(sessionDbId);


alessandropcostabr:
## Upstream synergy note

This drain complements `terminateSession()` added in v10.6.2 — your code handles the generator cleanup side (abort + zombie path fixes + restart-or-terminate invariant), this handles the pending message side.

The gap: when `completeByDbId()` calls `deleteSession()`, the generator is aborted via SIGTERM, but messages already enqueued in `pending_messages` are never claimed. `terminateSession()` in `worker-service.ts` does call `markAllSessionMessagesAbandoned()`, but the hook-triggered completion path (`SessionCompletionHandler.completeByDbId`) doesn't — that's what this PR adds.

Together they cover the full pipeline: generator abort → message drain → session complete. The `markAllSessionMessagesAbandoned()` method already exists and is battle-tested in `terminateSession()` — we're reusing it in the one path where it was missing.

Production evidence: 0 orphaned messages over 23 days with this fix active.

alessandropcostabr:
## Heads-up: overlap with #1553

@ousamabenyounes's #1553 also modifies `completeByDbId` — it adds `sessionStore.markSessionCompleted(sessionDbId)` **before** `deleteSession()` to persist session status to the DB.

Our change adds `markAllSessionMessagesAbandoned(sessionDbId)` **after** `deleteSession()` to drain the pending queue.

Both fixes are needed and complementary:
```
markSessionCompleted()     ← #1553: persist status to DB
deleteSession()            ← existing: abort generator, cleanup memory
markAllSessionMessagesAbandoned()  ← #1567: drain orphaned pending messages
broadcastSessionCompleted() ← existing: update UI
```

If both PRs merge, the merge should be straightforward — they touch different lines in the same method. Happy to rebase on top of #1553 if needed.

ousamabenyounes:
Nice catch @alessandropcostabr — 15 orphaned messages with ages up to 3 days is exactly the kind of silent accumulation that's hard to spot without production data. The 23-day evidence is compelling.

Your analysis of the overlap with #1553 is spot-on. The two changes are fully complementary:

```
markSessionCompleted()              ← #1553: persist status to DB
deleteSession()                     ← existing: abort generator
markAllSessionMessagesAbandoned()   ← #1567: drain orphaned queue  
broadcastSessionCompleted()         ← existing: update UI
```

**Proposed merge order:** #1553 first (it's the simpler DB-only change), then rebase #1567 on main — the diff will be clean since we touch different lines. Happy to give a heads-up here when #1553 lands so you can rebase immediately.

Issue #1551, opened 2026-04-01
`<persisted-output>` tags leak into $CMEM context injection


cheefbird:
# `<persisted-output>` tags leak into $CMEM context injection

## Problem

Claude Code's `<persisted-output>` system tags are not being stripped before storage, causing them to leak into the `$CMEM` context injection in subsequent sessions. These tags are ephemeral transport wrappers used by Claude Code for large tool outputs and should not be persisted in observations.

## Environment

- **Claude-mem**: 10.6.3
- **Claude Code**: 2.1.89
- **Node.js**: v20.20.1
- **Bun**: 1.3.10
- **OS**: macOS 26.3.1 (arm64)
- **Platform**: darwin

**Paths:**
- Plugin: `~/.claude/plugins/marketplaces/thedotmack`
- Data Directory: `~/.claude-mem`
- Database: `~/.claude-mem/claude-mem.db` (47.3 MB)

**Configuration:**
- Model: sonnet
- Worker Port: 37777
- Context Observations: 50
- Log Level: INFO

## Steps to Reproduce

1. Run a tool that produces large output (e.g., a grep across many files returning 40KB+ of results)
2. Claude Code wraps the result in `<persisted-output>...</persisted-output>` tags
3. claude-mem captures the tool response via the PostToolUse hook
4. `src/utils/tag-stripping.ts` `stripTagsInternal()` function processes the content but does not strip `persisted-output` tags — it only strips: `claude-mem-context`, `private`, `system_instruction`, `system-instruction`
5. The tag content (including the closing `</persisted-output>`) is stored in the observation
6. On the next session, the `$CMEM` context injection includes the leaked tag

## Expected Behavior

`<persisted-output>` tags should be stripped before storage, same as other system-level tags. These are ephemeral Claude Code transport wrappers and have no semantic value in stored observations.

## Actual Behavior

`<persisted-output>` tags and their content are preserved in observations and leak into future `$CMEM` context injections.

## Suggested Fix

Add `persisted-output` to the list of stripped tags in `stripTagsInternal()` function in `src/utils/tag-stripping.ts`.

cheefbird:
didn't know where to include this in the report, but there is this possibly relevant claude-code issue: https://github.com/anthropics/claude-code/issues/17407

PR #1568, opened 2026-04-02
feat: semantic context injection via Chroma on UserPromptSubmit

alessandropcostabr:
## Summary

Adds per-prompt semantic search that injects the most **relevant** past observations into context, replacing the current recency-based approach. Complements the file-read gate (#1564) — this is proactive injection at the prompt level, that is reactive interception at the tool level.

## Problem

Current context injection (SessionStart) selects the N most recent observations. If a user asks about authentication, they get CSS observations because those happened last. After `/clear`, all context is lost until a new session starts.

## Solution

On every `UserPromptSubmit`, query ChromaDB with the user's prompt text and inject the top-5 semantically similar observations as `additionalContext`.

### New endpoint

`GET /api/context/semantic?q=<prompt>&project=<project>&limit=5`

Returns formatted markdown with relevant past work via `SearchManager.search()` with `format: 'json'`.

### How it works

```
User types prompt
  → session-init handler fires
  → if prompt >= 20 chars AND CLAUDE_MEM_SEMANTIC_INJECT=true:
    → GET /api/context/semantic?q=<prompt>
      → SearchManager.search() → Chroma vector similarity
      → Format top-N as markdown
    → Return as hookSpecificOutput.additionalContext
  → Claude receives relevant past work, not just recent work
```

### Comparison with current approach

| | Before (SessionStart) | After (UserPromptSubmit) |
|---|---|---|
| Trigger | 1x per session | Every prompt |
| Selection | Recency (SQL ORDER BY) | Relevance (Chroma vectors) |
| Post-`/clear` | Lost | Re-injects on next prompt |
| Tokens | ~1,800 | ~800-1,200 |
| Topic switch | Stale context | Context follows topic |

### Complementary to file-read gate (#1564)

| | Semantic injection (this PR) | File-read gate (#1564) |
|---|---|---|
| Hook | UserPromptSubmit | PreToolUse |
| When | Before any action | When Claude tries to read a file |
| Action | Injects relevant context proactively | Blocks read, offers cached observations |
| Philosophy | Proactive | Reactive / progressive |

Together: Claude gets relevant context upfront (may avoid reads entirely) + gate catches remaining redundant reads.

## Configuration

| Setting | Default | Description |
|---|---|---|
| `CLAUDE_MEM_SEMANTIC_INJECT` | `true` | Enable/disable per-prompt injection |
| `CLAUDE_MEM_SEMANTIC_INJECT_LIMIT` | `5` | Max observations injected per prompt |

## Production Data (23 days, 3,400+ observations, 8 projects)

| Metric | Before (recency) | After (semantic) |
|--------|-------------------|------------------|
| Context selection | 8 most recent | 5 most relevant |
| Topic switching | Stale (previous topic) | Follows current topic |
| Post-`/clear` | Memory lost | Re-injects on next prompt |
| Token cost | ~1,800 per injection | ~800-1,200 per injection |
| Observations indexed | — | 3,400+ in Chroma |

### Real-world validation

In a fresh session, asked to analyze a service:
- Model **recalled** relevant past work ("Recalled 1 memory")
- **Cited** project rules from documentation
- **Asked** before acting instead of making assumptions
- Result: 0 user corrections (vs 5+ without semantic injection)

## Testing

- [x] Build compiles (worker-service.cjs + mcp-server.cjs + context-generator.cjs)
- [x] All 3 markers verified in compiled output
- [x] 23 days production validation
- [x] Graceful degradation when Chroma unavailable
- [x] Minimum prompt length prevents noise
- [x] Skips media-only prompts

## Related

- #1551 — `<persisted-output>` tags leak into context injection
- #1564 — File-read decision gate (complementary approach)

### DIFF
--- a/src/cli/handlers/session-init.ts
+++ b/src/cli/handlers/session-init.ts
@@ -87,17 +87,18 @@ export const sessionInitHandler: EventHandler = {
 
     // Skip SDK agent re-initialization if context was already injected for this session (#1079)
     // The prompt was already saved to the database by /api/sessions/init above —
-    // no need to re-start the SDK agent on every turn
-    if (initResult.contextInjected) {
+    // no need to re-start the SDK agent on every turn.
+    // Note: we do NOT return here — semantic injection below must run on every prompt.
+    const skipAgentInit = Boolean(initResult.contextInjected);
+    if (skipAgentInit) {
       logger.info('HOOK', `INIT_COMPLETE | sessionDbId=${sessionDbId} | promptNumber=${promptNumber} | skipped_agent_init=true | reason=context_already_injected`, {
         sessionId: sessionDbId
       });
-      return { continue: true, suppressOutput: true };
     }
 
     // Only initialize SDK agent for Claude Code (not Cursor)
     // Cursor doesn't use the SDK agent - it only needs session/observation storage
-    if (input.platform !== 'cursor' && sessionDbId) {
+    if (!skipAgentInit && input.platform !== 'cursor' && sessionDbId) {
       // Strip leading slash from commands for memory agent
       // /review 101 -> review 101 (more semantic for observations)
       const cleanedPrompt = prompt.startsWith('/') ? prompt.substring(1) : prompt;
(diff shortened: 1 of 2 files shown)


alessandropcostabr:
## Upstream synergy note

This PR uses the `workerHttpRequest()` helper introduced in v10.5.6 (replacing the old `getWorkerPort()` + `fetch()` pattern) for the semantic endpoint call in `session-init.ts`. Compatible with the current hook API.

Also complements your file-read gate (#1564): that PR intercepts at the tool level (PreToolUse → "you already have observations for this file"), this one injects at the prompt level (UserPromptSubmit → "here are relevant past observations for your question"). Different trigger points, same goal: surface existing knowledge before Claude re-reads files.

Per CodeRabbit review, endpoint changed from GET to POST to avoid prompt exposure in URL logs and URL-length limits. Handler accepts both body and query params for backwards compatibility.

alessandropcostabr:
## Dependency note: project filter in Chroma search (#1540)

@Sheldenshi's #1540 fixes a bug where `searchObservations()` ignores the `project` parameter in Chroma queries. This means our semantic injection endpoint could return observations from **other projects** when querying Chroma.

Our `handleSemanticContext` passes `project` to `SearchManager.search()`, but without #1540's fix, the Chroma vector query runs unscoped. The SQLite hydration path was already correct — only the Chroma path was missing the project filter.

If #1540 merges first, our semantic injection automatically gets correct project scoping. If ours merges first, the endpoint works but may return cross-project results until #1540 lands.

Recommend merging #1540 before or alongside this PR.

xkonjin:
## Quick Review: Semantic Context Injection

This PR adds Chroma-based semantic context injection. A few observations:

**Strengths:**
- Clean integration with the existing prompt flow
- Good use of vector search for relevant context retrieval
- Proper error handling when Chroma is unavailable

**Questions:**
- What's the performance impact on prompt latency? Consider adding metrics
- How does this behave with very large memory collections? Any pagination or limits?

**Suggestion:** Consider making the similarity threshold configurable rather than hardcoded.

Overall looks like a solid feature addition.

PR #1576, opened 2026-04-02
fix: sort timeline newest-first so preview shows recent observations

### LINKED ISSUE #1565: Session context preview shows stale entries due to ascending timeline sort
## Problem

When Claude Code starts a new session, the `$CMEM` context block is injected via the SessionStart hook. Claude Code displays a **"Preview (first 2KB)"** of the hook output to the user. Because the timeline is sorted **chronologically ascending** (oldest entries first), the preview gets truncated before reaching the most recent observations.

**Result:** The user sees observations from days ago (e.g. Mar 30) instead of the latest session's entries (e.g. Apr 2), even though the full data is correctly loaded into the LLM context.

## Reproduction

1. Have 50+ observations spanning multiple days
3. The preview shows old entries from the beginning of the timeline
4. The most recent entries are cut off by the 2KB preview limit

## Root Cause

Two sort operations enforce ascending (oldest-first) order:

**`src/services/context/ObservationCompiler.ts` line 247:**
```typescript
// Sort chronologically
timeline.sort((a, b) => {
  const aEpoch = a.type === 'observation' ? a.data.created_at_epoch : a.data.displayEpoch;
  const bEpoch = b.type === 'observation' ? b.data.created_at_epoch : b.data.displayEpoch;
  return aEpoch - bEpoch;  // ascending — old entries first
});
```

**`src/services/context/sections/TimelineRenderer.ts` line 34-37:**
```typescript
const sortedEntries = Array.from(itemsByDay.entries()).sort((a, b) => {
  const aDate = new Date(a[0]).getTime();
  const bDate = new Date(b[0]).getTime();
  return aDate - bDate;  // ascending — old days first
});
```

## Suggested Fix

Reverse both sort comparators to `bEpoch - aEpoch` / `bDate - aDate` so that the most recent observations appear first in the output. This way, the 2KB preview shown to the user contains the most relevant (recent) context.

Alternatively, a configuration option for sort order (`newestFirst: true/false`) would let users choose their preferred display.

## Environment

- claude-mem version: 10.6.3
- Claude Code CLI
- ~50 observations loaded, spanning 4 days

octo-patch:
Fixes #1565

## Problem

The `$CMEM` context block shown during `SessionStart` is truncated to a 2KB preview. Because both sort comparators used ascending (oldest-first) order, the preview always showed old observations from days ago instead of the most recent ones.

## Solution

Reverse both sort comparators to descending (newest-first):

- `src/services/context/ObservationCompiler.ts` — `buildTimeline()`: change `aEpoch - bEpoch` → `bEpoch - aEpoch`
- `src/services/context/sections/TimelineRenderer.ts` — `groupTimelineByDay()`: change `aDate - bDate` → `bDate - aDate`

With this change the 2KB preview limit will truncate after recent entries, not before them.

## Testing

- Manually verified sort order reversal with 50+ observations spanning multiple days
- Full context still passed to LLM context unchanged; only the preview display order is affected

### DIFF
--- a/src/services/context/ObservationCompiler.ts
+++ b/src/services/context/ObservationCompiler.ts
@@ -240,11 +240,11 @@ export function buildTimeline(
     ...summaries.map(summary => ({ type: 'summary' as const, data: summary }))
   ];
 
-  // Sort chronologically
+  // Sort chronologically (newest first so preview truncation shows recent entries)
   timeline.sort((a, b) => {
     const aEpoch = a.type === 'observation' ? a.data.created_at_epoch : a.data.displayEpoch;
     const bEpoch = b.type === 'observation' ? b.data.created_at_epoch : b.data.displayEpoch;
-    return aEpoch - bEpoch;
+    return bEpoch - aEpoch;
   });
 
   return timeline;
--- a/src/services/context/sections/TimelineRenderer.ts
+++ b/src/services/context/sections/TimelineRenderer.ts
@@ -30,11 +30,11 @@ export function groupTimelineByDay(timeline: TimelineItem[]): Map<string, Timeli
     itemsByDay.get(day)!.push(item);
   }
 
-  // Sort days chronologically
+  // Sort days newest-first so preview truncation shows recent entries
   const sortedEntries = Array.from(itemsByDay.entries()).sort((a, b) => {
     const aDate = new Date(a[0]).getTime();
     const bDate = new Date(b[0]).getTime();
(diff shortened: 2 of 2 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1917, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1588, opened 2026-04-03
docs: align session ID architecture guide with current runtime semantics

suyua9:
## Summary
- correct the architecture guide to say observations are keyed by the real `memory_session_id`, not `contentSessionId`
- document the current resume gate: `memorySessionId`, `lastPromptNumber > 1`, and `!forceInit`
- update the pitfalls and usage examples so future contributors do not reintroduce the older placeholder model

## Validation
- `git diff --check`
- unable to run `bun test tests/session_id_usage_validation.test.ts` in this environment because `bun` is not installed

### DIFF
--- a/docs/SESSION_ID_ARCHITECTURE.md
+++ b/docs/SESSION_ID_ARCHITECTURE.md
@@ -30,7 +30,7 @@ Claude-mem uses **two distinct session IDs** to track conversations and memory:
                            ↓
 ┌─────────────────────────────────────────────────────────────┐
 │ 3. First SDK message arrives with session_id                │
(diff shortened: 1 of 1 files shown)


suyua9:
Applied the two documentation follow-ups from CodeRabbit in commit `8c037042`:
- updated the flow diagram to use `ensureMemorySessionIdRegistered(...)`
- aligned the invariant wording with the current truthiness-based detection (`!!memorySessionId`)

Validation this round:
- `git diff --check`

I still could not run the `bun`-based tests in this environment because `bun` is not installed.

PR #1593, opened 2026-04-04
fix: add AbortSignal.timeout to isPortInUse on non-Windows platforms

### LINKED ISSUE #1587: Worker daemon hangs on startup: fetch() in isPortInUse has no timeout, hangs forever on Bun 1.3.11/Linux
## Bug Description

The worker daemon hangs indefinitely on startup when no existing worker is running. The root cause is `isPortInUse()` in `src/services/infrastructure/HealthMonitor.ts` calling `fetch()` with no timeout. On Bun 1.3.11 on Linux/WSL2, `fetch()` to a port with nothing listening hangs forever instead of throwing `ECONNREFUSED`.

## Steps to Reproduce

1. Ensure no worker is running (no process on port 37777, no PID file)
2. Run: `bun worker-service.cjs restart` (or `start`, or `--daemon`)
3. The process hangs forever with no output, no log entries, and port 37777 never binds

## Root Cause

`HealthMonitor.ts:40-49`:

```typescript
export async function isPortInUse(port: number): Promise<boolean> {
  try {
    // Note: Removed AbortSignal.timeout to avoid Windows Bun cleanup issue (libuv assertion)
    const response = await fetch(`http://127.0.0.1:${port}/api/health`);
    return response.ok;
  } catch (error) {
    return false;
  }
}
```

The `AbortSignal.timeout` was removed to work around a Windows Bun libuv assertion. However, on Bun 1.3.11 (Linux x64 / WSL2), `fetch()` to a closed port **never rejects** — it hangs indefinitely. This blocks the daemon startup in the CLI dispatcher's `default` case at `await Vl(r)` (the minified `isPortInUse`), before the `WorkerService` is ever constructed.

### Verification

```bash
# Hangs forever:
timeout 5 bun -e "fetch('http://127.0.0.1:37777/api/health').then(r => console.log(r.status)).catch(e => console.log(e.message))"
# EXIT: 124 (killed by timeout)

# Works with timeout:
timeout 5 bun -e "fetch('http://127.0.0.1:37777/api/health', { signal: AbortSignal.timeout(2000) }).then(r => console.log(r.status)).catch(e => console.log(e.message))"
# Error: The operation timed out.
# EXIT: 0
```

## Suggested Fix

Add `AbortSignal.timeout` back, conditionally skipping it on Windows if the libuv assertion is still an issue:

```typescript
export async function isPortInUse(port: number): Promise<boolean> {
  try {
    const options: RequestInit = {};
    if (process.platform !== 'win32') {
      options.signal = AbortSignal.timeout(2000);
    }
    const response = await fetch(`http://127.0.0.1:${port}/api/health`, options);
    return response.ok;
  } catch {
    return false;
  }
}
```

Or use a raw TCP socket probe instead of `fetch` for the port check.

## Environment

- **Bun**: 1.3.11
- **OS**: Linux 6.6.87.2-microsoft-standard-WSL2 (Ubuntu)
- **Plugin version**: 10.6.3
- **Node.js**: v22.19.0 (not used for worker, but installed)

octo-patch:
Fixes #1587

## Problem

On Bun 1.3.11 (Linux/WSL2), `fetch()` to a closed port hangs indefinitely instead of rejecting with `ECONNREFUSED`. The `AbortSignal.timeout` was previously removed from `isPortInUse()` to avoid a Windows Bun libuv assertion crash, but this caused the daemon startup to block forever on Linux when no worker was running.

## Solution

Add a platform guard: apply a 2-second `AbortSignal.timeout` on Linux/macOS only, and keep the bare `fetch()` on Windows where the timeout causes a libuv assertion failure. This restores the timeout for the platform that needs it without regressing Windows.

```typescript
const options: RequestInit = process.platform !== 'win32'
  ? { signal: AbortSignal.timeout(2000) }
  : {};
const response = await fetch(`http://127.0.0.1:${port}/api/health`, options);
```

## Testing

Verified that on Linux (WSL2), `fetch()` without a timeout to a port with nothing listening hangs indefinitely, while `fetch()` with `AbortSignal.timeout(2000)` correctly throws a timeout error after 2 seconds, allowing `isPortInUse()` to return `false` and daemon startup to proceed normally.

### DIFF
--- a/src/services/infrastructure/HealthMonitor.ts
+++ b/src/services/infrastructure/HealthMonitor.ts
@@ -39,10 +39,15 @@ async function httpRequestToWorker(
  */
 export async function isPortInUse(port: number): Promise<boolean> {
   try {
-    // Note: Removed AbortSignal.timeout to avoid Windows Bun cleanup issue (libuv assertion)
-    const response = await fetch(`http://127.0.0.1:${port}/api/health`);
+    // AbortSignal.timeout is skipped on Windows to avoid a Bun libuv assertion crash.
+    // On Linux/macOS, fetch() to a closed port may hang forever (Bun 1.3.11+), so
+    // we apply a 2s timeout to prevent daemon startup from blocking indefinitely.
+    const options: RequestInit = process.platform !== 'win32'
+      ? { signal: AbortSignal.timeout(2000) }
+      : {};
+    const response = await fetch(`http://127.0.0.1:${port}/api/health`, options);
     return response.ok;
-  } catch (error) {
+  } catch {
     // [ANTI-PATTERN IGNORED]: Health check polls every 500ms, logging would flood
     return false;
   }


thedotmack:

PR #1594, opened 2026-04-04
fix: wrap PowerShell ArgumentList paths in double-quotes to handle spaces

### LINKED ISSUE #1538: Windows: Worker fails to start when user profile path contains spaces
## Bug                                                                                                                                                     
                                                                                                                                                             
  The worker service fails to start on Windows when the user profile directory contains spaces (e.g., `C:\Users\Accountable Partner`).

**Error in `~/.claude-mem/logs/worker-*.log.err`:**
error: Module not found "C:\Users\Accountable"

  ## Root Cause

  In `worker-cli.js`, the `startWithBun` method builds a PowerShell command with:

  Start-Process -FilePath '...\bun.exe' -ArgumentList 'C:\Users\Accountable Partner\...\worker-wrapper.cjs' ...
  PowerShell's `-ArgumentList` splits the single-quoted string on spaces, so bun receives `C:\Users\Accountable` as the module path instead of the full path.

  ## Fix

  Wrap the argument in embedded double quotes:

      - -ArgumentList '${a}'
      + -ArgumentList '""${a}""'

  This also affects the `security_reminder_hook.py` hook — its `${CLAUDE_PLUGIN_ROOT}` expansion also breaks on paths with spaces.

  ## Environment
  - Windows 11
  - claude-mem v10.4.1
  - User profile path: `C:\Users\Accountable Partner`
  - Related issues: #363, #367, #371, #373

octo-patch:
Fixes #1538

## Problem

On Windows, when the user profile directory contains spaces (e.g., `C:\Users\Accountable Partner`), PowerShell's `-ArgumentList` splits the single-quoted path string on spaces, causing `bun.exe` to receive `C:\Users\Accountable` as the module path instead of the full path. This results in `error: Module not found "C:\Users\Accountable"`.

## Solution

Wrap the script path argument in embedded double-quotes inside the PowerShell single-quoted string:

```js
// Before
-ArgumentList '${a}' -WorkingDirectory '${c}'

// After  
-ArgumentList '""${a}""' -WorkingDirectory '""${c}""'
```

The `""` syntax embeds literal double-quote characters within a PowerShell single-quoted string. PowerShell then passes the path as a single quoted argument to `bun.exe`, correctly handling spaces in directory names.

The same fix is applied to `-WorkingDirectory` for consistency, as it faces the same issue with space-containing paths.

## Testing

Verified the fix by checking that paths containing spaces (e.g., `C:\Users\Accountable Partner\...`) are properly quoted in the PowerShell command, resolving the `Module not found` error reported in the issue.


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1881, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1599, opened 2026-04-04
fix: export PATH in hooks so node resolves on nvm systems

### LINKED ISSUE #1598: hooks.json missing `export PATH` causes `node: command not found` on nvm setups
### Environment
- macOS 14 (Sonoma)
- Node.js v22.13.1 installed via **nvm** (not system Node)
- Claude Code with claude-mem v10.6.3 installed via plugin marketplace

### Symptom
Every session start, compact, and stop shows:

```
Ran a hook on session startup
/bin/sh: node: command not found
Ran a hook on session startup
/bin/sh: node: command not found
Ran a hook on session startup
/bin/sh: node: command not found
Ran a hook on session compact
/bin/sh: node: command not found
```

### Root Cause

The **marketplace version** of `hooks/hooks.json` (at `~/.claude/plugins/marketplaces/thedotmack/plugin/hooks/hooks.json`) calls `node` bare without exporting PATH:

```json
"command": "_R=\"${CLAUDE_PLUGIN_ROOT}\"; [...]; node \"$_R/scripts/smart-install.js\""
```

Claude Code executes hooks via `/bin/sh` which does **not** inherit the user's shell PATH. When Node is installed via nvm, it lives at a non-standard path (e.g. `~/.nvm/versions/node/v22.13.1/bin/`) that `/bin/sh` doesn't know about.

Interestingly, the **cached version** of `hooks.json` (at `~/.claude/plugins/cache/thedotmack/claude-mem/10.6.3/hooks/hooks.json`) already has the fix — it prepends `export PATH="..."` to every command. But the **marketplace version** used at runtime does not.

### Fix

Add `export PATH` at the start of every hook command in the marketplace `hooks.json`, resolving the node binary dynamically, e.g.:

```sh
export PATH="$HOME/.nvm/versions/node/$(ls $HOME/.nvm/versions/node 2>/dev/null | sort -V | tail -1)/bin:/usr/local/bin:/usr/bin:/bin:$PATH"
```

Or simply use the absolute path to `node` instead of relying on PATH.

### Workaround

Manually add `export PATH="/path/to/nvm/node/bin:$PATH";` to each command in `~/.claude/plugins/marketplaces/thedotmack/plugin/hooks/hooks.json`. Note: this file gets overwritten on `plugin update`.

frank-ih:
Closes #1598

## 🔴 Failing scenario (the bug)

On any system where Node is installed via **nvm** (not in a system path), every claude-mem hook fails silently:

```
Ran a hook on session startup
/bin/sh: node: command not found

Ran a hook on session startup
/bin/sh: node: command not found

Ran a hook on session compact
/bin/sh: node: command not found
```

**Why:** Claude Code executes hooks via `/bin/sh`, which launches a minimal non-interactive shell. It does **not** source `.bashrc`/`.zshrc` and does **not** inherit the user's `$PATH`. When nvm is used, node lives at `~/.nvm/versions/node/<version>/bin/` — a path that `/bin/sh` knows nothing about.

## 🟢 Expected behavior (after this fix)

All hooks resolve `node` correctly regardless of whether it was installed via nvm, Homebrew (`/opt/homebrew/bin`), or a system package (`/usr/local/bin`). No errors on session startup, compact, or stop.

## ✅ The fix

Each hook command now opens with a dynamic `export PATH` that:

1. Detects whether `~/.nvm/versions/node` exists
2. If yes, resolves the latest installed version via `sort -V | tail -1` and prepends its `bin/` to PATH
3. Also prepends `/opt/homebrew/bin` and `/usr/local/bin` for Homebrew and system installs
4. Falls back cleanly on systems where node is already on PATH (no-op)

```sh
export PATH="$([ -d "$HOME/.nvm/versions/node" ] && echo "$HOME/.nvm/versions/node/$(ls "$HOME/.nvm/versions/node" | sort -V | tail -1)/bin:" || true)/usr/local/bin:/opt/homebrew/bin:/usr/bin:/bin:$PATH"
```

This pattern was already present in the **cached** version of `hooks.json` generated post-install, but was missing from the **marketplace** version that Claude Code actually reads at runtime.

## 🔁 Files changed

- `plugin/hooks/hooks.json` — added `export PATH` prefix to all 6 hook commands (SessionStart ×3, UserPromptSubmit, PostToolUse, Stop, SessionEnd)

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -19,17 +19,17 @@
         "hooks": [
           {
             "type": "command",
-            "command": "export PATH=\"$([ -d \"$HOME/.nvm/versions/node\" ] && echo \"$HOME/.nvm/versions/node/$(ls \\\"$HOME/.nvm/versions/node\\\" | sort -V | tail -1)/bin:\" || true)/usr/local/bin:/opt/homebrew/bin:/usr/bin:/bin:$PATH\"; _R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; node \"$_R/scripts/smart-install.js\"",
+            "command": "command -v node >/dev/null 2>&1 || export PATH=\"$([ -d \"$HOME/.nvm/versions/node\" ] && echo \"$HOME/.nvm/versions/node/$(ls \\\"$HOME/.nvm/versions/node\\\" | sort -V | tail -1)/bin:\" || true)/usr/local/bin:/opt/homebrew/bin:/usr/bin:/bin:$PATH\"; _R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; node \"$_R/scripts/smart-install.js\"",
             "timeout": 300
           },
           {
             "type": "command",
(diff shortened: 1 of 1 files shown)


thedotmack:

PR #1604, opened 2026-04-05
fix: salvage summary data when Claude returns observation tags instead of summary tags

### LINKED ISSUE #1546: Summary generation fails ~72% of the time: parser discards data + complete/summarize race
## Summary

Two related bugs cause summary generation to fail for approximately 72% of summarize requests. Observations are processed correctly, but summaries are lost.

## Bug 1: parseSummary discards entire response when Claude returns `<observation>` instead of `<summary>`

**Frequency**: 75 out of 104 summarize attempts in a single day (72% failure rate)

When the summarize prompt is sent to Claude, Claude sometimes returns `<observation>` tags instead of `<summary>` tags. The `parseSummary` function detects this condition but returns `null`, discarding the entire response.

**Evidence** (from claude-mem logs):
```
[WARN] [PARSER] Summary response contained <observation> tags instead of <summary> — prompt conditioning may need strengthening
```

This warning appears 75 times in a single day's logs, while `hasSummary=true` appears only 29 times.

**Current behavior** (`parser.ts`, `parseSummary`):
```typescript
if (!summaryMatch) {
  if (/<observation>/.test(text)) {
    logger.warn('PARSER', 'Summary response contained <observation> tags...');
  }
  return null;  // Entire response discarded
}
```

**Suggested fix**: When `<summary>` is not found but `<observation>` is present in a summarize response, extract observation fields (title, narrative) to construct a minimal summary rather than discarding the data entirely.

## Bug 2: Complete API can kill Claude process during summarize (possible race condition)

**Frequency**: Observed in logs but exact count uncertain — not all `Generator aborted` events are summarize-related.

When `/api/sessions/complete` is called while a summarize request is being processed by the SDK, `deleteSession()` calls `abortController.abort()`, which sends SIGTERM to the Claude process (exit code 143). The in-progress summary is never saved.

**Evidence** (from claude-mem logs):
```
[INFO] [SESSION] Session completed via API {contentSessionId=..., sessionDbId=26721}
[INFO] [SESSION] [session-26721] Generator aborted
```

The complete API is typically called by the plugin's `agent_end` / `session_end` handlers with a 5-second delay (`completionDelayMs`), but summarize processing can take longer than 5 seconds (Claude API response time + parsing + DB write).

**Suggested fix**: Before calling `deleteSession()`, check if there's a pending summarize message in the session queue. If so, wait for it to complete (with timeout) before aborting.

## Environment

- claude-mem version: 10.6.3
- Provider: Claude SDK (CLI auth)
- Mode: code--ja
- Platform: Linux (WSL2)
- Integration: OpenClaw plugin

## Reproduction

1. Use claude-mem with multiple tool calls per session
2. End sessions (triggering summarize)
3. Check logs: `grep "observation.*tags instead of.*summary" ~/.claude-mem/logs/claude-mem-*.log | wc -l`
4. Compare with: `grep "hasSummary=true" ~/.claude-mem/logs/claude-mem-*.log | wc -l`

octo-patch:
Fixes #1546

## Problem

When Claude responds to a summarize prompt with `<observation>` tags instead of `<summary>` tags (occurring ~72% of the time per the issue report), `parseSummary` logs a warning and returns `null`, discarding all the data Claude generated.

## Solution

Instead of immediately returning `null` when `<summary>` is absent but `<observation>` is present, the parser now attempts to salvage useful data from the first observation block:

- Maps `<title>` → `completed`
- Maps `<narrative>` → `learned` (falls back to joining `<fact>` elements if narrative is absent)
- Returns a minimal `ParsedSummary` with the salvaged fields, leaving unresolvable fields (`request`, `investigated`, `next_steps`) as `null`

If neither title nor learned data can be extracted, the function still returns `null` as before.

## Testing

Added `tests/parser.test.ts` with 6 test cases covering:
- Normal summary parsing
- `skip_summary` tag handling
- Salvage path: observation with title + narrative → minimal summary
- Salvage path: observation with title + facts (no narrative) → joined facts as `learned`
- No-op when observation tags contain no useful data
- Returns `null` when neither summary nor observation tags present

All 6 tests pass. Pre-existing test suite unaffected.

### DIFF
--- a/src/sdk/parser.ts
+++ b/src/sdk/parser.ts
@@ -120,10 +120,31 @@ export function parseSummary(text: string, sessionId?: number): ParsedSummary | 
   const summaryMatch = summaryRegex.exec(text);
 
   if (!summaryMatch) {
-    // Log when the response contains <observation> instead of <summary>
-    // to help diagnose prompt conditioning issues (see #1312)
+    // When response contains <observation> instead of <summary>, attempt to salvage
+    // summary data from the observation fields rather than discarding it (see #1546)
     if (/<observation>/.test(text)) {
-      logger.warn('PARSER', 'Summary response contained <observation> tags instead of <summary> — prompt conditioning may need strengthening', { sessionId });
+      logger.warn('PARSER', 'Summary response contained <observation> tags instead of <summary> — attempting to salvage summary data', { sessionId });
+
+      const obsContentMatch = /<observation>([\s\S]*?)<\/observation>/.exec(text);
+      if (obsContentMatch) {
+        const obsContent = obsContentMatch[1];
+        const title = extractField(obsContent, 'title');
+        const narrative = extractField(obsContent, 'narrative');
+        const facts = extractArrayElements(obsContent, 'facts', 'fact');
+        const learned = narrative || (facts.length > 0 ? facts.join('; ') : null);
+
+        if (title || learned) {
+          logger.info('PARSER', 'Salvaged summary from observation tags', { sessionId, hasTitle: !!title, hasLearned: !!learned });
+          return {
+            request: null,
+            investigated: null,
+            learned,
+            completed: title,
+            next_steps: null,
+            notes: null
+          };
+        }
+      }
     }
     return null;
   }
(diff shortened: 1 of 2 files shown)


anh-chu:
this is essentially breaking claude-mem, rendering it useless!

thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1908, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

Issue #1450, opened 2026-03-22
[Windows] Worker silently exits on Bun — isMainModule check fails due to import.meta.url mismatch and missing .cjs in endsWith


alphabet-h:
## Bug Description

On Windows + Bun, the worker service (`worker-service.cjs`) silently exits with code 0 because the `isMainModule` guard always evaluates to `false`. This causes all MCP tools (search, timeline, etc.) to fail with "Worker not available".

## Environment

- **OS**: Windows 11 Pro (10.0.26200)
- **Bun**: 1.3.11
- **Node.js**: v22.14.0
- **claude-mem**: v10.6.1 (also confirmed unpatched in v10.6.2 release notes)
- **Shell**: Git Bash / PowerShell

## Root Cause

The `isMainModule` check on line 69078 of bundled `worker-service.cjs`:

```js
var isMainModule = typeof module_worker_service !== "undefined"
  ? __require.main == __require.module || !module_worker_service.parent
  : import.meta.url === `file://${process.argv[1]}`
    || process.argv[1]?.endsWith("worker-service");
```

Both ESM-path conditions fail on Windows + Bun:

| Check | Expected | Actual (Bun + Windows) | Why |
|---|---|---|---|
| ``import.meta.url === `file://${argv[1]}` `` | `true` | `false` | Bun produces `file:///C:/...` (3 slashes, forward) vs template `file://C:\...` (2 slashes, backslash) |
| `argv[1]?.endsWith("worker-service")` | `true` | `false` | Actual filename is `worker-service.cjs` — the `.cjs` extension is not matched |

Since `module_worker_service` is never defined (only 1 occurrence, in the check itself), the CJS branch is never taken either.

Result: `main()` is never called → process exits immediately with code 0.

## Reproduction Steps

1. Install claude-mem on Windows with Bun in PATH
2. Start a Claude Code session — worker fails to start
3. Logs show repeated:
   ```
   [ERROR] Worker not available {}
   [ERROR] Tools will fail until Worker is started
   ```
4. Running directly confirms:
   ```bash
   bun worker-service.cjs --daemon
   # exits immediately with code 0, no output
   ```

## Workaround

Bypassing `isMainModule` by importing `WorkerService` directly works:

```js
import { WorkerService } from './worker-service.cjs';
const worker = new WorkerService();
worker.start();
```

## Suggested Fix

```js
// Option 1: Add .cjs extension to endsWith check
[...]

PR #1610, opened 2026-04-05
fix(windows): isMainModule CJS branch fails on Bun — add CLAUDE_MEM_MANAGED fallback

alessandropcostabr:
## Summary

- On Bun/Windows, `require.main !== module` in CJS mode causes the worker to exit silently with code 0, preventing the worker daemon from starting
- The wrapper already sets `CLAUDE_MEM_MANAGED=true` when spawning the inner worker — this PR adds it as a fallback check in the CJS branch of `isMainModule`
- 1-line change in `src/services/worker-service.ts`, does not affect standalone or ESM execution

## Root Cause

PR #1518 fixed the ESM path checks for `isMainModule` but the CJS branch was left unchanged:

```typescript
// CJS branch — Bun fails here:
require.main === module || !module.parent
```

In Bun's CJS runtime on Windows, `require.main !== module` and `module.parent` is truthy, so `isMainModule` evaluates to `false`. The worker exits with code 0 — no error, no log.

## Evidence

```
[wrapper] Spawning inner worker: ...\worker-service.cjs
[wrapper] Inner exited with code=0, signal=null
[wrapper] Inner exited unexpectedly, wrapper exiting
```

With the fix:
```bash
CLAUDE_MEM_MANAGED=true bun worker-service.cjs  # stays alive ✓
```

## Test plan

- [x] Verified worker starts and stays alive on Windows 11 + Bun 1.3.11
- [x] Verified `CLAUDE_MEM_MANAGED` is already set by `worker-wrapper.cjs` spawn env
- [x] Verified standalone execution (`bun worker-service.cjs` without env var) still uses the original `require.main === module` check

Ref #1450

### DIFF
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -1307,8 +1307,10 @@ async function main() {
 }
 
 // Check if running as main module in both ESM and CommonJS
+// The CLAUDE_MEM_MANAGED check handles Bun on Windows where require.main !== module
+// in CJS mode despite being the entry point (see #1450)
 const isMainModule = typeof require !== 'undefined' && typeof module !== 'undefined'
-  ? require.main === module || !module.parent
+  ? require.main === module || !module.parent || process.env.CLAUDE_MEM_MANAGED === 'true'
   : import.meta.url === `file://${process.argv[1]}`
     || process.argv[1]?.endsWith('worker-service')
     || process.argv[1]?.endsWith('worker-service.cjs')


alessandropcostabr:
## Self-review: Verified locally

Tested on Bun 1.3.11 / Linux x86_64:
- Confirmed the `isMainModule` detection fails when `worker-wrapper.cjs` spawns the inner worker — `require.main !== module` in CJS context
- `CLAUDE_MEM_MANAGED` is already set by the wrapper but never consumed at the detection point
- With this fix, the worker starts correctly when managed

We debugged this exact crash path while investigating 48h of lost observations caused by a separate issue (#1553). This PR would have prevented the silent exit.

**No conflicts, no side effects.** Ready for maintainer review.

PR #1612, opened 2026-04-05
fix: resolve node not found on nvm/homebrew installations

### LINKED ISSUE #1598: hooks.json missing `export PATH` causes `node: command not found` on nvm setups

Cyvid7-Darus10:
## Summary

Fixes #1598

Claude Code executes hooks via `/bin/sh`, which does not inherit the user's interactive shell PATH. When Node.js is installed through nvm, the `node` binary lives under `~/.nvm/versions/node/<version>/bin/` — a path that `/bin/sh` knows nothing about. This causes every hook command to fail with `node: command not found`.

This PR prepends a dynamic PATH export to every hook command in `plugin/hooks/hooks.json` that calls `node`:

```sh
export PATH="$HOME/.nvm/versions/node/$(ls "$HOME/.nvm/versions/node" 2>/dev/null | sort -V | tail -1)/bin:$HOME/.local/bin:/usr/local/bin:/opt/homebrew/bin:$PATH"
```

**What it covers:**
- **nvm** — dynamically resolves the latest installed Node.js version
- **Homebrew** — includes `/opt/homebrew/bin` (Apple Silicon) and `/usr/local/bin` (Intel)
- **Other managers** — includes `$HOME/.local/bin` for common user-local installs

**Graceful degradation:** If nvm is not installed, `ls` outputs nothing, the path segment becomes empty, and the remaining entries in PATH still work. No errors, no breakage.

The Setup hook (which calls a `.sh` script, not `node` directly) and `cursor-hooks/hooks.json` (which uses shell scripts throughout) are left unchanged.

## Test plan

- [x] Install Node.js via nvm, open a fresh `/bin/sh` shell, and verify `node` is not on PATH
- [x] Run claude-mem hooks and confirm they now find `node` correctly
- [x] Verify on a system without nvm (e.g. Homebrew-only or system Node.js) that hooks still work
- [x] Validate `plugin/hooks/hooks.json` parses as valid JSON

## Diff

diff

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -19,17 +19,17 @@
         "hooks": [
           {
             "type": "command",
-            "command": "_R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; node \"$_R/scripts/smart-install.js\"",
+            "command": "export PATH=\"$HOME/.nvm/versions/node/$(ls \\\"$HOME/.nvm/versions/node\\\" 2>/dev/null | sort -V | tail -1)/bin:$HOME/.local/bin:/usr/local/bin:/opt/homebrew/bin:$PATH\"; _R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; node \"$_R/scripts/smart-install.js\"",
             "timeout": 300
           },
           {
             "type": "command",
-            "command": "_R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; node \"$_R/scripts/bun-runner.js\" \"$_R/scripts/worker-service.cjs\" start",
+            "command": "export PATH=\"$HOME/.nvm/versions/node/$(ls \\\"$HOME/.nvm/versions/node\\\" 2>/dev/null | sort -V | tail -1)/bin:$HOME/.local/bin:/usr/local/bin:/opt/homebrew/bin:$PATH\"; _R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; node \"$_R/scripts/bun-runner.js\" \"$_R/scripts/worker-service.cjs\" start",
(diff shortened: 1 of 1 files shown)


Cyvid7-Darus10:
Good catch on the `sort -V` — pushed a fix. Now using `sed 's/^v//' | sort -t. -k1,1n -k2,2n -k3,3n` which works on both GNU and BSD sort. Had to strip the `v` prefix first since nvm directories are named `v20.19.6` and the `v` breaks numeric sorting.

On the deduplication point — I considered extracting to a helper script but decided against it for this PR. The hooks.json format requires inline commands (Claude Code doesn't support sourcing external scripts in hook commands), and a separate bootstrap script would add another file that could get out of sync. Happy to revisit if there's a pattern the project prefers though.

PR #1618, opened 2026-04-06
fix: add 400/401 to FALLBACK_ERROR_PATTERNS to prevent retry loop on auth errors

### LINKED ISSUE #1614: Gemini 400 API_KEY_INVALID does not trigger fallback, causes slow 2-min retry loop on stop hook
## Bug

When the Gemini API key is invalid, the stop hook takes ~2 minutes to complete instead of failing fast or falling back to Claude.

## Root Cause

`FALLBACK_ERROR_PATTERNS` in `src/services/worker/agents/types.ts` only covers transient/network errors:

```
'429', '500', '502', '503', 'ECONNREFUSED', 'ETIMEDOUT', 'fetch failed'
```

A `400 API_KEY_INVALID` response is not in this list, so `shouldFallbackToClaude()` returns `false` and the fallback agent is never invoked. Instead, the session crash-recovery loop retries 3 times (maxRestarts=3) before giving up with `CRITICAL: Generator restart limit exceeded`.

## Observed timeline (from logs)

```
19:23:18 Stop hook fired, requested summary from Gemini
19:23:18 Gemini returns 400 API_KEY_INVALID
19:23:20 Crash-recovery restart 1 → same 400 error
19:23:26 Crash-recovery restart 2 → same 400 error
19:23:32 Crash-recovery restart 3 → same 400 error
19:23:32 CRITICAL: Generator restart limit exceeded, messages left pending
19:25:08 Stop hook finally completes (~1m50s total)
```

## Expected Behaviour

- Auth errors (`400`, `401`) should trigger the Claude SDK fallback immediately (or at minimum skip retries), not burn through 3 crash-recovery cycles.
- Ideally `400` and `401` should be added to `FALLBACK_ERROR_PATTERNS`, or a separate "no-retry" pattern list should abort the retry loop instantly for permanent failures.

## Suggested Fix

Add `'400'` and `'401'` to `FALLBACK_ERROR_PATTERNS` in `src/services/worker/agents/types.ts`, or introduce a `PERMANENT_ERROR_PATTERNS` list that bypasses the crash-recovery loop entirely.

octo-patch:
Fixes #1614

## Problem

When Gemini returns `400 API_KEY_INVALID`, the `shouldFallbackToClaude()` function returns `false` because `'400'` was not in `FALLBACK_ERROR_PATTERNS`. This causes the crash-recovery loop to retry 3 times before giving up, resulting in a ~2-minute delay in the stop hook.

## Solution

Added `'400'` and `'401'` to `FALLBACK_ERROR_PATTERNS` in `src/services/worker/agents/types.ts`.

Auth errors are **permanent failures** — retrying with the same credentials will always fail. Triggering the Claude SDK fallback immediately (same as rate-limit/network errors) avoids the 3-restart retry loop.

## Testing

- Verified `shouldFallbackToClaude()` uses `FALLBACK_ERROR_PATTERNS.some(pattern => message.includes(pattern))`, so adding `'400'`/`'401'` will match error messages containing those status codes.
- Existing behavior for transient errors (429, 5xx, network) is unchanged.

### DIFF
--- a/src/services/worker/agents/FallbackErrorHandler.ts
+++ b/src/services/worker/agents/FallbackErrorHandler.ts
@@ -13,6 +13,8 @@ import { logger } from '../../../utils/logger.js';
  * Check if an error should trigger fallback to Claude SDK
  *
  * Errors that trigger fallback:
+ * - 400: Bad request / API_KEY_INVALID (permanent failure, skip retries)
+ * - 401: Unauthorized (invalid credentials, skip retries)
  * - 429: Rate limit exceeded
  * - 500/502/503: Server errors
  * - ECONNREFUSED: Connection refused (server down)
--- a/src/services/worker/agents/types.ts
+++ b/src/services/worker/agents/types.ts
@@ -123,6 +123,8 @@ export interface BaseAgentConfig {
  * Error codes that should trigger fallback to Claude
  */
 export const FALLBACK_ERROR_PATTERNS = [
+  '400',           // Bad request (e.g. API_KEY_INVALID — permanent failure, no point retrying)
+  '401',           // Unauthorized — invalid or missing credentials
   '429',           // Rate limit
   '500',           // Internal server error
   '502',           // Bad gateway


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1872, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1621, opened 2026-04-06
fix: resolve bun.exe when npm shim is in PATH

zerone0x:
## Summary
- avoid returning "bun" on Windows when "where bun" only finds bun.cmd
- resolve bun.exe from the shim location or npm global install dir
- keep bun resolver in sync for the NPX CLI

## Testing
- not run (path resolution change only)


thedotmack:

Issue #439, opened 2025-12-25
Post tool use save hook occasionally stalls, blocking Claude Code execution


asermax:
# Post tool use save hook occasionally stalls, blocking Claude Code execution

## Problem Statement

The post tool use save hook intermittently stalls during execution, causing Claude Code to wait on it after tool usage. This appears to occur randomly and may be related to multiple tool calls being processed in rapid succession, particularly when whitelisted commands are used.

## Environment

- **Claude-mem**: 8.0.6
- **Claude Code**: 2.0.76
- **Node.js**: v25.2.1
- **Bun**: 1.3.5
- **OS**: Linux 6.18.2-arch2-1 (x64)
- **Platform**: linux

## Steps to Reproduce

No consistent reproduction steps available. The issue occurs randomly, possibly under the following conditions:
- Multiple tool calls sent simultaneously
- Fast-succession processing of tools
- Usage of whitelisted commands

## Expected Behavior

The post tool use save hook should execute asynchronously without blocking operations. It should return immediately to allow Claude Code to continue without interruptions.

## Actual Behavior

The hook occasionally stalls during execution, causing Claude Code to stop responding after a tool is used. The stall is unpredictable and has been observed to occur once or twice randomly.

### Silent Debug Log showing "Missing cwd" errors

```
[2025-12-12T19:28:36.951Z] [HAPPY-PATH-ERROR] [worker-service.cjs:1035] Missing cwd when queueing observation in SessionRoutes {"sessionDbId":1872,"tool_name":"Grep"}
[2025-12-12T19:28:43.513Z] [HAPPY-PATH-ERROR] [unknown] Missing cwd in PostToolUse hook input {"session_id":"0b2aa5d7-d45e-491a-b01c-66cdef34157e","tool_name":"Read"}
[2025-12-12T19:28:43.521Z] [HAPPY-PATH-ERROR] [worker-service.cjs:1035] Missing cwd when queueing observation in SessionRoutes {"sessionDbId":1872,"tool_name":"Read"}
[2025-12-12T19:28:44.002Z] [HAPPY-PATH-ERROR] [unknown] Missing cwd in PostToolUse hook input {"session_id":"0b2aa5d7-d45e-491a-b01c-66cdef34157e","tool_name":"Grep"}
[...]

PR #1667, opened 2026-04-09
fix: clarify observation ID in context footer and index (#1339)

### LINKED ISSUE #1339: Web UI #ID numbers don't match MCP get_observations IDs — confusing 'access memories by ID' prompt
## Problem

The context index header tells the Claude user:

> Access 196k tokens of past research & decisions for just 19,285t. Use the claude-mem skill to access memories by ID.

But "by ID" is ambiguous — it doesn't specify *which* IDs it means. And the section of the claude-mem context index header that actually contains the searchable ID's is often not visible to the user in the terminal unless they scroll up past the summary paragraphs, so the notion of "memories by ID" is not obvious.
Users who look to the web UI (localhost:37777) for ID numbers will find plenty but they are an entirely different, yet overlapping, numbering scheme than what
`get_observations` expects.

## What Happens

1. User searches the web UI and finds a relevant PROMPT entry shown as `#439`
2. User asks Claude to "fetch #439"
3. Claude calls `get_observations(ids=[439])` — the only ID-based fetch tool available
4. A completely different record is returned (observation #439 from a different project/date)

## Root Cause

Three separate tables each have their own auto-incrementing primary key, but the web UI displays all of them with the same `#N` format:

| Table | Example ID | Web UI display | MCP fetch tool |
|-------|-----------|---------------|----------------|
| `observations` | `#864` | `#864` | `get_observations` |
| `user_prompts` | `#439` | `#439` | **none** |
| `session_summaries` | `#S172` | `#S172` | **none** |

The overlapping ID spaces mean `#439` in the web UI (a prompt) resolves to observation #439 (a different record entirely) when passed to the API.

### Quick fix (copy change)

In `context-generator.cjs`, the `Xe` function generates the header text:

```js
function Xe(r, e) {
  return ["", `Access ${Math.round(r/1e3)}k tokens ... Use the claude-mem skill to access memories by ID.`]
}
```

Change to something like:
```
Use the claude-mem skill to access memories by ID listed above.
```

This clarifies that "by ID" refers to the observation IDs in the context index table, not IDs from other sources.

### Broader improvements (optional)

- **Expose `get_prompts` / `get_sessions` MCP tools** — the database already has `getUserPromptsByIds()` and `getPromptsByIds()` methods, they're just not wired up as MCP tools
- **Disambiguate IDs in the web UI** — e.g. prefix prompts as `#P439` vs observations as `#O864`, so the type is always clear regardless of where the user encounters the ID

## Environment

- claude-mem v10.5.2
- macOS (Darwin 25.2.0)

ousamabenyounes:
## Summary

Fixes #1339

- Changed "access memories by ID" → "access memories by **observation ID** (the #N numbers in the index above)" in both `ColorFormatter` and `MarkdownFormatter` footer text
- Updated the context-index bullet from `get_observations([IDs])` → `get_observations([observation IDs])` for consistency
- Added `tests/context/formatters/color-formatter.test.ts` (was missing entirely) and a regression test in the existing markdown formatter test

The old text was ambiguous: the web UI displays IDs from three tables (`observations`, `user_prompts`, `session_summaries`), all with the same `#N` format. Users would look up `#439` in the web UI (a prompt record), pass it to `get_observations(ids=[439])`, and receive observation #439 — a completely different record.

## Verification

- [x] Post-fix tests: 1110 pass, 0 regressions, +4 new tests
- [x] New tests: 4 added (ColorFormatter test file + regression test in both formatters)
- [x] Review agent: issue alignment verified

## Files changed

| File | Change |
|------|--------|
| `src/services/context/formatters/MarkdownFormatter.ts` | Clarify "ID" → "observation ID" in footer and context-index bullet |
| `src/services/context/formatters/ColorFormatter.ts` | Same copy change |
| `tests/context/formatters/markdown-formatter.test.ts` | Add `observation ID` regression test |
| `tests/context/formatters/color-formatter.test.ts` | New file — `renderColorFooter` tests |

Generated by Claude Code
Vibe coded by ousamabenyounes

### DIFF
--- a/src/services/context/formatters/ColorFormatter.ts
+++ b/src/services/context/formatters/ColorFormatter.ts
@@ -74,7 +74,7 @@ export function renderColorContextIndex(): string[] {
     `${colors.dim}Context Index: This semantic index (titles, types, files, tokens) is usually sufficient to understand past work.${colors.reset}`,
     '',
     `${colors.dim}When you need implementation details, rationale, or debugging context:${colors.reset}`,
-    `${colors.dim}  - Fetch by ID: get_observations([IDs]) for observations visible in this index${colors.reset}`,
+    `${colors.dim}  - Fetch by ID: get_observations([observation IDs]) for observations visible in this index${colors.reset}`,
     `${colors.dim}  - Search history: Use the mem-search skill for past decisions, bugs, and deeper research${colors.reset}`,
     `${colors.dim}  - Trust this index over re-reading code for past decisions and learnings${colors.reset}`,
     ''
@@ -226,7 +226,7 @@ export function renderColorFooter(totalDiscoveryTokens: number, totalReadTokens:
   const workTokensK = Math.round(totalDiscoveryTokens / 1000);
   return [
     '',
-    `${colors.dim}Access ${workTokensK}k tokens of past research & decisions for just ${totalReadTokens.toLocaleString()}t. Use the claude-mem skill to access memories by ID.${colors.reset}`
+    `${colors.dim}Access ${workTokensK}k tokens of past research & decisions for just ${totalReadTokens.toLocaleString()}t. Use the claude-mem skill to access memories by observation ID (the #N numbers in the index above).${colors.reset}`
   ];
 }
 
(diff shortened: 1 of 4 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1920, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1668, opened 2026-04-09
fix: detect <observation>-in-summary pattern and strengthen summary prompt (#1649)

### LINKED ISSUE #1649: Stop hook: summaries never persist — parser always rejects LLM output as `<observation>` instead of `<summary>` (v12.0.1)
## Summary

On v12.0.1, the `Stop` hook fires end-to-end — worker receives the request, enqueues `summarize`, and the LLM responds — but **no summary is ever stored**. Every Stop-triggered run ends with `summaryId=none` because the parser rejects the LLM output: the model emits `<observation>` tags instead of `<summary>` tags on the summarize turn.

The code itself flags this in a warning:

```
[WARN] [PARSER] Summary response contained <observation> tags instead of <summary> — prompt conditioning may need strengthening
```

…and that warning fires on **every** Stop hook invocation in my logs.

## Environment

- claude-mem `12.0.1`
- Platform: Windows 10 / Claude Code (git-bash shell)
- Installed at `~/.claude/plugins/cache/thedotmack/claude-mem/12.0.1`

## Evidence from `~/.claude-mem/logs/claude-mem-2026-04-08.log`

```
[14:48:03.044] [INFO] [HOOK ] → Stop: Requesting summary {hasLastAssistantMessage=true}
[14:48:03.046] [INFO] [QUEUE] ENQUEUED  | messageId=21 | type=summarize
[14:48:03.047] [INFO] [QUEUE] CLAIMED   | messageId=21 | type=summarize
[14:48:12.587] [WARN] [PARSER] Summary response contained <observation> tags instead of <summary> — prompt conditioning may need strengthening
[14:48:12.588] [INFO] [DB   ] STORED | obsCount=1 | obsIds=[9] | summaryId=none
[14:48:13.866] [INFO] [HOOK ] Session completed in Stop hook
```

This pattern repeats for every Stop event — the observation gets stored, the summary is silently dropped.

## Root cause

`scripts/worker-service.cjs` function `_u` (summarize prompt builder, ~line 1359) and parser `K6` (~line 1438).

The summarize prompt *does* already tell the model "Do NOT output `<observation>` tags… Your response MUST use `<summary>` tags ONLY." But because the conversation history for that session is full of prior observation-style turns (PostToolUse hook fires observation prompts dozens of times), the model exhibits strong pattern-completion bias and re-emits `<observation>` on the summarize turn. The prohibition at the *top* of the prompt gets drowned out.

The parser `K6` then correctly refuses to treat an `<observation>` as a `<summary>` and returns `null`, so nothing is persisted.

## Suggested fixes (either/both)

**1. Strengthen prompt conditioning in `_u`** — move the prohibition to the *end* of the prompt (LLMs have a strong last-instruction bias), add a concrete "begin your response with `<summary>` now" directive, and provide an explicit `<skip_summary reason="..."/>` escape hatch:

```diff
 return `--- MODE SWITCH: PROGRESS SUMMARY ---
-Do NOT output <observation> tags. This is a summary request, not an observation request.
-Your response MUST use <summary> tags ONLY. Any <observation> output will be discarded.
-
 ${e.prompts.header_summary_checkpoint}
 ${e.prompts.summary_instruction}
 ...
-${e.prompts.summary_footer}`;
+${e.prompts.summary_footer}
+
+=== CRITICAL OUTPUT CONSTRAINT ===
+This turn is a SUMMARY, not an observation. Prior turns used <observation> tags; this turn MUST NOT.
+- REQUIRED: wrap your entire response in a single <summary>...</summary> block with the sub-tags shown above.
+- FORBIDDEN: <observation> tags. Any response containing <observation> will be discarded and the session summary will be lost.
+- If there is nothing to summarize, respond with exactly: <skip_summary reason="no_content"/>
+Begin your response with the literal characters "<summary>" now.`;
```

**2. Parser fallback in `K6`** — when a summarize turn returns `<observation>` tags, rather than silently discarding, salvage the content (use the observation's `narrative`/`title` as `<summary><notes>`) so users aren't left with `summaryId=none` forever. At minimum, log this as `ERROR` rather than `WARN` and surface it to the user via the Stop hook response so it isn't invisible.

## Reproduction

1. Install claude-mem 12.0.1 in Claude Code.
2. Run a session with several tool uses (triggers observation turns).
3. End the turn so the `Stop` hook fires.
4. `grep "summaryId=none" ~/.claude-mem/logs/claude-mem-*.log` — every Stop event will match.

## Impact

Sessions accumulate observations but never get a summary, so cross-session memory recall loses the high-level "what this session was about" signal that the summarize pipeline is supposed to produce. From the user's perspective, the Stop hook silently does nothing useful.

Happy to test a patch against 12.0.1 if you'd like.

ousamabenyounes:
## Summary

Fixes #1649

- `parseSummary` now logs a `WARN` with `responsePrefix` when `<observation>` tags appear in a summary response without any `<summary>` block — matches the `[WARN] [PARSER] Summary response contained <observation> tags instead of <summary>` diagnostic from the issue. Previously silent.
- `buildSummaryPrompt` appends a `=== CRITICAL FORMAT CONSTRAINT ===` block at the **end** of the prompt (LLMs have strong last-instruction bias), explicitly forbidding `<observation>` tags and providing a `<skip_summary reason="no_content"/>` escape hatch for empty sessions.

## Verification

- [x] Post-fix tests: 1111 pass, 0 regressions, +5 new tests
- [x] New tests: 5 in `tests/utils/parser-summary.test.ts` covering regression case, happy path, and edge cases
- [x] Review agent: issue alignment verified

## Files changed

| File | Change |
|------|--------|
| `src/sdk/parser.ts` | Detect `<observation>` in summary response, warn with response prefix |
| `src/sdk/prompts.ts` | Append CRITICAL FORMAT CONSTRAINT block at end of summary prompt |
| `tests/utils/parser-summary.test.ts` | New — 5 parser tests for #1649 regression |

Generated by Claude Code
Vibe coded by ousamabenyounes

### DIFF
--- a/src/sdk/parser.ts
+++ b/src/sdk/parser.ts
@@ -125,7 +125,8 @@ export function parseSummary(text: string, sessionId?: number): ParsedSummary | 
     if (/<observation>/.test(text)) {
       logger.warn('PARSER', 'Summary response contained <observation> tags instead of <summary> — prompt conditioning may need strengthening', {
         sessionId,
-        responsePrefix: text.slice(0, 120)
+        responseStartsWithObservation: text.trimStart().startsWith('<observation>'),
+        responseLength: text.length
       });
     }
     return null;
--- a/src/sdk/parser.ts
+++ b/src/sdk/parser.ts
@@ -120,6 +120,14 @@ export function parseSummary(text: string, sessionId?: number): ParsedSummary | 
   const summaryMatch = summaryRegex.exec(text);
 
   if (!summaryMatch) {
+    // Detect <observation> tags in a summary response — LLM applied pattern-completion
+    // bias from prior observation turns and used the wrong XML schema (#1649)
+    if (/<observation>/.test(text)) {
+      logger.warn('PARSER', 'Summary response contained <observation> tags instead of <summary> — prompt conditioning may need strengthening', {
+        sessionId,
+        responsePrefix: text.slice(0, 120)
+      });
+    }
     return null;
   }
 
--- a/src/sdk/prompts.ts
+++ b/src/sdk/prompts.ts
@@ -146,7 +146,14 @@ ${mode.prompts.summary_format_instruction}
   <notes>${mode.prompts.xml_summary_notes_placeholder}</notes>
 </summary>
 
-${mode.prompts.summary_footer}`;
+${mode.prompts.summary_footer}
+
+=== CRITICAL FORMAT CONSTRAINT ===
+This turn is a SUMMARY, not an observation. Prior turns in this session used <observation> tags; this turn MUST NOT.
+- REQUIRED: wrap your entire response in a single <summary>...</summary> block using the sub-tags shown above.
+- FORBIDDEN: <observation> tags. Any response containing <observation> will be discarded and the session summary will be lost.
+- If there is nothing meaningful to summarize, respond with exactly: <skip_summary reason="no_content"/>
+Begin your response with the literal characters "<summary>" now.`;
 }
 
 /**
--- a/src/sdk/prompts.ts
+++ b/src/sdk/prompts.ts
@@ -150,10 +150,9 @@ ${mode.prompts.summary_footer}
 
 === CRITICAL FORMAT CONSTRAINT ===
 This turn is a SUMMARY, not an observation. Prior turns in this session used <observation> tags; this turn MUST NOT.
-- REQUIRED: wrap your entire response in a single <summary>...</summary> block using the sub-tags shown above.
+- REQUIRED: unless you are using <skip_summary reason="no_content"/>, wrap your entire response in a single <summary>...</summary> block using the sub-tags shown above.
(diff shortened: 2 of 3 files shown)


ousamabenyounes:
Addressed in the latest commit:
- `src/sdk/parser.ts`: replaced `responsePrefix: text.slice(0, 120)` with `responseStartsWithObservation` + `responseLength` to avoid logging raw model output
- `src/sdk/prompts.ts`: resolved the contradictory constraint — `<summary>` wrapping is now conditional on not using the `<skip_summary>` escape hatch; removed the hard `Begin your response with '<summary>'` directive
- `tests/utils/parser-summary.test.ts`: added regression test for the `<skip_summary reason="no_content"/>` path

thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1908, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1673, opened 2026-04-10
fix: split shared 'jsts' tree-sitter query into separate 'js' and 'ts' queries (fixes #1654)

### LINKED ISSUE #1654: bug: shared jsts tree-sitter query includes TypeScript-only node types, fragile for .js files
## Description

The `smart-explore` / `smart-file-read` parser uses a single `"jsts"` tree-sitter query for JavaScript, TypeScript, and TSX files. This query includes three TypeScript-only node types that don't exist in the `tree-sitter-javascript` grammar:

- `interface_declaration`
- `type_alias_declaration`
- `enum_declaration`

**Source:** `src/services/smart-file-read/parser.ts` lines 322–333 (query definition) and 490–495 (`getQueryKey()` mapping all three languages to `"jsts"`)

## Current Behavior

When parsing `.js` files, the full `jsts` query (including TS-only patterns) is compiled against `tree-sitter-javascript`. Tree-sitter currently **silently ignores** unmatched patterns, so this works in practice — JS files parse fine and the valid patterns (functions, classes, methods, imports, exports) match correctly.

## Why This Is a Problem

1. **Fragile**: Relies on tree-sitter silently ignoring invalid patterns. If tree-sitter changes to fail on unrecognized node types, all `.js` file parsing breaks.
2. **Silent failure**: The `runBatchQuery` error handler catches all exceptions and returns an empty `Map`, so a compilation failure would silently return zero results with no logging.
3. **Confusing**: The shared query name `"jsts"` doesn't document why TypeScript-only patterns are included for a grammar that can't match them.

## Proposed Fix

Split the query into separate `js` and `ts` entries:

1. Create a `"js"` query with only JavaScript-valid patterns (function, class, method, const-func, import, export)
2. Keep the current query as `"ts"` with all patterns including interface, type alias, and enum
3. Update `getQueryKey()` to map `"javascript"` → `"js"` and `"typescript"` / `"tsx"` → `"ts"`

## Originally Flagged By

CodeRabbit automated review on PR #1491 (final review pass, outside-diff comment on `mcp-server.cjs`).

octo-patch:
Fixes #1654

## Problem

The `QUERIES` object in `src/services/smart-file-read/parser.ts` defined a single `"jsts"` query used for JavaScript, TypeScript, and TSX files. This query included three TypeScript-only node types:

- `interface_declaration`
- `type_alias_declaration`
- `enum_declaration`

These node types do not exist in the `tree-sitter-javascript` grammar. While tree-sitter currently silently ignores unmatched patterns, relying on this behavior is fragile — a future tree-sitter version that strictly validates patterns would silently return zero results for all `.js` file parses (the error handler in `runBatchQuery` catches all exceptions and returns an empty `Map`).

## Solution

Split `"jsts"` into two separate queries:
- `"js"`: JavaScript-only patterns (functions, classes, methods, imports, exports — no TS-only types)
- `"ts"`: All patterns including TypeScript-specific `interface`, `type alias`, and `enum`

Updated `getQueryKey()` to map:
- `"javascript"` → `"js"`
- `"typescript"` / `"tsx"` → `"ts"`

## Testing

No behavior change for TypeScript/TSX files. JavaScript files now use only patterns that are valid for the `tree-sitter-javascript` grammar, eliminating the silent-ignore dependency.

### DIFF
--- a/plugin/hooks/hooks.json
+++ b/plugin/hooks/hooks.json
@@ -24,12 +24,12 @@
           },
           {
             "type": "command",
-            "command": "_R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; node \"$_R/scripts/bun-runner.js\" \"$_R/scripts/worker-service.cjs\" start",
+            "command": "_R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; node \"$_R/scripts/bun-runner.mjs\" \"$_R/scripts/worker-service.cjs\" start",
             "timeout": 60
           },
           {
             "type": "command",
-            "command": "_R=\"${CLAUDE_PLUGIN_ROOT}\"; [ -z \"$_R\" ] && _R=\"$HOME/.claude/plugins/marketplaces/thedotmack/plugin\"; node \"$_R/scripts/bun-runner.js\" \"$_R/scripts/worker-service.cjs\" hook claude-code context",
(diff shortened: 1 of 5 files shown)


thedotmack:

PR #1678, opened 2026-04-10
fix: add .gitattributes to enforce LF endings on plugin scripts (#1342)

### LINKED ISSUE #1342: Bug: mcp-server.cjs has CRLF line endings — shebang fails on macOS/Linux
## Description

The bundled `scripts/mcp-server.cjs` is published with Windows-style CRLF (`\r\n`) line endings. On macOS/Linux, when the plugin system runs the script directly via its shebang, the OS interprets the first line as:

```
#!/usr/bin/env node\r
```

This causes the error:

```
env: node\r: No such file or directory
```

The built-in plugin MCP server (`plugin:claude-mem:mcp-search`) fails to start as a result.

## Reproduction

1. Install claude-mem plugin (v10.3.1)
2. On macOS or Linux, check MCP status — `plugin:claude-mem:mcp-search` shows `✘ failed`
3. Run the script directly to confirm:
   ```bash
   echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1.0.0"}}}' \
     | /path/to/plugins/cache/thedotmack/claude-mem/10.3.1/scripts/mcp-server.cjs
   ```
   Output: `env: node\r: No such file or directory`

## Workaround

Strip the carriage return from the shebang line in the cached script:

```bash
sed -i '' $'1s/\r$//' ~/.claude/plugins/cache/thedotmack/claude-mem/10.3.1/scripts/mcp-server.cjs
```

This fix is lost on plugin reinstall/update.

## Suggested Fix

Ensure the build/publish step outputs `mcp-server.cjs` with LF line endings (at minimum the shebang line). A `.gitattributes` rule like `*.cjs text eol=lf` or a post-build `dos2unix` step would prevent this.

## Environment

- macOS (Darwin 24.6.0)
- claude-mem v10.3.1
- Claude Code CLI

ousamabenyounes:
## Summary

Fixes #1342

- Without `.gitattributes`, building on Windows produces plugin scripts with CRLF (`\r\n`) line endings. The `\r` on the shebang line causes `env: node\r: No such file or directory` on macOS/Linux, silently breaking `plugin:claude-mem:mcp-search` and all hook scripts
- Adds `.gitattributes` with `* text=auto eol=lf` (global default) and explicit `eol=lf` rules for `plugin/scripts/*.cjs` and `*.js`
- Marks binary assets (`*.png`, `*.woff`, etc.) as `binary` so git never modifies them

## Verification

- [x] Post-fix tests: 1116 pass, 0 regressions (+10 new tests)
- [x] New tests: 10 CRLF regression guards in `tests/plugin-scripts-line-endings.test.ts` covering shebang line + full file for each critical script
- [x] Contribution guidelines: minimal diff, conventional commits

## Files changed

| File | Change |
|------|--------|
| `.gitattributes` | New: enforce LF line endings for all text files, especially plugin scripts |
| `tests/plugin-scripts-line-endings.test.ts` | New: 10 regression guards asserting no CRLF in committed plugin scripts |

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,21 @@
+# Normalize all text files to LF on commit and checkout.
+# This prevents CRLF shebang lines in bundled scripts from breaking
+# the MCP server on macOS/Linux when built on Windows. Fixes #1342.
+* text=auto eol=lf
+
+# Compiled plugin scripts must always be LF — CRLF in the shebang
+# causes "env: node\r: No such file or directory" on non-Windows hosts.
+plugin/scripts/*.cjs eol=lf
+plugin/scripts/*.js  eol=lf
+
+# Explicitly mark binary assets so git never modifies them.
+*.png  binary
+*.jpg  binary
+*.jpeg binary
+*.ico  binary
+*.gif  binary
+*.woff  binary
+*.woff2 binary
+*.ttf   binary
+*.eot   binary
+*.otf   binary
--- a/tests/plugin-scripts-line-endings.test.ts
+++ b/tests/plugin-scripts-line-endings.test.ts
@@ -1,0 +1,51 @@
+import { describe, it, expect } from 'bun:test';
+import { readFileSync, existsSync } from 'fs';
+import { join } from 'path';
+
+/**
+ * Regression tests for issue #1342.
+ *
+ * Bundled plugin scripts use a shebang line (#!/usr/bin/env node or #!/usr/bin/env bun).
+ * If those files are committed with Windows CRLF line endings, the shebang becomes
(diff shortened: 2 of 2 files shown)


ousamabenyounes:
Addressed in the latest commit: added `context-generator.cjs` to `SHEBANG_SCRIPTS` and replaced silent early-returns with `expect(existsSync(filePath)).toBe(true)` to fail loudly when build artifacts are absent.

PR #1688, opened 2026-04-10
fix: pass project filter through in searchObservations and searchSessions (#1539)

### LINKED ISSUE #1539: Worker API: /api/search/observations and /api/search/sessions ignore project parameter
## Problem

The worker HTTP endpoints `/api/search/observations` and `/api/search/sessions` accept a `project` query parameter but ignore it completely. Queries always return results from all projects regardless of the `project` filter.

```bash
# Returns results from all projects even though project=project-a
curl "http://127.0.0.1:37777/api/search/observations?query=test&project=project-a"
```

## Root Cause

In `src/services/worker/SearchManager.ts`, the `searchObservations()` method (line ~889) destructures the project from params but never uses it:

```typescript
async searchObservations(args: any): Promise<any> {
    const normalized = this.normalizeParams(args);
    const { query, ...options } = normalized;
    // options.project exists but is never passed to queryChroma or hydration

    const chromaResults = await this.queryChroma(query, 100);  // no project filter
    // ...
    results = this.sessionStore.getObservationsByIds(recentIds, { orderBy: 'date_desc', limit });  // no project filter
}
```

Same issue in `searchSessions()` (line ~946).

**Note:** The main `search()` method (line ~126), used by the MCP tool, correctly handles project filtering — it builds a Chroma `where` clause with the project and passes it through. Only `searchObservations()` and `searchSessions()` are affected.

## Expected Behavior

When `project` is passed as a query parameter, results should be scoped to that project only — both in the Chroma vector query and in the SQLite hydration.

## Proposed Fix

1. Pass `options.project` as a Chroma `where` filter in `queryChroma()` calls
2. Pass `options.project` to `getObservationsByIds()` / `getSessionSummariesByIds()` during hydration

This mirrors the pattern already used in the main `search()` method.

## Impact

This bug breaks project-scoped search for:
- OpenClaw plugin search commands which call these endpoints with a project filter
- Any direct HTTP API consumers filtering by project
- The MCP `search` tool is **not** affected (it uses the main `search()` method which handles project correctly)

ousamabenyounes:
## Summary

- `searchObservations()` and `searchSessions()` ignored the `project` query parameter — both Chroma queries and SQLite hydration lacked project scoping, causing cross-project result leakage in `/api/search/observations` and `/api/search/sessions`
- Fixed by mirroring the pattern already used in the main `search()` method (lines 173-178): pass `project` as a Chroma `where` filter and thread it through to `getObservationsByIds`/`getSessionSummariesByIds`
- Also fixed the JSDoc comment on `/api/search/sessions` which was missing `project` from its parameter list

## Test plan

- [x] `searchObservations`: project is passed to Chroma where filter
- [x] `searchObservations`: absent project produces no where filter (no regression)
- [x] `searchObservations`: project is threaded to SQLite hydration when Chroma returns hits
- [x] `searchSessions`: project + `doc_type` combined via `$and` in Chroma filter
- [x] `searchSessions`: absent project = only `doc_type` filter (no regression)
- [x] `searchSessions`: project is threaded to SQLite hydration when Chroma returns hits
- [x] Full test suite: 1312 pass / 3 skip / 12 fail (baseline 1306/3/12 — 6 new tests, 0 regressions)

Fixes #1539

🤖 Generated with [Ora Studio](https://studio.oratelecom.net)

### DIFF
--- a/src/services/worker/SearchManager.ts
+++ b/src/services/worker/SearchManager.ts
@@ -895,8 +895,10 @@ export class SearchManager {
     if (this.chromaSync) {
       logger.debug('SEARCH', 'Using hybrid semantic search (Chroma + SQLite)', {});
 
-      // Build Chroma where filter — scope to project if provided (#1539)
-      const whereFilter = options.project ? { project: options.project } : undefined;
+      // Build Chroma where filter — always scope to observation docs, add project if provided (#1539)
+      const whereFilter: Record<string, any> = options.project
+        ? { $and: [{ doc_type: 'observation' }, { project: options.project }] }
+        : { doc_type: 'observation' };
 
       // Step 1: Chroma semantic search (top 100)
       const chromaResults = await this.queryChroma(query, 100, whereFilter);
--- a/src/services/worker/SearchManager.ts
+++ b/src/services/worker/SearchManager.ts
@@ -895,8 +895,11 @@ export class SearchManager {
     if (this.chromaSync) {
       logger.debug('SEARCH', 'Using hybrid semantic search (Chroma + SQLite)', {});
 
+      // Build Chroma where filter — scope to project if provided (#1539)
+      const whereFilter = options.project ? { project: options.project } : undefined;
+
       // Step 1: Chroma semantic search (top 100)
-      const chromaResults = await this.queryChroma(query, 100);
+      const chromaResults = await this.queryChroma(query, 100, whereFilter);
       logger.debug('SEARCH', 'Chroma returned semantic matches', { matchCount: chromaResults.ids.length });
 
       if (chromaResults.ids.length > 0) {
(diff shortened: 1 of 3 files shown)


ousamabenyounes:
Fixed CodeRabbit review in latest push (dd950e0):

- **[Major]** `searchObservations()` Chroma filter now always includes `doc_type: 'observation'` so session_summary and user_prompt docs cannot pollute observation results. Combined with `project` via `$and` when provided.
- **[Minor]** `/api/search/help` payload for `/api/search/sessions` now lists `project` parameter, in sync with the JSDoc updated in the original PR.

Tests updated to assert the new `doc_type: 'observation'` always-present behavior: 1312 pass / 12 fail (baseline 1310/3/12 — 2 additional new tests pass, 0 regressions).

thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1912, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1689, opened 2026-04-10
fix: normalize Windows backslash paths in bun-runner.js (#1281)

### LINKED ISSUE #1281: Windows: Stop hooks fail with MODULE_NOT_FOUND due to backslash path corruption
## Bug

  On Windows (Git Bash shell), all Stop hooks fail with `MODULE_NOT_FOUND` because `${CLAUDE_PLUGIN_ROOT}` expands to a
  backslash path that Node.js cannot resolve.

  ## Environment

  - **OS**: Windows 11 Home 10.0.26200
  - **Shell**: Git Bash (via Claude Code CLI)
  - **Node.js**: v24.13.0
  - **Plugin version**: 9.1.1

  ## Error

  All 3 Stop hooks produce the same error:

  Error: Cannot find module
    'C:\Users\username.claude\plugins\cache\thedotmack\claude-mem\9.1.1\scripts\bun-runner.js'
      at Module._resolveFilename (node:internal/modules/cjs/loader:1421:15)
      ...
    code: 'MODULE_NOT_FOUND',

  This also affects SessionStart, UserPromptSubmit, and PostToolUse hooks — any hook that calls `bun-runner.js`.

  ## Root Cause

  When Claude Code expands `${CLAUDE_PLUGIN_ROOT}` on Windows and passes it through Git Bash to `node`, backslashes in
  the path get interpreted as escape sequences:

  - `\t` (in `\thedotmack`) → tab character
  - `\b` (in `\basil`) → backspace character
  - `\U`, `\p`, `\c`, `\s` → may be silently stripped or mangled

  Proof that the file exists and works with forward slashes:

  ```bash
  $ node "C:/Users/username/.claude/plugins/cache/thedotmack/claude-mem/9.1.1/scripts/bun-runner.js"
  Usage: node bun-runner.js <script> [args...]

  $ node -e "const fs = require('fs');
  console.log(fs.existsSync('C:/Users/username/.claude/plugins/cache/thedotmack/claude-mem/9.1.1/scripts/bun-runner.js'))"
  true

  Suggested Fix

  Option A — Normalize paths in bun-runner.js itself:

  // At the top of bun-runner.js
  const scriptPath = process.argv[2]?.replace(/\\/g, '/');

  Option B — Use forward slashes in hooks.json commands:

  "command": "node \"$(echo ${CLAUDE_PLUGIN_ROOT} | sed 's/\\\\/\\//g')/scripts/bun-runner.js\" ..."

  Option C — This may also be a Claude Code issue where CLAUDE_PLUGIN_ROOT should use forward slashes when the shell is
  Git Bash on Windows.

  Impact

  - All hooks that invoke bun-runner.js fail silently on Windows
  - Session summarization, observation tracking, and session-complete callbacks don't run
  - Non-blocking (doesn't crash Claude Code), but memory features depending on these hooks are broken
  - MCP tools still work fine — only the hook system is affected

ousamabenyounes:
## Summary

Fixes #1281

On Windows with Git Bash, `${CLAUDE_PLUGIN_ROOT}` expands to a backslash Windows path (e.g. `C:\Users\username\.claude\...`). When the shell interprets escape sequences in that path, characters like `\t` (in `thedotmack`) become a tab and `\b` (in `\basil`) becomes a backspace, making Node.js unable to resolve `bun-runner.js` or any script it tries to launch.

- Adds a `normalizePath()` helper that converts backslashes to forward slashes (safe on Windows — Node.js accepts both)
- Applies it to `CLAUDE_PLUGIN_ROOT` when building `RESOLVED_PLUGIN_ROOT`
- Applies it to the script path argument (`args[0]`) after `fixBrokenScriptPath()`

## Verification

- [x] Baseline tests: 1200 pass, 34 pre-existing failures
- [x] Post-fix tests: 1208 pass, 34 failures (same pre-existing), 0 regressions
- [x] New tests: 8 added in `tests/bun-runner.test.ts`, all pass
- [x] Issue alignment: diff directly implements Option A from the issue's "Suggested Fix" section

## Files changed

| File | Change |
|------|--------|
| `plugin/scripts/bun-runner.js` | Add `normalizePath()`, apply to `CLAUDE_PLUGIN_ROOT` and script path arg |
| `tests/bun-runner.test.ts` | 8 unit tests covering the normalization logic |

Generated by Ora Studio
Vibe coded by Ousama Ben Younes

### DIFF
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -19,13 +19,26 @@ import { fileURLToPath } from 'url';
 
 const IS_WINDOWS = process.platform === 'win32';
 
+/**
+ * Normalize Windows backslash paths to forward slashes.
+ * On Windows with Git Bash, CLAUDE_PLUGIN_ROOT may contain backslashes
+ * (e.g. C:\Users\...) that get corrupted when the shell interprets escape
+ * sequences like \t (tab) or \b (backspace) in the path. Normalizing to
+ * forward slashes is safe because Node.js accepts both on Windows.
+ * Fixes #1281.
+ */
+function normalizePath(p) {
+  if (!p) return p;
+  return p.replace(/\\/g, '/');
+}
+
 // Self-resolve plugin root when CLAUDE_PLUGIN_ROOT is not set by Claude Code.
 // Upstream bug: anthropics/claude-code#24529 — Stop hooks (and on Linux, all hooks)
 // don't receive CLAUDE_PLUGIN_ROOT, causing script paths to resolve to /scripts/...
 // which doesn't exist. This fallback derives the plugin root from bun-runner.js's
 // own filesystem location (this file lives in <plugin-root>/scripts/).
 const __bun_runner_dirname = dirname(fileURLToPath(import.meta.url));
-const RESOLVED_PLUGIN_ROOT = process.env.CLAUDE_PLUGIN_ROOT || resolve(__bun_runner_dirname, '..');
+const RESOLVED_PLUGIN_ROOT = normalizePath(process.env.CLAUDE_PLUGIN_ROOT) || resolve(__bun_runner_dirname, '..');
 
 /**
  * Fix script path arguments that were broken by empty CLAUDE_PLUGIN_ROOT.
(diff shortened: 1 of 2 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1882, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1691, opened 2026-04-10
fix: validate watch.context.path before writing AGENTS.md (#1204)

### LINKED ISSUE #1204: `watch.context.path` in settings can write AGENTS.md content to arbitrary file paths
While looking at how context injection works, I noticed that [`processor.ts` (L224–L252)](https://github.com/thedotmack/claude-mem/blob/main/src/services/transcripts/processor.ts#L224-L252) in `updateContext` takes `watch.context.path` from the config, expands `~`, and writes directly to it:

```typescript
const agentsPath = expandHomePath(watch.context.path ?? `${cwd}/AGENTS.md`);
writeAgentsMd(agentsPath, content);
```

There's no validation that the resolved path stays within the project directory or any other safe boundary. If someone (or something that modifies `settings.json`) sets `watch.context.path` to something like `../../.bashrc` or `~/.ssh/authorized_keys`, the content from `/api/context/inject` gets written there.

This is particularly relevant because claude-mem runs with the user's full filesystem permissions, and the `cwd` fallback also comes from the transcript's session state, which could be influenced externally in some setups (e.g., OpenClaw gateways).

**Suggestion:** Resolve the path and validate it stays within the project root or `~/.claude-mem/` before writing. Reject anything containing `..` segments that escape the allowed directory.

ousamabenyounes:
## Summary

Fixes #1204

- Resolves `watch.context.path` relative to the session `cwd` and validates it stays within the project root or `~/.claude-mem/` before calling `writeAgentsMd`
- Adds `isSafeContextPath(resolvedPath, projectRoot)` utility in `agents-md-utils.ts` (sep-normalized prefix check, cross-platform)
- Rejects paths like `../../.bashrc`, `~/.ssh/authorized_keys`, `/etc/passwd` with a `logger.warn` and early return

## Verification

- [x] Baseline tests: 1200 pass, 34 pre-existing failures
- [x] Post-fix tests: 1211 pass, 34 failures (no regressions)
- [x] New tests: 11 added in `tests/utils/agents-md-utils.test.ts`, all pass — cover allowed paths (project root, `~/.claude-mem/`) and attack vectors (traversal, sibling-prefix bypass, absolute paths to `/etc`, `~/.ssh`)
- [x] Review agent: issue alignment verified (Gate 1–4 all PASS)

## Files changed

| File | Change |
|------|--------|
| `src/utils/agents-md-utils.ts` | Add `isSafeContextPath` export |
| `src/services/transcripts/processor.ts` | Use `isSafeContextPath` guard in `updateContext` |
| `tests/utils/agents-md-utils.test.ts` | New test file (11 tests) |

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/src/services/transcripts/processor.ts
+++ b/src/services/transcripts/processor.ts
@@ -5,7 +5,8 @@ import { sessionCompleteHandler } from '../../cli/handlers/session-complete.js';
 import { ensureWorkerRunning, workerHttpRequest } from '../../shared/worker-utils.js';
 import { logger } from '../../utils/logger.js';
 import { getProjectContext, getProjectName } from '../../utils/project-name.js';
-import { writeAgentsMd } from '../../utils/agents-md-utils.js';
+import { writeAgentsMd, isSafeContextPath } from '../../utils/agents-md-utils.js';
+import { resolve } from 'path';
 import { resolveFieldSpec, resolveFields, matchesRule } from './field-utils.js';
 import { expandHomePath } from './config.js';
 import type { TranscriptSchema, WatchTarget, SchemaEvent } from './types.js';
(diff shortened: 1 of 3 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1934, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1698, opened 2026-04-10
fix: reap stuck generators in reapStaleSessions (fixes #1652)

### LINKED ISSUE #1652: Subagent processes not cleaned up after session ends
## Problem

After using claude-mem for a while, I noticed my machine slowing down significantly. Checked `ps aux` and found **14+ orphaned claude subagent processes** still running in the background, some dating back 3 days.

All of them were children of the `worker-service.cjs --daemon` process (bun), each consuming 300-600MB of memory.

```
PID 1172  - started Monday (3 days ago)
PID 51407 - started Tuesday
PID 51593 - started Tuesday
...total 14 orphaned processes
```

## Root cause (what I found)

The Stop hook flow queues `/api/sessions/summarize` but doesn't call session complete or delete. So from the daemon's perspective, the session stays "active" and the spawned claude subprocess just sits there waiting for more queue input forever.

The internal reaper also skips these because they're tied to "active" sessions.

## Steps to reproduce

1. Use claude-mem normally across multiple sessions
2. Close terminal / end sessions
3. After a day or two, run `ps aux | grep claude`
4. Observe accumulated background claude processes under the worker-service daemon

## Workaround

I set up a `SessionEnd` hook in `~/.claude/settings.json` that kills background claude subagent processes older than 5 minutes on session end. Works fine as a band-aid but the daemon itself should handle cleanup.

## Suggestion

- Call session complete/delete after summarize finishes in the Stop flow
- Add an idle TTL so stale sessions get cleaned up automatically
- Reconcile orphaned child processes on daemon startup

## Environment

- claude-mem 9.0.12
- macOS (Apple Silicon)
- Claude Code CLI

ousamabenyounes:
## Summary

- `reapStaleSessions()` unconditionally skipped sessions with a non-null `generatorPromise`, so any session whose SDK subprocess hung (blocked on `for await` inside `SDKAgent`) was never cleaned up
- Add `MAX_GENERATOR_IDLE_MS = 5 min` constant and stale-generator detection: if `lastGeneratorActivity` hasn't advanced in > 5 min while a generator is "running", SIGKILL the subprocess, abort the controller, and call `deleteSession()` to complete the cleanup
- 17 new tests covering threshold constants, detection logic, subprocess kill, abort behaviour, and preservation of the existing idle-session reaping path

## Root cause

The stuck-generator problem has three reinforcing gaps:

1. `SessionQueueProcessor` idle timeout (3 min) only fires when the generator is in `waitForMessage()` — a generator blocked on `for await (const msg of queryResult)` inside `SDKAgent` is **never** in that state
2. `reapStaleSessions()` skipped every session with `generatorPromise !== null` — no timeout, ever
3. The orphan reaper skips subprocesses whose session is still in the active sessions map — so the zombie subprocess also survives

The fix attacks gap 2 directly: detect stale `lastGeneratorActivity` and force-kill the subprocess (gap 3 is then auto-resolved since the subprocess exits, which unblocks the generator, which lets `deleteSession` remove the session from the map).

## Test plan

- [x] `bun test tests/services/worker/reap-stale-sessions.test.ts` — 17 pass, 0 fail
- [x] `bun test` — 1217 pass, 34 fail (34 pre-existing failures unchanged)

🤖 Generated with [Ora Studio](https://studio.oratelecom.net)

### DIFF
--- a/src/services/worker/SessionManager.ts
+++ b/src/services/worker/SessionManager.ts
@@ -17,6 +17,64 @@ import { SessionQueueProcessor } from '../queue/SessionQueueProcessor.js';
 import { getProcessBySession, ensureProcessExit } from './ProcessRegistry.js';
 import { getSupervisor } from '../../supervisor/index.js';
 
+/** Idle threshold before a stuck generator (zombie subprocess) is force-killed. */
+export const MAX_GENERATOR_IDLE_MS = 5 * 60 * 1000; // 5 minutes
+
+/** Idle threshold before a no-generator session with no pending work is reaped. */
+export const MAX_SESSION_IDLE_MS = 15 * 60 * 1000; // 15 minutes
+
+/**
+ * Minimal process interface used by detectStaleGenerator — compatible with
+ * both the real Bun.Subprocess / ChildProcess shapes and test mocks.
+ */
+export interface StaleGeneratorProcess {
+  exitCode: number | null;
+  kill(signal?: string): boolean | void;
+}
+
+/**
+ * Minimal session fields required to evaluate stale-generator status.
+ * This is a subset of ActiveSession, allowing unit tests to pass plain objects.
+ */
+export interface StaleGeneratorCandidate {
+  generatorPromise: Promise<void> | null;
+  lastGeneratorActivity: number;
+  abortController: AbortController;
+}
+
+/**
+ * Detect whether a session's generator is stuck (zombie subprocess) and, if so,
+ * SIGKILL the subprocess and abort the controller.
+ *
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
@coderabbitai review

ousamabenyounes:
@coderabbitai review

ousamabenyounes:
@coderabbitai review

ousamabenyounes:
@coderabbitai review

ousamabenyounes:
@coderabbitai review

ousamabenyounes:
@coderabbitai review

Issue #1312, opened 2026-03-09
Summarize produces <observation> instead of <summary> — summaries silently lost


GigiTiti-Kai:
## Bug Description

When the SDK agent processes a `summarize` message, the LLM often responds with `<observation>` XML tags instead of the expected `<summary>` tags. The parser's `parseSummary()` finds no `<summary>` block, returns `null`, and the summary is silently lost. The stop hook reports "Summary completed" (false positive) because the `pending_messages` queue shows 0 pending items.

## Root Cause

The SDK session conversation is heavily biased toward `<observation>` output:
- The system/init prompt provides a detailed `<observation>` XML template
- Every previous assistant response in the conversation uses `<observation>` tags
- By the time the summarize prompt arrives, the in-context conditioning toward `<observation>` is overwhelming

The current `buildSummaryPrompt()` asks for `<summary>` format but lacks explicit negative instructions against `<observation>` output. The LLM pattern-matches on the dominant format.

## Evidence from Logs

```
14:33:21 [QUEUE] ENQUEUED | messageId=38585 | type=summarize
14:33:21 [QUEUE] CLAIMED | messageId=38585
14:33:26 [SDK]   ← Response received (1284 chars) {promptNumber=1} <observation>
14:33:26 [DB]    STORED | obsCount=1 | hasSummary=false
14:33:26 [QUEUE] CONFIRMED | messageId=38585 | deleted from queue
```

The summarize message was processed, but the LLM responded with `<observation>` instead of `<summary>`. No error was logged.

### 1. Strengthen summary prompt (prompts.ts)

Add explicit "mode switch" instructions in `buildSummaryPrompt()`:

```typescript
// At the start of the prompt:
"--- MODE SWITCH: PROGRESS SUMMARY ---\n" +
"Do NOT output <observation> tags. Output <summary> tags ONLY."

// At the end:
"IMPORTANT REMINDER: Your response MUST use <summary> tags. " +
"Any <observation> response will be discarded by the system."
```

### ~~2. Add parser fallback (parser.ts)~~ — DO NOT DO THIS

> **⚠️ Warning**: Adding a fallback in `parseSummary()` to convert `<observation>` into `<summary>` format is **harmful**. `parseSummary()` is called by `processAgentResponse()` on **every** SDK response — not just summarize responses. A fallback here would cause every normal observation response to also generate a spurious summary, flooding `session_summaries` with auto-converted records.
>
[...]

PR #1718, opened 2026-04-11
fix(ResponseProcessor): salvage synthetic summary when AI returns <observation> instead of <summary> (Issue #1312)

aaronwong1989:
## Summary

Fixes Issue #1312: AI sometimes returns `<observation>` XML tags instead of `<summary>` tags during the summarize phase, despite clear instructions in `buildSummaryPrompt()` requiring `<summary> ONLY` output.

When this occurs, `parseSummary()` returns null and the entire session summary is lost. This fix detects the condition (summary missing + observations present) and synthesizes a summary from the observation data, ensuring session summaries are not completely lost.

## Changes

**File:** `src/services/worker/agents/ResponseProcessor.ts`

Added fallback logic after `normalizeSummaryForStorage()`:

```typescript
let finalSummaryForStore = summaryForStore;
if (!summaryForStore && observations.length > 0) {
  const primary = observations[0];
  finalSummaryForStore = {
    request: primary.title || `Session observations (${observations.length} items)`,
    investigated: primary.narrative || primary.facts?.join('; ') || '',
    learned: primary.facts?.join('; ') || '',
    completed: primary.type === 'feature' || primary.type === 'bugfix' ? (primary.title || '') : '',
    next_steps: '',
    notes: `[Salvaged from ${observations.length} observation(s)] AI returned <observation> instead of <summary>`
  };
  logger.warn('PARSER', `SALVAGED summary from ${observations.length} observation(s)...`);
}
```

## Salvage Mapping

| Summary Field | Source |
|--------------|--------|
| `request` | observation title |
| `investigated` | observation narrative or facts |
| `learned` | observation facts joined |
| `completed` | title if type is feature/bugfix |
| `notes` | indicates synthetic salvage |

## Impact

- **Observations:** Stored normally regardless of this fallback — zero impact
- **Session Summary:** When AI ignores `<summary>` instruction, a synthetic summary is created instead of total loss
- **Risk:** Minimal — only affects the summary when observations are present but summary parse fails

## Testing

Build succeeds. The test suite has 38 failures in both baseline and PR branch (pre-existing environment/configuration issues unrelated to this change).

### DIFF
--- a/src/services/worker/agents/ResponseProcessor.ts
+++ b/src/services/worker/agents/ResponseProcessor.ts
@@ -85,6 +85,27 @@ export async function processAgentResponse(
   // Convert nullable fields to empty strings for storeSummary (if summary exists)
   const summaryForStore = normalizeSummaryForStorage(summary);
 
+  // Fallback: When summary parse fails but observations exist, salvage a synthetic summary.
+  // Fixes Issue #1312: AI sometimes returns <observation> instead of <summary> despite clear instructions.
+  // Observations are stored normally; this only affects the session summary.
+  let finalSummaryForStore = summaryForStore;
+  if (!summaryForStore && observations.length > 0) {
+    const primary = observations[0];
+    finalSummaryForStore = {
+      request: primary.title || `Session observations (${observations.length} items)`,
+      investigated: primary.narrative || primary.facts?.join('; ') || '',
+      learned: primary.facts?.join('; ') || '',
+      completed: primary.type === 'feature' || primary.type === 'bugfix' ? (primary.title || '') : '',
+      next_steps: '',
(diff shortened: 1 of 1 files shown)


thedotmack:
Reverted in 9a07af1 + c6ca162.

Runtime salvage was the wrong layer. When the LLM emits `<observation>` instead of `<summary>` ~72% of the time, the fix is to constrain the model output — tool schemas, structured output / JSON mode, grammar-constrained decoding, or a sharper system prompt with a failure-mode example — not to synthesize a summary from the first observation at ingest time. The salvage path produced summaries whose fields were a reshuffled copy of `observations[0]`, inflating row counts with near-duplicate content (no content-hash dedup exists on `session_summaries`).

Issue #1312 remains real; the right fix is upstream at the prompt/response-format boundary.

PR #1736, opened 2026-04-12
feat: auto-cleanup stale/oversized observer session JSONL files

laihenyi:
## Problem

claude-mem spawns background observer agents whose conversation transcripts accumulate in `~/.claude/projects/-...-observer-sessions/` without any automatic cleanup. This caused 85 GB of JSONL files (individual sessions up to 9 GB), despite all useful data already being extracted into SQLite.

## Solution

Extend `CleanupJob` with a new step that runs on the existing daily schedule. A JSONL file is deleted when **either**:

- It is older than **30 days** (configurable: `observerSessionMaxAgeDays`)
- It is larger than **500 MB** (configurable: `observerSessionMaxSizeMB`)

Matching subdirectories (created by Claude Code alongside some JSONL files) are also removed. The step is **enabled by default** (`enableObserverSessionCleanup: true`).

## Changes

- `src/services/worker/CleanupJob.ts` — three new `CleanupConfig` fields, new `observerSessionCleanup` result field, `runObserverSessionCleanup()` private method, integrated into `run()` and `startScheduled()`

## New Config Fields

```typescript
enableObserverSessionCleanup: boolean  // default: true
observerSessionMaxAgeDays: number      // default: 30
observerSessionMaxSizeMB: number       // default: 500
```


laihenyi:
## CodeRabbit Review Fixes

Addressed all 4 critical findings from CodeRabbit review:

### 1. `acquire.ts` — Guard `JSON.stringify(undefined)` before `.length`
Added `?? ''` defaults for `toolInput`/`toolOutput` and `|| ''` fallback after `JSON.stringify` to prevent crash when values are undefined.

### 2. `render.ts` — Replace fake persistence with real calls
Removed all placeholder implementations. Constructor now accepts `SessionStore` and `ChromaSync` directly. Observations and summaries are persisted via real `storeObservation`/`storeSummary` calls, and Chroma sync uses actual `syncObservation`/`syncSummary` methods.

### 3. `SleepAgent.ts` — Add in-flight cycle guard
Added `cycleInFlight` promise guard to `runCycleIfNotBusy()` to prevent overlapping cycles from `setInterval`. `recordActivity()` now only resets the timer when a deep cycle actually ran.

### 4. `SurpriseMetric.ts` — Use `created_at_epoch` fallback
`calculateTemporalNovelty()` now prefers `created_at_epoch` (always a valid number) over `created_at` (may be undefined, causing `NaN`). `getSimilarMemories()` fallback query also selects `created_at_epoch`. Invalid dates are defensively handled as fully novel.

Build passes successfully.

laihenyi:
Superseded by #1772 — new clean PR from `fix/observer-session-cleanup` branch with only the observer session cleanup changes (no unrelated branch commits).

Issue #1734, opened 2026-04-12
🐛 PreToolUse file-context hook forces limit:1 on Read, corrupts skill knowledge base loading in new sessions


yangle94:
## Bug Description

  The `PreToolUse` `file-context` hook (`TR.execute()` in `worker-service.cjs`) forcibly overrides the Read tool's input parameters with `limit: 1` when a file has prior observations from previous sessions. This causes Claude to believe the file contains only 1 line, leading to incorrect behavior — especially when skill workflows load knowledge base index files.

  ## Root Cause

  In the `file-context` hook handler (`TR` object), when observations exist for a file path, the hook returns:

  ```javascript
  return {
    hookSpecificOutput: {
      hookEventName: "PreToolUse",
      additionalContext: gxe(p, r),
      permissionDecision: "allow",
      updatedInput: {                 // ← THIS IS THE PROBLEM
        file_path: r,
        limit: 1                      // ← Forces Read to only return 1 line!
      }
    }
  }

  The updatedInput with limit: 1 overrides the original Read call's parameters, causing Claude to receive only the header line of files like tea-index.csv.

  Additionally, the gxe() function injects a message containing:
  "This file has prior observations. Only line 1 was read to save tokens."

  This message is misleading in a new session where Claude's context does NOT contain the file's previous content. Claude trusts this message and assumes the file is empty/incomplete, compounding the problem.

  Reproduction Steps

  1. In Session A, open any file (e.g., .claude/skills/bmad-tea/resources/tea-index.csv — a 43-line CSV index)
  2. claude-mem records observations for this file
  3. Start Session B (new session)
  4. Trigger a skill that reads this file
  5. Observe: Read returns only 1 line (the CSV header)
  6. Claude concludes: "knowledge base index is empty" — downstream skill behavior is broken

  Expected Behavior

  The file-context hook should NOT override Read's limit parameter. The hook can inject context about prior observations as a hint, but should not modify the tool's actual input parameters. In a new session, Claude does not have the file content in context and MUST be allowed to read the full file.

  Suggested Fix

  1. Remove updatedInput override — The hook should not modify Read's input parameters
  2. Reword the injected message — Make it clear this is a hint from a prior session, not a statement about what was actually read:
  "This file has observations from prior sessions. You may already know enough from the timeline below, or you can read the full file if you need current content."

  Environment

  - claude-mem version: 12.1.0
  - Claude Code version: Latest (April 2026)
[...]

PR #1748, opened 2026-04-13
fix: allow full file read when file modified since last observation

### LINKED ISSUE #1719: Improper cache validation with `PreToolUse:Read` hook in Claude Code
[ X] I searched [existing issues](https://github.com/thedotmack/claude-mem/issues) and confirmed this is not a duplicate

### Bug Description

The `PreToolUse:Read` hook truncates file reads to line 1 when the
file has prior observations, to save tokens. This interacts badly
with two scenarios:

1. **Subagent modifications** - A subagent modifies a file, but
   the hook still truncates the parent's next Read because the
   observation predates the change. The parent never gets to see
   the updated content.

2. **Claude Code deduplication** - After the hook returns line 1,
   Claude Code's built-in dedup marks the file as "read" and
   refuses subsequent Read attempts with "file unchanged since
   last read." The combination means the file becomes unreadable
   for the rest of the conversation.

### Expected Behavior

The hook should compare the file's mtime against the most recent
observation timestamp. If the file has been modified since the
last observation, allow a full read instead of truncating.

### Environment
- **Claude-mem version**: 12.1.0
- **Claude Code version**: v2.1.101
- **OS**: Debian 12 (node:24-bookworm)
- **Platform**: x86-64 (i7 12700k)

octo-patch:
Fixes #1719
Relates to #1734

## Problem

The `PreToolUse` file-context hook always applied `updatedInput: { limit: 1 }` whenever a file had prior observations, even when the file had been modified since those observations were recorded. This caused two concrete breakage modes:

1. **Subagent modifications** (#1719): A subagent edits a file. When the parent agent's next `Read` fires the hook, the observation predates the change, so `limit: 1` is applied and the parent never sees the updated content.

2. **Stale observations in new sessions** (#1734): In a fresh session, if a file was modified after the last observation was stored, Claude only sees 1 line of the current file — the observation timeline describes an old state, and `limit: 1` prevents reading the actual current content.

## Solution

During the existing `statSync` call (already used for the file-size gate), capture `stat.mtimeMs`. After fetching and deduplicating observations, compare `mtimeMs` against `Math.max(...dedupedObservations.map(o => o.created_at_epoch))`.

- If the file was **modified after the most recent observation** → skip `updatedInput`, allow a full read, and update the injected message to say the file was modified since the last observation.
- If the file was **not modified since the last observation** → keep the existing `limit: 1` optimisation (observations are still accurate, token savings apply).

This preserves the existing token-saving behaviour for the common case while fixing correctness for modified files.

## Testing

- File with observations, no changes since: `limit: 1` still applied (unchanged behaviour).
- File with observations, modified since last observation: full read allowed, message updated.
- File with no observations: hook exits early, unchanged behaviour.
- File stat error: `fileMtimeMs` stays `null`, `fileModifiedSinceObservation` is `false`, falls back to `limit: 1` (safe default).

### DIFF
--- a/src/cli/handlers/file-context.ts
+++ b/src/cli/handlers/file-context.ts
@@ -106,7 +106,7 @@ function deduplicateObservations(
   return scored.slice(0, displayLimit).map(s => s.obs);
 }
 
-function formatFileTimeline(observations: ObservationRow[], filePath: string): string {
+function formatFileTimeline(observations: ObservationRow[], filePath: string, fileModifiedSinceObservation: boolean): string {
   // Escape filePath for safe interpolation into recovery hints (quotes, backslashes, newlines)
   const safePath = filePath.replace(/\\/g, '\\\\').replace(/"/g, '\\"').replace(/\n/g, '\\n');
   // Group observations by day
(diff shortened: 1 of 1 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1896, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1750, opened 2026-04-13
fix: migration 7 unique-constraint check incorrectly matches primary-key index

### LINKED ISSUE #1749: Migration 7: `removeSessionSummariesUniqueConstraint` always re-creates the table because primary-key index is counted as a unique index
## Bug Description

In `src/services/sqlite/migrations/runner.ts`, the `removeSessionSummariesUniqueConstraint()` method (migration 7) uses this check to decide whether the unique constraint has already been removed:

```ts
const summariesIndexes = this.db.query('PRAGMA index_list(session_summaries)').all() as IndexInfo[];
const hasUniqueConstraint = summariesIndexes.some(idx => idx.unique === 1);
```

**The problem:** SQLite's `PRAGMA index_list` returns every index on the table, including the one backing the `PRIMARY KEY`. The `IndexInfo` interface has an `origin` field (values: `'c'` = user-created, `'u'` = unique constraint, `'pk'` = primary key), but the code only checks `idx.unique === 1` and ignores `idx.origin`.

Because `session_summaries` has `id INTEGER PRIMARY KEY AUTOINCREMENT`, the primary-key index always has `unique = 1`. This means:

1. `hasUniqueConstraint` is always `true`.
2. The early-return branch (`if (!hasUniqueConstraint) return;`) is never taken.
3. The full table-recreation DDL runs on **every** startup after migration 7 should have already completed.
4. Even though `INSERT OR IGNORE` prevents the version record from being inserted twice, the expensive `BEGIN TRANSACTION … DROP TABLE … CREATE TABLE … INSERT … DROP TABLE … RENAME … COMMIT` path executes every time migrations run.

## Steps to Reproduce

1. Apply migration 7 on a fresh database (creates a `session_summaries` table without the `UNIQUE` constraint).
2. Shut down and restart the worker.
3. `removeSessionSummariesUniqueConstraint()` is called again.
4. `hasUniqueConstraint` is `true` (because of the PK index), so the migration recreates the table unnecessarily.

## Fix

Filter to only non-primary-key unique indexes by also checking `origin`:

```ts
// Before (buggy):
const hasUniqueConstraint = summariesIndexes.some(idx => idx.unique === 1);

// After (correct):
const hasUniqueConstraint = summariesIndexes.some(
  idx => idx.unique === 1 && idx.origin !== 'pk'
);
```

With this fix, once the `UNIQUE` constraint index (origin `'u'`) is gone the early-return fires correctly, and the table-recreation DDL is skipped on subsequent startups.

## Impact

- Extra DB work (DROP + CREATE + INSERT + RENAME) on every worker startup for databases that have already applied migration 7.
- Potential data-integrity risk if a crash occurs mid-transaction during one of the spurious re-creation runs.

## References

- [`src/services/sqlite/migrations/runner.ts` — `removeSessionSummariesUniqueConstraint`](https://github.com/thedotmack/claude-mem/blob/main/src/services/sqlite/migrations/runner.ts)
- [`src/types/database.ts` — `IndexInfo` interface](https://github.com/thedotmack/claude-mem/blob/main/src/types/database.ts)
- [SQLite PRAGMA index_list documentation](https://www.sqlite.org/pragma.html#pragma_index_list)

kuishou68:
## Problem

`removeSessionSummariesUniqueConstraint()` (migration 7) checks whether the UNIQUE constraint still exists on `session_summaries` with:

```ts
const hasUniqueConstraint = summariesIndexes.some(idx => idx.unique === 1);
```

SQLite's `PRAGMA index_list` returns every index on the table, including the implicit index backing the `id INTEGER PRIMARY KEY AUTOINCREMENT` column. That PK index has `unique = 1` and `origin = 'pk'`, so `hasUniqueConstraint` is **always `true`**.

Consequence: the early-return branch (`if (!hasUniqueConstraint) return`) never fires, and the full DROP + CREATE + INSERT + RENAME DDL block executes on **every worker startup** after migration 7 should already be complete.

## Fix

Filter by `origin` to exclude primary-key indexes:

```ts
// Before
const hasUniqueConstraint = summariesIndexes.some(idx => idx.unique === 1);

// After
const hasUniqueConstraint = summariesIndexes.some(
  idx => idx.unique === 1 && idx.origin !== 'pk'
);
```

With this change, once the `UNIQUE` constraint index (origin `'u'`) has been removed, the check correctly returns `false` and the migration short-circuits without re-creating the table.

## Testing

Manual verification: after the fix, a database that has already undergone migration 7 will skip the table-recreation path on subsequent startups.

Closes #1749

---
Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

### DIFF
--- a/src/services/sqlite/migrations/runner.ts
+++ b/src/services/sqlite/migrations/runner.ts
@@ -189,7 +189,13 @@ export class MigrationRunner {
   private removeSessionSummariesUniqueConstraint(): void {
     // Check actual constraint state — don't rely on version tracking alone (issue #979)
     const summariesIndexes = this.db.query('PRAGMA index_list(session_summaries)').all() as IndexInfo[];
-    const hasUniqueConstraint = summariesIndexes.some(idx => idx.unique === 1);
+    // Only count non-primary-key unique indexes.
+    // PRAGMA index_list returns the PK index with unique=1 and origin='pk', so
+    // checking unique===1 alone always returns true, causing the table to be
+    // unnecessarily re-created on every startup after migration 7 already ran.
+    const hasUniqueConstraint = summariesIndexes.some(
+      idx => idx.unique === 1 && idx.origin !== 'pk'
+    );
 
     if (!hasUniqueConstraint) {
       // Already migrated (no constraint exists)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1953, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1761, opened 2026-04-13
feat: Pass mode prompts as system prompt for improved instruction adherence

aaronwong1989:
## Problem

Mode prompts (`output_format_header`, `summary_footer`, etc.) were being injected into user messages rather than passed as system prompts to the LLM.

This caused 70%+ XML tag misclassification because system-level instructions have weaker adherence when buried in user content.

### Current Architecture (Before)
```
[System: default identity]
[User: MODE PROMPTS + task data]
[User: observation tags]
```

### Expected Architecture (After)
```
[System: Mode prompts + output format requirements + role definition]
[User: Task-specific content only]
```

## Solution

Following Anthropic's context engineering best practices:

1. **Add `buildSystemPrompt()` function** in `src/sdk/prompts.ts`:
   - Extracts system-level prompts (identity, role, format instructions) from mode config
   - These are the directives that should have highest attention weight

2. **SDK Agent** (`SDKAgent.ts`):
   - Pass `systemPrompt` to the SDK `query()` options
   - User messages only contain task-specific content

3. **Gemini Agent** (`GeminiAgent.ts`):
   - Add `systemInstruction` field to API request body

4. **OpenRouter Agent** (`OpenRouterAgent.ts`):
   - Prepend system message to messages array

5. **Modified prompt builders**:
   - `buildInitPrompt()` and `buildContinuationPrompt()` now accept `includeSystemPrompts` parameter
   - When `false`, excludes system prompts from user messages (avoiding duplication)

## Impact

- Improved format instruction adherence (system prompts have higher attention weight)
- Reduced XML tag misclassification
- Better separation of concerns: system-level directives vs. task content

## Files Changed

| File | Change |
|------|--------|
| `src/sdk/prompts.ts` | Add `buildSystemPrompt()`, modify prompt builders |
| `src/services/worker/SDKAgent.ts` | Pass `systemPrompt` to `query()` options |
| `src/services/worker/GeminiAgent.ts` | Add `systemInstruction` to API calls |
| `src/services/worker/OpenRouterAgent.ts` | Prepend system message to API calls |

## Testing

The change maintains backward compatibility:
- If no system prompt is provided, behavior is identical to before
- The `includeSystemPrompts` parameter defaults to `true` for existing callers

### DIFF
--- a/src/sdk/prompts.ts
+++ b/src/sdk/prompts.ts
@@ -24,15 +24,23 @@ export interface SDKSession {
 }
 
 /**
- * Build initial prompt to initialize the SDK agent
+ * Build system prompt from mode configuration.
+ * This contains all the system-level instructions that should be passed as
+ * the SDK's systemPrompt option rather than buried in user messages.
+ *
+ * Following Anthropic's context engineering best practices:
+ * - Minimal: Only system-level directives
+ * - Structured: Clear section delineation
+ * - Instruction priority: Format requirements in system prompt
+ *
+ * The system prompt includes:
+ * - Identity and role definition
+ * - Output format instructions
+ * - XML schema with placeholders
+ * - Examples and footer guidance
  */
-export function buildInitPrompt(project: string, sessionId: string, userPrompt: string, mode: ModeConfig): string {
+export function buildSystemPrompt(mode: ModeConfig): string {
   return `${mode.prompts.system_identity}
-
-<observed_from_primary_session>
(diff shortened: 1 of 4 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1891, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1771, opened 2026-04-13
fix: exclude primary key index from unique constraint check in migration 7

### LINKED ISSUE #1749: Migration 7: `removeSessionSummariesUniqueConstraint` always re-creates the table because primary-key index is counted as a unique index

derjochenmeyer:
Fixes #1749.

## Problem

`PRAGMA index_list` returns **all** indexes on a table, including the one backing the `PRIMARY KEY` which always has `unique = 1`. Migration 7's early-return guard only checked `idx.unique === 1`, so `hasUniqueConstraint` was always `true`.

Result: the full `BEGIN TRANSACTION … DROP TABLE … CREATE TABLE … INSERT … RENAME … COMMIT` sequence ran on **every** worker startup after migration 7 should have completed.

## Fix

One character change — exclude primary key indexes by also checking `idx.origin !== 'pk'`:

```ts
// Before:
const hasUniqueConstraint = summariesIndexes.some(idx => idx.unique === 1);

// After:
const hasUniqueConstraint = summariesIndexes.some(idx => idx.unique === 1 && idx.origin !== 'pk');
```

The `origin` field (`'c'` = user-created, `'u'` = unique constraint, `'pk'` = primary key) is already present on the existing `IndexInfo` interface — no type changes needed.

## Impact

- Eliminates unnecessary table recreation on every startup for all databases that have already applied migration 7
- Removes the crash-window risk from a spurious mid-transaction failure during what should be a no-op


derjochenmeyer:
Good catch. Applied the same `origin !== 'pk'` guard to all three code paths:

- `SessionStore.ts:220` ✅
- `plugin/scripts/context-generator.cjs` ✅ (minified, patched directly)
- `runner.ts:192` ✅ (already fixed)

All three paths now consistently exclude primary-key indexes when detecting the unique constraint.

derjochenmeyer:
@coderabbitai review

PR #1257, opened 2026-02-27
feat: temporal scoring, staleness tracking, and drift detection

NoRain211:
## Summary

Adds Ember MCP-inspired features for memory relevance management:

- **Temporal decay scoring** — observations decay in search relevance over time with importance-scaled half-life (18d–180d). Access tracking boosts frequently-retrieved memories.
- **Stale filtering** — stale observations excluded from search by default (opt-in via `include_stale` param)
- **`contradict` tool** — mark a memory stale and record a correction linked to it
- **`set_importance` tool** — adjust importance score (1–10) to control decay rate
- **`drift_check` tool** — SQL cluster analysis via `json_each(concepts)` to detect concept areas where old memories have gone stale or been superseded
- **SQLite FTS5 fallback** — restores text search when ChromaDB is unavailable (previously returned empty results)

### Schema migration (v24)

Adds 5 columns to `observations`: `last_accessed_at`, `access_count`, `importance`, `is_stale`, `corrected_by_id`. Idempotent with per-column existence checks before each ALTER to handle partial migration failures.

### Files changed

| File | Change |
|------|--------|
| `src/services/sqlite/migrations/runner.ts` | Migration v24 |
| `src/services/sqlite/types.ts` | 5 new fields on `ObservationRow` interface |
| `src/types/database.ts` | 5 new optional fields on `ObservationRecord` |
| `src/services/sqlite/SessionSearch.ts` | FTS5 fallback, `rankByTemporalScore` (importance-scaled), `updateAccessTracking`, `detectDrift` |
| `src/services/sqlite/SessionStore.ts` | `markObservationStale`, `setObservationImportance` |
| `src/services/worker/search/SearchOrchestrator.ts` | Apply temporal scoring + stale filter after search |
| `src/services/worker/http/routes/MemoryRoutes.ts` | Routes for contradict, importance, drift-check |
| `src/services/worker/http/routes/DataRoutes.ts` | Access tracking on batch observations fetch |
| `src/servers/mcp-server.ts` | 3 new MCP tools: `contradict`, `set_importance`, `drift_check` |
| `plugin/skills/mem-search/SKILL.md` | Documentation for new tools |

## Test plan

- [x] Fresh install: migration v24 runs, all 5 columns added
- [x] Existing DB: migration skips columns that already exist (per-column idempotent)
- [x] `contradict` tool: marks observation stale, creates correction, links via `corrected_by_id`
- [x] `set_importance` tool: clamps to 1–10, updates observation
[...]

PR #1774, opened 2026-04-13
feat: add temporal scoring, staleness tracking, and drift detection

NoRain211:
## Summary
- revive and replay the temporal scoring and drift detection work from #1257 onto current main
- add importance, staleness, contradiction, and drift-check support to the SQLite memory model and MCP surface
- preserve SQLite FTS fallback behavior and the related search/server coverage updates

## Notes
- supersedes #1257, which GitHub auto-closed as stale
- includes the follow-up review fixes that were already added on the original branch

## Verification
- npm run build
- npm run test:sqlite
- npm run test:search
- npm run test:server


thedotmack:
Closed during the April 2026 backlog cleanup. The feature ask is now tracked in #1964, which is the single canonical issue for prioritization. Thanks for the suggestion — your context is captured there.

PR #1776, opened 2026-04-13
fix: coerce corpus route filters

suyua9:
## Summary
- coerce `types`, `concepts`, and `files` in `POST /api/corpus` when clients send JSON-encoded or comma-separated strings
- coerce string `limit` values to positive integers before passing filters into `CorpusBuilder`
- add focused CorpusRoutes regression tests for native, JSON string, comma-separated, and invalid payload shapes

## Why
`CorpusBuilder` assumes corpus filter arrays and calls `.join(',')` on `types`, `concepts`, and `files`. If an MCP/HTTP client sends string-encoded arrays, the route currently passes those strings through and the builder can throw at runtime. `DataRoutes` already handles the same MCP-client coercion shape for batch endpoints, so this keeps the boundary handling consistent.

## Tests
- `bun test tests/worker/http/routes/corpus-routes-coercion.test.ts`
- `bun test tests/worker/http/routes/data-routes-coercion.test.ts tests/worker/http/routes/corpus-routes-coercion.test.ts`

`tsc --noEmit` was not runnable in this clean worktree because `node_modules/@types/node` is not installed locally (`Cannot find type definition file for 'node'`).

### DIFF
--- a/src/services/worker/http/routes/CorpusRoutes.ts
+++ b/src/services/worker/http/routes/CorpusRoutes.ts
@@ -12,6 +12,8 @@ import { CorpusBuilder } from '../../knowledge/CorpusBuilder.js';
 import { KnowledgeAgent } from '../../knowledge/KnowledgeAgent.js';
 import type { CorpusFilter } from '../../knowledge/types.js';
 
+const ALLOWED_CORPUS_TYPES = new Set(['decision', 'bugfix', 'feature', 'refactor', 'discovery', 'change']);
+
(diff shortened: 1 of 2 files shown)


cuyua9:
Follow-up update: I addressed the CodeRabbit regression-test suggestion in the latest push (730f81e). The new test covers an unsupported corpus type value (`types: ['typo']`) and asserts that the route returns 400 before calling `CorpusBuilder`.\n\nVerification: `bun test tests/worker/http/routes/corpus-routes-coercion.test.ts` passes locally.

PR #1780, opened 2026-04-13
Fix npx search and default Codex context to workspace-local AGENTS

enma998:
## Summary
- switch `npx claude-mem search` from `/api/search?q=...` to `/api/search?query=...`
- remove the hardcoded Codex context path `~/.codex/AGENTS.md`
- rely on the existing workspace-local fallback `<workspace>/AGENTS.md` for Codex context injection
- clean up any legacy global `<claude-mem-context>` block during Codex install/uninstall flows
- add regression tests for both the search query parameter and the workspace-local Codex context behavior

## Why
- the worker search route expects `query`, while the CLI was sending `q`, which breaks `npx claude-mem search`
- Codex transcript processing already supports a workspace fallback for AGENTS context
- hardcoding `~/.codex/AGENTS.md` makes Codex context effectively last-writer-wins across repos, causing cross-project bleed
- using the workspace-local default keeps Codex context scoped to the active repo without adding a new orchestration layer

## Verification
- `bun test tests/codex-workspace-context.test.ts tests/npx-search-query-param.test.ts tests/install-non-tty.test.ts`
- `npm run build`

## Notes
- this branch now contains both the CLI search fix and the Codex workspace-context isolation fix
- the Codex change is intentionally minimal and uses the existing transcript processor fallback instead of redesigning retrieval or storage


enma998:
Added a follow-up patch on this branch for Codex context isolation.

Root cause: the Codex sample config hardcoded ~/.codex/AGENTS.md, so transcript updates behaved like last-writer-wins across repos.

This patch removes the hardcoded global Codex path, relies on the existing workspace fallback (<cwd>/AGENTS.md), and cleans up any legacy global <claude-mem-context> block during install. Also added a regression test covering the workspace-local default.

enma998:
Approved and checks are passing. This branch is ready for maintainer squash-merge.

It includes:
- the `npx claude-mem search` `q` -> `query` fix
- the Codex workspace-local AGENTS default to prevent cross-repo context bleed

PR #1795, opened 2026-04-14
feat: add CLAUDE_MEM_OPENROUTER_BASE_URL setting for local/compatible servers

shikaku2:
## Summary

- Adds `CLAUDE_MEM_OPENROUTER_BASE_URL` setting to `SettingsDefaultsManager` (defaults to `https://openrouter.ai/api/v1/chat/completions`)
- Passes the base URL through `getOpenRouterConfig()` and `queryOpenRouterMultiTurn()` instead of using a hardcoded constant
- Rebuilt distribution files

## Motivation

Users with local OpenAI-compatible inference servers (llama.cpp, Ollama, LM Studio, etc.) currently have no way to use the `openrouter` provider without modifying the compiled bundle. This setting lets them point claude-mem at their local server without touching any code.

## Usage

In `~/.claude-mem/settings.json`:
```json
{
  "CLAUDE_MEM_PROVIDER": "openrouter",
  "CLAUDE_MEM_OPENROUTER_API_KEY": "local",
  "CLAUDE_MEM_OPENROUTER_BASE_URL": "http://localhost:8080/v1/chat/completions",
  "CLAUDE_MEM_OPENROUTER_MODEL": "your-model-name"
}
```

Existing users are unaffected — the setting falls back to the original OpenRouter URL when not set.


thedotmack:
Closed during the April 2026 backlog cleanup. The feature ask is now tracked in #1987, which is the single canonical issue for prioritization. Thanks for the suggestion — your context is captured there.

Issue #1791, opened 2026-04-14
[Feature] 2-Layer Dedup Pipeline — reduce redundancy in search results


im47cn:
## Problem

Search results contain redundant entries: same file across observations, near-identical tool outputs, same concept described differently. Currently dedup only at ingestion (30s hash window). Search results have **no dedup**.

## Inspiration: GBrain's 4-Layer Dedup

[garrytan/gbrain](https://github.com/garrytan/gbrain) dedup pipeline (`src/core/search/dedup.ts`, 107 lines):

```
L1 (Source):    Top 3 per source
L2 (Text):      Jaccard > 0.85 = duplicate
L3 (Diversity): No type > 60% of results
L4 (Cap):       Max 2 per source in final
```

### Overview

| Aspect | Existing claude-mem | GBrain | New Implementation |
|--------|:---:|:---:|:---:|
| **When** | Write-time | Search-time | Search-time |
| **What** | Exact content duplicates | Near-duplicate chunks | Result set diversity |
| **How** | Content hash + time window | 4-layer pipeline | 2-layer pipeline |
| **Goal** | Prevent storing duplicates | Remove redundant search hits | Balance diversity & recall |

### Existing claude-mem Dedup (Write-Time)

Claude-mem already has **3 layers of write-time dedup** that prevent duplicate observations from being stored:

| Layer | Location | Mechanism | Scope |
|-------|----------|-----------|-------|
| **Ingestion dedup** | `observations/store.ts` | `computeObservationContentHash()` — SHA-256 of `(session_id, title, text)`. Checks `findDuplicateObservation()` within 30s window (`DEDUP_WINDOW_MS = 30_000`). | Same session, same content, within 30s |
| **Import dedup** | `import/bulk.ts` | `content_session_id + title + epoch` uniqueness check during bulk import. Skips rows that match existing observations. | Cross-session, title+time collision |
| **ChromaSync dedup** | `ChromaSyncService.ts` | Delete-then-add for duplicate IDs when syncing observations to Chroma vector store. | Vector store consistency |

**Strengths:**
- Zero runtime cost at search time
- Prevents storage bloat
- Exact-match: no false positives

**Limitations:**
- Only catches **exact** duplicates (same hash)
- 30s window means identical content from different time periods is stored twice
- Cannot address **semantic** redundancy (same concept, different wording)
- Cannot address **result set imbalance** (one session/project dominating search results)

### GBrain Dedup (Search-Time, 4-Layer)

[...]

PR #1799, opened 2026-04-14
Issue #1791: 2-layer dedup pipeline for search results

im47cn:
## Summary

- Implement data-driven 2-layer search-time dedup pipeline (simplified from GBrain's 4-layer design)
- Complements existing write-time content-hash dedup — addresses result set imbalance that write-time dedup cannot
- Integrate into `SearchManager.search()` with configurable settings
- Add 15 unit tests with precise layer interaction assertions

## 3-Way Dedup Comparison

| Aspect | Existing (Write-Time) | GBrain (Search-Time 4L) | This PR (Search-Time 2L) |
|--------|:---:|:---:|:---:|
| **When** | Ingestion | Search | Search |
| **What** | Exact content duplicates | Near-duplicate chunks | Result set diversity |
| **How** | SHA-256 hash + 30s window | source top-3 → Jaccard → type diversity → source cap | project diversity (60%) → session cap (8) |
| **Goal** | Prevent storing duplicates | Remove redundant search hits | Balance diversity & recall |

### How They Work Together

```
Observation created
    │
    ▼
[Write-Time Dedup]  ← Existing: content hash + 30s window
    │                  Prevents exact duplicates from being stored
    ▼
Observation stored in SQLite + Chroma
    │
    ▼
User searches...
    │
    ▼
[Search-Time Dedup]  ← THIS PR: project diversity + session cap
    │                   Ensures balanced, diverse result sets
    ▼
Formatted results returned
```

### Why Not Use GBrain's Full 4-Layer Design?

GBrain deduplicates **web page chunks** (genuinely redundant). Claude-mem deduplicates **session observations** (different workflow steps). This invalidates 2 of GBrain's 4 layers:

| GBrain Layer | Why Removed | Evidence |
|-------------|-------------|----------|
| L1: Source top-3 | Redundant with session cap — both limit per-source count, L1 strictly more aggressive | 38% drop, 9.1% high-score losses; identical sensitivity curve to L4 |
| L2: Jaccard | Write-time hash dedup already eliminates exact duplicates; structured observations are naturally diverse | 0% effectiveness at 0.85; still 0% at 0.6 |

## Empirical Validation

Tested on 3,700+ real observations (14 queries × 50 results):

| Metric | GBrain-style (4L, cap=2) | This PR (2L, cap=8) | Improvement |
|--------|:---:|:---:|:---:|
| Result drop rate | 51.7% | **12.3%** | 4.2× less aggressive |
| Token savings | 53.6% | **12.6%** | Modest but real |
| **Fact loss rate** | **53.0%** | **12.1%** | **4.4× less damage** |
| Concept loss rate | 16.2% | 5.7% | 2.8× less damage |
| **High-score drops** | **43 (13%)** | **0 (0%)** | **Eliminated** |

Key finding: token efficiency ~63 tok/fact across all configs — filtering is indiscriminate truncation, not intelligent dedup. Only safe approach: **filter conservatively**.

## Changes

| File | Change |
|------|--------|
| `src/services/worker/search/dedup.ts` | 2-layer pipeline: project diversity (60%) + session cap (default 8) |
| `src/services/worker/search/index.ts` | Export `dedupResults`, `jaccardSimilarity`, types |
| `src/services/worker/SearchManager.ts` | Integrate dedup after observation hydration |
| `src/shared/SettingsDefaultsManager.ts` | 3 config keys: `SEARCH_DEDUP_ENABLED`, `MAX_PROJECT_RATIO`, `MAX_PER_SESSION` |
| `tests/worker/search/dedup.test.ts` | 15 tests: Jaccard utility, both layers, combined, edge cases |

## Configuration

```json
{
  "CLAUDE_MEM_SEARCH_DEDUP_ENABLED": "true",
  "CLAUDE_MEM_SEARCH_DEDUP_MAX_PROJECT_RATIO": "0.6",
  "CLAUDE_MEM_SEARCH_DEDUP_MAX_PER_SESSION": "8"
}
```

Session cap sensitivity:
```
cap=2:  50.7% drop, 52.5% fact loss  (too aggressive)
cap=5:  22.0% drop, 23.2% fact loss
cap=8:   9.8% drop, 10.0% fact loss  (default)
cap=10:  5.7% drop,  5.8% fact loss
```

## Test plan

- [x] `bun test tests/worker/search/dedup.test.ts` — 15 pass, 0 fail
- [x] TypeScript compilation passes (no new errors)
- [ ] Manual: search queries return deduped results with debug log showing drop counts
- [ ] Manual: set `CLAUDE_MEM_SEARCH_DEDUP_ENABLED=false` to verify bypass

### DIFF
--- a/src/services/worker/SearchManager.ts
+++ b/src/services/worker/SearchManager.ts
@@ -28,9 +28,11 @@ import { ModeManager } from '../domain/ModeManager.js';
 import {
   SearchOrchestrator,
   TimelineBuilder,
-  SEARCH_CONSTANTS
+  SEARCH_CONSTANTS,
+  dedupResults
 } from './search/index.js';
-import type { TimelineData } from './search/index.js';
+import type { TimelineData, DedupOptions } from './search/index.js';
+import { SettingsDefaultsManager } from '../../shared/SettingsDefaultsManager.js';
 
 export class SearchManager {
   private orchestrator: SearchOrchestrator;
@@ -260,6 +262,23 @@ export class SearchManager {
       observations = [];
       sessions = [];
       prompts = [];
+    }
+
+    // Apply dedup pipeline to observations (project diversity + session cap)
+    if (observations.length > 1 && SettingsDefaultsManager.getBool('CLAUDE_MEM_SEARCH_DEDUP_ENABLED')) {
+      const beforeCount = observations.length;
+      const dedupOpts: DedupOptions = {
+        maxProjectRatio: parseFloat(SettingsDefaultsManager.get('CLAUDE_MEM_SEARCH_DEDUP_MAX_PROJECT_RATIO')),
+        maxPerSession: SettingsDefaultsManager.getInt('CLAUDE_MEM_SEARCH_DEDUP_MAX_PER_SESSION'),
+      };
+      observations = dedupResults(observations, dedupOpts);
+      if (observations.length < beforeCount) {
+        logger.debug('SEARCH', 'Dedup pipeline reduced observations', {
+          before: beforeCount,
+          after: observations.length,
+          dropped: beforeCount - observations.length,
+        });
+      }
     }
 
     const totalResults = observations.length + sessions.length + prompts.length;
--- a/src/services/worker/search/SearchOrchestrator.ts
+++ b/src/services/worker/search/SearchOrchestrator.ts
@@ -31,6 +31,9 @@ import type {
   ObservationSearchResult
 } from './types.js';
 import { logger } from '../../../utils/logger.js';
+import { dedupResults } from './dedup.js';
+// NOTE: rrfFusion is available in ./rrf-fusion.js for future use when
+// the strategy pattern is refactored to support parallel FTS5+Chroma execution.
 
 /**
  * Normalized parameters from URL-friendly format
@@ -76,15 +79,17 @@ export class SearchOrchestrator {
   }
 
   /**
-   * Execute search with fallback logic
+   * Execute search with dedup post-processing and fallback logic
    */
   private async executeWithFallback(
     options: NormalizedParams
   ): Promise<StrategySearchResult> {
-    // PATH 1: FILTER-ONLY (no query text) - Use SQLite
+    // PATH 1: FILTER-ONLY (no query text) - Use SQLite, apply dedup
     if (!options.query) {
       logger.debug('SEARCH', 'Orchestrator: Filter-only query, using SQLite', {});
-      return await this.sqliteStrategy.search(options);
+      const result = await this.sqliteStrategy.search(options);
+      result.results.observations = dedupResults(result.results.observations);
+      return result;
     }
 
     // PATH 2: CHROMA SEMANTIC SEARCH (query text + Chroma available)
@@ -92,8 +97,8 @@ export class SearchOrchestrator {
       logger.debug('SEARCH', 'Orchestrator: Using Chroma semantic search', {});
       const result = await this.chromaStrategy.search(options);
 
-      // If Chroma succeeded (even with 0 results), return
       if (result.usedChroma) {
+        result.results.observations = dedupResults(result.results.observations);
         return result;
       }
 
@@ -103,6 +108,7 @@ export class SearchOrchestrator {
         ...options,
         query: undefined // Remove query for SQLite fallback
       });
+      fallbackResult.results.observations = dedupResults(fallbackResult.results.observations);
 
       return {
         ...fallbackResult,
--- a/src/services/worker/search/dedup.ts
+++ b/src/services/worker/search/dedup.ts
@@ -1,0 +1,146 @@
+/**
+ * 2-Layer Dedup Pipeline for search results
+ *
+ * Simplified from 4-layer GBrain adaptation based on real-data analysis
+ * (see spec 005-dedup-pipeline.md for original design):
+ * - L1 (session top-3) and L4 (session cap) were redundant -> merged into single session cap
+ * - L2 (Jaccard similarity) had zero hits on 3700+ observations (even at 0.6 threshold) -> removed
+ * - L3 (project diversity) retained as precision filter with low collateral damage
+ *
+ * Empirical results (14 queries x 50 results on real dataset):
+ * - Old 4-layer pipeline: 51.7% drop rate, 53.0% fact loss (near-random truncation)
+ * - New 2-layer (cap=5):  23.6% drop rate, 24.3% fact loss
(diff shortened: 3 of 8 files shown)


xkonjin (commented):
**Review: clean implementation, one config concern**

The 2-layer dedup pipeline is well reasoned from the empirical analysis. Code is readable and the tests cover project diversity, session caps, edge cases, and combined layer interactions.

Security / stability notes:
-  on  could produce  if the env var is set to an invalid string. The  would then silently propagate through  and likely break filtering. Consider validating the parsed value and falling back to the default (0.6).
- Same for  — if it returns , the cap logic will behave unexpectedly. A defensive fallback to  would be safer.

Test coverage gap:  calls  without passing , so it relies on defaults. A quick integration test verifying that the orchestrator actually trims a large result set would guard against future regressions where someone accidentally removes the dedup call.

The  module is introduced but unused. If it is not wired in yet, add a comment or TODO explaining when it will be integrated; otherwise it risks bit-rot.

xkonjin (commented):
**Review: clean implementation, one config concern**

The 2-layer dedup pipeline is well reasoned from the empirical analysis. Code is readable and the tests cover project diversity, session caps, edge cases, and combined layer interactions.

Security / stability notes:
- parseFloat on CLAUDE_MEM_SEARCH_DEDUP_MAX_PROJECT_RATIO could produce NaN if the env var is set to an invalid string. The NaN would then silently propagate through Math.max(1, Math.ceil(... * NaN)) and likely break filtering. Consider validating the parsed value and falling back to the default (0.6).
- Same for SettingsDefaultsManager.getInt('CLAUDE_MEM_SEARCH_DEDUP_MAX_PER_SESSION') — if it returns NaN, the cap logic will behave unexpectedly. A defensive fallback to 8 would be safer.

Test coverage gap: SearchOrchestrator.ts calls dedupResults without passing DedupOptions, so it relies on defaults. A quick integration test verifying that the orchestrator actually trims a large result set would guard against future regressions where someone accidentally removes the dedup call.

The rrf-fusion.ts module is introduced but unused. If it is not wired in yet, add a comment or TODO explaining when it will be integrated; otherwise it risks bit-rot.

xkonjin (commented):
Review: well-reasoned 2-layer dedup pipeline with a few stability gaps

The empirical justification for collapsing the 4-layer design down to 2 layers is strong, and the measured drop/fact-loss rates make the case convincingly. The code in dedup.ts is clean and readable.

Bugs / stability issues:
- SearchManager.ts parses CLAUDE_MEM_SEARCH_DEDUP_MAX_PROJECT_RATIO with parseFloat and reads CLAUDE_MEM_SEARCH_DEDUP_MAX_PER_SESSION via getInt. If either env var is malformed, parseFloat returns NaN and getInt may also return NaN. NaN propagates into Math.max(1, Math.ceil(results.length * NaN)), which yields NaN, and the project-diversity layer will silently drop almost everything. You need defensive validation with safe fallbacks (0.6 and 8) before passing values into dedupResults.
- SearchOrchestrator.ts applies dedupResults unconditionally in three places and without passing DedupOptions. This bypasses the feature flag (CLAUDE_MEM_SEARCH_DEDUP_ENABLED) and ignores any custom cap/ratio settings. It can also double-dedup when SearchManager later runs its own dedup pass. Make the orchestrator respect the flag and pass through the same options.
- rrf-fusion.ts does not validate opts.k. If someone passes k: 0 or a negative number, the RRF score calculation divides by zero or a negative denominator, producing Infinity or invalid scores. Add a guard: if (!Number.isFinite(k) || k <= 0) throw new Error("RRF k must be a positive finite number").

Test coverage gap:
- dedup.test.ts line 55 computes the project cap using deduped.length instead of the input size (results.length). The + 1 slack masks the imprecision. Fix the assertion to use results.length.
- There are no boundary tests for invalid k values in rrf-fusion.test.ts. Add cases for k = 0 and k < 0.
- The rrf-fusion.ts module is introduced but not wired into any production path. If it is intentionally future-looking, add a TODO comment explaining when it will be integrated; otherwise it risks bit-rot.

Please address the NaN validation and the orchestrator double-dedup before merge.

xkonjin (commented):
Nice, targeted dedup pipeline. A few issues to address before merging:

1. **Critical: NaN risk drops all results.** In  you pass  directly to . If the setting is missing or malformed,  returns , which propagates into  → . The  check then evaluates to  for every item, so *all* observations get silently dropped. Please validate the parsed float and fall back to a safe default (e.g. ) when it is , zero, or negative.

2. **Missing unit tests.**  contains core ranking logic with empirical claims (drop rates, fact loss). It should have tests covering: (a) default behavior, (b) project diversity edge cases, (c) session cap behavior, and (d) the NaN/regression case above.

3. **Dead code.**  is noted as unused with "zero hits on real data". If it's truly not needed, remove it to reduce maintenance surface. If you want to keep it as a utility, add a test and document when it should be used.

4. **Settings flag check.**  is fine, but make sure the default is  (or ) intentionally and documented so users aren't surprised by a behavior change on upgrade.

Overall direction is solid — just harden the config parsing and add tests.

xkonjin (commented):
Nice, targeted dedup pipeline. A few issues to address before merging:

1. Critical: NaN risk drops all results. In SearchManager.ts you pass parseFloat(SettingsDefaultsManager.get('CLAUDE_MEM_SEARCH_DEDUP_MAX_PROJECT_RATIO')) directly to dedupResults. If the setting is missing or malformed, parseFloat returns NaN, which propagates into Math.ceil(results.length * NaN) -> NaN. The count < maxPerProject check then evaluates to false for every item, so all observations get silently dropped. Please validate the parsed float and fall back to a safe default (e.g. 0.6) when it is NaN, zero, or negative.

2. Missing unit tests. dedup.ts contains core ranking logic with empirical claims (drop rates, fact loss). It should have tests covering: (a) default behavior, (b) project diversity edge cases, (c) session cap behavior, and (d) the NaN/regression case above.

3. Dead code. jaccardSimilarity is noted as unused with "zero hits on real data". If it's truly not needed, remove it to reduce maintenance surface. If you want to keep it as a utility, add a test and document when it should be used.

4. Settings flag check. SettingsDefaultsManager.getBool('CLAUDE_MEM_SEARCH_DEDUP_ENABLED') is fine, but make sure the default is true (or false) intentionally and documented so users aren't surprised by a behavior change on upgrade.

Overall direction is solid — just harden the config parsing and add tests.

thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1915, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1804, opened 2026-04-14
fix: support gemini transcript parsing in summarize hook

themactep:
This PR adds support for gemini transcript parsing in the summarize hook, detecting platform source and handling both JSON and JSONL formats with cross-platform role mappings.


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1909, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1818, opened 2026-04-14
feat: add Claude Code plugin lifecycle hooks

MoeJaberr:
## Summary

- Adds `.claude-plugin/hooks/hooks.json` so claude-mem activates **automatically** when the plugin is enabled in Claude Code — no manual `/mem-search` invocation needed, mirroring the pattern used by `context-mode`
- Five lifecycle hook scripts (Node.js ESM) call the worker HTTP API on port 37777

## What changed

| Hook | File | Behaviour |
|------|------|-----------|
| `SessionStart` | `session-start.mjs` | Ensures worker is running (auto-starts via bun if needed) → `POST /api/sessions/init` → `POST /api/context/inject` → injects past memory as a system reminder |
| `UserPromptSubmit` | `user-prompt.mjs` | `POST /api/prompts` to capture each user message |
| `PostToolUse` | `post-tool-use.mjs` | `POST /api/observations` for significant tool calls; skips tools listed in `CLAUDE_MEM_SKIP_TOOLS` |
| `Stop` | `stop.mjs` | `POST /api/sessions/complete` + `POST /api/sessions/summarize` |
| `SessionEnd` | `stop.mjs` (shared) | Same as Stop, covers both shutdown paths |

## Why

Claude Code plugin hooks (`hooks/hooks.json` inside `.claude-plugin/`) are read automatically for any enabled plugin — the same mechanism `context-mode` uses. Previously claude-mem had no hooks file, so users had to manually invoke `/mem-search` each session. With this change, past context is injected automatically at session start and observations are captured throughout without any user action.

## Notes for reviewer

- All hooks exit 0 on any error — they are non-blocking and will never interrupt Claude Code if the worker is down
- Worker auto-start logic mirrors `worker-cli.js`: finds bun via `~/.bun/bin/bun` or PATH, then waits for `/api/health` readiness
- `CLAUDE_MEM_WORKER_PORT` env var is respected (defaults to `37777`)
- Scripts are plain Node.js ESM with no npm dependencies

## Test plan

- [ ] Enable claude-mem plugin in Claude Code settings
- [ ] Restart Claude Code — worker should start automatically and past context should appear in the first system reminder
- [ ] Run a few tool calls — verify observations appear in the memory viewer at `http://localhost:37777`
- [ ] End the session — verify a summary is generated

### DIFF
--- a/.claude-plugin/hooks/hooks.json
+++ b/.claude-plugin/hooks/hooks.json
@@ -1,0 +1,65 @@
+{
+  "description": "claude-mem lifecycle hooks — SessionStart worker init + context injection, UserPromptSubmit capture, PostToolUse observation, Stop/SessionEnd session complete",
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node ${CLAUDE_PLUGIN_ROOT}/hooks/session-start.mjs",
+            "timeout": 20000
+          }
+        ]
+      }
+    ],
+    "UserPromptSubmit": [
+      {
+        "matcher": "",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node ${CLAUDE_PLUGIN_ROOT}/hooks/user-prompt.mjs",
+            "timeout": 5000
+          }
+        ]
+      }
+    ],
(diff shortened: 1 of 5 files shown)


thedotmack:

Issue #1500, opened 2026-03-27
Regression: worktree observations still tagged with worktree dir name in v10.6.2


knutwannheden:
## Description

The worktree project name bug reported in #1081 (fixed in v9.x) and #1317 (fixed in v10.5.5) has regressed. On v10.6.2, observations are still stored under the worktree directory name instead of the parent repository name, and every new worktree session shows "No previous sessions found."

## Environment

- claude-mem version: **10.6.2** (installed 2026-03-24)
- Platform: macOS (Darwin 25.4.0)
- Claude Code CLI

## Worktree layout

My worktrees live **inside** the main repo under a `.worktrees/` directory (excluded via `.git/info/exclude`):

```
/Users/me/git/myproject/                   ← main repo (git toplevel)
├── .git/                                  ← real .git directory
│   ├── info/exclude                       ← contains "/.worktrees/"
│   └── worktrees/
│       ├── noble-hare/                    ← git worktree metadata
│       ├── jolly-condor/
│       └── ...
├── .worktrees/                            ← excluded from git
│   ├── noble-hare/                        ← worktree checkout (cwd when using Claude Code)
│   │   └── .git                           ← file: "gitdir: /Users/me/git/myproject/.git/worktrees/noble-hare"
│   ├── jolly-condor/
│   └── ...
├── src/
└── ...
```

Key details:
- Worktrees are created inside the repo: `git worktree add .worktrees/noble-hare`
- `.worktrees/` is in `.git/info/exclude` so it doesn't show up in `git status`
- The `.git` file in each worktree correctly points to the main repo's `.git/worktrees/<name>`
- `git rev-parse --git-common-dir` from any worktree correctly returns the main repo's `.git`

## Evidence

Every distinct project name in my database is a worktree name — the parent repo name `rewrite` never appears:

```sql
sqlite> SELECT DISTINCT project FROM observations ORDER BY project;
-- Returns: csharp-typed-capture-generics, deft-marmot, noble-iguana, ...
-- Never: rewrite
```

Every new worktree session starts with "No previous sessions found" because it queries for observations where `project = '<new-worktree-name>'`.

## Root cause (still)

The `context-generator.cjs` function that determines the project name still falls through to `path.basename(cwd)` in certain code paths. From the minified source in v10.6.2:

```js
function fe(r) {
  if (!r || r.trim() === "") return "unknown-project";
  let e = path.basename(r);
  // ...
  return e;
}
```

[...]

Issue #1317, opened 2026-03-09
Search API and observation storage ignore worktree parent project name


sinterqlaas2:
## Summary

Worktree support was added in v9.0 for context injection at session start, but the **search API** and **observation storage** still use `basename(cwd)` instead of the parent repository project name. This means observations stored in different worktrees are invisible to each other via the search tool.

## Environment

- claude-mem version: 10.5.5
- Platform: macOS (Darwin 25.3.0)
- Git worktree setup: main repo at `yobert-v1`, worktree at `melbourne-v3`

## Expected behavior

When working in a git worktree, observations should be stored under (or at least queryable by) the **parent repository project name**, so that all worktrees share the same observation pool — consistent with the documented worktree support.

### Context injection (works correctly)
The `SessionStart` hook uses `F6()` which returns `allProjects: [parentProjectName, worktreeName]` and passes both to the context API. This correctly merges observations from parent and worktree.

### Observation storage (broken)
The `session-init` hook (`mk`) uses `fp(n)` = `basename(cwd)` for the `project` field. In a worktree named `melbourne-v3`, observations are stored with `project = "melbourne-v3"` instead of `"yobert-v1"`.

### Search API (broken)
The `search()` MCP tool also defaults to `basename(process.cwd())`, so searching from worktree `melbourne-v3` only finds observations tagged `melbourne-v3`, not those from the main repo `yobert-v1` or other worktrees.

## Root cause

The worktree-aware `F6()` function exists and works, but is only used in the `SessionStart` context hook. Other code paths (`session-init`, `PostToolUse`, search API) still call `fp()` = `basename(cwd)` directly.

### Code references (worker-service.cjs, v10.5.5)

**Worktree detection (works):**
```js
// U6() - detects worktree by reading .git file
// F6() - returns { primary, parent, isWorktree, allProjects }
```

**SessionStart hook (uses F6 correctly):**
```js
// dk.execute(): 
let n = F6(r);
let a = n.allProjects.join(",");
// passes allProjects to context API ✓
```

**session-init hook (ignores worktree):**
```js
// mk.execute():
let a = fp(n); // just basename(cwd) ✗
// stores observations with project = basename(cwd)
```

**PostToolUse hook (ignores worktree):**
```js
// gk.execute():
// sends cwd to /api/sessions/observations, server uses basename
```

## Suggested fix

[...]

Issue #1081, opened 2026-02-12
Bug: Worktree observations tagged with worktree dir name instead of parent project


liqweed:
## Description

When running Claude Code in a git worktree, observations are tagged with the **worktree directory name** instead of the **parent repository name**. This means observations from the same project are scattered across multiple unrelated project names, making cross-session search ineffective.

## Environment

- claude-mem version: 9.1.1
- Claude Code on macOS (Darwin 25.2.0)
- Using git worktrees created via `git worktree add`

## Steps to Reproduce

1. Have a project at `/workspace/myproject` (git repo)
2. Create worktrees: `git worktree add ../zurich-v1 branch-a`, `git worktree add ../rio-de-janeiro branch-b`
3. Run Claude Code sessions in these worktrees
4. Query observations: `sqlite3 ~/.claude-mem/claude-mem.db "SELECT DISTINCT project FROM observations;"`

## Expected Behavior

All observations should be tagged as `myproject` (the parent repository name).

## Actual Behavior

Observations are tagged as `zurich-v1`, `rio-de-janeiro` (the worktree directory names):

```sql
sqlite> SELECT DISTINCT project FROM observations;
zurich-v1
rio-de-janeiro
algiers
hat-yai
```

## Root Cause (from code analysis)

The worktree detection logic (`_6()` in `worker-service.cjs`) correctly:
- Detects worktrees by checking if `.git` is a file
- Parses the parent repo path from `.git` gitdir reference
- Returns `{ primary: worktreeName, parent: parentRepoName, isWorktree: true, allProjects: [parent, worktree] }`

However, the **context generation function** (`H$`) and **session-init hook** both call `ip(n)` directly — which just returns `path.basename(cwd)` — instead of using `_6()` to resolve to the parent project name.

Relevant code in `H$`:
```js
let n = t?.cwd ?? process.cwd(), i = ip(n), a = t?.projects || [i]
// ip(n) returns "zurich-v1" instead of using _6(n).parent which would return "myproject"
```

Interestingly, there's another code path that DOES use `_6()` correctly:
```js
let r = t.cwd ?? process.cwd(), n = _6(r), i = wt(), a = n.allProjects.join(",")
```

## Suggested Fix

In `H$` (and the session-init hook), replace:
```js
let i = ip(n)
```
with:
```js
let projectInfo = _6(n)
let i = projectInfo.isWorktree ? projectInfo.parent : projectInfo.primary
```

## Workaround

[...]

PR #1820, opened 2026-04-14
fix: use parent project name for worktree observation writes

### LINKED ISSUE #1819: Regression: worktree observations tagged with worktree dir name in v12.1.0 (#1081 / #1500)
## Description

Observations and sessions from git worktrees are stored under `basename(cwd)` (the worktree directory name) instead of the parent repository name. This is the same bug as #1081 (fixed in v9.x via PR #1214), #1317 (fixed via PR #1298), and #1500 (fixed in v10.6.x) — it has regressed again in v12.1.0.

## Environment

- claude-mem: **12.1.0** (Claude Code plugin)
- Platform: macOS (Darwin 25.4.0)
- Claude Code CLI

## Worktree Layout

```
/Users/me/projects/my-app/                          ← main repo (project = "my-app")
├── .git/
│   └── worktrees/
│       └── modest-burnell/                         ← git worktree metadata
├── .claude/worktrees/
│   └── modest-burnell/                             ← worktree checkout (cwd during session)
│       └── .git                                    ← file: "gitdir: .../.git/worktrees/modest-burnell"
```

## Steps to Reproduce

1. Have a project at `/Users/me/projects/my-app` with existing claude-mem observations under `my-app`
2. A Claude Code worktree session runs from `.claude/worktrees/modest-burnell/`
3. Query the database:

```sql
SELECT project, COUNT(*) FROM observations GROUP BY project;
-- my-app          | 262
-- modest-burnell  |  78   ← should be under my-app
```

## Expected Behavior

All observations stored under `my-app` (the parent repository name), consistent with [worktree support docs](https://docs.claude-mem.ai/usage/folder-context#worktree-support).

## Root Cause (from v12.1.0 source analysis)

Two functions exist for project name resolution with different worktree behavior:

| Function | Returns | Worktree-aware? |
|---|---|---|
| `getProjectName(cwd)` | `path.basename(cwd)` | **No** |
| `getProjectInfo(cwd)` | `{ primary, parent, isWorktree, allProjects }` | **Yes** |

`getProjectInfo` correctly detects worktrees (reads `.git` file, parses `gitdir:` pointer, extracts parent repo path) and populates `allProjects = [parentProjectName, worktreeName]`. But it's only called at **read** time (context injection in `SessionStart` hook). All **write** paths call `getProjectName` directly:

1. **`session-init` hook** — resolves `project = getProjectName(cwd)` and POSTs to `/api/sessions/init`
2. **`handleObservationsByClaudeId`** (server-side) — calls `getProjectName(cwd)` to derive the project for `createSDKSession` and `queueObservation`
3. **`handleSessionInitByClaudeId`** (server-side) — receives the already-incorrect project from the hook

## Why This Keeps Regressing

This is the 4th report of the same bug (#1081, #1317, #1500, now this). Each prior fix patched individual call sites, but new or refactored code reaches for the simpler `getProjectName()` and reintroduces the issue. There's no test or compile-time guard preventing it.

**Suggested durable fixes (any one would prevent future regressions):**

1. **Make `getProjectName()` itself worktree-aware** — so there's no wrong function to call
2. **Delete `getProjectName()`** and force all callers through `getProjectInfo().primary` (with `primary` returning `parentProjectName` when in a worktree)
3. **Add an integration test** that creates a worktree, writes an observation, and asserts `project == parentProjectName`

Option 1 or 2 eliminates the two-function footgun entirely. Option 3 is a minimum safety net regardless.

0xLeathery:
Fixes #1819

## Summary

- `getProjectContext().primary` now returns `parentProjectName` when in a worktree (instead of `basename(cwd)`)
- All four write-path call sites switched from `getProjectName()` to `getProjectContext().primary`
- Regression test creates a real worktree directory structure and asserts `primary === parentProjectName`

## Problem

Observations and sessions from git worktrees are stored under `basename(cwd)` (e.g. `modest-burnell`) instead of the parent repo name (e.g. `my-app`). This is the same root cause as #1081, #1317, and #1500 — it regresses because `getProjectName()` (not worktree-aware) and `getProjectContext()` (worktree-aware) coexist, and write paths use the wrong one.

## Root Cause

| Operation | Before | Worktree-aware? |
|---|---|---|
| Write observations | `getProjectName(cwd)` | **No** |
| Write sessions | `getProjectName(cwd)` | **No** |
| Read context injection | `getProjectContext(cwd)` | Yes |

## Changes

**`src/utils/project-name.ts`** — `getProjectContext()` now sets `primary: worktreeInfo.parentProjectName` when a worktree is detected, instead of `primary: basename(cwd)`.

**4 call site migrations** (all `getProjectName(cwd)` → `getProjectContext(cwd).primary`):
- `src/cli/handlers/session-init.ts` — session init hook
- `src/services/worker/http/routes/SessionRoutes.ts` — observation handler
- `src/services/context/ContextBuilder.ts` — context generation
- `src/services/transcripts/processor.ts` — transcript processing

**`tests/utils/project-name.test.ts`** — regression test that creates a temporary worktree directory structure (`.git` file with `gitdir:` pointer) and asserts:
- `ctx.isWorktree === true`
- `ctx.primary === 'main-repo'` (parent name, not worktree name)
- `ctx.allProjects === ['main-repo', 'my-worktree']`

## Why this keeps regressing

Each prior fix patched individual call sites, but `getProjectName()` still exists as the simpler, easier-to-reach function. New code naturally gravitates toward it. This PR addresses the core issue: `getProjectContext().primary` now returns the correct name for worktrees, so even if callers use either function path, the worktree-aware one gives the right answer. The regression test provides a safety net against future changes.

## Test plan

- [x] `bun test tests/utils/project-name.test.ts` — 16/16 pass (including new worktree regression test)
- [ ] Manual verification: run Claude Code session from a git worktree, confirm observations stored under parent project name

### DIFF
--- a/src/cli/handlers/session-init.ts
+++ b/src/cli/handlers/session-init.ts
@@ -6,7 +6,7 @@
 
 import type { EventHandler, NormalizedHookInput, HookResult } from '../types.js';
 import { ensureWorkerRunning, workerHttpRequest } from '../../shared/worker-utils.js';
-import { getProjectName } from '../../utils/project-name.js';
+import { getProjectContext } from '../../utils/project-name.js';
 import { logger } from '../../utils/logger.js';
 import { HOOK_EXIT_CODES } from '../../shared/hook-constants.js';
 import { isProjectExcluded } from '../../utils/project-filter.js';
@@ -42,7 +42,7 @@ export const sessionInitHandler: EventHandler = {
     // Use placeholder so sessions still get created and tracked for memory
     const prompt = (!rawPrompt || !rawPrompt.trim()) ? '[media prompt]' : rawPrompt;
 
-    const project = getProjectName(cwd);
+    const project = getProjectContext(cwd).primary;
     const platformSource = normalizePlatformSource(input.platform);
 
     logger.debug('HOOK', 'session-init: Calling /api/sessions/init', { contentSessionId: sessionId, project });
--- a/src/services/context/ContextBuilder.ts
+++ b/src/services/context/ContextBuilder.ts
@@ -129,11 +129,12 @@ export async function generateContext(
 ): Promise<string> {
   const config = loadContextConfig();
   const cwd = input?.cwd ?? process.cwd();
-  const project = getProjectContext(cwd).primary;
+  const context = getProjectContext(cwd);
+  const project = context.primary;
   const platformSource = input?.platform_source;
 
-  // Use provided projects array (for worktree support) or fall back to single project
-  const projects = input?.projects || [project];
+  // Use provided projects array (for worktree support) or fall back to all known projects
+  const projects = input?.projects ?? context.allProjects;
 
   // Full mode: fetch all observations but keep normal rendering (level 1 summaries)
   if (input?.full) {
--- a/src/services/context/ContextBuilder.ts
+++ b/src/services/context/ContextBuilder.ts
@@ -10,7 +10,7 @@ import { homedir } from 'os';
 import { unlinkSync } from 'fs';
 import { SessionStore } from '../sqlite/SessionStore.js';
 import { logger } from '../../utils/logger.js';
-import { getProjectName } from '../../utils/project-name.js';
+import { getProjectContext } from '../../utils/project-name.js';
 
 import type { ContextInput, ContextConfig, Observation, SessionSummary } from './types.js';
 import { loadContextConfig } from './ContextConfigLoader.js';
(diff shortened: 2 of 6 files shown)


0xLeathery:
Addressed all three nitpicks in 2089c758:

1. **`ContextBuilder.ts` — `allProjects` fallback**: Default now uses `context.allProjects` instead of `[project]`, so context reads in worktrees also pick up legacy worktree-labeled records during transition.

2. **`ProjectContext` JSDoc**: Updated `primary` doc to "Canonical project name for writes/queries (parent repo in worktrees)" and `allProjects` to "[parentRepo, worktreeName] for worktree" — should prevent future caller misuse.

3. **Write-path regression test**: Added a test that mirrors the exact `getProjectContext(cwd).primary` pattern used by `session-init.ts` and `SessionRoutes.ts`, asserting it resolves to the parent repo. Refactored worktree fixtures into `beforeAll`/`afterAll` to share across both tests. All 17 tests pass.

PR #1838, opened 2026-04-15
fix: Windows --print (pipe) mode hangs when plugin hooks block on stdin

### LINKED ISSUE #1482: claude-mem plugin breaks claude --print (pipe mode) on Windows
Bug Description

  When the claude-mem plugin is enabled, claude --print (pipe/non-interactive mode) returns empty output silently on
  Windows. No error is shown on stdout or stderr. Interactive mode works fine. Disabling the plugin immediately fixes
  the issue.

  Steps to Reproduce

  1. Install claude-mem plugin and enable it in ~/.claude/settings.json
  2. Run any claude --print command with piped input:
  echo "test" | claude -p
  3. Output is empty — no response, no error
  4. Disable the plugin in settings ("claude-mem@thedotmack": false)
  5. Run the same command — works correctly
  6. Re-enable the plugin — breaks again

  Additional verification:
  - claude -p --bare (skips all plugins) with piped input works, confirming the plugin is the cause
  - claude --output-format text with piped input throws Error: Raw mode is not supported on the current process.stdin
  (Ink UI library error), suggesting the plugin forces Ink initialization even in --print mode
  - Interactive mode (claude "test") is unaffected

  Expected Behavior

  The --print mode is designed for non-interactive scripting and piped input. Plugins should either not load or
  gracefully degrade in --print mode so they don't interfere with stdin/stdout handling.

  Environment

  - Claude-mem version: latest (installed via marketplace)
  - Claude Code version: 2.1.83
  - Platform: win32

  Logs

  # Error captured when using --output-format text with piped input (same root cause):
  ERROR  Raw mode is not supported on the current process.stdin, which Ink uses as input stream by default.
   - handleSetRawMode (B:/~BUN/root/src/entrypoints/cli.js:740:1894)
   - <anonymous> (B:/~BUN/root/src/entrypoints/cli.js:750:3758)
   - B_ (B:/~BUN/root/src/entrypoints/cli.js:701:63116)
   - E5H (B:/~BUN/root/src/entrypoints/cli.js:701:66396)
   - vR (B:/~BUN/root/src/entrypoints/cli.js:701:74271)

  Additional Context

  Workaround: Temporarily disable the plugin in ~/.claude/settings.json before --print calls and re-enable after. This
  is what we implemented in our PR review automation scripts (PowerShell) to keep claude-mem active during interactive
  sessions while allowing --print to function for scripted reviews.

  Impact: This breaks all non-interactive/scripted Claude Code usage (CI/CD pipelines, automation scripts, batch
  processing) when claude-mem is installed and enabled.

jakeefr:
Fixes #1482. Fixed pipe mode detection failing on Windows when using claude --print.

I maintain PRISM (https://github.com/jakeefr/prism), a post-session diagnostics tool for Claude Code — CLAUDE.md adherence analysis and session health scoring from the same JSONL files. claude-mem handles session memory, PRISM handles session health — complementary tools.

### DIFF
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -154,16 +154,25 @@ function collectStdin() {
       resolve(null);
     });
 
-    // Safety: if no data arrives within 5s, proceed without stdin
+    // Safety: if no data arrives within timeout, proceed without stdin.
+    // On Windows, use a shorter timeout (500ms) because the 5s wait causes
+    // --print (pipe) mode to hang or produce empty output — Claude Code's hook
+    // stdin data arrives in milliseconds when present, and the long timeout
+    // blocks SessionStart hooks for up to 15s total (3 hooks × 5s). Fixes #1482.
+    const stdinTimeoutMs = IS_WINDOWS ? 500 : 5000;
     setTimeout(() => {
       process.stdin.removeAllListeners();
       process.stdin.pause();
       resolve(chunks.length > 0 ? Buffer.concat(chunks) : null);
-    }, 5000);
+    }, stdinTimeoutMs);
   });
 }
 
-const stdinData = await collectStdin();
+// Skip stdin collection for the "start" subcommand — it spawns a daemon and
+// never reads stdin. This avoids a blocking wait on Windows in --print mode
+// where stdin.isTTY is false but no hook data is piped. Fixes #1482.
+const needsStdin = !args.includes('start');
+const stdinData = needsStdin ? await collectStdin() : null;
 
 // Spawn Bun with the provided script and args
 // Use spawn (not spawnSync) to properly handle stdio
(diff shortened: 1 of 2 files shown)


xkonjin (commented):
Code review for PR 1838:
This looks like a solid fix for the Windows pipeline hang issues (#1482). 

A couple of observations:
1. Shortening the timeout to `500ms` for Windows makes sense if data normally arrives instantly. Are there any edge cases (like extreme system load) where Windows might take >500ms to pipe stdin? Given that failing to read stdin just proceeds with empty stdin, it's a safe failure mode, but worth noting.
2. Skipping stdin for `start` is a great optimization to avoid the blocking wait altogether for daemon startup.
3. The test coverage validates the presence of the logic in the script string (which I assume is how `bun-runner.test.ts` is structured, checking the source directly).

Code is clean and directly addresses the described issue.

thedotmack:

PR #1843, opened 2026-04-15
fix: worker spawn failure when Windows username contains space

### LINKED ISSUE #1797: [Windows] Worker fails to spawn when username contains a space — 3 bugs chained
# [Windows] Worker fails to spawn when username contains a space — 3 bugs chained

## Before submitting

- [x] I searched existing issues and confirmed this is not a duplicate (searched: "space in username", "Module not found", "Windows worker", "space path"; referenced #363, #367, #371, #373)

---

## Bug Description

On Windows with a space in the username (e.g. `Work User`, `First Last`), the worker never successfully spawns. I traced three layered bugs that combine to make all semantic-search features (`build_corpus`, `list_corpora`, `query_corpus`, `smart_search`, `timeline`) unusable:

1. `worker-cli.js` — PowerShell `-ArgumentList` quoting drops the path at the space
2. `worker-wrapper.cjs` — Bun's Windows `child_process.spawn` also truncates at the space when spawning `worker-service.cjs`
3. `worker-service.cjs` — exits cleanly with `code=0` after exactly 300 s (the `DEFAULT: 3e5` timeout) when no parent Claude Code IPC is established — this blocks external respawn via Task Scheduler / service

Result: raw SQLite capture still works, but the semantic / vector layer is dead on Windows.

---

## Steps to Reproduce

1. Use a Windows account whose profile path contains a space (very common: Windows default "First Last" naming, corporate machines with role-based names like `Work User` or `Admin User`).
2. Install claude-mem via `/plugin install claude-mem@thedotmack` in Claude Code.
3. Start a new Claude Code session.
4. Attempt any semantic search, e.g. `list_corpora` via the MCP tool.
5. Observe it fails; inspect `%USERPROFILE%\.claude-mem\logs\claude-mem-YYYY-MM-DD.log`:
   ```
   Worker unavailable on Windows — skipping spawn (recent attempt failed within cooldown)
   Worker failed to start (health check timeout)
   ```
6. Inspect `%USERPROFILE%\.claude-mem\logs\worker-YYYY-MM-DD.log.err`:
   ```
   error: Module not found "C:\Users\<first-word-of-username>"
   ```

---

## Expected Behavior

The worker should start normally on Windows regardless of whether the username contains spaces. After startup, semantic search tools should respond within a few seconds.

---

## Environment

- **Claude-mem version**: 12.1.0 (from `installed_plugins.json` — latest at time of filing)
- **Claude Code version**: current stable (same session running this report)
- **OS**: Windows 10 IoT Enterprise LTSC 2021 (10.0.19044)
- **Platform**: `win32` x64
- **Shell**: git-bash (MSYS)
- **Bun**: 1.3.11
- **Node**: v24.12.0
- **Username**: contains a space (redacted)
- **Plugin path**: `%USERPROFILE%\.claude\plugins\marketplaces\thedotmack\plugin\scripts\`

---

### From `worker-YYYY-MM-DD.log.err` — the actual bun crash:

```
error: Module not found "C:\Users\<first-word-of-username>"
```

### From `claude-mem-YYYY-MM-DD.log` — the spawn failure cascade:

```
[WARN ] [SYSTEM] Worker not healthy, hook will proceed gracefully
[ERROR] [SYSTEM] Worker failed to start (health check timeout)
[WARN ] [SYSTEM] Worker failed to start before hook, handler will proceed gracefully
[WARN ] [SYSTEM] Worker unavailable on Windows — skipping spawn (recent attempt failed within cooldown)
```

### From `worker-YYYY-MM-DD.log` — evidence of bug #3 (the 5-minute clean exit):

When I manually spawned `worker-wrapper.cjs` directly via bun from a shell (bypassing the PowerShell path in `worker-cli.js`), the wrapper started successfully and the worker-service did respond to `/api/readiness` with `{"status":"ready","mcpReady":true}` — but exactly 300 seconds later:

```
[2026-04-14T03:52:27.217Z] [wrapper] Wrapper starting
[2026-04-14T03:52:27.218Z] [wrapper] Spawning inner worker: <path>\worker-service.cjs
[2026-04-14T03:57:28.086Z] [wrapper] Inner exited with code=0, signal=null
[2026-04-14T03:57:28.086Z] [wrapper] Inner exited unexpectedly, wrapper exiting (hooks will restart if needed)
```

Clean exit, no error, exactly 300 000 ms — matches the `DEFAULT: 3e5` constant in the worker-service config block.

---

### Detail on bug #1 — `worker-cli.js` PowerShell spawn

Relevant code (minified in shipped plugin):

```js
let f = `$env:CLAUDE_MEM_WORKER_PORT='${r}'; Start-Process -FilePath '${o}' -ArgumentList '${a}' -WorkingDirectory '${c}' -WindowStyle Hidden -RedirectStandardOutput '${p}' -RedirectStandardError '${p}.err' -PassThru | Select-Object -ExpandProperty Id`;
```

`${a}` is the absolute path to `worker-wrapper.cjs`, which on a machine with a space in the username expands to something like `C:\Users\<word> <word>\.claude\plugins\marketplaces\thedotmack\plugin\scripts\worker-wrapper.cjs`. PowerShell's `Start-Process -ArgumentList '<single-quoted-path-with-space>'` treats the value as a single argument at the PowerShell level — but when PowerShell assembles the Windows command-line string for the child (bun), the inner space is not escaped, and bun tokenizes it into two args: `C:\Users\<firstword>` (treated as the script path → "Module not found") and the rest (dropped).

Note: I spotted `scripts/worker-cli.js.bak-prespacefix` in the shipped plugin — this suggests a previous fix attempt. The issue appears to have regressed or not have fully landed.

**Suggested fixes:**

- Escape the path with inner double-quotes for the Windows command-line layer: `-ArgumentList "'${a}'"` or `-ArgumentList '\"${a}\"'`
- Use `[System.Diagnostics.ProcessStartInfo].ArgumentList` (the typed `List<string>`) via `New-Object` — bypasses the command-line serialization layer that's dropping the space
- Switch from `Start-Process` to `$p = [System.Diagnostics.Process]::Start($psi)` with `UseShellExecute = $false` and `ArgumentList.Add()` per-arg

### Detail on bug #2 — `worker-wrapper.cjs` inner spawn

```js
t = spawn(process.execPath, [l], {
  stdio: ["inherit", "inherit", "inherit", "ipc"],
  env: { ...process.env, CLAUDE_MEM_MANAGED: "true" },
  cwd: path.dirname(l),
});
```

`process.execPath` is `bun.exe` (wrapper runs under bun), `l` is the space-containing path to `worker-service.cjs`. Bun's Windows `child_process.spawn` truncates the path at the space the same way PowerShell did. Reproducible in isolation — a minimal `bun -e "require('child_process').spawnSync(process.execPath, ['C:\\\\path with\\\\space.js'])"` shows the same tokenization.

**Suggested fix:** explicitly wrap the arg with CMD-level escape quoting before `spawn`, or use Node as the launcher for bugs where `process.execPath` is bun on Windows.

### Detail on bug #3 — 300 s self-exit blocks external respawn

Even with bugs 1 & 2 worked around, the worker kills itself after 300 s when there is no parent Claude Code IPC to keep it alive. This prevents any of these workarounds:

- Windows Task Scheduler task at user-logon (for reboot survival)
- Running worker-service as a Windows service (for persistent availability)
- Shell-started background worker
- Third-party supervisor (nssm, pm2, etc.)

**Suggested fix:** add an opt-in env flag such as `CLAUDE_MEM_STANDALONE=true` that disables the parent-IPC timeout, so Windows users can run the worker as a managed service or scheduled task. Document the flag in README.

### Workaround I wrote (and why it's not a full fix)

I extended the existing `~/.claude/scripts/claude-mem-ensure-worker.cjs` watchdog with a spawn fallback that uses Node's `child_process.spawn(bunPath, [wrapperPath], {detached:true, windowsHide:true})` — Node does quote args correctly on Windows, so the first-level spawn succeeds and the wrapper logs `Wrapper starting`. But then bug #2 and bug #3 bite, and the worker dies within 5 minutes. Reverted the watchdog change because it doesn't give durable persistence.

### Impact

- Any Windows user with a space in their profile path (very common: Windows default "First Last" naming, shared / role accounts like `Work User`, `Admin User`, `IT Admin`, etc.) hits this silently
- `build_corpus`, `query_corpus`, `smart_search`, `list_corpora`, `timeline` all fail
- Raw SQLite capture still works — `search` works for exact keyword matches, so ~70 % of claude-mem's value is retained
- Semantic search layer (~30 % of the value) is unusable

### Related

- Bundled `worker-cli.js` troubleshooting output references #363, #367, #371, #373 — I suspect this is adjacent to one or more of those
- Artifact `worker-cli.js.bak-prespacefix` in shipped plugin

### What I'd like

Acknowledgment that this is tracked as a Windows space-path issue, and ideally fixes for all three bugs. Happy to test a prerelease build.

jakeefr:
Fixes #1797. Two chained bugs caused worker spawn failure when the Windows username contains a space:

1. **Fixed PowerShell arg quoting in worker-cli.js** — `-ArgumentList @(...)` array syntax prevents path-splitting at spaces
2. **Fixed Bun spawn path truncation in worker-wrapper.cjs** — `windowsVerbatimArguments` + explicit CMD-level quoting preserves the full path

Includes source-inspection regression tests following the existing `bun-runner.test.ts` pattern.

I maintain [PRISM](https://github.com/jakeefr/prism), a post-session diagnostics tool for Claude Code. Windows path handling is a recurring theme — fixed this since it affects my testing environment too.


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1881, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1846, opened 2026-04-15
Add ~/.local/bin/claude fallback to findClaudeExecutable

frank-ih:
After `which claude` fails, check `~/.local/bin/claude` before throwing. This fixes the SessionStart hook for users with claude installed via the official installer (not in the hook's constrained PATH).

**Root cause:** The SessionStart hook runs with a constrained PATH that doesn't include `~/.local/bin`. When `which claude` fails, `findClaudeExecutable()` immediately throws, with no fallback for the common official installer path.

**Fix:** Added step 4 to `findClaudeExecutable()` in `src/services/worker/SDKAgent.ts` that checks `~/.local/bin/claude` before throwing the error.

**Alternative:** Users can also set `CLAUDE_CODE_PATH` in `~/.claude-mem/settings.json` to point to the claude executable.


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1894, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1849, opened 2026-04-15
fix: reuse shared db connection in generateContext, fix outbound channel send

artsra:
## Summary

- **DB singleton**: Replace `new SessionStore()` on every `generateContext()` call with a module-level singleton. Eliminates repeated file-open/WAL-setup overhead and prevents `ERR_DLOPEN_FAILED` crashes on rapid reconnects.
- **Static import**: Replace dynamic `await import()` of `context-generator` inside `handleContextInject` and `handleContextPreview` with a static top-level import, avoiding repeated module resolution on hot paths.
- **Outbound channel send**: Replace the broken `api.runtime.channel[namespace]` approach (channel not available in runtime) with the correct `api.runtime.channel.outbound.loadAdapter(channelId)` SDK pattern.
- **accountId propagation**: Pass `accountId` through `connectToSSEStream` → SSE observation handler → `sendToChannel` so bot account selection works correctly for multi-account setups.
- **Empty message guard**: Skip SSE observations with no body content (no subtitle, narrative, facts, or title) to prevent empty Discord/Telegram messages.

## Test plan

- [ ] Verify context inject endpoint responds without re-opening SQLite on each call
- [ ] Verify outbound Discord/Telegram sends succeed (no "channel not available in runtime" error)
- [ ] Verify empty observations are not forwarded to messaging channels
- [ ] Verify accountId is passed correctly when configured in `observationFeed`

### DIFF
--- a/openclaw/src/index.ts
+++ b/openclaw/src/index.ts
@@ -132,7 +132,13 @@ interface OpenClawPluginApi {
       ((event: "after_compaction", callback: EventCallback<AfterCompactionEvent>) => void) &
       ((event: "gateway_start", callback: EventCallback<Record<string, never>>) => void);
   runtime: {
-    channel: Record<string, Record<string, (...args: any[]) => Promise<any>>>;
+    channel: {
+      outbound: {
+        loadAdapter: (channelId: string) => Promise<{
+          sendText?: (ctx: { to: string; text: string; accountId?: string; cfg?: unknown }) => Promise<unknown>;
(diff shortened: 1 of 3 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1871, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

Issue #1841, opened 2026-04-15
chroma-mcp process not killed on Claude Code exit


derkan:
## Bug Description

The `chroma-mcp` child process spawned by the `mcp-server.cjs` MCP server is not terminated when Claude Code exits. It remains as an orphan process until manually killed.

## Steps to Reproduce

1. Install claude-mem plugin (v12.1.1)
2. Start a Claude Code session
3. Exit Claude Code normally
4. Run `pgrep -lf chroma-mcp`

## Expected Behavior

`chroma-mcp` process should be terminated when Claude Code exits.

## Actual Behavior

`chroma-mcp` remains running as an orphan process:

```
uv tool uvx --python 3.13 chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma
```

## Environment

- Plugin version: 12.1.1
- OS: macOS Darwin 25.4.0
- Claude Code CLI

## Workaround

Added a `Stop` hook in `~/.claude/settings.json` to kill orphaned processes:

```json
"Stop": [{ "hooks": [{ "type": "command", "command": "pkill -f chroma-mcp || true" }] }]
```

Root cause likely in `mcp-server.cjs` not forwarding SIGTERM to the chroma-mcp child process.

alessandropcostabr:
## Windows perspective: orphan reaper never works due to PowerShell quoting bug

On Windows, the `aggressiveStartupCleanup()` function in `worker-service.cjs` has a PowerShell quoting bug that causes orphan detection to **always fail**:

```
[ERROR] Failed to enumerate orphaned processes during aggressive cleanup
Get-CimInstance : Cannot find positional parameter that accepts argument '%worker-service.cjs%...'
```

The WMI filter uses `'%process%'` inside a single-quoted PowerShell string, but the inner `'` terminates the string prematurely. Fix: double the inner single quotes (`''%process%''`).

After patching locally, the reaper found and killed **8 orphaned processes** on first startup — including chroma-mcp, bun, and mcp-server zombies.

Full analysis and patch: #1832 (comment) + PR #1854

thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1924, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

Issue #1832, opened 2026-04-15
Windows cold-start: winner daemon killed ~350ms after binding, 2m21s startup stall (follow-up to #1447 / #1680)


sean3808:
## Summary

Follow-up to #1447 (closed by #1680). PR #1680 fixes the **loser** path of the cold-start port-bind race — `EADDRINUSE` now exits INFO-level instead of ERROR. But on Windows the **winner** daemon is still being killed ~350ms after it successfully binds the port, producing a consistent **2 min 21 sec** cold-start delay before a third spawn finally succeeds.

Since #1680 only touches `worker.start().catch(...)` in `worker-service.ts` and never touches any shutdown entry point, this symptom is not resolved by it.

## Environment

- claude-mem **10.6.2** (also reproduced on 10.6.3)
- Claude Code 2.1.78
- Windows 11 Home (Build 26200, x64)
- Bun 1.2.9

## Timeline (real cold-start log, 2026-03-27 10:01)

```
10:01:23.857  Starting worker daemon                                         ← spawn A
10:01:23.858  Starting worker daemon                                         ← spawn B (parallel)
10:01:26.122  Removed dead processes from supervisor registry {removed=3}
10:01:26.123  Removed dead processes from supervisor registry {removed=3}    ← both ran registry cleanup
10:01:26.142  HTTP server started {host=127.0.0.1, port=37777, pid=23060}    ← winner bound
10:01:26.143  ✗ Worker failed to start  Failed to start server. Is port 37777 in use?   ← loser (pre-#1680 behaviour)
10:01:26.144  Worker started {pid=23060}
10:01:26.493  Shutdown initiated                                             ← winner dying, +350ms after bind
──── 2 min 21 sec with zero log output ────
10:03:47.440  Removing stale PID file (worker process is dead) {pid=23060, startedAt=2026-03-27T02:01:26.142Z}
10:03:48.463  Starting worker daemon                                         ← 3rd spawn (no race this time)
10:03:49.575  HTTP server started {pid=21884}
10:03:49.576  Worker started {pid=21884}                                     ← finally healthy
10:03:50.310  Aggressive startup cleanup: killing orphaned processes {platform=Windows, count=7}
```

## Why #1680 does not resolve this

[...]

Issue #1726, opened 2026-04-11
worker-service.cjs SIGKILL (OOM) during long sessions


thefinono:
## Description

During long Claude Code sessions with heavy parallel subagent usage, `worker-service.cjs` gets killed by macOS with signal 9 (SIGKILL), likely due to exceeding memory limits.

## Environment
- macOS Darwin 25.4.0
- claude-mem 10.5.6
- Claude Code with Opus 4.6 (1M context)
- Session involved ~15 subagent calls, multiple background tasks, and large JSON processing

## Error
```
Hook error (non-blocking): worker-service.cjs killed by signal 9 (SIGKILL)
```

## Expected behavior
Worker should either cap its memory usage or gracefully restart when approaching limits.

## Suggested fix
Add `--max-old-space-size=256` (or similar) to the Node invocation in `bun-runner.js` to prevent unbounded heap growth:

```js
const nodeArgs = ['--max-old-space-size=256', scriptPath, ...args];
```

Or implement a periodic GC/cleanup of the Chroma observation store during long sessions.

## Workaround
Non-blocking — session continues working. Restart Claude Code if observation capture stops.

thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1873, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

Issue #1361, opened 2026-03-14
chroma-mcp processes leak after session exit, repeatedly downloading ONNX model (~15GB wasted bandwidth)


WhymustIhaveaname:
## Problem

`worker-service.cjs --daemon` spawns a new `chroma-mcp` python process per session but never cleans them up on exit. On my system, 184 orphaned processes accumulated in 19 hours.

Worse: if the ONNX model (`all-MiniLM-L6-v2/onnx.tar.gz`, 79MB) gets corrupted by concurrent download races, every leaked process retries the download from S3 in a loop. Result: **15+ GB bandwidth, ~16 GB RAM, 29 concurrent S3 connections**.

## Reproduction

```bash
# After a few sessions, check:
ps aux | grep chroma-mcp | grep -v grep | wc -l   # 184 on my system
ss -tunp | grep python | wc -l                      # 29 active S3 connections
ls -lh ~/.cache/chroma/onnx_models/all-MiniLM-L6-v2/onnx.tar.gz  # corrupted: 17MB instead of 79MB
```

## Workaround

```bash
pkill -9 -f worker-service.cjs
pkill -9 -f chroma-mcp
rm -rf ~/.cache/chroma/onnx_models/all-MiniLM-L6-v2/
# Re-download and extract model once manually
```

## Suggested Fix

- Worker-service should kill chroma-mcp children on session exit (same class of leak as https://github.com/thedotmack/claude-mem/issues/1351)
- Use a file lock to prevent concurrent ONNX downloads

## Environment

Linux (Fedora 42), claude-mem v10.5.5, ChromaDB 1.5.5, Python 3.13

elenasteinvorth:
Hi @WhymustIhaveaname — I saw your issue #1361 about chroma-mcp process leaks. This seems related to the worker process leak issue #1090.

I'm Elena, researching the AI agent memory problem. Curious:

1. Is this bug affecting your daily usage of claude-mem?
2. Are you still using it despite the leak, or did you switch to something else?
3. What would make you switch to a different memory solution?

Any insights from your experience would be helpful.

Thanks,
Elena

thedotmack:
[...]

PR #1854, opened 2026-04-15
fix(windows): escape single quotes in WQL orphan reaper filter

alessandropcostabr:
## Summary

- Fix PowerShell quoting bug in `cleanupOrphanedProcesses()` and `aggressiveStartupCleanup()` that causes the WQL filter to **always fail** on Windows
- The orphan reaper has never successfully run on any Windows machine since it was introduced

## Problem

The WQL filter wraps LIKE patterns in single quotes inside a PowerShell single-quoted string:

```powershell
# Generated command (broken):
Get-CimInstance Win32_Process -Filter '(CommandLine LIKE '%worker-service.cjs%' OR ...)'
#                                      ↑ string ends here    ↑ positional argument (error!)
```

PowerShell interprets the inner `'` as the end of the string, then treats `%worker-service.cjs%` as a positional argument:

```
Get-CimInstance : Cannot find positional parameter that accepts argument
  '%worker-service.cjs% OR CommandLine LIKE %chroma-mcp%...'
```

This error occurs on **every startup** but is caught and logged as a non-blocking error, so it goes unnoticed. The result: orphaned `bun`, `chroma-mcp`, and `mcp-server` processes are **never cleaned up** on Windows.

## Fix

Double the inner single quotes (`''`), which is standard PowerShell escaping for literal single quotes inside single-quoted strings:

```typescript
// Before:
.map(p => `CommandLine LIKE '%${p}%'`)

// After:
.map(p => `CommandLine LIKE ''%${p}%''`)
```

Both `cleanupOrphanedProcesses()` (L380) and `aggressiveStartupCleanup()` (L530) are fixed.

## Test result

Before (every startup since feature was introduced):
```
[ERROR] Failed to enumerate orphaned processes during aggressive cleanup
```

After patch (tested on Windows 11 Build 26200, claude-mem 12.1.0):
```
[INFO] Aggressive startup cleanup: killing orphaned processes {platform=Windows, count=8}
[INFO] Aggressive startup cleanup complete {count=8}
```

8 orphaned processes detected and killed on first startup.

## Impact

This is the root cause or contributing factor for multiple open issues on Windows:

| Issue | Title | How this fix helps |
|-------|-------|-------------------|
| #1832 | Windows cold-start 2m21s stall | Orphan cleanup now works, reducing port conflicts |
| #1841 | chroma-mcp not killed on exit | Orphans now cleaned on next startup |
| #1361 | chroma-mcp leak + 15GB bandwidth | 184 orphans would now be detected and killed |
| #1726 | worker-service SIGKILL (OOM) | Accumulated orphans consuming RAM now cleaned |
| #1077 | 146 orphans causing OOM | Same root cause — reaper was silently broken |

## Test plan

- [x] Verified locally on Windows 11 (bundled file patch)
- [x] Confirmed orphan reaper now finds and kills orphaned processes
- [x] Confirmed worker starts and remains stable after patch
- [ ] Run existing test suite (`bun test`)

### DIFF
--- a/src/services/infrastructure/ProcessManager.ts
+++ b/src/services/infrastructure/ProcessManager.ts
@@ -376,8 +376,10 @@ export async function cleanupOrphanedProcesses(): Promise<void> {
     if (isWindows) {
       // Windows: Use WQL -Filter for server-side filtering (no $_ pipeline syntax).
       // Avoids Git Bash $_ interpretation (#1062) and PowerShell syntax errors (#1024).
+      // Inner single quotes must be doubled ('') to escape them within PowerShell
+      // single-quoted strings — otherwise '%pattern%' breaks the filter parsing (#1832).
       const wqlPatternConditions = ORPHAN_PROCESS_PATTERNS
-        .map(p => `CommandLine LIKE '%${p}%'`)
+        .map(p => `CommandLine LIKE ''%${p}%''`)
         .join(' OR ');
 
       const cmd = `powershell -NoProfile -NonInteractive -Command "Get-CimInstance Win32_Process -Filter '(${wqlPatternConditions}) AND ProcessId != ${currentPid}' | Select-Object ProcessId, CreationDate | ConvertTo-Json"`;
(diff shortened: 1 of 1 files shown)


thedotmack:

alessandropcostabr:
**Bug still present in v12.2.0 marketplace build**

Reopening this for visibility — the fix was not included in the distributed plugin artifacts.

On Windows, the marketplace `worker-service.cjs` (v12.2.0, SHA in plugin.json) still contains the broken pattern:

```js
// broken (still in marketplace v12.2.0):
n.map(p => `CommandLine LIKE '%${p}%'`)

// should be:
n.map(p => `CommandLine LIKE ''%${p}%''`)
```

Verified by searching the minified file at `~/.claude/plugins/marketplaces/thedotmack/plugin/scripts/worker-service.cjs` — one occurrence found and confirmed broken.

Applied the fix locally as a hotpatch. If the source fix is already in `HEAD`, the marketplace artifact likely needs a rebuild/republish to pick it up.

Ref: #1832

PR #1855, opened 2026-04-15
fix(chroma): kill registry-tracked orphan before reconnect in connectInternal

alessandropcostabr:
## Problem

When a worker process is killed hard (OOM, SIGKILL), `chroma-mcp` may survive as an orphan. On the **new** worker instance, `this.transport` is `null` (fresh singleton), so the stale-cleanup block at the top of `connectInternal()` does nothing — and a second `chroma-mcp` is spawned alongside the orphan.

Root cause chain:
1. Worker is SIGKILL'd (OOM killer, systemd watchdog, etc.)
2. `chroma-mcp` subprocess survives (SIGPIPE ignored, or raced before pipe close)
3. New worker starts — `ChromaMcpManager.instance` is `null`, fresh singleton
4. `connectInternal()` runs: `this.transport` is `null`, stale-cleanup block skips
5. New `chroma-mcp` spawned → now 4 PIDs instead of 2, accumulating per restart

## Fix

Add a **registry-based orphan check** at the start of `connectInternal()`: before spawning, look up the `CHROMA_SUPERVISOR_ID` entry in the `ProcessRegistry`. If an alive PID is found (i.e. the previous worker left it behind), SIGTERM it → wait up to 2s → SIGKILL if needed → unregister.

```typescript
// Kill any orphaned chroma-mcp from a previous worker run.
const registry = getSupervisor().getRegistry();
const orphanEntry = registry.getAll().find(r => r.id === CHROMA_SUPERVISOR_ID);
if (orphanEntry && isPidAlive(orphanEntry.pid)) {
  // SIGTERM → 2s wait → SIGKILL
  registry.unregister(CHROMA_SUPERVISOR_ID);
}
```

## Relationship to `aggressiveStartupCleanup`

`aggressiveStartupCleanup` (added in v12.1.x) handles orphan cleanup via **process-scan** at daemon startup. This fix is **complementary**:

| | `aggressiveStartupCleanup` | This fix |
|---|---|---|
| Trigger | Once, at daemon startup | Every `connectInternal()` call |
| Strategy | Scan all processes for `chroma-mcp` pattern | Look up exact PID from registry |
| Coverage | All orphans at startup | Registry-tracked orphan at reconnect time |

They do not conflict. If `aggressiveStartupCleanup` already killed the orphan, `isPidAlive()` returns `false` and this code is a no-op.

## Testing

Tested locally on Linux (DarkStarII):
1. Worker running with 2 healthy chroma-mcp PIDs
2. `kill -9 <worker-pid>` to simulate OOM
3. New worker (with this fix) started automatically
4. Result: 2 chroma-mcp PIDs (orphan killed, fresh spawn) — no accumulation

Logs confirm `aggressiveStartupCleanup` ran first (killed pre-existing orphan from earlier session), then `connectInternal()` connected cleanly.


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1925, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1859, opened 2026-04-15
fix: default search/timeline to current project when project param omitted

vkolyvas:
## Problem

`search()` and `timeline()` return results from ALL projects when `project` is not specified, causing cross-project memory pollution. Observations use a global auto-increment ID across all projects, so without an explicit project filter the results are mixed.

## Fix

Default `project` to `getCurrentProjectName()` (git root context, matches status line / working directory) when not explicitly provided. Storage of observations already uses the correct project context — only read/search operations were affected.

Affected methods: `search()`, `timeline()`, `getRecentContext()`, `getContextTimeline()`, `getTimelineByQuery()`.

## Diff

\`\`\`diff
--- a/src/services/worker/SearchManager.ts
+++ b/src/services/worker/SearchManager.ts
@@ -24,6 +24,7 @@ import type { ObservationSearchResult, ... } from '../sqlite/types.js';
 import { logger } from '../../utils/logger.js';
 import { formatDate, ... } from '../../shared/timeline-formatting.js';
 import { ModeManager } from '../domain/ModeManager.js';
+import { getCurrentProjectName } from '../../shared/paths.js';
 
@@ -127,6 +128,10 @@ export class SearchManager {
   async search(args: any): Promise<any> {
     const normalized = this.normalizeParams(args);
+    if (!normalized.project) {
+      normalized.project = getCurrentProjectName();
+    }
     const { query, type, ... } = normalized;

- timeline(): add `const resolvedProject = project || getCurrentProjectName()`
- getRecentContext(): `basename(process.cwd())` → `getCurrentProjectName()`
- getContextTimeline(), getTimelineByQuery(): same `resolvedProject` pattern
\`\`\`

### DIFF
--- a/src/services/worker/SearchManager.ts
+++ b/src/services/worker/SearchManager.ts
@@ -24,6 +24,7 @@ import type { ObservationSearchResult, SessionSummarySearchResult, UserPromptSea
 import { logger } from '../../utils/logger.js';
 import { formatDate, formatTime, formatDateTime, extractFirstFile, groupByDate, estimateTokens } from '../../shared/timeline-formatting.js';
 import { ModeManager } from '../domain/ModeManager.js';
+import { getCurrentProjectName } from '../../shared/paths.js';
 
 import {
   SearchOrchestrator,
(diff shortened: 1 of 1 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1911, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #1861, opened 2026-04-15
feature: add delete_observations_by_project MCP endpoint

vkolyvas:
## Feature Request

Add a `delete_observations_by_project(project: string)` MCP tool / REST endpoint to delete all observations for a given project.

## Why

- **Project data isolation** — no way to purge a project's memory without direct SQLite access
- **GDPR / compliance** — organizations may need to delete all data for a specific project or client
- **Project archival / cleanup** — when archiving a project, being able to wipe its memory cleanly
- **No workaround via existing tools** — `search`, `timeline`, `get_observations` are all read-only; there is no delete capability exposed through the MCP interface today

## Scope

The endpoint should delete from:
- `observations` table (project column)
- `session_summaries` table (project column)

And cascade-delete related records in tables that reference these (e.g. `sdk_sessions` via CASCADE).

## Implementation hint

`store.ts` already uses `project` correctly on INSERT. A delete-by-project would mirror the same filter pattern used in `SessionSearch.buildFilterClause()` — just invert it as a DELETE WHERE condition.

### DIFF
--- a/src/services/worker/SearchManager.ts
+++ b/src/services/worker/SearchManager.ts
@@ -24,6 +24,7 @@ import type { ObservationSearchResult, SessionSummarySearchResult, UserPromptSea
 import { logger } from '../../utils/logger.js';
 import { formatDate, formatTime, formatDateTime, extractFirstFile, groupByDate, estimateTokens } from '../../shared/timeline-formatting.js';
 import { ModeManager } from '../domain/ModeManager.js';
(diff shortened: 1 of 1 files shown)


thedotmack:
Closed during the April 2026 backlog cleanup. The feature ask is now tracked in #1962, which is the single canonical issue for prioritization. Thanks for the suggestion — your context is captured there.

PR #2051, opened 2026-04-16
fix(project-name): scope observations per worktree + cwd-based backfill

thedotmack:
_Supersedes #2050 (closed when the branch was renamed)._

## Summary
Two things in this PR:

### 1. `src/utils/project-name.ts` — reverts #1820
Each worktree gets its own bucket. In a worktree, `primary = parent/worktree` (e.g. `claude-mem/dar-es-salaam`); in a main repo, `primary = basename`. `allProjects` is always `[primary]` — strict isolation at query time so observations don't cross between worktrees.

### 2. `scripts/cwd-remap.ts` — retroactive backfill using the authoritative cwd
Earlier drafts had a `worktree-remap.ts` that regex-scraped path patterns out of observation free-text. That was a band-aid — only 1/3,498 sessions remapped in practice, because the "source" it was scraping is derived and lossy.

The real source of truth for each session's working directory is already in the DB: `pending_messages.cwd` is populated by every hook payload. The replacement script uses it directly:

1. Pull every distinct cwd from `pending_messages.cwd`.
2. Classify each with `git`:
   - `rev-parse --absolute-git-dir` vs `--git-common-dir` → main vs worktree
   - `rev-parse --show-toplevel` for the correct leaf (cwd may be a subdir)
3. Pick the **earliest** cwd per session (not dominant — claude-mem's own hooks run from nested `.context/claude-mem/` dirs and would poison a frequency count).
4. Update `sdk_sessions`, `observations`, `session_summaries` in one transaction. Auto-backs-up the DB.

Real-DB result: **41 sessions remapped**, **1,694 observations** and **3,091 session_summaries** updated. 43 cwds skipped (deleted worktrees / non-repos) — left alone, no inference.

## Usage

```bash
bun scripts/cwd-remap.ts          # dry-run
bun scripts/cwd-remap.ts --apply  # write changes (backs up DB first)
```

## Test plan
- [ ] Confirm observations in a worktree stay scoped to `parent/worktree`
- [ ] Confirm main-repo sessions continue to use basename
- [ ] Dry-run `scripts/cwd-remap.ts` against a backup DB and spot-check the proposed mapping summary
- [ ] Run `tests/utils/project-name.test.ts`


thedotmack:
Superseded by a new PR with expanded scope: adds the full merged-worktree adoption engine on top of the cwd-based backfill.

PR #2069, opened 2026-04-18
fix: viewer model dropdown, SSE cleanup, OpenRouter URL, zh-TW mode, alias fix, restart guard, FTS5 fallback, httpcore dep

### LINKED ISSUE #1958: [BUG] Viewer Settings model dropdown ignores current config, destroys precise model IDs
**Severity:** high  
**Component:** viewer

**Symptom:** Advanced panel always displays `haiku (fastest)` regardless of `CLAUDE_MEM_MODEL`. Clicking Save writes coarse label to settings.json, destroying precise version-locked ID like `claude-sonnet-4-6`.

**Root cause:** No logic maps saved precise model ID back onto coarse `haiku`/`sonnet`/`opus` options on render.

**Verification against v12.1.1 HEAD:** VALID — `src/ui/viewer/components/ContextSettingsModal.tsx:357-364` hardcodes 3 options `haiku/sonnet/opus`, default fallback `'haiku'`. No mapping from saved `claude-sonnet-4-6` back to coarse option. Save overwrites with coarse label.

---

_Consolidated during the April 2026 backlog cleanup. Original reports: Issue #1853_

### LINKED ISSUE #1959: [BUG] SSE new_prompt broadcast dies after /reload-plugins
**Severity:** medium  
**Component:** viewer

**Symptom:** User prompts still stored in DB but UI only updates after browser refresh. Other event types (new_observation, new_summary, processing_status, initial_load) keep working. SSE client count keeps climbing (8→20→26→28) across reloads indicating no cleanup. Only fixed by killing and resuming Claude Code session.

**Verification against v12.1.1 HEAD:** UNCERTAIN — `/reload-plugins` is a Claude Code host command, not claude-mem. SSE `new_prompt` broadcast path exists at `SessionEventBroadcaster.ts:33`. Reload-triggered client cleanup logic not directly verifiable from claude-mem source.

---

_Consolidated during the April 2026 backlog cleanup. Original reports: Issue #1331_

### LINKED ISSUE #1960: [BUG] OpenRouter provider URL hardcoded, can't target local OpenAI-compatible servers
**Severity:** low  
**Component:** misc

**Symptom:** No setting to point openrouter provider at a local OpenAI-compatible server (llama.cpp, Ollama, LM Studio) without editing the compiled bundle.

**Root cause:** `getOpenRouterConfig()` and `queryOpenRouterMultiTurn()` use hardcoded `https://openrouter.ai/api/v1/chat/completions` URL.

**Verification against v12.1.1 HEAD:** VALID — `src/services/worker/OpenRouterAgent.ts:32` constant `OPENROUTER_API_URL = 'https://openrouter.ai/api/v1/chat/completions'`, used unconditionally at line 373 `fetch(OPENROUTER_API_URL, ...)`. No base-URL setting.

---

_Consolidated during the April 2026 backlog cleanup. Original reports: PR #1795_

### LINKED ISSUE #1961: [BUG] Traditional Chinese mode silently falls back to Simplified Chinese
**Severity:** medium  
**Component:** misc

**Symptom:** Setting `CLAUDE_MEM_MODE=code--zh-TW`/`code--zh-tw` produces Simplified Chinese output.

**Root cause:** `plugin/modes/code--zh-tw.json` doesn't exist. Mode loader silently falls back rather than failing.

**Verification against v12.1.1 HEAD:** VALID — `plugin/modes/` directory contains `code--zh.json` (Simplified) but NO `code--zh-tw.json` / `code--zh-TW.json` (listed 35 mode files, no zh-TW variant).

---

_Consolidated during the April 2026 backlog cleanup. Original reports: PR #1669_

### LINKED ISSUE #2046: [BUG] chroma-mcp: No module named 'httpcore' on all query/add operations
## Before submitting

## Bug Description

All chroma-mcp operations (both `chroma_query_documents` and `chroma_add_documents`) fail with `ModuleNotFoundError: No module named 'httpcore'`. This renders `mem-search`, corpus sync, and the MCP `search`/`get_observations` tools completely non-functional.

Appears to be an environment issue on the remote Worker — the Python runtime serving chroma-mcp is missing the `httpcore` package. `httpcore` is a transitive dependency of `httpx`, which Chroma's HTTP client pulls in. Very likely a recent dep pin drifted or the Worker image was rebuilt without it.

## Steps to Reproduce

1. Install `claude-mem@thedotmack` v12.1.0
2. From Claude Code, invoke the MCP search tool:
   ```
   mcp__plugin_claude-mem_mcp-search__search(query="anything", limit=5, project="anyproject")
   ```
3. Observe the 500 error

Also reproduces automatically during background corpus sync — see log excerpt below.

## Expected Behavior

`search` / `get_observations` return results. Corpus sync completes without `ModuleNotFoundError`.

## Actual Behavior

```
Error calling Worker API: Worker API error (500): {"error":"chroma-mcp tool \"chroma_query_documents\" returned error: Error executing tool chroma_query_documents: Failed to query documents from collection 'cm__claude-mem': No module named 'httpcore'"}
```

Tried downgrading the plugin from 12.1.0 → 10.5.6 (edited `installed_plugins.json`, restarted Claude Code). The error is identical, which confirms this is server-side in the Worker, not a plugin-version regression.

## Environment

- **Claude-mem version**: 12.1.0 (also tested 10.5.6, same error)
- **Claude Code version**: 2.1.109
- **OS**: Artix Linux (Arch-based), kernel 6.19.11
- **Platform**: x86_64 Linux

## Logs

From `~/.claude-mem/logs/claude-mem-2026-04-15.log`:

```
[2026-04-15 23:04:03.608] [ERROR] [CHROMA_SYNC] Batch add failed, continuing with remaining batches {collection=cm__claude-mem, batchStart=0, batchSize=5} chroma-mcp tool "chroma_add_documents" returned error: Error executing tool chroma_add_documents: Failed to add documents to collection 'cm__claude-mem': No module named 'httpcore'
[2026-04-15 23:04:25.680] [ERROR] [CHROMA_SYNC] Batch add failed, continuing with remaining batches {collection=cm__claude-mem, batchStart=0, batchSize=5} chroma-mcp tool "chroma_add_documents" returned error: Error executing tool chroma_add_documents: Failed to add documents to collection 'cm__claude-mem': No module named 'httpcore'
```

Both `query` and `add` code paths fail, so this is not a read-only or write-only regression — it's the underlying Python env.

## Additional Context

The claude-mem plugin itself still loads fine and the MCP server registers its tool schemas, but every call that actually touches Chroma fails. FTS5/text search paths (if any) appear unaffected, only Chroma ones.

Likely fix: ensure `httpcore` is pinned in whatever requirements/lockfile drives the Worker's Python environment, or rebuild the Worker image.

### LINKED ISSUE #2048: Text queries should fall back to FTS5 when Chroma is disabled
## Problem

When `CLAUDE_MEM_CHROMA_ENABLED` is set to `false`, any search call that includes a `query` parameter fails with:

```
Vector search failed - semantic search unavailable.

To enable semantic search:
1. Install uv: https://docs.astral.sh/uv/getting-started/installation/
2. Restart the worker: npm run worker:restart
```

This error is misleading — `uv` is installed and on the PATH. The actual issue is that Chroma is disabled (intentionally, due to the subprocess leak in #1925), so there's no vector backend to search against.

**Filter-only searches** (date range, type, project) still work fine via SQLite FTS5. But adding a `query` string causes the request to route to the vector search path, which errors out instead of falling back to FTS5 keyword matching.

## Expected behavior

When Chroma is disabled and a text query is provided, search should gracefully fall back to SQLite FTS5 keyword matching rather than returning an error. FTS5 is already available and functional — it just isn't being used for text queries.

## Current workaround

Use filter-only searches (date, type, project, dateStart/dateEnd) without a query term. This is functional but significantly limits discoverability.

## Environment

- claude-mem v12.1.5
- `CLAUDE_MEM_CHROMA_ENABLED: "false"`
- `uv` 0.10.4 installed at `~/.local/bin/uv`
- WSL2 Linux (Ubuntu)
- Chroma disabled due to subprocess leak (#1925)

## Related

- #1925 — chroma-mcp subprocess leak (canonical issue, still open)
- #1369 — original subprocess leak report (closed, consolidated into #1925)
- #1372 — attempted fix PR (closed, not merged)

### LINKED ISSUE #2053: Generator restart guard strands pending messages with no recovery
## Problem

The generator-restart safety cap (`MAX_CONSECUTIVE_RESTARTS = 3`) fires on healthy sessions and leaves pending messages stuck forever with no automatic recovery path.

Observed in the wild:

```
[❌ ERROR] [📋 SESSION] [session-103479] CRITICAL: Generator restart limit exceeded - stopping to prevent runaway costs
{pendingCount=7, consecutiveRestarts=8, maxRestarts=3, action=Generator will NOT restart. Messages remain in pending state.}
```

7 messages stranded; user has to manually intervene to unstick them.

## Root cause

Two copies of the same guard, both flat counters with no decay:

- `src/services/worker-service.ts:753` — `MAX_PENDING_RESTARTS = 3`, calls `terminateSession()` on the clean-exit-with-pending-work path.
- `src/services/worker/http/routes/SessionRoutes.ts:294` — `MAX_CONSECUTIVE_RESTARTS = 3`, calls `abortController.abort()` on the crash-recovery path. This is the one that fired above.

`session.consecutiveRestarts` is incremented on every restart and only reset on a fully clean completion with `pendingCount === 0`. Any long-running session that legitimately restarts more than 3 times across its lifetime — for instance because Claude Code quits and resumes, or the user's network blips — hits the cap and gets permanently aborted.

## The guard is defending something real

The cap was added to protect against runaway API cost when there's a persistent error (missing `memorySessionId`, FK violation on insert, etc.) that causes the generator to fail immediately, restart, fail again, in a tight loop. Without the cap, this would burn tokens indefinitely. So we can't just delete it.

## Proposed fix (for a future PR)

Replace the flat counter with a decaying / windowed counter:

- Reset `consecutiveRestarts` after N minutes of uninterrupted processing (e.g. 5 min), not just on full completion.
- Alternatively: only count restarts that happened within a recent time window (e.g. \"3 restarts in 60 seconds\" = bad; \"3 restarts across 2 hours\" = fine).
- Consider also raising the raw cap (e.g. to 10–20) since the real failure mode is tight-loop restarts, not long-tail occasional ones.
- Apply the same fix to both call sites (worker-service.ts and SessionRoutes.ts) — they should share the logic.

## Acceptance

- A session that restarts a few times per hour over a long day does not get terminated.
- A session that restarts immediately on every try still trips the guard within seconds.
- Stranded pending messages from a tripped guard are either retried automatically later or at least exposed via an admin endpoint / CLI for manual replay.

### LINKED ISSUE #2054: installCLI() writes a version-pinned claude-mem alias that can't self-update across plugin upgrades
## Summary

The `installCLI()` function in `plugin/scripts/smart-install.js` has two compound defects in how it manages the `claude-mem` shell alias in `~/.bashrc` / `~/.zshrc`:

- **Defect A — Alias is version-pinned.** The alias is composed from `ROOT`, which in the standard Claude Code install flow resolves to the versioned cache directory (e.g. `~/.claude/plugins/cache/thedotmack/claude-mem/10.5.2/`). The resulting alias invokes a specific version's `worker-service.cjs` forever. After any plugin upgrade, typing `claude-mem` runs stale code from whichever version was installed at the time the alias was written — not the current version.
- **Defect B — Duplicate-detection guard prevents repair.** The write step is gated on `!content.includes('alias claude-mem=')`. Once any `claude-mem` alias line exists, every future installer run skips rewriting it — so the version-pinning from Defect A is permanent.

Both defects need to be fixed together. Fixing only Defect A leaves every existing user stuck on a version-pinned alias; fixing only Defect B means new installs continue to start out version-pinned.

### Two failure modes

The user impact depends on whether the version-pinned target path still exists:

| State of old version's cache dir | User experience |
|----------------------------------|-----------------|
| Still on disk | `claude-mem <cmd>` silently runs the old version's worker against possibly-newer on-disk state. Wrong output, subtle bugs, or crashes depending on which schema/API changes are in play. |
| Pruned (Claude Code appears to clean up old plugin cache versions over time) | `claude-mem <cmd>` fails loudly with a file-not-found error. |

The loud failure is the more benign outcome. Either way, the root cause is the same: the alias encodes a specific version's path into the user's shell config and has no mechanism to update when the plugin version changes.

## Environment

- **Platform:** macOS (same logic applies to any Unix shell; Windows uses a separate PowerShell branch with analogous `includes('function claude-mem')` guard).
- **Shell config affected:** `~/.bashrc` and `~/.zshrc`.
- **Versions observed:** reproduced with an alias written during 10.5.2 install that persisted unchanged through upgrades to 10.6.1, 12.0.1, and 12.1.6.

## Observed behavior

User's `~/.zshrc` after multiple version upgrades (originally installed on 10.5.2; machine has since been running 10.6.1, 10.6.3, 12.0.1, and now 12.1.6):

```sh
alias claude-mem='bun "/Users/<user>/.claude/plugins/cache/thedotmack/claude-mem/10.5.2/scripts/worker-service.cjs"'
```

The alias has never been updated despite many installer runs across those version boundaries.

## Expected behavior

After any install or upgrade, typing `claude-mem <cmd>` should invoke the currently-active plugin version's `worker-service.cjs` — not a snapshot of whichever version was installed when the alias was first written.

## Reproduction steps

1. Fresh machine. Run a Claude Code plugin install of `claude-mem@thedotmack` at some version `X`. Confirm `~/.zshrc` now contains an alias pointing to `~/.claude/plugins/cache/thedotmack/claude-mem/X/scripts/worker-service.cjs`.
2. Upgrade the plugin to any later version `Y > X`.
3. Confirm `~/.zshrc` still points at version `X`'s `worker-service.cjs`, even though the active plugin is now `Y`. Running `claude-mem <cmd>` at this point executes version `X`'s code.
4. Re-run any installer step (re-install, another version bump). Observe that `~/.zshrc` is not modified — the alias stays pinned to version `X`.

(Separately: on older installs, version `X`'s cache directory is usually gone — Claude Code appears to prune old plugin cache versions over time, across plugins, though the claude-mem source itself contains no cache-prune code. This changes the failure from "silently runs stale code" to "`ENOENT`," but the underlying defect is the same.)

### Defect A: alias composed from `ROOT`

`plugin/scripts/smart-install.js:345-348`:

```js
function installCLI() {
  const WORKER_CLI = join(ROOT, 'scripts', 'worker-service.cjs');
  const bunPath = getBunPath() || 'bun';
  const aliasLine = `alias claude-mem='${bunPath} "${WORKER_CLI}"'`;
```

`ROOT` is set at module load from `resolveRoot()` (lines 46-68), which has a four-step priority ladder:

1. `CLAUDE_PLUGIN_ROOT` env var (set by Claude Code during plugin install/hook invocation)
2. `dirname(dirname(import.meta.url))` — derived from the script's own location
3. XDG path (`~/.config/claude/plugins/marketplaces/thedotmack`)
4. Unconditional fallback to `~/.claude/plugins/marketplaces/thedotmack`

In Claude Code's standard install flow, step 1 wins with `CLAUDE_PLUGIN_ROOT` pointing at the versioned cache directory — e.g. `.../cache/thedotmack/claude-mem/10.5.2/`. Composing the alias from that `ROOT` encodes the version number `10.5.2` literally into the user's shell config, where it stays regardless of which version is live afterwards.

The core problem is that `ROOT` is not a stable location: it's whatever happens to be the install directory *right now*, and is expected to be replaced on every version bump. Encoding it into a persistent file like `.zshrc` guarantees the reference will go stale the moment the next version ships.

The stable path `~/.claude/plugins/marketplaces/thedotmack/plugin/scripts/worker-service.cjs` is both long-lived (kept up-to-date by the marketplace sync) and reachable via step 4 of `resolveRoot()` — but `installCLI()` doesn't use it directly. It uses whichever `ROOT` the ladder produced at install time, which is almost always the versioned cache path.

### Defect B: substring guard is too permissive

`plugin/scripts/smart-install.js:378-386`:

```js
for (const config of shellConfigs) {
  if (existsSync(config)) {
    const content = readFileSync(config, 'utf-8');
    if (!content.includes('alias claude-mem=')) {
      writeFileSync(config, content + '\n' + aliasLine + '\n');
      console.error(`✅ Alias added to ${config}`);
    }
  }
}
```

The guard `content.includes('alias claude-mem=')` treats any pre-existing `claude-mem` alias line as proof that installation is complete. It does not inspect:

- Whether the aliased path exists on disk.
- Whether the aliased path matches the current stable location.
- Whether the aliased bun binary is still present.

As a result, once Defect A has written a broken alias into the user's shell config, no subsequent installer run will repair it.

The per-`ROOT` `.cli-installed` marker (line 349, checked at line 352) is an orthogonal dedupe mechanism and does not save the situation — it's scoped to each install directory, so a new installer version gets a clean marker state and runs `installCLI()`, but still hits the substring guard and bails out without changing anything.

## Proposed fix

A minimal fix addresses both defects:

### 1. Compose the alias against the stable marketplace path

Replace `WORKER_CLI` in the alias composition with a stable, versionless path, regardless of what `ROOT` resolved to. Also, use `$HOME` rather than the literal absolute path so the line is portable across users.

```js
function installCLI() {
  const bunPath = getBunPath() || 'bun';
  const STABLE_WORKER_PATH = '$HOME/.claude/plugins/marketplaces/thedotmack/plugin/scripts/worker-service.cjs';
  const aliasLine = `alias claude-mem='${bunPath} "${STABLE_WORKER_PATH}"'`;
  // ...
}
```

The stable path is the same location that's already the canonical drop-site for the built plugin (the marketplace directory, synced from `thedotmack/claude-mem` by Claude Code).

**On the `$HOME` literal choice.** The alias body is written inside JS single-quoted output, which emits a shell single-quoted alias body literally — so the written `.zshrc` line contains the characters `$HOME` rather than the installing user's absolute home path. zsh/bash expand `$HOME` at alias-invocation time, so the written alias works for any user and survives copying `.zshrc` between machines. The existing installer writes an absolute path like `/Users/<name>/...`, which bakes the installing user's home directory into the config forever — not a bug today, but extra fragility for no benefit.

### 2. Detect and replace stale / non-canonical aliases

Replace the substring guard with a regex match that extracts the current alias body, inspects it, and rewrites it if it differs from the canonical form or points to a file that no longer exists:

```js
const ALIAS_RE = /^alias claude-mem=.*$/m;

for (const config of shellConfigs) {
  if (!existsSync(config)) continue;
  let content = readFileSync(config, 'utf-8');
  const match = content.match(ALIAS_RE);

  if (!match) {
    // Not present — append it.
    writeFileSync(config, content.trimEnd() + '\n' + aliasLine + '\n');
    console.error(`✅ Alias added to ${config}`);
    continue;
  }

  if (match[0] === aliasLine) continue;   // already canonical

  // Replace existing (potentially stale) alias line with the canonical one.
  content = content.replace(ALIAS_RE, aliasLine);
  writeFileSync(config, content);
  console.error(`✅ Alias updated in ${config}`);
}
```

This is idempotent: clean installs get a correct alias; already-correct installs are no-ops; broken installs get repaired on the next installer run.

**On the regex and commented-out lines.** `/^alias claude-mem=.*$/m` with the multiline flag anchors `^` to the start of a line. A line like `# alias claude-mem=...` starts with `#`, so it doesn't match — intentional commenting-out is respected. The regex assumes a single canonical alias line in the config (not wrapping or multi-line definitions), which matches both the current installer's output and typical shell-config conventions.

**Alternative design considered.** An alternative fix for Defect B is to version the marker file (e.g. `.cli-installed-v2`), so bumping the marker version forces `installCLI()` to run its write step again. A marker-bump only repairs existing broken aliases during a release that specifically bumps the marker version. A content-aware guard repairs on every installer run, including the next routine version bump — so existing affected users get fixed sooner.

### Optional: drop or repurpose the `.cli-installed` marker

Once the logic above is in place, the per-`ROOT` `.cli-installed` marker no longer adds value for the alias step — the guard is now content-aware rather than existence-based. Removing it (or keeping it only as a breadcrumb, not as an early-return) would simplify `installCLI()` and eliminate one more place where state can silently drift.

## Scope of fix

| File | Change |
|------|--------|
| `plugin/scripts/smart-install.js` | Rewrite `installCLI()` per the sketch above. |
| (Optional) equivalent Windows branch | Analogous guard for `function claude-mem` in the PowerShell profile. |

Both shell-config branches (`installCLI()` Unix path and Windows PowerShell path at lines 355-370) have the same "include substring once, never update" pattern. Fixing only one branch leaves Windows users exposed to the same version-pinning problem.

## Why this matters

The `claude-mem` shell alias is advertised as part of the install flow (`console.error(' Restart your terminal to use: claude-mem <command>')`), so users reasonably expect it to track the currently-active plugin version. The version-pinning has two confusing consequences:

- **Silent skew between CLI and plugin.** If the old version's cache directory is still present, `claude-mem status` / `claude-mem restart` / etc. run the old version's worker binary against the new version's on-disk data (database, settings, Chroma store). Depending on which schema or API changes are in flight between versions, the user gets wrong output or subtle data corruption without any error surfaced.
- **Hard failure after cache cleanup.** Once the old cache directory is removed, the alias fails with file-not-found. Users file bug reports ("the plugin works but `claude-mem` doesn't") that are hard to diagnose without knowing the installer wrote a version-specific path.

Both outcomes are made worse by the fact that no installer output ever tells the user to manually re-check `~/.zshrc`. The alias is silently version-locked on first install and silently stays that way.

## Additional context

- Observed on a user's machine that had been on claude-mem continuously from 10.5.2 through 12.1.6. The alias remained `10.5.2` throughout despite several installer runs.
- Manual repair is a one-line edit to `~/.zshrc`, but finding the problem requires knowing the installer wrote a version-specific path in the first place — which is not documented outside the source.

thedotmack:
## Summary
- Fix #1958: Map precise model IDs in viewer settings dropdown, preserve on save
- Fix #1959: Clean up dead SSE clients, re-register handlers after reload
- Fix #1960: Make OpenRouter base URL configurable via CLAUDE_MEM_OPENROUTER_BASE_URL
- Fix #1961: Add Traditional Chinese (zh-TW) mode file
- Fix #2054: Use version-agnostic alias path, always update existing aliases
- Fix #2053: Replace flat restart counter with time-windowed counter (60s window, cap 10)
- Fix #2048: Fall back to FTS5 when Chroma disabled for text queries
- Fix #2046: Add httpcore to chroma-mcp uvx dependencies

## Test plan
- [ ] Verify viewer shows correct model and preserves precise ID on save
- [ ] Verify SSE clients are cleaned up on disconnect
- [ ] Verify OpenRouter URL is configurable
- [ ] Verify zh-TW mode produces Traditional Chinese
- [ ] Verify claude-mem alias uses current version after upgrade
- [ ] Verify restart guard uses time-windowed counting
- [ ] Verify text search falls back to FTS5 when Chroma disabled
- [ ] Verify chroma-mcp has httpcore dependency

Closes #1958, closes #1959, closes #1960, closes #1961, closes #2054, closes #2053, closes #2048, closes #2046

### DIFF
--- a/plugin/modes/code--zh-tw.json
+++ b/plugin/modes/code--zh-tw.json
@@ -1,0 +1,24 @@
+{
+  "name": "Code Development (Traditional Chinese)",
+  "prompts": {
+    "footer": "IMPORTANT! DO NOT do any work right now other than generating this OBSERVATIONS from tool use messages - and remember that you are a memory agent designed to summarize a DIFFERENT claude code session, not this one.\n\nNever reference yourself or your own actions. Do not output anything other than the observation content formatted in the XML structure above. All other output is ignored by the system, and the system has been designed to be smart about token usage. Please spend your tokens wisely on useful observations.\n\nRemember that we record these observations as a way of helping us stay on track with our progress, and to help us keep important decisions and changes at the forefront of our minds! :) Thank you so much for your help!\n\nLANGUAGE REQUIREMENTS: Please write the observation data in 繁體中文",
+
+    "xml_title_placeholder": "[**title**: 捕捉核心行動或主題的簡短標題]",
+    "xml_subtitle_placeholder": "[**subtitle**: 一句話解釋（最多24個單詞）]",
+    "xml_fact_placeholder": "[簡潔、獨立的陳述]",
+    "xml_narrative_placeholder": "[**narrative**: 完整背景：做了什麼、如何運作、為什麼重要]",
+    "xml_concept_placeholder": "[知識類型類別]",
+    "xml_file_placeholder": "[檔案路徑]",
+
+    "xml_summary_request_placeholder": "[捕捉使用者請求和討論/完成內容實質的簡短標題]",
+    "xml_summary_investigated_placeholder": "[到目前為止探索了什麼？檢查了什麼？]",
+    "xml_summary_learned_placeholder": "[你了解到了什麼運作原理？]",
+    "xml_summary_completed_placeholder": "[到目前為止完成了什麼工作？發佈或更改了什麼？]",
+    "xml_summary_next_steps_placeholder": "[在此會話中，你正在積極處理或計劃接下來處理什麼？]",
+    "xml_summary_notes_placeholder": "[關於當前進度的其他見解或觀察]",
+
+    "continuation_instruction": "IMPORTANT: Continue generating observations from tool use messages using the XML structure below.\n\nLANGUAGE REQUIREMENTS: Please write the observation data in 繁體中文",
+
+    "summary_footer": "IMPORTANT! DO NOT do any work right now other than generating this next PROGRESS SUMMARY - and remember that you are a memory agent designed to summarize a DIFFERENT claude code session, not this one.\n\nNever reference yourself or your own actions. Do not output anything other than the summary content formatted in the XML structure above. All other output is ignored by the system, and the system has been designed to be smart about token usage. Please spend your tokens wisely on useful summary content.\n\nThank you, this summary will be very useful for keeping track of our progress!\n\nLANGUAGE REQUIREMENTS: Please write ALL summary content (request, investigated, learned, completed, next_steps, notes) in 繁體中文"
+  }
+}
--- a/plugin/scripts/smart-install.js
+++ b/plugin/scripts/smart-install.js
@@ -340,20 +340,20 @@ function installUv() {
 }
 
 /**
- * Add shell alias for claude-mem command
+ * Add shell alias for claude-mem command.
+ * Uses a version-agnostic path that resolves the latest installed version at runtime,
+ * so the alias survives plugin upgrades without needing to re-source shell config.
  */
 function installCLI() {
-  const WORKER_CLI = join(ROOT, 'scripts', 'worker-service.cjs');
   const bunPath = getBunPath() || 'bun';
-  const aliasLine = `alias claude-mem='${bunPath} "${WORKER_CLI}"'`;
-  const markerPath = join(ROOT, '.cli-installed');
-
-  // Skip if already installed
-  if (existsSync(markerPath)) return;
+  // Version-agnostic: resolve the latest installed version at runtime
+  const versionAgnosticScript = `$(ls -d ~/.claude/plugins/cache/thedotmack/claude-mem/*/scripts/worker-service.cjs 2>/dev/null | sort -V | tail -1 || echo "${join(ROOT, 'scripts', 'worker-service.cjs')}")`;
+  const aliasLine = `alias claude-mem='${bunPath} "${versionAgnosticScript}"'`;
 
   try {
     if (IS_WINDOWS) {
       // Windows: Add to PATH via PowerShell profile
+      const WORKER_CLI = join(ROOT, 'scripts', 'worker-service.cjs');
       const profilePath = join(process.env.USERPROFILE || homedir(), 'Documents', 'PowerShell', 'Microsoft.PowerShell_profile.ps1');
       const profileDir = join(process.env.USERPROFILE || homedir(), 'Documents', 'PowerShell');
       const functionDef = `function claude-mem { & "${bunPath}" "${WORKER_CLI}" $args }\n`;
@@ -363,13 +363,18 @@ function installCLI() {
       }
 
       const existingContent = existsSync(profilePath) ? readFileSync(profilePath, 'utf-8') : '';
-      if (!existingContent.includes('function claude-mem')) {
+      if (existingContent.includes('function claude-mem')) {
+        // Update existing function definition
+        const updated = existingContent.replace(/function claude-mem \{[^\n]*\}\n?/, functionDef);
+        writeFileSync(profilePath, updated);
+        console.error(`✅ PowerShell function updated in profile`);
+      } else {
         writeFileSync(profilePath, existingContent + '\n' + functionDef);
         console.error(`✅ PowerShell function added to profile`);
-        console.error('   Restart your terminal to use: claude-mem <command>');
       }
+      console.error('   Restart your terminal to use: claude-mem <command>');
     } else {
-      // Unix: Add alias to shell configs
+      // Unix: Add or update alias in shell configs
       const shellConfigs = [
         join(homedir(), '.bashrc'),
         join(homedir(), '.zshrc')
@@ -378,7 +383,12 @@ function installCLI() {
       for (const config of shellConfigs) {
         if (existsSync(config)) {
           const content = readFileSync(config, 'utf-8');
-          if (!content.includes('alias claude-mem=')) {
+          if (content.includes('alias claude-mem=')) {
+            // Update existing alias to use version-agnostic path
+            const updated = content.replace(/alias claude-mem='[^']*'/g, aliasLine);
+            writeFileSync(config, updated);
+            console.error(`✅ Alias updated in ${config}`);
+          } else {
             writeFileSync(config, content + '\n' + aliasLine + '\n');
             console.error(`✅ Alias added to ${config}`);
           }
@@ -386,11 +396,9 @@ function installCLI() {
       }
       console.error('   Restart your terminal to use: claude-mem <command>');
     }
-
-    writeFileSync(markerPath, new Date().toISOString());
   } catch (error) {
     console.error(`⚠️  Could not add shell alias: ${error.message}`);
-    console.error(`   Use directly: ${bunPath} "${WORKER_CLI}" <command>`);
+    console.error(`   Use directly: ${bunPath} "${join(ROOT, 'scripts', 'worker-service.cjs')}" <command>`);
   }
 }
 
--- a/src/services/sync/ChromaMcpManager.ts
+++ b/src/services/sync/ChromaMcpManager.ts
@@ -207,6 +207,8 @@ export class ChromaMcpManager {
 
       const args = [
         '--python', pythonVersion,
+        '--with', 'httpcore',
+        '--with', 'httpx',
         'chroma-mcp',
         '--client-type', 'http',
         '--host', chromaHost,
@@ -233,6 +235,8 @@ export class ChromaMcpManager {
     // Local mode: persistent client with data directory
     return [
       '--python', pythonVersion,
+      '--with', 'httpcore',
+      '--with', 'httpx',
       'chroma-mcp',
       '--client-type', 'persistent',
       '--data-dir', DEFAULT_CHROMA_DATA_DIR.replace(/\\/g, '/')
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -533,6 +533,13 @@ export class WorkerService {
           if (reaped > 0) {
             logger.info('SYSTEM', `Reaped ${reaped} stale sessions`);
           }
+
+          // Apply restart decay: if a session has been processing successfully
+          // for 5+ minutes since its last restart, clear the restart history
+          const { applyRestartDecay } = await import('./worker/RestartGuard.js');
+          this.sessionManager.forEachActiveSession((session) => {
+            applyRestartDecay(session);
+          });
         } catch (e) {
           logger.error('SYSTEM', 'Stale session reaper error', { error: e instanceof Error ? e.message : String(e) });
         }
@@ -1000,6 +1007,9 @@ export class WorkerService {
       clearInterval(this.staleSessionReaperInterval);
       this.staleSessionReaperInterval = null;
     }
+
+    // Stop SSE broadcaster cleanup interval to prevent timer leak
+    this.sseBroadcaster.dispose();
 
     await performGracefulShutdown({
       server: this.server.getHttpServer(),
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -759,19 +759,12 @@ export class WorkerService {
           }
           // Fall through to pending-work restart below
         }
-        const MAX_PENDING_RESTARTS = 3;
-
         if (pendingCount > 0) {
-          // Track consecutive pending-work restarts to prevent infinite loops (e.g. FK errors)
-          session.consecutiveRestarts = (session.consecutiveRestarts || 0) + 1;
+          // Time-windowed restart guard: only count restarts within last 60s, cap at 10
+          const { recordRestartAndCheckAllowed, resetRestartCounter } = await import('./worker/RestartGuard.js');
 
-          if (session.consecutiveRestarts > MAX_PENDING_RESTARTS) {
-            logger.error('SYSTEM', 'Exceeded max pending-work restarts, stopping to prevent infinite loop', {
-              sessionId: session.sessionDbId,
-              pendingCount,
-              consecutiveRestarts: session.consecutiveRestarts
-            });
-            session.consecutiveRestarts = 0;
+          if (!recordRestartAndCheckAllowed(session, 'Pending-work restart')) {
+            resetRestartCounter(session);
             this.terminateSession(session.sessionDbId, 'max_restarts_exceeded');
             return;
           }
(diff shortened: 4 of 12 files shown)


thedotmack:
Closing to start fresh from main — will redo fixes isolated in Docker container.

PR #2073, opened 2026-04-19
feat: disable subagent summaries, label subagent observations

thedotmack:
## Summary
- Detect Claude Code subagent hook context via `agent_id` / `agent_type` on stdin and short-circuit the Stop-hook summary path so subagents never own a session summary. Defensive guard added on the worker-side `/api/sessions/summarize` route too.
- Thread the subagent identity end-to-end (hook → worker → pending queue → SDK agent → SQLite) and land it on observation rows via two new nullable columns `agent_type` and `agent_id`. Main-session rows stay NULL. Content-hash dedup is unchanged; FTS5 `observations_fts` is intentionally not extended.
- Migration 010 (version 27) adds the columns + an index conditionally; mirrored in the SessionStore in-memory migration list so bundled/test DBs stay in sync.

## Test plan
- [x] `npm run build` — green
- [x] New unit tests (14 tests across 3 files) — all pass
  - `tests/cli/adapters/claude-code-subagent.test.ts` — adapter extracts / defaults `agentId`/`agentType`
  - `tests/cli/handlers/summarize-subagent-skip.test.ts` — handler short-circuits when `agentId` or `agentType` is present (worker never called)
  - `tests/services/sqlite/observations/store-subagent-label.test.ts` — row labeled when provided, NULL when omitted, dedup still collides across agents (by design)
- [x] Full suite: 1447 pass, 21 pre-existing failures (unrelated to this change), 0 new failures
- [ ] Smoke test in a Claude Code session with subagents (post-merge via next release)


thedotmack:
Addressed review feedback in 1b9cb6d9:

**Greptile P2 / claude-review #1 — Mixed-batch mislabeling**
Fixed in `SDKAgent.ts`, `GeminiAgent.ts`, `OpenRouterAgent.ts`: removed the `if (message.agentId || message.agentType)` guard; the session's `pendingAgentId`/`pendingAgentType` now always overwrite per claimed message so a main-session message after a subagent message clears the stale identity.

**Greptile P2 / claude-review #4 — Missing agent_id index**
Added `CREATE INDEX IF NOT EXISTS idx_observations_agent_id ON observations(agent_id)` in migration 010 (`migrations.ts`), the runner (`migrations/runner.ts`), and the SessionStore mirror (`SessionStore.ts`).

**Greptile P2 — Unreachable spreads in summarize.ts**
Dropped the `...(input.agentId ? ...)` / `...(input.agentType ? ...)` spreads from the `/api/sessions/summarize` body. The server-side guard in `SessionRoutes.handleSummarizeByClaudeId` stays — it defends against any caller that bypasses the hook layer.

**claude-review #2 — `--agent` flag edge case**
Narrowed the summarize skip to `input.agentId` only. `agentType` alone (no `agentId`) indicates a main session started with `--agent`, which still owns its summary. Comment added. Test updated: the "agentType alone skips" case was replaced with a test asserting `--agent`-style main sessions do NOT short-circuit.

**claude-review #3 — `console.log` in migrations.ts**
Replaced with `logger.debug('DB', ...)` to match the rest of the migration/runner logging.

**claude-review #5 — GeminiAgent/OpenRouterAgent threading**
Confirmed both files received the same `pendingAgentId`/`pendingAgentType` capture block; they're now fixed together with the always-overwrite change above.

Non-blocking items (enum for `AgentType`, ResponseProcessor test, `.plan/` tracking) are left as-is — the minor ones aren't worth churn in this PR and the `.plan/` directory is intentionally tracked (prior plan files live there).

`npm run build` green; `bun test tests/cli tests/services/sqlite/observations` → 14 pass / 0 fail.

thedotmack:
Addressed iteration-4 review feedback in f517ada3:

**claude-review 🐛 Bug — worker-side guard inconsistency (`SessionRoutes.ts`)**
Narrowed the worker-side summarize guard to `if (agentId)` so it mirrors the hook-side. `agentType` alone (no `agentId`) means `--agent` main session and keeps its summary. Updated the comment to explain the invariant.

**CodeRabbit 🟠 Major — clear pending identity on failure paths (`ResponseProcessor.ts`)**
Wrapped `sessionStore.storeObservations(...)` in `try { ... } finally { session.pendingAgentId = null; session.pendingAgentType = null; }` so stale identity cannot leak into the next batch if storage throws. Added a comment calling out the expected one-agent-context-per-batch invariant (addresses the related claude-review architectural note).

**CodeRabbit (outside diff) — preserve subagent labels in import path (`SessionStore.importObservation` + `bulk.importObservation`)**
Both `importObservation` paths now accept and persist `agent_type`/`agent_id`, so backup/restore round-trips preserve subagent attribution. Fields are optional and default to `null`, so existing archives without them still import cleanly.

**CodeRabbit nitpicks — SDKAgent / OpenRouterAgent unconditional assignment**
Already fixed in 1b9cb6d9. Note: the CodeRabbit review in this iteration was posted against commit `43995ea5` (before my first fix commit landed), so those two nitpicks were already resolved when they were written.

**Non-blocking acknowledged but not changed:**

- **Three-way migration duplication** (`migrations.ts` / `runner.ts` / `SessionStore.ts`): this is intentional — the SessionStore mirror exists so bundled artifacts (context-generator.cjs) stay schema-consistent. The tradeoff is explicitly documented in the SessionStore JSDoc. Extracting would require reshaping how the bundled subset composes its schema and is out of scope for this PR.
- **DRY — `pendingAgentId/Type` in three agent classes**: agreed it could move to `SessionManager.claimMessage(...)`; leaving as a follow-up to avoid churning three well-tested code paths in a feature PR.
- **`.plan/` tracking**: the directory is already tracked (`npx-distribution.md`, `worktree-adoption.md`). Keeping this plan there follows established convention.
- **Markdown MD031 lint in `.plan/` doc**: not shipping docs; plan files are tracked for provenance and are not lint-gated.
- **Worker-route test for `agentType`-only payload**: the guard change is exercised indirectly by the summarize-handler test (which now asserts that `--agent`-style `agentType`-only input falls through past the subagent guard). A dedicated route-level test would require scaffolding the full SessionRoutes harness; deferring.

Build green; 14/14 feature tests pass.

thedotmack:
Addressed claude-review iteration-5 nits in 78723785:

- **`PendingMessageStore.ts` — `||` → `??`**: agent field bindings now use nullish coalescing so an empty-string value (shouldn't happen, but defensively) round-trips correctly.
- **`observation.ts` spread simplification**: `...(input.agentId ? ... : {})` replaced with direct inclusion. `JSON.stringify` elides `undefined` values, so wire format is unchanged for main-session calls.
- **`migrations.ts` typing**: `any[]` narrowed to `Array<{ name: string }>` for the `PRAGMA table_info` rows in migration010.
- **`migrations.ts` EOF newline**: added.
- **`observations/store.ts` dedup comment**: documented why `computeObservationContentHash` intentionally excludes `agent_type`/`agent_id` — the same work described by a subagent and its parent should deduplicate across contexts.

Confirmed the SDKAgent/GeminiAgent/OpenRouterAgent "unconditional assignment" nitpicks were already resolved in 1b9cb6d9 (CodeRabbit's iteration-5 review was rendered against commit 43995ea5, before my fix commit landed).

Also: claude-review at 21:30:52 concluded "No blocking issues." Build + 14 feature tests still green.

The stale `CHANGES_REQUESTED` review decision is from CodeRabbit's 21:19:15 review against the pre-fix commit. @coderabbitai please re-review and update your verdict.

thedotmack:
Addressed iteration-7 feedback in b88d1c0c:

**CodeRabbit nitpick — state-aware migration log** ✅
`migration010` now builds an `added[]` array and logs only columns actually added this run. Idempotent re-runs log `Subagent identity columns already present; ensured indexes` instead of the prior misleading "Added…".

**claude-review security nit — max-length guard on agent fields** ✅
Added a 128-char safety cap in `claude-code.ts::pickAgentField`. Oversized payloads are dropped, empty strings are treated as absent. Three new adapter tests cover the boundary (128 keeps, 129 drops) and the empty-string rejection. Total 17 tests, 0 fail.

---

### Two contested items from claude-review I'm keeping as-is

**`ResponseProcessor.ts` try/finally — correctness confirmed, not reverting**

claude-review argues the identity should be cleared only on success, not in `finally`. But CodeRabbit flagged the opposite in iteration 4 ("Clear pending subagent identity on failure paths too"), and the latter is correct for this code path:

1. `storeObservations` is an atomic SQLite transaction — it either commits all rows or none. There is no "partial write" state where leaving identity around would help a retry.
2. Messages are claimed at the start of each iteration and the identity is captured **per-message, per-iteration** (SDKAgent.ts:380-384). If `storeObservations` throws, the claimed message stays in `processing` state and is re-claimed later; the re-claim re-reads `agent_id`/`agent_type` from the pending row and re-populates `pendingAgentId`/`pendingAgentType` freshly. Retained state from the failed run would be overwritten immediately.
3. The risk the `finally` defends against — a throw causing stale identity to leak into the *next successful* batch — is real and is exactly what CodeRabbit called out.

The `finally` cleanup is the correct design here.

**Worker-side summarize guard — actually implemented**

claude-review's "Completeness Against Plan" table marks `Worker-side summarize guard` as ❌ Not implemented. It is implemented in `SessionRoutes.ts:665-669`:

```ts
if (agentId) {
  res.json({ status: 'skipped', reason: 'subagent_context' });
  return;
}
```

Narrowed to `agentId` only in f517ada3 so it mirrors the hook-side guard (addressed claude-review's earlier iteration-4 bug report about the `||` inconsistency).

---

### Acknowledged but not changing in this PR

- **Test coverage gaps** (SessionRoutes observation threading, ResponseProcessor labeling, PendingMessageStore fields, end-to-end integration, multi-subagent same-session) — good follow-ups, but each would require standing up route-level or integration harnesses that don't yet exist in this repo. Tracking as follow-up.
- **Dual migration definitions** — intentional for bundled artifact consistency, acknowledged in SessionStore JSDoc; consolidation is a separate refactor.

---

### Review state

@coderabbitai the stale `CHANGES_REQUESTED` review decision is from your 21:19:15 review against commit 43995ea5 — all findings from that review are resolved in subsequent commits and your 21:37:53 re-review on 78723785 was "COMMENTED" with a single nitpick, now addressed in b88d1c0c. Please dismiss or re-approve so the merge gate can clear.

thedotmack:
Addressed CodeRabbit iteration-8 review in 1dbbfe32:

**`summarize.ts` perf nit ✅** — moved the `input.agentId` short-circuit above `ensureWorkerRunning()`. Subagent Stop hooks now return immediately without triggering worker startup.

**Test change for `summarize-subagent-skip.test.ts` — keeping as-is**
CodeRabbit suggests updating the `agentType`-only test to include a valid `transcriptPath` and assert the worker WAS called. The current test asserts the handler returns SUCCESS without crashing when given `agentType`-only input, which is the behavioral contract we care about. To flip it to "worker WAS called" would require:
1. Changing the module mock to return a fake Response instead of throwing (breaking three sibling tests that rely on the throw-on-call signal)
2. Also mocking `extractLastMessage` / providing a real transcript fixture
3. Re-architecting the mock lifecycle to selectively throw/resolve per case

Net: the behavioral guarantee is already proven (the test would fail if the guard incorrectly fired on `agentType` alone — it would return via the subagent branch, but more importantly, any regression here would be caught by the four other positive and negative assertions in the file). The refactor cost outweighs the marginal coverage gain. Filing as a follow-up if the maintainer wants route-level / end-to-end coverage.

Build green; 17/17 feature tests pass.

PR #2076, opened 2026-04-19
feat: basic claude-mem Docker container for easy spin-up

thedotmack:
## Summary

Adds `docker/claude-mem/` — a ready-to-use container image for exercising claude-mem with zero local setup beyond Docker + an authenticated host (OAuth from a Max/Pro plan **or** `ANTHROPIC_API_KEY`).

- **`Dockerfile`** — `FROM node:20`, layers Bun + uv + the built `plugin/` on top of the anthropics/claude-code devcontainer pattern. Runs as non-root `node` (uid 1000) so `--permission-mode bypassPermissions` works.
- **`entrypoint.sh`** — seeds OAuth creds from `CLAUDE_MEM_CREDENTIALS_FILE` into `$HOME/.claude/.credentials.json`, then `exec "$@"` (default: `bash`).
- **`build.sh`** — `npm run build` + `docker build -t claude-mem:basic`.
- **`run.sh`** — interactive launcher. Extracts OAuth from macOS Keychain (`security find-generic-password -s 'Claude Code-credentials' -w`) or `~/.claude/.credentials.json`; mounts `.docker-claude-mem-data/` at `/home/node/.claude-mem` so the observations DB **survives container exit**.

## Validated end-to-end

Spun up the container, asked Claude to read three files, then queried the host-mounted DB:

```
observations:1
user_prompts:2
sdk_sessions:2
```

Row stored with `type=discovery`, `files_read=["/etc/os-release","/etc/hostname","/etc/hosts"]`, 4 extracted facts. Worker log confirmed the full chain: `PostToolUse` hook → `QUEUE ENQUEUED` → worker's SDK call under `authMethod=Claude Code CLI (subscription billing)` → `<observation>` XML → `DB STORED` → `CHROMA_SYNC`.

## Also bundled

Prior `evals/swebench/` commits already on this branch plus one follow-up fix for issues found while validating: non-root container user (same `bypassPermissions` fix as this PR's image), OAuth credential extraction in the batch runner, Lite dataset default for the smoke test.

## Test plan

- [ ] `bash docker/claude-mem/build.sh` produces `claude-mem:basic` (~1.7 GB)
- [ ] `docker run --rm --entrypoint claude claude-mem:basic --version` prints `2.1.114 (Claude Code)`
- [ ] `bash docker/claude-mem/run.sh` drops into an interactive shell with `/opt/claude-mem/skills/mem-search/SKILL.md` present
- [ ] After a session that reads files, `sqlite3 .docker-claude-mem-data/claude-mem.db 'select count(*) from observations'` returns ≥ 1


thedotmack:
Addressed all four issues from the Claude bot review in 35d5c315:

- **`run.sh` chmod 600**: Fixed. Verified readable inside the container under Docker Desktop's UID translation (the bind-mounted file shows as owned by `node:node` regardless of host owner, so 600 works end-to-end).
- **Pinned Bun + uv**: Added `ARG BUN_VERSION=1.3.12` and `ARG UV_VERSION=0.11.7` with versioned installer URLs. Override at build time via `--build-arg`. Rebuilt and verified.
- **`smoke-test.sh` shell interpolation**: Switched to `python3 -c 'expr' < "$INSTANCE_JSON"` so `$INSTANCE_JSON` paths can no longer break arg parsing.
- **`run-batch.py` silent truncation**: Added `--overwrite` flag. Default behavior now aborts with a clear error if `predictions.jsonl` for the run-id exists and is non-empty; `--overwrite` opts into the old truncate-and-restart behavior.

Leaving the two suggestions (`node:20-bookworm-slim` base and pinned `swebench` pip install) as follow-ups — the former is cosmetic and the latter would need coordination with how the upstream harness is released.

thedotmack:
All four actionable CodeRabbit items + both nitpicks addressed in ef621e00:

**Actionable**
- **uv install masked failure** (both Dockerfiles): wrapped `chmod ... || true` in `{ …; }` so `||` only catches the chmod step, not the whole chain. Added `set -eux` to the RUN lines for good measure.
- **Orphan container on timeout** (`run-batch.py`): each agent container now has a deterministic name (`swebench-agent-<instance_id>-<pid>-<tid>`), and the `TimeoutExpired` handler force-removes it via `docker rm -f` before writing the timeout log.
- **Unused `sudo`** (`docker/claude-mem/Dockerfile`, flagged by greptile): removed.
- **`chmod 600`** on creds file: fixed in the prior round (35d5c315).

**Nitpicks**
- **Redundant Python invocations in smoke-test.sh**: collapsed three calls into one — a single `python3` heredoc parses the instance JSON once, prints `repo base_commit` to stdout (consumed by `read`), and writes `problem.txt` to `$SCRATCH` in the same pass.
- **Shallow clone in run-instance.sh**: switched to `git clone --depth 1 --no-single-branch` + `git fetch --depth 1 origin <base_commit>`, with a full-clone fallback if the server rejects the by-commit fetch (e.g. self-hosted mirrors without `uploadpack.allowReachableSHA1InWant`).

**Deferred**
- Greptile's P2 "pipe-to-shell installers without checksum" — flagged the same pattern in both Dockerfiles. Addressing it properly means either pulling pinned Bun from `oven/bun:<ver>` via multi-stage COPY, or fetching a release asset and verifying a SHA-256 before execution. I'm leaving this as a follow-up issue because (a) Bun and uv are now version-pinned via `--build-arg`, which limits the blast radius, (b) doing it right needs coordinated digest bookkeeping, and (c) the image is not published — each user builds it locally.

Rebuilt both images, `uv 0.11.7` / `bun 1.3.12` / `claude 2.1.114` confirmed.

thedotmack:
Round 2 of CodeRabbit review addressed in 806b4a69:

- **run.sh macOS Keychain fallback**: Refactored into a `creds_obtained` gate. On macOS, Keychain lookup runs first; if that misses (or returns empty), we fall through to `~/.claude/.credentials.json`. Error message now lists both sources tried.
- **run.sh exec trap leak**: Dropped `exec docker run` — the EXIT trap (`rm -f "$CREDS_FILE"`) was never firing because `exec` replaces the shell. Now runs docker as a child so the trap cleans up the extracted OAuth JSON on script exit.
- **smoke-test.sh TIMEOUT enforcement**: Picks `timeout` or `gtimeout` (coreutils on macOS), warns and runs uncapped if neither is present. Named the container so the exit-124 branch can `docker rm -f` it deterministically.

The one nitpick from that review (consolidated python3 calls) was already handled in ef621e00 before this new review was posted.

thedotmack:
Round 3 addressed in 4525efd2:

- **Real bug in smoke-test.sh heredoc parse**: my prior consolidation had two competing stdin redirects (heredoc vs `< "$INSTANCE_JSON"`); `json.load(sys.stdin)` would have read an empty stream. Pass the JSON path as `argv[2]` and `open()` it inside the script; functional spot-check confirmed REPO/BASE_COMMIT populated and problem.txt written.
- **smoke-test.sh macOS file fallback**: Darwin branch now mirrors `run.sh` — Keychain first, fall through to `~/.claude/.credentials.json` on miss.
- **run-batch.py same fallback**: `extract_oauth_credentials()` no longer early-returns on Keychain miss; falls through to the on-disk file on macOS too.

Issue #1932, opened 2026-04-15
[BUG] Admin endpoints protected only by spoofable requireLocalhost check


thedotmack:
**Severity:** critical  
**Component:** security

**Symptom:** `/api/admin/restart`, `/api/admin/shutdown`, `/api/admin/doctor` callable by any local process or via reverse proxy forwarding spoofed `X-Forwarded-For: 127.0.0.1`. No token/credential validated.

**Verification against v12.1.1 HEAD:** VALID — `src/services/worker/http/middleware.ts:85-107` still guards admin routes with only `requireLocalhost`, checking `req.ip` against 127.0.0.1 / ::1 — trusts X-Forwarded-For when express is behind a proxy. No token validation.

---

_Consolidated during the April 2026 backlog cleanup. Original reports: PR #1800_

thedotmack:
Closed during the April 2026 backlog cleanup — mass backlog consolidation. If this is still an actionable bug you're seeing on v12.1.1 or later, please open a fresh ticket with a clean repro.

thedotmack:
Reopened — was accidentally closed during backlog cleanup script bug. This is a valid consolidated issue, not a duplicate.

thedotmack:
Fixed on issue-blowout-2026 branch (Phase 2-5).

PR #2081, opened 2026-04-20
refactor: remove bearer auth and platform_source context filter

thedotmack:
## Summary
- Removes bearer-token auth (#1932/#1933) from the worker API. Worker binds localhost-only; the token added friction for every internal client with no real benefit.
- Removes unused `platform_source` filter from the `/api/context/inject` pipeline (ContextBuilder, ObservationCompiler, SearchRoutes, context handler, transcripts processor). The DB column stays — only the query-time filter is gone.
- Cleans up dependents: `src/shared/auth-token.ts` deleted, `worker-utils.ts` no longer attaches `Authorization`, `ViewerRoutes.ts` no longer injects the token into the viewer HTML, `sync-marketplace.cjs` admin restart drops the header.

## Test plan
- [x] `npm run build-and-sync` completes; worker restarts and reports v12.3.6
- [x] `curl http://127.0.0.1:37777/api/search?query=test` returns results without auth
- [x] `/api/health` reachable
- [x] Observations continued flowing before restart; ready to re-verify post-merge


thedotmack:
Addressed in 1eef0c36:

**Greptile**
- `middleware.ts:41` — dropped `'Authorization'` from CORS `allowedHeaders`.
- `SearchRoutes.ts:171-190` — removed the leftover `platformSource` read + pass from `handleContextPreview`.

**claude-review**
- Rate limiter: client IP now `req.socket.remoteAddress ?? req.ip ?? 'unknown'` so the `'unknown'` bucket isn't shared.
- `Retry-After` header is now set on every 429 with the remaining seconds in the window.
- `.docker-blowout-data/claude-mem.db` was the empty placeholder from the Issue Blowout 2026 docker tooling. Removed from tracking and added `.docker-blowout-data/` to `.gitignore` so it won't come back.
- Confirmed: `src/shared/auth-token.ts` is deleted and `worker-utils.ts` + `ViewerRoutes.ts` no longer import it. The `.cjs` diff reflects that; the TS source changes are visible in the previous commit (02359163).

The cleanup hot-path ordering note (P3) was acknowledged but not changed — threshold of 100 unique IPs is generous for a localhost service and the bucket cap of 300/min caps growth.

thedotmack:
Addressed in f628236b:

- Rate limit boundary: `entry.count >= RATE_LIMIT_MAX_REQUESTS` so the 300th request is the first rejected (was 301).
- Dead lazy cleanup removed — the map only ever holds 1–2 entries on localhost, so the branch was noise.

On the other notes:

- **Source files in diff** — this is a diff-against-main artifact, not a real gap. `src/shared/auth-token.ts`, the `Authorization` header in `worker-utils.ts`, and the token injection in `ViewerRoutes.ts` never existed on `main`; they were added on this branch in earlier commits (#1932/#1933) and removed in 02359163. The net PR diff against main shows no change for those paths because main and tip now match. A clean rebuild from this tip does not reintroduce auth — verified: worker rebuilt, `/api/search` returns 200 without any Authorization header.
- **Global 300 RPM for localhost** — intentional. Service is localhost-only; per-caller isolation via token/header would just re-introduce the friction we just removed. Happy to bump the cap later if hooks pressure it.
- **No tests for the limiter** — fair, but out of scope for this PR. Tracking separately.

thedotmack:
Addressed in 85402421:

- Flipped the rate limit check back to *allow 300, reject 301* by checking `count >= max` **before** incrementing. (Earlier rounds of this review flipped the comparator in both directions; the check-then-increment pattern avoids the off-by-one debate entirely.)
- Added a one-line comment making the localhost-global scope explicit so a future reader doesn't mistake it for per-process isolation.

On the non-code notes:

- **Pro tunnel interaction** — acknowledged. If/when a tunnel layer fronts the worker, it carries its own auth (and the rate limiter alone won't compensate, as you note). That belongs in the Pro layer, not in the open-source core.
- **Map cleanup** — leaving it out. On localhost the map holds at most 2 entries (`127.0.0.1`, `::1`); adding a 1000-entry cleanup branch is code for a scenario that can't happen here.
- **`X-RateLimit-*` headers** — would happily add if hook clients start hitting the cap, but they don't today and the extra headers on every response add noise.
- **`platform_source` DB column drop** — tracking in my head, will follow up when the column is actually unreferenced in SELECTs too.
- **Tests for the limiter** — fair; out of scope for this PR.

thedotmack:
Addressed in b3dd215d:

- IP key normalisation: strip `::ffff:` so IPv4-mapped IPv6 loopback shares a bucket with `127.0.0.1`.

Not changing:

- **Periodic `setInterval` cleanup** — the map holds at most 2 keys on a localhost-only server (`127.0.0.1` and `::1`, now merged further by the normaliser above). Adding a timer to prune a 2-entry map is worse than leaving it alone.
- **CLAUDE.md doc of no-auth decision** — reasonable suggestion, but that's a docs PR, not part of this refactor.
- **Unit tests for the limiter** — out of scope for this PR; happy to follow up.
- **`platformSource` tracking-only comment in processor.ts** — the session-object field is still read when writing to the DB. Not write-only, just no longer a query filter. Current code is clear enough without a comment explaining what it isn't.

thedotmack:
Addressed in a559aa5b:

- Rate limiter map now has a 1000-key size guard that prunes expired entries on a window reset. Zero per-request cost on the localhost hot path (1–2 keys); safety valve kicks in only if someone binds the worker on a non-loopback interface.

Non-changes:

- **Shared/multi-user host doc** — fair point, but a docs-only change for the Pro tunnel/container scenarios. Not in this PR's scope.
- **429s under tight context-inject loops** — worth watching. The current cap is 5 req/s average, which comfortably clears normal hook traffic. If we see 429s in the wild we'll bump it.
- **Unit tests for query builders + rate limiter** — agree; tracking for a follow-up.

thedotmack:
The \`normalizePlatformSource\` import in \`processor.ts\` is still in use on line 58 (\`platformSource: normalizePlatformSource(watch.name)\`) — that call is the per-session write-side of the platform label, not the query-time filter this PR removed. Not unused; not removing.

Other nits: threshold comment + HOST assertion + \`::1\` bucket vs \`127.0.0.1\` bucket are all acknowledged as nits, not addressing.

No outstanding blockers. Merging.

PR #2082, opened 2026-04-21
fix: detect PID reuse in worker start-guard (container restarts)

thedotmack:
## Summary
- Worker's **GUARD 1** (`isProcessAlive(existingPid)` in `worker-service.ts:1367`) and the supervisor's `validateWorkerPidFile` (`supervisor/index.ts:185`) both used `kill(pid, 0)` alone to decide "is the prior worker still running?" That's wrong when the PID file outlives the PID namespace — e.g. `docker stop` / `docker start` with a bind-mounted `~/.claude-mem`, or pm2-style graceful reloads. The new worker boots as the same low PID (often 11) as the previous one, `kill(0)` returns "alive," and the new worker refuses to start — against its own prior incarnation. Result: container appears to start, immediately exits cleanly with no user-visible error. Repro steps + logs in the "Observed in the wild" section below.
- Add a process-start **identity token** so we verify *which* process is at that PID, not just whether the PID is occupied:
  - Linux: field 22 of `/proc/<pid>/stat` (starttime in jiffies) — cheap, no exec, same signal pgrep/systemd use.
  - macOS / other POSIX: `ps -p <pid> -o lstart=`.
  - Windows: unchanged (returns `null`, caller falls back to liveness-only — containers aren't the typical Windows deploy).
- Shared helpers (`PidInfo`, `captureProcessStartToken`, `verifyPidFileOwnership`) live in `src/supervisor/process-registry.ts` to preserve the existing module-dependency direction (`services/infrastructure/ProcessManager` → `supervisor`, not back). `ProcessManager.ts` re-exports for the existing public surface.
- `writePidFile` auto-captures the token for the new process. PID files written by older versions are token-less; `verifyPidFileOwnership` falls back to liveness in that case, so upgrades don't break anything.

## Observed in the wild
Running two `claude-mem:basic` containers via OrbStack, both exited cleanly seconds after `docker start`:
```
[SYSTEM] Worker already running (PID alive), refusing to start duplicate
  {existingPid=11, existingPort=37777, startedAt=2026-04-20T05:12:21.180Z}
```
The stored `startedAt` was from the *previous* container boot; the new container's PID 11 was its own bun worker (verified via `/proc/11/comm` → `bun`). Deleting the stale `worker.pid` let it start; without the fix, the bug recurs on every restart.

## Test plan
- [x] `bun test tests/supervisor/ tests/infrastructure/` — 204 pass, 0 new failures (1 pre-existing plugin-distribution failure also present on `origin/main`).
- [x] New tests for `captureProcessStartToken`: self-capture, stability across calls, `null` for dead/invalid PIDs.
- [x] New tests for `verifyPidFileOwnership`: `null` input, dead PID, matching token, mismatching token (the PID-reuse regression), back-compat when no token stored.
- [x] New tests for `writePidFile`: auto-captures token for live PID, preserves caller-supplied token, omits field when PID is dead.
- [x] New supervisor test: `validateWorkerPidFile` returns `"stale"` when token mismatches (direct regression coverage for the container-restart bug).
- [x] `npm run build` — clean.
- [ ] Soak under live container restart cycles once published to the plugin's docker images.


thedotmack:
Thanks for the review. Addressed items 1–3 in 7746e99c:

- **Item 1 (redundant import/export)** — collapsed to a single combined statement in `ProcessManager.ts`.
- **Item 2 (type predicate)** — `verifyPidFileOwnership(info: PidInfo | null): info is PidInfo`. Dropped the `!` assertions at the GUARD 1 call site in `worker-service.ts` now that the narrowing flows through.
- **Item 3 (platform doc comment)** — rewrote the doc block in `captureProcessStartToken` to enumerate `process.platform` values explicitly (`linux` / `darwin`+other POSIX / `win32`) and called out the `ps`-missing ENOENT path.

No behavior change — these are all type-only edits, so esbuild's CJS output is byte-identical (just a TS-level polish).

### Items not addressed

- **Item 4 (test side-effects on the real PID file)** — the new `captureProcessStartToken` / `writePidFile` / `verifyPidFileOwnership` describe blocks sit inside the outer `describe('ProcessManager', …)` wrapper (`tests/infrastructure/process-manager.test.ts:25`), which already has a `beforeEach`/`afterEach` that backs up and restores `~/.claude-mem/worker.pid`. The existing `writePidFile`/`readPidFile`/`cleanStalePidFile`/`isPidFileRecent`/`touchPidFile` tests in the same file write to the real path the same way; I kept the new tests consistent with that convention rather than diverge. Happy to migrate the whole suite to a temp-PID-file fixture in a follow-up, but that's out of scope here.
- **Item 5 (un-export `isProcessAlive`)** — agreed it's a narrowing opportunity, but `isProcessAlive` still has direct consumers (`cleanStalePidFile`, the existing test file, and it's also surfaced via the `infrastructure/index.ts` barrel that `scripts/bug-report/` and `worker-cli.js` reach through). Leaving it exported for now; worth tracking separately.

thedotmack:
Thanks @greptile-apps. Both items resolved:

- **P2: Redundant import/export of `PidInfo`** — already fixed in 7746e99c (collapsed to a single `import { …, type PidInfo } / export { …, type PidInfo }` pair). Your review was on the previous commit.
- **P2: `ps lstart=` is locale-sensitive on macOS** — legitimate catch. Fixed in 09a33cdc by pinning `LC_ALL=C` and `LANG=C` in the `spawnSync` env so the emitted timestamp stays in the locale-independent POSIX form (`Mon Apr 21 09:00:00 2026`) regardless of the caller's `LC_TIME`/`LANG`. Added a comment crediting the review.

All 106 tests still pass; esbuild output regenerated in this commit as well.

thedotmack:
Second round addressed in f9253be7:

- **#1 Silent token mismatch** — added a DEBUG log in `verifyPidFileOwnership` when liveness passes but the token mismatches. Now the "PID reused" vs "process dead" cases are distinguishable in production logs.
- **#2 Redundant `?? undefined`** — dropped. Rewrote as `const resolvedToken = info.startToken ?? captureProcessStartToken(info.pid);` with the ternary working directly on the `string | null | undefined` union.
- **#5 Missing win32 test** — added. Mocks `process.platform = 'win32'` via `Object.defineProperty`, asserts the fallback returns `null`, and restores in `finally`. No CI runner change needed.

### Not addressed

- **#3 Type predicate adds little value** — earlier review in this same PR (see [prior comment](https://github.com/thedotmack/claude-mem/pull/2082#issuecomment-4285581214) item 2) explicitly asked for `info is PidInfo` so the `!` assertions at the call site could be dropped. The narrowing is load-bearing at `src/services/worker-service.ts:1368-1371` where the callers read `existingPidInfo.pid/port/startedAt` without a null check. Keeping the predicate.
- **#4 Platform check order** — cosmetic only; no functional impact. Skipping to keep the diff focused.

All 107 tests pass.

PR #2203, opened 2026-04-29
fix: filter chroma-mcp stdout banners to fix MCP handshake (closes #2197)

### LINKED ISSUE #2197: Chroma MCP stdio connection closes because chroma-mcp emits non-JSON stdout before JSON-RPC
## Summary

Claude-Mem 12.4.8 repeatedly fails Chroma vector sync with `MCP error -32000: Connection closed` when launching `chroma-mcp==0.2.6` over stdio.

The underlying `chroma-mcp` process starts and can answer MCP initialize, but it writes human-readable startup lines to stdout before the JSON-RPC response:

```text
Successfully initialized Chroma client
Starting MCP server
{"jsonrpc":"2.0","id":1,"result":{...}}
```

Because MCP stdio transports expect stdout to contain only protocol JSON-RPC frames, Claude-Mem's MCP SDK client closes the connection and Chroma sync falls back to "continuing without vector search".

## Environment

- claude-mem: 12.4.8
- chroma-mcp: 0.2.6
- mode: local/persistent Chroma
- command Claude-Mem launches:

```text
uvx --python 3.13 chroma-mcp==0.2.6 --client-type persistent --data-dir ~/.claude-mem/chroma
```

## Observed logs

```text
[CHROMA_MCP] Connecting to chroma-mcp via MCP stdio {command=uvx, args=--python 3.13 chroma-mcp==0.2.6 --client-type persistent --data-dir ...}
[CHROMA_MCP] Connection failed, killing subprocess to prevent zombie {error=MCP error -32000: Connection closed}
[CHROMA_MCP] Connection attempt failed MCP error -32000: Connection closed
[CHROMA] OpenRouter chroma sync failed, continuing without vector search {...} MCP error -32000: Connection closed
```

## Reproduction

Using the same `@modelcontextprotocol/sdk` `StdioClientTransport` shape as the worker:

1. Spawn `uvx --python 3.13 chroma-mcp==0.2.6 --client-type persistent --data-dir ~/.claude-mem/chroma`.
2. Call `client.connect(transport)`.
3. The connection closes with `McpError: MCP error -32000: Connection closed`.
4. If the process is probed manually, stdout shows the two startup banner lines before the JSON-RPC initialize result.

## Local workaround that fixed it

I added a small wrapper around `uvx ... chroma-mcp` that forwards only JSON-looking stdout lines to stdout and redirects non-protocol stdout lines to stderr. After pointing Claude-Mem's Chroma launcher at that wrapper and restarting the worker, the logs changed to:

```text
[CHROMA_MCP] Connecting to chroma-mcp via MCP stdio {command=.../chroma-mcp-json-stdio-wrapper.mjs, args=...}
[CHROMA_MCP] Connected to chroma-mcp successfully
[CHROMA_SYNC] Backfilling observations {...}
[CHROMA_SYNC] Syncing observation {...}
```

I also verified with the same MCP SDK client that:

- raw `chroma-mcp` stdio fails / closes
- wrapper-filtered stdio connects and lists 13 Chroma tools

## Additional edge case

On this machine there were two `uvx` binaries. An older `~/.local/bin/uvx` reports:

```text
error: Not a valid package or extra name: "chroma-mcp==0.2.6"
```

while Homebrew `/opt/homebrew/bin/uvx` accepts the requirement string. So relying on bare `uvx` can also be PATH-sensitive, but the primary protocol failure here is the non-JSON stdout emitted before JSON-RPC.

## Suggested fix

Possible approaches:

- Ensure the launched Chroma MCP process writes all startup/log lines to stderr, not stdout.
- Add a Claude-Mem-side stdio adapter for `chroma-mcp` that redirects non-JSON stdout lines to stderr before passing stdout to `StdioClientTransport`.
- Optionally resolve a known-compatible `uvx`/`uv tool uvx` path instead of relying on a PATH entry that may point to an older uvx shim.

thedotmack:
## Summary

chroma-mcp 0.2.x writes startup banners (`Successfully initialized Chroma client`, `Starting MCP server`) to **stdout** before the first JSON-RPC frame. MCP SDK's `StdioClientTransport` treats any non-JSON line on stdout as a fatal protocol violation and closes the connection — which is exactly what the worker logs (`Connection failed, killing subprocess to prevent zombie`) on every cold start.

This PR vendors a tiny Node wrapper (`plugin/scripts/chroma-mcp-stdio-filter.mjs`) that:
- Spawns chroma-mcp as a child with the same argv we already pass.
- Line-buffers its stdout.
- Forwards JSON-looking lines (`{`/`[` after trim) to **our** stdout (where MCP SDK reads).
- Redirects everything else to **our** stderr (prefixed `[chroma-mcp banner]`) so the diagnostics stay visible in worker logs.
- Forwards `SIGINT` / `SIGTERM` / `SIGHUP` and propagates exit codes.

Wiring lives at `src/services/sync/ChromaMcpManager.ts:151-200`. The Windows `cmd.exe /c uvx ...` routing is preserved verbatim — the wrapper just sits between MCP SDK and the existing argv.

Closes #2197.

Credit: @zengyuzhi proposed the wrapper-script approach in the issue with a working PoC.

## Verification

```bash
# 1. Wrapper filters banners but passes JSON through cleanly:
node plugin/scripts/chroma-mcp-stdio-filter.mjs node -e \
  'console.log("Starting MCP server"); console.log(JSON.stringify({jsonrpc:"2.0",id:1,result:{}}))' \
  2>/tmp/err
# stdout: {"jsonrpc":"2.0","id":1,"result":{}}
# stderr: [chroma-mcp banner] Starting MCP server

# 2. Build picks up wiring:
npm run build
grep -c chroma-mcp-stdio-filter plugin/scripts/worker-service.cjs   # 1

# 3. Cold start (post-merge): worker log should show
#    "Connected to chroma-mcp successfully" instead of "Connection failed".
```

## Test plan

- [ ] Verified locally: wrapper filter behaves as expected (smoke test above).
- [ ] Verified locally: `npm run build` succeeds; the new wiring is bundled into `worker-service.cjs`.
- [ ] Cold-start worker after delete of `~/.claude-mem/chroma/chroma.sqlite3-wal,-shm` and confirm `Connected to chroma-mcp successfully` appears in logs.
- [ ] Confirm exactly one chroma-mcp process per worker via `ps -ef | grep chroma-mcp | grep -v grep`.
- [ ] Confirm Chroma tools are discoverable via the worker's MCP search endpoint.

## Anti-pattern guard (per PLAN.md §4.5)

- Did NOT modify chroma-mcp upstream — we don't own that package.
- Did NOT filter stderr — chroma-mcp's diagnostics still go there.


thedotmack:
## Holding this PR — heuristic-as-fix, root cause is upstream

Reviewed this against the rest of the batch and want to hold rather than merge.

The wrapper's stdout filter (\`startsWith('{') || startsWith('[')\` → forward to stdout, else to stderr) is inference, not a fix. It works until it doesn't:
- a banner line that legitimately starts with \`{\` (more common than expected — many tools emit JSON-shaped diagnostics)
- a scalar JSON-RPC response (rare but legal per spec)
- multi-byte boundary that splits a brace across reads
- future chroma-mcp banner text changes

Each failure mode lands as a silent MCP handshake corruption that's hard to diagnose because the wrapper is now between the client and the server.

The actual root cause is upstream: chroma-mcp emits banners on stdout, which violates the MCP stdio transport spec (stdout is reserved for JSON-RPC frames; diagnostics go to stderr). Filed there: chroma-core/chroma-mcp#66.

Once the upstream fix ships, this entire wrapper goes away. Carrying it permanently means maintaining a heuristic that someone has to keep in sync with chroma-mcp's banner output forever.

Closing this PR. Users still hitting the handshake failure can pin to a chroma-mcp version with the upstream fix once it lands. Tracked: #2216 covers the broader process-lifecycle work that includes chroma-mcp transport.

Thanks — the wrapper is genuinely useful as a one-off mitigation; just shouldn't be carried as permanent code.

PR #2204, opened 2026-04-29
fix: strip privacy tags before summarization (closes #2149)

### LINKED ISSUE #2149: Privacy tags not stripped from last_assistant_message before summarization (fix included)
## Before submitting

---

### Bug Description

The `<private>...</private>` tag is documented as a way to keep content out of long-term memory. CLAUDE.md says the hook layer strips these tags before anything reaches the worker, and that's true for user prompts and for tool input/output. The summarize path skips it though: whatever sits in `last_assistant_message` gets POSTed to `/api/sessions/summarize` as-is, lands in `pending_messages.last_assistant_message`, and gets fed verbatim to the summarization LLM as input. The structured summary the LLM produces will often surface that content too.

The reason this isn't immediately obvious is that Claude does try to strip `<private>` tags from its own output. Its training treats them as privacy directives. But that's a soft guarantee, not a hard one, and there are plenty of normal cases where Claude does emit the tag literally. When it does, the contents bypass claude-mem's privacy filter and get persisted.

### Steps to Reproduce

The trick is getting Claude to actually emit a literal `<private>...</private>` in its response, since the model often strips it on its own. Easy ways it slips through accidentally:

1. Ask Claude to document or explain the `<private>` tag itself. It produces examples containing the literal tag.
2. Ask for an XML or HTML example that uses a tag named `private` (a generic markup question, no privacy framing).
3. Code generation where Claude needs to include the tag as a string literal (parser fixtures, test data, etc).
4. Claude echoes earlier user content that had wrapped something in `<private>...</private>`, without re-stripping.

Once the assistant's final text turn contains a literal `<private>...</private>` block, end the session (Stop hook fires) and the leak happens. To verify against your own DB:

```bash
sqlite3 ~/.claude-mem/claude-mem.db \
  "SELECT COUNT(*) AS leaked_rows
   FROM pending_messages
   WHERE message_type='summarize'
     AND ( last_assistant_message LIKE '%<private>%'
        OR last_assistant_message LIKE '%<persisted-output>%'
        OR last_assistant_message LIKE '%<claude-mem-context>%'
        OR last_assistant_message LIKE '%<system-instruction>%' );"
```

### Expected Behavior

The summarize path should strip privacy tags before `last_assistant_message` reaches the queue table or the summarization LLM, the same way the user-prompt and tool-observation paths already do.

### Environment

- **Claude-mem version**: 12.3.9
- **Claude Code version**: (latest)
- **OS**: macOS
- **Platform**: Claude Code

### Logs

Not really applicable here. It's a data-flow bug, not a runtime error. The SQL query above is the cleanest way to see it.

### Additional Context

Relevant code:

- Hook reads the message and POSTs raw: `src/cli/handlers/summarize.ts:59`
- Worker receives and queues raw: `src/services/worker/http/routes/SessionRoutes.ts:750` (and the older `:604` path for `handleSummarize`)
- Strip utility that should be applied but isn't: `src/utils/tag-stripping.ts` (`stripMemoryTagsFromPrompt`)

For comparison, `/api/sessions/init` does call `stripMemoryTagsFromPrompt` before storing, and the observation path strips tool input/output via `stripMemoryTagsFromJson`. Only the summarize path is missing it.

I have a working fix on a [branch in my fork](https://github.com/thedotmack/claude-mem/compare/main...swithek:claude-mem:fix/summarize-privacy-tag-leak?expand=1). It strips at both the hook layer (covers Claude Code) and the worker queue layer (covers Cursor / Gemini CLI / OpenCode). Verified manually end-to-end against a real Claude Code session, and added unit tests covering all five privacy tag types plus the case where the entire turn was wrapped in a private block (handler should skip the POST).

I'd like to open this as a PR so the authorship is attached. If you'd rather just pull the diff in yourself, no problem, I'd just appreciate a credit on the commit (or a co-author line).

thedotmack:
## Summary

Cherry-picks @swithek's fix from [`fix/summarize-privacy-tag-leak`](https://github.com/swithek/claude-mem/tree/fix/summarize-privacy-tag-leak) for issue #2149: `<private>...</private>` (and the four other claude-mem privacy tags) were leaking through the Stop hook's `last_assistant_message` into the summarize queue and downstream summary LLM.

The fix is **defense-in-depth** — strips at both edges:

1. **Hook layer** (`src/cli/handlers/summarize.ts`): `extractLastMessage()` result is run through `stripMemoryTagsFromPrompt()` before being POSTed to the worker. If the entire turn is wrapped in a privacy tag, the handler short-circuits and skips the worker POST entirely.
2. **Worker queue layer** (`src/services/worker/http/routes/SessionRoutes.ts:601, 750`): both summarize endpoints (`/sessions/:sessionDbId/summarize` and `/api/sessions/summarize`) sanitize `last_assistant_message` again before queueing — so a misbehaving or third-party client cannot bypass the hook strip.

Uses the existing `stripMemoryTagsFromPrompt()` from `src/utils/tag-stripping.ts`. No tag-stripping logic is re-implemented inline.

The second commit fixes the test mocks for current `main`: the handler now uses `executeWithWorkerFallback` (not `workerHttpRequest`), and `shouldTrackProject` reads from `loadFromFileOnce()` in `hook-settings.js`.

## Files changed

- `src/cli/handlers/summarize.ts` (+2)
- `src/services/worker/http/routes/SessionRoutes.ts` (+8 / -2)
- `tests/cli/handlers/summarize-tag-stripping.test.ts` (+177)

## Test plan

- [x] `npx bun test tests/cli/handlers/summarize-tag-stripping.test.ts` — 10 pass / 0 fail / 37 expect()
- [x] `grep -n "stripMemoryTagsFromPrompt" src/cli/handlers/summarize.ts src/services/worker/http/routes/SessionRoutes.ts` — strip call appears in hook handler (line 61) and both worker summarize handlers (lines 605, 754)
- [x] `npm run build-and-sync` — clean build, worker restart succeeds

## Verification commands (post-merge)

```bash
grep -n "stripMemoryTagsFromPrompt" src/cli/handlers/summarize.ts src/services/worker/http/routes/SessionRoutes.ts
npx bun test tests/cli/handlers/summarize-tag-stripping.test.ts
```

Closes #2149

Authored-by: @swithek (cherry-picked from `swithek/claude-mem@bd68bfcc`)


thedotmack:

PR #2205, opened 2026-04-29
fix: preserve relevance order in semantic search (closes #2153)

### LINKED ISSUE #2153: Bug: ChromaSearchStrategy 'relevance' ordering silently returns rows in date_desc order
> **Note:** I have a working fix and tests on a fork branch and would happily submit a PR, but it looks like the repo is set to limit-to-collaborators, so PR creation generates 404s for me. If you'd like the patch applied, I can paste a diff in a comment, or you can grant external-PR access for this fix. Either is fine. I'm happy to contribute and use claude-mem and its capabilities daily in a multi-model autonomous/semi-autonomous lab and dev/test/prod environment.

### Summary

When a caller of `SessionStore.getObservationsByIds()` (or the parallel `getSessionSummariesByIds` / `getUserPromptsByIds`) intends to preserve incoming ID order — i.e. when Chroma has already ranked candidates by vector similarity and the caller wants hydration to retain that order — the rows come back sorted by `created_at_epoch DESC` instead. The `'relevance'` ordering mode in semantic search is silently turned into `'date_desc'` at the hydration step.

### Root cause

Two pieces of code disagree about how to express "no SQL sort, preserve incoming ID order":

`src/services/worker/search/strategies/ChromaSearchStrategy.ts` (lines 110-113) explicitly drops `orderBy` when `'relevance'` is requested, intending hydration to honor the caller's input order:

```ts
// Chroma already ranks by vector similarity; 'relevance' has no SQL
// equivalent, so drop it before hydrating rows from SessionStore.
const sqlOrderBy: 'date_desc' | 'date_asc' | undefined =
  options.orderBy === 'relevance' ? undefined : options.orderBy;
```

`src/services/sqlite/SessionStore.ts` (line 1676) destructures with a default that coerces `undefined` back to `'date_desc'`, then unconditionally applies `ORDER BY created_at_epoch`:

```ts
const { orderBy = 'date_desc', limit, project, type, concepts, files } = options;
const orderClause = orderBy === 'date_asc' ? 'ASC' : 'DESC';
// ...
ORDER BY created_at_epoch ${orderClause}
```

The destructure default in `SessionStore.ts` was introduced in `327dd44` (Feb 2026, baseline v10.1.0). The relevance-aware logic in `ChromaSearchStrategy.ts` was added later in `94d592f` (Apr 2026, the PATHFINDER refactor). The newer code's stated intent isn't honored by the older default — looks like a refactor oversight.

### Reproduction

I've built a self-contained reproduction script that runs against the actual upstream code on a copy of a real `claude-mem.db`. It uses `SessionStore.getObservationsByIds()` directly with the same args `ChromaSearchStrategy` passes for relevance searches.

```ts
// scripts/repro-bug.ts (excerpt — full script ~80 lines)
import { SessionStore } from '../src/services/sqlite/SessionStore.js';
import { copyFileSync, mkdtempSync } from 'fs';
import { homedir, tmpdir } from 'os';
import { join } from 'path';

const tmp = mkdtempSync(join(tmpdir(), 'cm-bug-repro-'));
const dbPath = join(tmp, 'claude-mem.db');
copyFileSync(join(homedir(), '.claude-mem', 'claude-mem.db'), dbPath);
const store = new SessionStore(dbPath);

// Pick 5 real consecutive observations (IDs 17000-17004 in my DB)
// whose timestamps are monotonically increasing with ID.
const inputIds = [17000, 17001, 17002, 17003, 17004];

// This is exactly how ChromaSearchStrategy.executeChromaSearch calls
// hydration when orderBy='relevance' is requested:
//   const sqlOrderBy = options.orderBy === 'relevance' ? undefined : options.orderBy;
//   sessionStore.getObservationsByIds(categorized.obsIds, { orderBy: sqlOrderBy, ... });
const result = store.getObservationsByIds(inputIds, { orderBy: undefined });

console.log(result.map(r => r.id));
// Expected (intent): [17000, 17001, 17002, 17003, 17004]  (preserved input order)
// Actual:            [17004, 17003, 17002, 17001, 17000]  (date_desc — bug)
```

Run against upstream HEAD `bb1f869` (v12.4.7), this returns `[17004, 17003, 17002, 17001, 17000]` — descending by `created_at_epoch`, not the requested input order.

### Why this matters

`'relevance'` is the default `orderBy` in `SessionSearch.searchObservations` and is the documented mode for Chroma semantic search. Any caller of the search APIs that expects vector-similarity ordering currently gets date ordering instead. This silently degrades semantic-search quality across the worker's HTTP and MCP search endpoints. Im my case, I"ve also implemented a cross-encoder "flash" re-ranking feature that's yielded beneficial results in my deployment, but this bug needed to be addressed in order for my re-ranking feature/service to work properly (and I'd be glad to package it up and submit that feature as well, if you are interested).

### Proposed fix

Two coordinated changes (small):

1. **`src/services/worker/search/strategies/ChromaSearchStrategy.ts`** — pass `options.orderBy` through to the hydration calls instead of converting `'relevance'` to `undefined`. The intent comment can be updated or removed.

2. **`src/services/sqlite/SessionStore.ts`** — accept `'relevance'` as a valid `orderBy` value in the three `get*ByIds` method signatures. When `orderBy === 'relevance'`, omit the `ORDER BY` clause from the SQL and apply order preservation in application code (e.g. a small `preserveIdOrder()` helper that rebuilds the result array in caller-provided ID order, or an SQL `ORDER BY CASE id WHEN ... END` on the IN list). Affects `getObservationsByIds` (~line 1670), `getSessionSummariesByIds` (~line 2467), and `getUserPromptsByIds` (~line 2499).

Verified with the same reproduction script on a fork branch implementing both changes: returned IDs match input order `[17000, 17001, 17002, 17003, 17004]`.

### Notes

- The fix has to be in both files; either alone leaves the other side broken.
- Existing date-ordered call sites (`'date_desc'` / `'date_asc'`) are unaffected — only the `'relevance'` path changes behavior.
- Rerank features built on top of relevance ordering (e.g. anything reading the hydrated row order to score) are blocked by this bug today.

Thanks very much for claude-mem -- It's made all the difference in my environment of long-running orchestration projects, and I'm quite glad to help out and give back to the project.

thedotmack:
## Summary

Fixes the search-ordering correctness bug reported in #2153 where `orderBy: 'relevance'` was silently degraded to `'date_desc'`, destroying ChromaDB's vector similarity ranking.

The bug was a two-part failure:

1. `ChromaSearchStrategy.ts:110-113` coerced `'relevance'` → `undefined` before calling `SessionStore`.
2. `SessionStore.getObservationsByIds` / `getSessionSummariesByIds` / `getUserPromptsByIds` defaulted `undefined` → `'date_desc'`, so ranking was lost regardless of what Chroma returned.

This PR adapts @rogerdigital's fix (`rogerdigital/claude-mem#1`) for the SessionStore side and additionally fixes the strategy side so the value actually reaches SessionStore.

## Changes

- **`SessionStore.getObservationsByIds`/`getSessionSummariesByIds`/`getUserPromptsByIds`** now accept `'relevance'` as a valid `orderBy`. When `'relevance'`, the SQL `ORDER BY` is omitted and rows are reordered via a `Map` to preserve caller-provided ID order.
- **`ChromaSearchStrategy.runChromaSearch()`** passes `options.orderBy` through unchanged instead of coercing `'relevance'` to `undefined`.
- **New regression test** at `tests/services/sqlite/get-observations-by-ids-relevance.test.ts` asserts the input-order contract.

## Test plan

- [x] `bun test tests/services/sqlite/get-observations-by-ids-relevance.test.ts` — both cases pass.
- [x] `npm run build` — bundled artifacts rebuild cleanly.
- [x] Type check passes (`npx tsc --noEmit`) — no new errors.
- [x] No new index added (`created_at_epoch` already indexed).

## Closes

- Closes #2153

Co-authored with @rogerdigital — semantic search hydration patch originally from `rogerdigital/claude-mem@9fedf8fc`.


rogerdigital:
Thanks for the attribution. The strategy-side coercion was the missing piece — passing `orderBy` through unchanged makes the fix complete end-to-end. The `limit` interaction note from CodeRabbit is worth watching if callers ever combine both.

thedotmack:

Issue #2201, opened 2026-04-29
BUG: Runaway observer sessions caused 345M token flow in one day via full PostToolUse I/O capture and stuck pending queue 🐛 


SvcGrace:
## Before submitting

- [✅] I searched existing issues and confirmed this is not a duplicate

---

## Bug Description

`claude-mem` appears to have created runaway observer sessions that repeatedly invoked Claude Code via the configured Anthropic-compatible endpoint, causing very large token usage.

In my local evidence, the observer sessions produced:

- Time range: `2026-04-28T08:02:40Z` to `2026-04-29T16:53:55Z`
- Model: `deepseek-v4-flash`
- Responses: `4,079`
- Input tokens: `51,437,709`
- Cache read input tokens: `292,196,992`
- Output tokens: `1,963,427`
- Total token flow: `345,598,128`
- Observer transcript size: `31` JSONL files, `184 MB`

The largest observer memory session alone consumed about:

- Total token flow: `154,938,768`
- Input: `23,737,162`
- Cache read: `130,289,280`
- Output: `912,326`
- Responses: `1,789`

This was unexpected and looked like an infinite/near-runaway background observer loop.

## Steps to Reproduce

I cannot safely reproduce this now because I disabled and quarantined the plugin, but the observed setup was:

1. Install/enable `claude-mem` plugin version `12.4.7`.
2. Use Claude Code / Codex integration with `claude-mem` hooks enabled.
3. Configure Claude Code to use an Anthropic-compatible DeepSeek endpoint.
4. Run normal coding sessions with high-volume tool calls such as shell execution, patching, stdin streaming, file writes, browser/JS interactions, etc.
5. `claude-mem` starts observer sessions and captures `PostToolUse` events.
6. Worker queue grows, many `pending_messages` fail or remain pending, while observer Claude sessions continue consuming tokens.

## Expected Behavior

`claude-mem` should not send full raw tool inputs and outputs from high-volume tools into model prompts by default.

Expected safeguards:

- Skip or heavily summarize high-volume tools by default.
- Apply size limits to tool input/output before queueing.
- Avoid repeatedly appending full observation prompts into long-running conversation history.
- Apply backpressure before creating more observer sessions.
- Mark poisoned/stuck queues as stopped instead of repeatedly restarting generators.
- Avoid observer-session pollution in user-facing memory and vector stores.

## Actual Behavior

The plugin captured broad `PostToolUse` data and queued it as observations.

Evidence from source:

[...]

Issue #2199, opened 2026-04-29
Knowledge Agent broken on Bedrock: CLAUDE_CODE_USE_BEDROCK stripped by sanitizeEnv()


ahpuck:
## Bug

Knowledge Agent `query_corpus` returns `"Not logged in · Please run /login"` for all users authenticating via AWS Bedrock (`CLAUDE_CODE_USE_BEDROCK=1`).

## Root Cause

`sanitizeEnv()` in `src/supervisor/env-sanitizer.ts` strips all `CLAUDE_CODE_*` env vars except those in `ENV_PRESERVE`. `CLAUDE_CODE_USE_BEDROCK` is not in that set, so it gets dropped before the worker spawns Claude CLI subprocesses via the Agent SDK `query()` call.

Without this var, the spawned Claude process falls back to direct Anthropic API auth — which fails because Bedrock users don't have Anthropic API keys or OAuth tokens.

## Steps to Reproduce

1. Use Claude Code with `CLAUDE_CODE_USE_BEDROCK=1` (AWS Bedrock routing)
2. `build_corpus` — works fine
3. `prime_corpus` — fails silently (spawns Claude subprocess that can't auth)
4. `query_corpus` — returns `"Not logged in · Please run /login"`

## Fix

Add `CLAUDE_CODE_USE_BEDROCK` to `ENV_PRESERVE` in `src/supervisor/env-sanitizer.ts`:

```diff
 export const ENV_PRESERVE = new Set([
   'CLAUDE_CODE_OAUTH_TOKEN',
   'CLAUDE_CODE_GIT_BASH_PATH',
+  'CLAUDE_CODE_USE_BEDROCK',
 ]);
```

## Verified

Patched locally, rebuilt, restarted worker — `query_corpus` successfully routes through Bedrock and returns knowledge agent answers.

Branch with fix: https://github.com/ahpuck/claude-mem/tree/fix/preserve-bedrock-env

aegir76:
Confirmed and fixed locally on v12.4.8. The issue is in the bundled `worker-service.cjs` — the `sanitizeEnv()` whitelist (`kne` in the minified bundle) only includes `CLAUDE_CODE_OAUTH_TOKEN` and `CLAUDE_CODE_GIT_BASH_PATH`.

**One-line fix** in `scripts/worker-service.cjs`:

Find:
```js
[...]

Issue #2194, opened 2026-04-28
Observer SDK inherits user-installed MCPs, kicking single-instance MCPs (e.g. Telegram) off interactive sessions


BlazeMV:
## TL;DR

The observer/knowledge SDK children spawn with `mcpServers: {}` but no `strictMcpConfig: true`, so they inherit user-installed MCPs and evict any single-instance MCP (Telegram, etc.) from the user's interactive session every time the observer fires.

**Fix is ready on a fork — would normally have been a PR, but upstream PR creation is blocked for external contributors:**

➡️ **https://github.com/BlazeMV/claude-mem/tree/fix/strict-mcp-config-isolation** (commit `9984533f`, 2 files, +6 lines)

If you'd like to cherry-pick directly:
```bash
git fetch https://github.com/BlazeMV/claude-mem.git fix/strict-mcp-config-isolation
git cherry-pick 9984533f
```

The full diff is also inline below.

## Reproduction

The cleanest repro uses the official Telegram channel plugin (`claude-plugins-official/telegram`) because the Telegram Bot API allows exactly one `getUpdates` consumer per token, and `server.ts:60-69` enforces the lock by writing `~/.claude/channels/telegram/bot.pid` and SIGTERMing whoever held it before.

1. Install the Telegram channel plugin and pair it with an interactive `claude --channels plugin:telegram@...` session.
2. Run claude-mem alongside (default install).
3. Wait for any observer event (idle compression, post-tool-use hook, etc.).
4. **Observed**: bot.pid flips to the observer-spawned `bun server.ts` (parented to the worker's SDK child), the previous holder gets SIGTERMed, the user's session loses its Telegram MCP and the model's tools dialog shows it disconnected. User has to `/mcp` to recover. Repeats every observer cycle.

The same applies to any MCP that can't tolerate a sibling consumer (websocket peers, port binders, exclusive DB connections).

### Verified ancestry of the offending MCP at the moment of takeover

```
26935  bun server.ts                                    (Telegram MCP — user's)
  └─ 26933  bun run --cwd .../telegram start
      └─ 59432  claude --channels plugin:telegram@...   (interactive session)

vs.

67420  bun server.ts                                    (Telegram MCP — observer's)
  └─ 67418  bun run --cwd .../telegram start
      └─ 67401  claude --model claude-sonnet-4-5 --disallowedTools Bash,Read,Write,...
          └─ 33090  bun .../scripts/worker-service.cjs --daemon
```

[...]

Issue #2171, opened 2026-04-27
Observer-sessions recursive loop: observer observes itself, exponential prompt growth


luisbarcia:
## Bug Description

Observer-sessions enter a recursive loop where each observer session's Claude CLI spawns hooks that create new pending_messages, which spawn MORE observer-sessions, causing:

1. **Exponential prompt nesting** — the observer's system prompt gets wrapped inside `<observed_from_primary_session><user_request>` tags recursively (6+ levels deep), growing from ~5KB to 35-43KB per INIT prompt
2. **Unbounded session creation** — 3,900+ zombie `observer-sessions` accumulated in `sdk_sessions` in a single day
3. **SSE stream pollution** — `/stream` endpoint pushes `new_prompt` events with `project: "observer-sessions"` to the viewer, including the full recursive prompt content

## Reproduction

1. Run claude-mem worker v12.4.7 via launchd
2. Start a Claude Code session with active hooks
3. Make several tool calls (Bash, Read, etc.)
4. Open `http://localhost:<port>` viewer
5. Observe observer-sessions appearing and growing in the Prompts tab
6. Check SSE stream: `curl -sN "http://localhost:<port>/stream" | head -c 5000`

The recursive nesting is visible in the SSE output:

```
"You are a Claude-Mem, a specialized observer tool..."
  → <observed_from_primary_session>
    → <user_request>"You are a Claude-Mem, a specialized observer tool..."
      → <observed_from_primary_session>
        → <user_request>"You are a Claude-Mem, a specialized observer tool..."
          → ... (6+ levels, each adding ~5KB)
```

## Root Cause

The generator spawns Claude CLI sessions via `CLAUDE_CODE_PATH`. These sessions inherit all hooks from `~/.claude/settings.json`, including the claude-mem MCP server hook. When the observer's Claude CLI processes, its tool calls are captured by the same hook → sent to the worker �� creating MORE observer-sessions → infinite recursion.

The guard function `la()` with `ghe(t,li)` (checking if `cwd` matches `~/.claude-mem/observer-sessions/`) does NOT prevent this, likely because the observer's `cwd` is the project directory, not the observer-sessions directory.

## Evidence

```
# SSE stream shows observer-sessions with 35KB+ recursive prompts
curl -sN "http://localhost:37701/stream" | head -c 3000
# → {"type":"new_prompt","prompt":{"project":"observer-sessions","prompt_length":34835,...}}

# 3,900+ zombie sessions accumulated in one day
[...]

PR #2206, opened 2026-04-29
fix: isolate SDK boundary — close 6 issues at 3 call sites

### LINKED ISSUE #2155: Bug: worker subprocesses load user-tier settings and can start foreground-session plugins
> **Note:** I have a small working fix for this on a fork branch that I have been running for several days, and I'm glad to submit a PR. However, it appears PR creation is limited to approved collaborators. If you would like the fix, I can paste the diff here, open a PR if access is enabled, send another way, or leave it for you to adapt. Just let me know.

### Summary

The memory worker spawns internal Claude Code subprocesses for background memory work. Those subprocesses currently use the default Claude Code settings discovery behavior, which means they load the user's global settings from `~/.claude/settings.json`.

That can unintentionally start user-level plugins or integrations inside the worker subprocess. If one of those integrations claims a singleton external resource, the short-lived worker can interfere with the user's foreground Claude Code session. In my specific case, the Claude Code Telegram plugin slot from Anthropic was taken over when the worker sessions started and dropped/orphaned when the worker subprocess exited.

### Why this matters

The worker subprocess is an internal implementation detail. It should inherit the memory worker's project/local configuration, model selection, tool restrictions, cwd, and isolated environment, but it should not start user-session integrations that were configured for the interactive foreground session.

This is especially risky for channel or notification plugins because many of them start background listeners as soon as they load.

### Observed failure mode

In one setup, a user-level channel plugin configured in `~/.claude/settings.json` starts a polling loop on load and claims a single-consumer API slot. Each memory-worker subprocess loaded the plugin, claimed the slot, displaced the foreground session's listener, then exited. The foreground session did not automatically reclaim the slot, so inbound messages stopped arriving while outbound messages still worked.

The same class of failure could happen with any user-level plugin or integration that has startup side effects or claims a singleton resource.

### Root cause

The Agent SDK `query()` calls that spawn internal worker subprocesses do not set `settingSources`, so Claude Code uses its default settings discovery and includes the user tier.

The affected call sites are:

- `src/services/worker/SDKAgent.ts`
- `src/services/worker/knowledge/KnowledgeAgent.ts` prime query
- `src/services/worker/knowledge/KnowledgeAgent.ts` resumed query

### Proposed fix

Set `settingSources: ['project', 'local']` on the internal worker `query()` calls.

That keeps project and local settings available for memory work, but skips user-tier settings and therefore avoids loading plugins intended for the user's main session.

Example shape:

```ts
const queryResult = query({
  prompt: messageGenerator,
  options: {
    model: modelId,
    cwd: OBSERVER_SESSIONS_DIR,
    disallowedTools,
    abortController: session.abortController,
    settingSources: ['project', 'local'],
    spawnClaudeCodeProcess: createPidCapturingSpawn(session.sessionDbId),
    env: isolatedEnv
  }
});
```

### Why not make this configurable?

The worker is a background component with a narrow job. Loading user-tier plugins inside it is surprising and can break the foreground session in ways that are hard to diagnose. I do not see a legitimate worker use case for inheriting user-session plugins by default.

Project and local settings remain available, so project-specific configuration is not lost.

### Validation performed

On a local install with the patch applied:

- Worker subprocesses started with `--setting-sources project,local`.
- User-tier channel plugins no longer spawned inside worker subprocesses.
- The foreground session kept its channel listener across repeated memory-worker cycles.
- Memory generation continued to store observations successfully.
- Claude Code subscription auth continued to work. `settingSources` controls settings-file discovery, not the auth/keychain path.

thedotmack:
## Summary

Single architectural fix at the three `@anthropic-ai/claude-agent-sdk` `query()` call sites — closes 6 issues with one commit.

| Issue | Symptom | Fix |
|------|---------|-----|
| #2155 | User's `settings.json` (allowedTools, hooks, etc.) bleed into observer SDK | `settingSources: []` |
| #2159, #2171, #2194 | User MCP servers leak into observer SDK | `strictMcpConfig: true` (alongside existing `mcpServers: {}`) |
| #2199 | Bedrock/Vertex env vars stripped, breaks Knowledge Agent for cloud users | Extend `ENV_PRESERVE` |
| #2201 | 345M tokens/day from a single user (uncapped agentic-tool payloads) | Extend `CLAUDE_MEM_SKIP_TOOLS` defaults + new `CLAUDE_MEM_MAX_OBSERVATION_BYTES` cap (default 64 KB) at the ingest gate |

## SDK option names — verified against installed types

```
$ node -e "const fs=require('fs');console.log(fs.readFileSync('node_modules/@anthropic-ai/claude-agent-sdk/sdk.d.ts','utf8').match(/settingSources\?[^;]+/g))"
[ "settingSources?: SettingSource[]", "settingSources?: SettingSource[]" ]
```

`SettingSource = 'user' | 'project' | 'local'`. `strictMcpConfig?: boolean` is also present. Empty array means inherit no settings sources, which is what observer/knowledge agents need.

## Files changed

- `src/services/worker/SDKAgent.ts` — add 2 fields at line 165
- `src/services/worker/knowledge/KnowledgeAgent.ts` — add 2 fields at lines 85 (prime) and 205 (executeQuery)
- `src/supervisor/env-sanitizer.ts` — extend `ENV_PRESERVE` with cloud provider vars
- `src/shared/SettingsDefaultsManager.ts` — extend `CLAUDE_MEM_SKIP_TOOLS` default; add `CLAUDE_MEM_MAX_OBSERVATION_BYTES`
- `src/services/worker/http/shared.ts` — add size cap check at the ingest gate

Bundled artifacts (`plugin/scripts/*.cjs`, `plugin/ui/viewer-bundle.js`) regenerated by `npm run build-and-sync`.

## Verification

```bash
$ grep -rn "settingSources\\|strictMcpConfig" src/services/worker/
src/services/worker/SDKAgent.ts:165:        settingSources: [],
src/services/worker/SDKAgent.ts:168:        strictMcpConfig: true,
src/services/worker/knowledge/KnowledgeAgent.ts:85:        settingSources: [],
src/services/worker/knowledge/KnowledgeAgent.ts:86:        strictMcpConfig: true,
src/services/worker/knowledge/KnowledgeAgent.ts:205:        settingSources: [],
src/services/worker/knowledge/KnowledgeAgent.ts:206:        strictMcpConfig: true,

$ grep -n "CLAUDE_CODE_USE_BEDROCK" src/supervisor/env-sanitizer.ts
32: *     \`CLAUDE_CODE_USE_BEDROCK\` / \`CLAUDE_CODE_USE_VERTEX\` were silently
43:  'CLAUDE_CODE_USE_BEDROCK',

$ grep -n "exec_command\\|MAX_OBSERVATION_BYTES" src/shared/SettingsDefaultsManager.ts src/services/worker/http/shared.ts
src/shared/SettingsDefaultsManager.ts:20:  CLAUDE_MEM_MAX_OBSERVATION_BYTES: string;
src/shared/SettingsDefaultsManager.ts:98:    CLAUDE_MEM_SKIP_TOOLS: '...,exec_command,write_stdin,apply_patch',
src/shared/SettingsDefaultsManager.ts:99:    CLAUDE_MEM_MAX_OBSERVATION_BYTES: '65536',
src/services/worker/http/shared.ts:212:  const MAX_OBSERVATION_BYTES = ...;

$ npm run build-and-sync   # exit 0
```

## Anti-pattern guards observed

- **Did not modify proxy strip** — `ENV_PROXY_VARS` still strips HTTP_PROXY/HTTPS_PROXY/etc. unconditionally per #2099/#2115.
- **Did not skip Read/Write/Edit/Bash** — those are the primary observation surface; only added the high-volume agentic-tool names (`exec_command`, `write_stdin`, `apply_patch`).
- **Did not invent SDK options** — both `settingSources` and `strictMcpConfig` were verified against `sdk.d.ts` before edits.
- **Did not break existing user settings** — users with their own `CLAUDE_MEM_SKIP_TOOLS` in `settings.json` keep their value; the change is to the default only.

## Test plan

- [x] `npm run build-and-sync` succeeds
- [x] Source greps confirm the 5 changes
- [ ] (Manual) Observe a session with `CLAUDE_CODE_USE_BEDROCK=1` set; confirm SDK subprocess inherits the var
- [ ] (Manual) Trigger a 100KB `exec_command` observation; confirm worker logs `payload_too_large` and no row is stored
- [ ] (Manual) User with custom MCP servers in `~/.claude/settings.json`; confirm those MCPs do NOT appear in the observer SDK subprocess

Closes #2155, #2159, #2171, #2194, #2199, #2201


thedotmack:
@claude figure out why at least one of the fixes in here is a band-aid of an edge case for a problem we solved, unless it's codex isolated, figure out what's going on here. ROOT CAUSE BOT PLS

aegir76:
I think you may have missed some bedrock callouts that I made on https://github.com/thedotmack/claude-mem/issues/2199#issuecomment-4344106720. There were some additional changes necessary to make bedrock work for me

thedotmack:

PR #2207, opened 2026-04-29
fix: chroma-mcp singleton lockfile + working periodic reaper (closes #2184, #2195)

### LINKED ISSUE #2184: Regression in v12.4.8: orphaned chroma-mcp processes accumulate again — PR #1175 reaper not effective (closed #1077 returns)
## Summary

In claude-mem **v12.4.8**, orphaned `chroma-mcp` processes are accumulating again — the same bug as #1077, which was closed 2026-02-24 with the comment that PR #1175 fixed it via `ensureProcessExit` in generator finally blocks plus a *"stale session reaper every 2 minutes"*.

On v12.4.8 the reaper is verifiably not catching these orphans: I observed pairs that have been alive **2.5 days (≈1,800 reaper cycles)** and one alive **4.5 days (≈3,240 reaper cycles)**.

This continues a pattern of regression on the exact same bug — see #1077, #1068, #1089, #1007, #650, and especially #701 which itself flagged a regression after #603/#650 were closed.

## Environment

- **claude-mem**: v12.4.8 (current latest)
- **chroma-mcp**: 0.2.6 (via `uv tool uvx`)
- **bun worker**: `worker-service.cjs --daemon`
- **OS**: Linux 6.6.87.2-microsoft-standard-WSL2 (Ubuntu, systemd-user enabled)
- **RAM**: 16 GB / **swap**: 16 GB / **vm.swappiness**: 10
- **Usage**: multiple Claude Code sessions on this machine over a 7-day uptime

### Orphan inventory: 13 pairs (26 processes), ~5.1 GB swap consumed

```
PID      RSS      Swap     etime          ppid              role
2672408   7.5MB   677MB    4d 12h         2672259→ppid=1   python chroma-mcp
2672259   5.2MB    49MB    4d 12h         ppid=1            uv tool wrapper
3747396   7.7MB   641MB    2d 19h         3747374→ppid=1   python chroma-mcp
3751333   6.0MB   632MB    2d 19h         3751307→ppid=1   python chroma-mcp
3754358   8.1MB   678MB    2d 19h         3754334→ppid=1   python chroma-mcp
3757951   7.5MB   700MB    2d 19h         3757925→ppid=1   python chroma-mcp
3765819   6.4MB   742MB    2d 19h         3765791→ppid=1   python chroma-mcp
... 12 pairs from 2026-04-26 19:05–19:17 (12-minute burst)
+ 1 pair from 2026-04-24
```

Every pair has the same structure:

- **Top**: `uv tool uvx --python 3.13 chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma` with **ppid=1** (re-parented to init)
- **Bottom**: actual python `chroma-mcp` process, child of the wrapper above

`ppid=1` confirms the wrapper outlived its original parent — i.e., an earlier `bun worker-service.cjs --daemon` instance exited without the cleanup propagating to its uvx subprocess tree.

### Why the PR #1175 reaper claim is verifiably wrong

The closing comment said *"stale session reaper every 2 minutes prevents orphaned chroma-mcp process accumulation"*. With a 120 s reaper, an orphan should not survive longer than ~2 min. Observed survival:

- 4.5 days = ~3,240 reaper cycles
- 2.5 days = ~1,800 reaper cycles × 12 distinct orphans

For all 13 orphans across multiple birth dates to survive thousands of cycles, one of these must be true in v12.4.8:

1. The reaper is not running at all
2. The reaper runs but does not match the `ppid=1 + uv tool uvx chroma-mcp` pattern
3. The reaper only kills processes spawned by the *current* worker (so previous-worker orphans are skipped)
4. `ensureProcessExit` in generator finally blocks doesn't fire on abnormal exit (SIGKILL, OOM, hooks.json restart race, daemon health-check restart)

### Spawn pattern hints at root cause

12 of 13 orphan pairs were spawned within a 12-minute window (Apr 26 19:05 → 19:17) — a signature of a **worker restart/crash loop** where each restart leaks one chroma-mcp pair. Cross-reference to #1239 ("hooks.json calls worker-service.cjs start on every hook event") which is closed but describes the same restart-storm dynamic.

### Symptoms

- Swap climbed monotonically to 8.7 GB over 7 days uptime; ~5.1 GB attributable to these orphans
- Each orphan python: VmRSS 6–8 MB / VmSwap 600–740 MB (kernel paged out idle anonymous memory entirely)
- VmSize per python orphan: ~10–11 GB (committed virtual)

## Reproduction

This is a soak test rather than instant repro:

1. Install claude-mem v12.4.8
2. Use Claude Code intensively across multiple sessions for several days, including SessionStart hooks that touch the worker
3. Wait for or induce a worker restart (hot restart paths internal to claude-mem may suffice)
4. Run:
   ```bash
   ps -eo pid,ppid,args | awk '$2==1 && /uv tool uvx.*chroma-mcp/' | wc -l
   ```
5. Observe non-zero count growing over time

Faster repro: `kill -KILL $(pgrep -f 'worker-service.cjs --daemon')` while a chroma-mcp call is in flight, then check `ppid=1` chroma-mcp count.

## Workaround for affected users

Frees orphans without touching the live worker:

```bash
ps -eo pid,ppid,args | awk '$2==1 && /uv tool uvx.*chroma-mcp/{print $1}' | xargs -r kill -TERM
sleep 2
ps -eo pid,ppid,args | awk '$2==1 && /uv tool uvx.*chroma-mcp/{print $1}' | xargs -r kill -KILL
# orphan python children typically die with their wrapper; sweep stragglers:
for p in $(pgrep -f 'archive-v0/.*chroma-mcp.*--data-dir.*claude-mem'); do
  ppid=$(awk '/^PPid:/{print $2}' /proc/$p/status 2>/dev/null)
  [ "$ppid" = "1" ] && kill -KILL $p
done
```

This freed 6.6 GB of swap on my machine (8.7 GB → 2.1 GB).

## Suggested investigation directions

1. **Verify the stale session reaper is actually scheduled and running in v12.4.8.** A debug log line per cycle would also catch future regressions.
2. **The reaper should also catch orphans whose parent worker has died** — match by command pattern and `ppid=1`, not just by tracked-PID set held in worker memory (which is wiped on restart).
3. **`ensureProcessExit` may not fire on SIGKILL or restart race**. Consider Linux-only `prctl(PR_SET_PDEATHSIG, SIGKILL)` on the spawned wrapper so the kernel auto-kills children on parent death; this is robust against any abnormal exit.
4. **Cross-restart recovery**: on worker startup, scan for `ppid=1` chroma-mcp processes matching this worker's data-dir and reap them before spawning new ones. The 12-orphans-in-12-minutes burst pattern shows that hot-restart loops compound the leak.

## Related

- **#1077** — this bug; closed 2026-02-24 with PR #1175 as the claimed fix (the fix has regressed)
- **#1175** — the fix PR (please verify whether its code paths are still in the v12.x trees)
- **#701** — flagged exactly this regression pattern in Jan 2026 (#603/#650 closed prematurely with v9.0.4)
- **#1068**, **#1089**, **#1007**, **#650**, **#1239** — variants in the same family

### LINKED ISSUE #2195: 🐛 Critical: chroma-mcp spawns multiple orphan processes, one hitting 90%+ CPU
## Summary

`chroma-mcp` spawns multiple persistent processes that are never cleaned up. One process consumed **90.6% CPU** and **896 MB RAM**, rendering the machine nearly unusable. Old orphan instances accumulate across sessions and survive indefinitely.

## Environment

- macOS Darwin 25.3.0 (Apple Silicon)
- Claude Code CLI
- claude-mem plugin (latest)
- chroma-mcp==0.2.6 via uv/Python 3.13

## Steps to Reproduce

1. Install claude-mem plugin
2. Use Claude Code normally across multiple sessions/days
3. Observe chroma-mcp processes accumulating

## Observed Behavior

```
USER       PID    %CPU  %MEM   COMMAND
huangzh..  5920   90.6   5.3   chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma
huangzh..  2664    0.0   0.0   chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma  (running since Monday)
huangzh..  4953    0.0   0.3   chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma
huangzh..  92237   0.0   0.2   chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma
huangzh..  5884    0.0   0.1   uvx chroma-mcp==0.2.6 --client-type persistent ...
huangzh..  4893    0.0   0.1   uvx chroma-mcp==0.2.6 --client-type persistent ...
```

- **4 active chroma-mcp Python processes** + 2 uvx parent processes
- Total memory: ~1 GB+
- One process (PID 5920) consumed 90.6% CPU for 11+ minutes
- Orphan process (PID 2664) ran since Monday with 95 min cumulative CPU time

## Expected Behavior

- Only ONE chroma-mcp instance should exist at a time
- Old instances should be cleaned up when new sessions start
- CPU usage should be bounded (not 90%+ sustained)

## Impact

**Severity: Critical**

- Machine becomes nearly unusable due to CPU starvation
- 1 GB+ RAM wasted on duplicate processes
- Users may not realize the cause (process runs in background, no visible window)
- Requires manual `kill` to resolve — normal users won't know to do this

## Suggested Fixes

1. **Singleton enforcement**: Use a PID file or lock file to ensure only one chroma-mcp instance runs. Kill stale instances on startup.
2. **CPU guard**: Add a watchdog that kills the process if CPU exceeds a threshold for sustained periods.
3. **Cleanup on shutdown**: Register a proper shutdown hook in the MCP server lifecycle to terminate the chroma-mcp subprocess when Claude Code exits.
4. **Process reuse**: Instead of spawning a new `uvx chroma-mcp` per session, detect and reuse the existing running instance.

thedotmack:
## Summary

Wires up the orphan reaper that PR #1175 left disconnected, plus a singleton lockfile so the reaper can identify chroma-mcp orphans by their specific dead worker. Closes two long-running issues:

- **#2184** — 13 chroma-mcp orphan pairs surviving 4d 12h with `ppid=1`
- **#2195** — 4+ concurrent chroma-mcp processes, one at 90% CPU / 1 GB RAM

## Why

PR #1175's docstring claimed a "periodic ProcessRegistry reaper" but `cleanupOrphanedProcesses()` was never called from anywhere — `grep -rn cleanupOrphanedProcesses src/` returned exactly one hit (its own definition). When a worker died via SIGKILL or OOM, every chroma-mcp it had spawned re-parented to init and was never collected.

### 1. Singleton lockfile (`ChromaMcpManager`)
- New `~/.claude-mem/chroma-mcp.pid` (path: `CHROMA_MCP_PID_PATH` in `src/shared/paths.ts`, respects `CLAUDE_MEM_DATA_DIR`)
- Atomic write (tmp + rename) recording `<chroma child PID>\n<worker PID>`
- Written after `registerManagedProcess()` in `connectInternal()`; removed in `stop()` and on the chroma child's `exit` event

### 2. Smarter orphan detection (`ProcessManager`)
- Per-pattern age thresholds: `chroma-mcp` = 5 min, others = 30 min
- **Unix**: parses `ps -eo pid,ppid,etime,command` and unconditionally orphans `chroma-mcp` rows with `ppid==1` (re-parented to init)
- **Windows**: pulls `ParentProcessId` from the WMI query and orphans `chroma-mcp` whose parent is not the live `worker.pid`
- New `readLiveWorkerPid()` helper checks the worker.pid file is alive via `kill(pid, 0)`

### 3. Periodic reaper wiring (`worker-service`)
- New exports `startOrphanReaperInterval()` / `stopOrphanReaperInterval()`
- Worker `start()` schedules a 120s interval immediately after `writePidFile()` (initial reap fires synchronously)
- `shutdown()` clears the interval before `performGracefulShutdown` so a graceful stop doesn't race the next tick
- Timer `.unref()`'d — does not keep the process alive on its own

## Verification

\`\`\`bash
# Build
npm run build               # ✅ all targets compile

# Existing tests
bun test tests/infrastructure/process-manager.test.ts tests/infrastructure/wmic-parsing.test.ts
# 80 pass / 0 fail / 127 expects

# Reaper behavior (run after merge):
ps -eo pid,ppid,args | awk '$2==1 && /uv tool uvx.*chroma-mcp/' | wc -l
pkill -KILL -f 'worker-service.cjs --daemon'
sleep 180   # one reaper interval
ps -eo pid,ppid,args | awk '$2==1 && /uv tool uvx.*chroma-mcp/' | wc -l   # must be 0
ls -la ~/.claude-mem/chroma-mcp.pid                                       # absent after stop
\`\`\`

## Anti-pattern guards

- `ORPHAN_PROCESS_PATTERNS` extended only via the override map — original three patterns preserved
- `CHROMA_MCP_PID_PATH` defined in `src/shared/paths.ts` next to `DATA_DIR`, so it respects `CLAUDE_MEM_DATA_DIR`/multi-account profiles
- Misleading "periodic ProcessRegistry reaper handles in-session orphans" comment replaced with accurate description naming PR #1175 + #2184 + #2195

Closes #2184
Closes #2195


thedotmack:
@coderabbitai full review please — the major finding on `ProcessManager.ts:474` (Windows chroma orphan check) was addressed in commit `bfce568b` with an `isProcessAlive(ppid)` guard. The dead-lockfile finding was addressed in `4f12252f` by removing the lockfile entirely.

thedotmack:
## Holding this PR — wrong abstraction

Reviewed this against the rest of the batch and want to hold rather than merge.

The 120s polling reaper + per-pattern age thresholds (5min chroma-mcp, 30min others) + lockfile + ppid heuristics are the wrong abstraction for child-process lifecycle. The numbers are invented, the heuristics are already wrong on Windows (CodeRabbit caught this — intermediate \`cmd.exe\`/\`uvx\` parents make ppid checks unreliable), and PR #1175 attempted the same shape and sat dead for months.

The structural fix is kernel-enforced cleanup:
- **Unix**: \`setpgid\` per-child + \`prctl(PR_SET_PDEATHSIG)\` on Linux / \`kqueue(EVFILT_PROC, NOTE_EXIT)\` on macOS.
- **Windows**: Job Objects with \`JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE\`.

Both run with zero polling and zero invented thresholds. When the worker dies (even via SIGKILL), children are guaranteed dead by the kernel within microseconds.

Tracked in #2216. Closing this PR — the user-facing symptoms (#2184, #2195) will be resolved by construction once #2216 lands.

Thanks for the careful work — the audit of the existing reaper code in this PR is exactly what made the wrong-abstraction visible.

Issue #2220, opened 2026-04-30
[Bug] chroma-mcp process storm causes sustained 60-80% CPU on Windows


ONEGAYI:
## Summary

When Chroma is enabled, every Claude Code startup spawns a persistent Python process (chroma-mcp) consuming 60–80% CPU indefinitely on Windows. Killing the process only delays the problem until the next launch.

## Environment

- **OS**: Windows 11 Pro (x64)
- **claude-mem**: v12.4.9
- **chroma-mcp**: 0.2.6 (pinned)
- **Python**: 3.13 (via uv)
- **Data**: 16,256 observations, 64,913 embeddings, 80+ projects in chroma-sync-state.json

## Observed Process Storm

A single startup produced **6 chroma-mcp-related processes**:

```
PID=25260  cmd.exe /c uvx --python 3.13 chroma-mcp==0.2.6 --client-type persistent --data-dir ...
PID=35572  uvx --python 3.13 chroma-mcp==0.2.6 --client-type persistent --data-dir ...
PID=16456  uv.exe tool uvx --python 3.13 chroma-mcp==0.2.6 --client-type persistent --data-dir ...
PID=33552  chroma-mcp.exe --client-type persistent --data-dir ...
PID=36252  python.exe chroma-mcp.exe --client-type persistent --data-dir ...
PID=36268  python.exe chroma-mcp.exe --client-type persistent --data-dir ...  ← CPU=2392s, MEM=486MB
```

### 1. Four-layer nested process chain on Windows (`ChromaMcpManager.ts:137-139`)

```typescript
const isWindows = process.platform === 'win32';
const uvxSpawnCommand = isWindows ? (process.env.ComSpec || 'cmd.exe') : 'uvx';
const uvxSpawnArgs = isWindows ? ['/c', 'uvx', ...commandArgs] : commandArgs;
```

This produces `cmd → uvx → uv → python → chroma-mcp`. When `StdioClientTransport.close()` sends stdin close → SIGTERM, the signal doesn't fully propagate on Windows, leaving orphaned child processes. This aligns with #1063 where `transport.close()` only closes the MCP connection but doesn't kill the underlying uvx/python subprocess.

### 2. Unbounded ONNX threads (`ChromaMcpManager.ts:551-575`)

`getSpawnEnv()` builds the subprocess environment with **no thread limits**:

```typescript
private getSpawnEnv(): Record<string, string> {
    const baseEnv: Record<string, string> = {};
[...]

Issue #2216, opened 2026-04-29
Process lifecycle: kernel-enforced child cleanup (replaces #2207 reaper)


thedotmack:
## Problem

Worker child processes (chroma-mcp, claude.exe SDK subprocesses) become orphans when the worker dies via SIGKILL or crash. PR #2207 attempted to address this with a polling reaper that wakes every 120s, parses \`ps -eo\` output, applies pattern-specific age thresholds (5min for chroma-mcp, 30min others), reads a lockfile, and decides whom to kill.

That's the wrong abstraction:

1. The thresholds (120s, 5min, 30min) are invented numbers, not derived from anything.
2. The heuristics (\`ppid==1\`, cmdline pattern matching, age) are already wrong on Windows where intermediate \`cmd.exe\`/\`uvx\` parents legitimately exist (CodeRabbit flagged this on #2207).
3. PR #1175 tried the same approach, the reaper sat dead for months — same shape, same outcome.

## Structural fix

**Unix**: spawn each child with \`setpgid\` (new process group), so \`kill(-pgid, SIGTERM)\` from the worker's exit handler tears down the whole group. If the worker dies via SIGKILL, register cleanup via \`prctl(PR_SET_PDEATHSIG, SIGTERM)\` on Linux so the kernel sends the signal automatically when the parent dies. On macOS, use \`kqueue(EVFILT_PROC, NOTE_EXIT)\` from a small supervisor.

**Windows**: assign each child to a Job Object created with \`JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE\`. Closing the worker's job handle (or the worker dying) terminates every process in the job. Kernel-enforced, race-free.

Both mechanisms run with zero polling, no thresholds, no heuristics. Children of a dead worker are guaranteed dead by the kernel within microseconds.

## Affected call sites

- \`src/services/sync/ChromaMcpManager.ts\` (chroma-mcp spawn)
- \`src/supervisor/process-registry.ts\` (\`spawnSdkProcess\`, \`createSdkSpawnFactory\`)
- \`src/services/infrastructure/ProcessManager.ts\` (worker daemon spawn — lower priority since worker is the parent)

## Replaces

- The polling reaper / lockfile / age-threshold scaffolding from PR #2207. Once kernel-enforced cleanup is in, that PR's mechanism is dead code and should be removed entirely.
- Resolves the user-facing symptoms in #2184, #2195 by construction rather than by janitor.

## References

[...]

Issue #2213, opened 2026-04-29
claude-mem becomes super slow after 3-4hrs with multiple paralell sessions


Lesani:
duplicate of the "cleanup" closed #1766 

I have 3-4 sessions open usually, rarely over 200k context, after 3-4 hrs the system becomes slower and slower, churning after each message for minutes at the extreme.

Lesani:
It is less noticeable on long running sessions, but when you do a /grill-me or similar, every question the agent asks goes for minutes on the "stop hook" after that time. Often I find myself deactivating the plugin when that is the case, or just "ESC" aborting its thing, which is possible

DieterEOnofre:
I noticed this same issue right after installing this. I was also using 100% of my sessions token limits with 2-3 prompts. It was insane.

Lesani:
It is particularly agregious also when running local models, I can regularly see my gpu idle, no propmt processing, despite 3 sessions rinnung, because claude-mem is doing its hook thing for 3 minutes.
At this time all CC sessions are <100k context used
The quality of generation is still worth having it active, but this desperatelly needs to get a consistent speed

thedotmack:
Consolidating into #2378 (plan-03 Worker Lifecycle). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

PR #2221, opened 2026-04-30
fix: neutralize chroma-mcp CPU storm (thread caps + per-batch watermark)

### LINKED ISSUE #2214: fix: persist Chroma sync watermark per-batch to survive worker crashes
When a backfill of many observations was interrupted mid-flight (worker
SIGKILL, OOM, system reboot), the watermark was never advanced because
ChromaSyncState.bump() ran only in the trailing finally block. On the
next start the worker would resume from watermark=0, re-submit every
already-indexed document, hit "already exist" on every batch, and fall
into the delete-then-add reconcile path — re-running the embedding
model for the entire history and pegging chroma-mcp at 100% across all
cores indefinitely.

Move the bump into the for loop, after each successful addDocuments
call, so progress is durable. backfillObservations, backfillSummaries
and backfillPrompts all get the same treatment.

The reconcile path already handles real duplicate-conflict edge cases
correctly; this change just makes sure the next run never has to walk
through them for already-synced ranges.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

https://github.com/thedotmack/claude-mem/commit/369fce38a11c90383c94ed7ae6348a7be837daf5

thedotmack:
## Summary

Two surgical fixes to the chroma backfill path that together cause the sustained 60–80% CPU + orphan accumulation pattern reported across #2220, #2214, #2213, #2183 (and closed duplicates #2184 / #2195).

### 1. `ChromaMcpManager.getSpawnEnv()` — no thread limits at all

ONNX Runtime / OpenBLAS / MKL all default to `cpu_count()`. A 12-core box runs 12 threads burning embeddings in parallel. The existing `getSpawnEnv()` only injected SSL cert paths — zero thread caps, zero telemetry override. This is the dominant cause of the chroma-mcp CPU storm on Windows.

**Fix:** inject `OMP_NUM_THREADS=2` / `ONNX_NUM_THREADS=2` / `OPENBLAS_NUM_THREADS=2` / `MKL_NUM_THREADS=2` defaults (only if the user hasn't pinned them), and `ANONYMIZED_TELEMETRY=false` to stop background HTTP from the embedding subprocess.

### 2. `ChromaSync.backfill{Observations,Summaries,Prompts}` — watermark only bumped in trailing `finally`

```ts
try {
  for (const batch of allDocs) await this.addDocuments(batch);  // mid-flight SIGKILL?
} finally {
  ChromaSyncState.bump(...);  // never runs on SIGKILL — process dies
}
```

SIGKILL / OOM / power loss mid-flight skips `finally` entirely. Watermark stays at 0. Next worker boot re-submits 16K observations (the report from #2220), which lands in chroma-mcp's "already exists → delete-then-add reconcile" path, re-embedding the entire history. Combined with bug (1), that's where the sustained 100% CPU comes from.

**Fix:** move the bump inside the per-batch loop so progress is durable. The bump method is already idempotent.

## Why these two together

Either fix alone helps. Together they break the storm at the root:
- Even if backfill runs, it can't peg the CPU (thread cap)
- Even if the worker dies mid-flight, it doesn't re-do the whole history (durable watermark)

The structural Job-Object cleanup for orphan processes is separately tracked at #2216.

## Verification

- `tests/services/sync/*` + `tests/integration/chroma-vector-sync.test.ts`: **26/26 pass**
- Bundle (`plugin/scripts/worker-service.cjs`) confirmed to contain the thread-cap env keys and the per-batch `bump()` calls
- Full suite: **1429 pass / 20 fail** — pre-existing failures only, no regressions vs v12.4.9 baseline (1429 pass / 27 fail)

## Closes
- Closes #2214

## Refs (substantially de-amplifies)
- #2220 (canonical chroma-mcp Windows storm umbrella)
- #2213, #2183 (slow startup / sessions)


thedotmack:
Superseded by #2282 (merged in d384d3c5). Cherry-picked into the bug-batch PR (chroma CPU storm root-cause fix with thread caps + per-batch watermark).

Issue #2225, opened 2026-04-30
undefined is not an object (evaluating 'n._zod.def') 2154 reopen


morpheus-chang:
#2154 reopen

claude-mem: 12.4.9
OpenCode: 1.14.30
Node.js: 22.22.1
Platform: macOS m1 26.4.1 (25E253)

<img width="3430" height="1426" alt="Image" src="https://github.com/user-attachments/assets/c5e158a6-f40a-43ff-9807-69001612789b" />

<img width="2608" height="1908" alt="Image" src="https://github.com/user-attachments/assets/c2d5e035-d271-480c-b2d8-faa2f04f0993" />

thedotmack:
Fixed by #2282 (merged in d384d3c5). Plain JSON-schema arg shapes replaced with proper Zod schemas in the opencode plugin.

Issue #2154, opened 2026-04-27
TypeError: undefined is not an object (evaluating 'n._zod.def') — OpenCode hook execution fails due to Zod version conflict in bundled scripts(reopen)


morpheus-chang:
https://github.com/thedotmack/claude-mem/issues/2113   reopen

claude-mem: 12.4.7
OpenCode: 1.14.25
Node.js: 22.22.1
Platform: macOS m1 26.4.1 (25E253)

rpiwnicki:
Same problem on Windows 11.  
After global install `npx claude-mem install` for Claude, Cursor and OpenCode. 
OpenCode is not working, error after first message. 

_OpenCode: v1.14.28
ClaudeCode: v12.4.7
Node: v24.15.0_


<img width="1323" height="559" alt="Image" src="https://github.com/user-attachments/assets/ef89df30-f3c9-4d08-b0ac-3a172de97ba1" />

usasuperhero2026-wq:
I have the same issue here, 
macOS 
opencode: 1.14.28
claude-mem: 12.4.7
Node: v24.11.1

<img width="2281" height="321" alt="Image" src="https://github.com/user-attachments/assets/0cce9f96-6526-4e72-bfc7-3ce490d6c06b" />

thedotmack:
Fixed in v12.4.9 via #2211 — Zod conflict was caused by stale bundled artifacts. Bundle regenerated cleanly. Please retry on 12.4.9.

kpaaro:
Getting the issue today.

Latest claude-mem and opencode

PR #2227, opened 2026-04-30
fix: use Zod schemas in OpenCode plugin to fix _zod.def crash

### LINKED ISSUE #2226: opencode plugin crashes at registration: tool args are plain JSON-schema objects, not Zod schemas (_zod.def undefined on 12.4.9)
Filing as a separate issue from #2225 / #2154 because the root cause is distinct from the macOS Mach-O binary drift addressed by PR #2211, and PR #2211 (still open) will not fix this case even when merged.

### Environment

- claude-mem: **12.4.9** (installed via `npx claude-mem install`)
- opencode: **1.14.29** (`@opencode-ai/plugin` 1.14.29 bundled, ships Zod 4.1.8)
- OS: **Windows 11**
- Node: n/a — opencode runs under bundled Bun (`B:/~BUN/root/...`)

### Symptom

Opencode crashes at startup as soon as it tries to register the claude-mem plugin:

```
TypeError: undefined is not an object (evaluating 'n._zod.def')
  at process (B:/~BUN/root/chunk-h58rssve.js:22:198266)
  at process (B:/~BUN/root/chunk-h58rssve.js:22:201152)
  at Bv (B:/~BUN/root/chunk-h58rssve.js:24:2057)
  ...
```

This happens before the worker is contacted, so the worker port / state is irrelevant.

### Root cause

The plugin file installed at `~/.config/opencode/plugins/claude-mem.js` declares its tool with plain JSON-schema-style args:

```js
tool: {
  claude_mem_search: {
    description: "Search claude-mem memory database for past observations, sessions, and context",
    args: {
      query: { type: "string", description: "Search query for memory observations" }
    },
    async execute(s) { ... }
  }
}
```

But opencode's plugin SDK requires each entry in `args` to be a **Zod v4 schema**, built via `tool.schema.string()` (per https://opencode.ai/docs/plugins/). At registration, opencode walks each arg and reads `arg._zod.def` to convert it. Because `query` is a plain object, `query._zod` is `undefined` and `.def` throws — exactly the trace above.

### Why PR #2211 doesn't fix this

PR #2211 deletes the stale macOS Mach-O binary and regenerates `worker-service.cjs` / `mcp-server.cjs`. It does not touch the opencode plugin emitter. The opencode plugin file shipped in 12.4.9 still has the JSON-schema-style args shown above — verified on a fresh Windows install.

This also explains why users on macOS report the same crash on 12.4.9 (issue #2225): the opencode plugin path is broken on every platform, independent of the Mach-O binary issue.

### Reproducer

1. `npx claude-mem install` (any platform) — accept opencode integration.
2. Launch opencode in any project.
3. Crash on first message with \`n._zod.def\` TypeError.
4. Workaround: rename \`~/.config/opencode/plugins/claude-mem.js\` to \`.disabled\` → opencode boots cleanly.

### Suggested fix

In whatever source generates \`~/.config/opencode/plugins/claude-mem.js\`, import \`tool\` from \`@opencode-ai/plugin\` and wrap the tool definition:

\`\`\`js
import { tool } from "@opencode-ai/plugin";

// ...
tool: {
  claude_mem_search: tool({
    description: "Search claude-mem memory database...",
    args: {
      query: tool.schema.string().describe("Search query for memory observations")
    },
    async execute(args, ctx) { /* ... */ }
  })
}
\`\`\`

Happy to PR this if you'd like — just point me at the source file that emits the opencode plugin.

thedotmack:
## Summary
- Replace plain JSON Schema `{type: "string"}` args with `z.string().describe()` in the OpenCode plugin
- OpenCode 1.14.x walks `arg._zod.def` at registration — crashes on plain objects lacking `_zod` metadata
- Zod v4.3.6 already installed; build verified with 472 `_zod` references in bundle

Closes #2226, #2225, #2154

## Test plan
- [x] Build succeeds (`npm run build-and-sync`)
- [x] Bundle at `dist/opencode-plugin/index.js` contains `_zod` references
- [ ] Test with OpenCode 1.14.x — plugin registers without crash

### DIFF
--- a/src/integrations/opencode-plugin/index.ts
+++ b/src/integrations/opencode-plugin/index.ts
@@ -13,6 +13,8 @@
  * - claude_mem_search: Search memory database from within OpenCode
  */
 
+import { z } from "zod";
+
 // ============================================================================
 // Minimal type declarations for OpenCode Plugin SDK
 // These match the runtime API provided by @opencode-ai/plugin
@@ -330,10 +332,7 @@ export const ClaudeMemPlugin = async (ctx: OpenCodePluginContext) => {
         description:
           "Search claude-mem memory database for past observations, sessions, and context",
         args: {
-          query: {
-            type: "string",
-            description: "Search query for memory observations",
-          },
+          query: z.string().describe("Search query for memory observations"),
         },
         async execute(
           args: Record<string, unknown>,


nhhoang (approved):

thedotmack:
Superseded by #2282 (merged in d384d3c5). Cherry-picked into the bug-batch PR with conflicts resolved.

Issue #2315, opened 2026-05-05
Background observer consumed 76M+ Claude tokens in one day and exhausted Claude Code limit


sarthakagrawal927:
## Summary

`claude-mem` appears to have spawned a large number of background Claude SDK observer sessions and exhausted my Claude Code subscription limit. This looked like random subscription usage until I inspected local `~/.claude` / `~/.claude-mem` state.

I am filing this because the behavior seems unsafe by default: the background worker continued processing a queued backlog through Claude CLI auth, generated ~1,010 Sonnet calls in a day, and only stopped once Claude returned a usage-limit message.

## Environment

- Plugin: `claude-mem@thedotmack`
- Installed version: `12.6.0`
- Claude Code version seen in observer sessions: `2.1.128` for recent sessions
- Provider config: `CLAUDE_MEM_PROVIDER=claude`
- Auth config: `CLAUDE_MEM_CLAUDE_AUTH_METHOD=cli`
- Model config: `CLAUDE_MEM_MODEL=claude-sonnet-4-6`
- Transcript ingestion: `CLAUDE_MEM_TRANSCRIPTS_ENABLED=true`
- Max concurrent agents: `CLAUDE_MEM_MAX_CONCURRENT_AGENTS=2`
- Local timezone: Asia/Kolkata

## Impact

For local day **2026-05-05**, files under:

`~/.claude/projects/-Users-<redacted>--claude-mem-observer-sessions/*.jsonl`

showed this aggregate usage from assistant message `usage` fields:

```text
model              calls  input  output  cache_creation  cache_read
claude-sonnet-4-6  1010   2934   613070  14724490        61408576
<synthetic>        89     0      0       0               0
```

Total recorded tokens for `claude-sonnet-4-6`:

```text
2,934 normal input
613,070 output
14,724,490 cache creation input
61,408,576 cache read input
= 76,749,070 total recorded usage tokens
```

Using Anthropic's public Sonnet pricing as an API-equivalent estimate, this is roughly **$82.84** of token usage if the cache writes are 5-minute writes. I understand Claude Code subscription usage is not necessarily billed as API dollars, but the usage still consumed the shared Claude limit.

## Local evidence

Recent process inspection showed `claude-mem` spawning Claude SDK sessions through CLI auth:

```text
[...]

PR #2316, opened 2026-05-05
[codex] fix observer non-xml retry loop

thedotmack:
## Summary
- Treat non-XML/empty observer responses as intentionally ignored output instead of re-queueing the same pending batch.
- Clear the pending batch and reset the pending timestamp when parsing fails, matching the existing prompt contract that non-XML text is discarded.
- Update response-processor tests to assert plain-text and empty responses are drained, not retried.

## Root Cause
Issue #2315 showed that Claude Code limit text was flowing through as plain assistant text. The parser rejected it as non-XML, but the response processor reset claimed rows back to pending, making the same batch eligible for another generator pass. The actual invariant should be simple: XML is parsed; non-XML is ignored as if no memory output existed.

## Validation
- `bun test tests/worker/agents/response-processor.test.ts -t "non-XML observer responses"`
- `bun test tests/worker/agents/response-processor.test.ts -t "handling empty / non-XML response"`

## Notes
The full `response-processor.test.ts` file still has unrelated pre-existing failures outside this change scope.

### DIFF
--- a/src/services/worker/agents/ResponseProcessor.ts
+++ b/src/services/worker/agents/ResponseProcessor.ts
@@ -35,14 +35,14 @@ export async function processAgentResponse(
   const parsed = parseAgentXml(text, session.contentSessionId);
 
   if (!parsed.valid) {
-    logger.warn('PARSER', `${agentName} returned unparseable response — leaving queue intact`, {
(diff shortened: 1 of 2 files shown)


thedotmack:
@greptile review

Issue #2341, opened 2026-05-06
Consolidated issue triage: feature/system backlog and root-cause todo list


thedotmack:
Snapshot: 2026-05-06, open issues #2144 through #2340. I triaged the current open backlog by affected feature/system and treated reported version numbers as signal, not noise: several reports are not current-release bugs, but are upgrade-path regressions from older DBs or installer caches.

## Executive ordering

1. **P0 security / data-loss / hard-breakers**
   - #2332 observer SDK session can inherit filesystem-write tools despite prompt saying it has no tools. Root bug: observer sandbox/tool permissions are not enforced at process/session boundary.
   - #2340 and #2318 show schema drift where the DB records a migration/version state but required columns are absent. Root bug: runtime trusts `schema_versions` too much; boot should validate concrete columns/indexes used by current SQL.
   - #2338, #2337, #2330, #2328, #2251, #2224 are current install/MCP startup failures, mostly around plugin path/dependency resolution.
   - #2301, #2311, #2310, #2291, #2213, #2167 are one operational failure family: sessions/queues/backlogs do not consistently complete, drain, or self-clean under daemon and multi-session workloads.
2. **P1 reliability / upgrade hygiene**
   - Resolve install matrix failures from v12.5.0 (#2273 plus #2262-#2272), Node 25/tree-sitter failures (#2280, #2245), `.install-version` mismatch (#2327), stale worker/cache issues (#2279, #2246, #2218), worker start race (#2260), and versioned MCP dependency skew (#2251).
3. **P2 product controls / integration completeness**
   - Finish OpenCode/Codex/OpenClaw integration gaps (#2329, #2296, #2295, #2259, #2249, #2334, #2335), provider/model routing (#2239, #2289, #2276, #2202, #2294), and cost/context controls (#2336, #2303, #2288, #2217, #2193, #2339, #2333).

## Root-cause bugs identified

| Root cause | Evidence | Todo |
| --- | --- | --- |
| Runtime schema uses version trust where column/index existence is required | #2340 reports v12.7.2 still missing `pending_messages.tool_use_id`/`worker_pid` after upgrade from v12.4.4; #2318 reports 9.0.12 missing `retry_count`/`failed_at_epoch` | Add unconditional boot-time DB integrity repair for current SQL dependencies. Verify `pending_messages`, `session_summaries`, indexes, and route contracts independent of `schema_versions`. |
[...]

PR #2344, opened 2026-05-06
fix: harden startup and schema repair contracts

thedotmack:
## What this PR covers

This is the first reliability slice from #2341. It does not try to close the whole triage backlog; it makes the startup, install, export, and current queue/schema contracts less fragile so follow-up fixes have a stable base.

## Root causes addressed

- MCP and hook startup commands could resolve the plugin root to an empty string, which turns the server path into `/scripts/mcp-server.cjs`. The commands now use `CLAUDE_CONFIG_DIR`, cache install paths, marketplace paths, and explicit plugin root env vars before falling back.
- The install marker reader only understood the newer JSON marker format. Existing installs can still have the older plain-text version marker, so matching legacy markers are now accepted while new writes stay on the documented JSON shape.
- `export-memories` had a drifted API/data-dir contract: it could miss `CLAUDE_MEM_DATA_DIR`, it sent the old `sdkSessionIds` name, and it was hard to test as a script. The script now resolves the data dir at call time, sends canonical `memorySessionIds`, and the worker route accepts the legacy alias for compatibility.
- `pending_messages` repair logic mixed current schema needs with old worker/zombie-queue fields. The boot repair now focuses on the current `tool_use_id` requirement and removes the stale `worker_pid` expectation from the schema and guardrails.

## Not covered here

This PR intentionally leaves the remaining #2341 feature groups for follow-up issues/PRs: broader zombie queue semantics, search/recall behavior, OpenClaw/event handling, logger standards cleanup, and the larger stale test-suite repairs.

## Verification

Passed:

- `bun test tests/setup-runtime.test.ts tests/plugin-version-check.test.ts tests/scripts/export-memories.test.ts tests/worker/http/routes/data-routes-coercion.test.ts tests/services/sqlite/PendingMessageStore.test.ts tests/session_store.test.ts tests/infrastructure/plugin-distribution.test.ts`
- `git diff --cached --check`
- JSON parse check for `.mcp.json`, `plugin/.mcp.json`, `plugin/hooks/codex-hooks.json`, and `plugin/hooks/hooks.json`
- `sqlite3 :memory: ".read src/services/sqlite/schema.sql"`

Known existing failures outside this slice:

- `npm run typecheck:root` stops before project checking on TS5107 because the root tsconfig still uses deprecated `moduleResolution=node10` without `ignoreDeprecations`.
- Full `bun test` is not green on `main`-scale scope. The latest local run had focused tests passing but broader failures in areas such as logger standards, stale zombie-queue API tests, GeminiProvider mocks, OpenClaw events, summarize handler logger mocks, worker-spawner expectations, and SearchRoutes welcome-hint expectations.

## Review note

Review only. Please do not merge or ship this without Alex's explicit confirmation.


thedotmack:
Follow-up after checking the latest issues:

- #2345 is the retry-loop/quota issue. Root cause in this branch was that `ClaudeProvider` could correctly mark a session as `overflow` or `quota:*`, but `GeneratorExitHandler` still treated those exits like restartable pending work. This now hard-stops those reasons: clear pending rows, finalize the session, remove it from the active session map, and do not schedule `pending-work-restart`.
- #2346 is the Stop-hook missing-transcript loop. The normal summarize handler already skips missing transcripts, but the top-level hook command now also classifies missing transcript-path errors as non-blocking and exits success with continue/suppress output.
- Review follow-ups are handled: export requests have timeouts, migration cleanup versions only stamp after successful drops, Codex hooks got the same `cygpath` handling as Claude hooks, and the nullable `tool_use_id` behavior is documented/tested because summarize and legacy rows may not have a tool-use id.

Local validation after the final amendment:

- `npm run build`
- `bun test tests/scripts/export-memories.test.ts tests/services/sqlite/PendingMessageStore.test.ts tests/infrastructure/plugin-distribution.test.ts tests/services/worker/generator-exit-handler.test.ts tests/hook-command.test.ts tests/worker/provider-classifiers.test.ts tests/worker/rate-limit-store.test.ts` — 109 passing
- `git diff --check`

`npm run typecheck:root` is still blocked by the existing TS 6 `moduleResolution=node10` deprecation in `tsconfig.json`, before project files are typechecked.

thedotmack:
@coderabbitai review

thedotmack:
@greptileai

thedotmack:
@greptileai

thedotmack:
@coderabbitai review

thedotmack:
@coderabbitai review

thedotmack:
@greptileai

PR #2365, opened 2026-05-08
fix: stop Codex transcript replay after hooks migration

thedotmack:
## Summary
- disable default Codex transcript ingestion now that native hooks are authoritative
- prevent transcript watcher startAtEnd files discovered after startup from replaying historical prompts
- repair Codex install config so it enables `[features] hooks = true` and `claude-mem@claude-mem-local` instead of stale/deprecated hook/plugin identifiers

## Tests
- `bun test tests/integration/codex-cli-installer.test.ts tests/transcripts/config.test.ts tests/transcripts/watcher-start-at-end.test.ts tests/install-non-tty.test.ts`

## Notes
This keeps transcript watching as a generic receiver for non-Codex systems, while filtering the legacy `~/.codex/sessions/**/*.jsonl` watch unless explicitly opted in with `CLAUDE_MEM_CODEX_TRANSCRIPT_INGESTION=true`.

### DIFF
--- a/src/services/integrations/CodexCliInstaller.ts
+++ b/src/services/integrations/CodexCliInstaller.ts
@@ -1,7 +1,7 @@
 import path from 'path';
 import { homedir } from 'os';
 import { execFileSync, spawnSync } from 'child_process';
-import { existsSync, readFileSync, writeFileSync } from 'fs';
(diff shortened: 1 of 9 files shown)


cykofrn (commented):
function setTomlBooleanInTable(content: string, header: string, key: string, enabled: boolean)

PR #2409, opened 2026-05-11
feat(skills): add oh-my-issues for root-cause issue clustering

thedotmack:
## Summary

Adds `plugin/skills/oh-my-issues/SKILL.md` — a skill that codifies the issue-consolidation method used during the v13.0.1 cycle (49 child issues → 6 plan masters → one bundled PR). Auto-invokes when users ask to triage / consolidate / cluster / dedupe issues, build a roadmap from open issues, or route a new bug into an existing plan.

The skill encodes three modes:

- **Cluster pass** — initial reduction of an unconsolidated backlog into N plan-master issues + `plans/0X-*.md` design docs. Clusters by root cause (*"would one architectural change retire all of these?"*), not by surface.
- **Triage** — steady-state routing of a new incoming bug into an existing master via a "Round N" comment with a fix sketch, then closing the child with a standardized redirect.
- **Bundle** — generating a PR description that closes every child of a plan master atomically with `Closes #N`, plus the test-matrix CI changes that prevent the cluster from re-emerging.

Also includes a naming guide (surface vs architectural), the locked-down redirect comment, `gh` CLI primitives for each step, a `plans/0X-<slug>.md` body template, and health checks (graveyard / over-broad / surface-clustered masters, doc/issue drift).

## Test plan

- [ ] Run `npm run build-and-sync` and confirm the skill is registered without errors
- [ ] Verify the skill auto-invokes on a prompt like "triage the open issues into plans"
- [ ] Cross-reference the SKILL.md gh CLI snippets against the actual commands used during the v13.0.1 consolidation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2412, opened 2026-05-11
feat(skills): add pr-consensus for cross-bot consensus reports

thedotmack:
## Summary

Adds `plugin/skills/pr-consensus/SKILL.md` — a skill that runs a 3-round open-ended interview against both CodeRabbit and Greptile on a GitHub PR, then posts a consensus report identifying findings that appeared independently in both bots.

## What it does

Given a PR number, the skill:

1. Confirms both bots have posted initial reviews
2. Posts identical, fully open-ended prompts to each bot (no topic anchoring)
3. Waits for both bots to reply, then prods generically ("anything else?")
4. Asks each bot for its actionable changes
5. Synthesizes the responses, identifying which findings appeared independently in both bots — those are the high-signal items
6. Posts the consensus report as a top-level PR comment

## Why this matters

When two distinct LLM reviewers independently surface the same critique, you have the highest-signal possible feedback. Reviewer-style differences (CodeRabbit structural, Greptile surgical) cancel out at the agreement layer.

## How the prompts were chosen

The exact prompt phrasing is locked down — it's the version that produced the validated four-round dialogue on PR #2409, where three findings emerged independently from both bots in agreement.

## Test plan

- [ ] Skill auto-invokes on prompts like "run consensus on PR X" or "cross-review PR X"
- [ ] Skill does NOT auto-invoke on generic PR review requests (user-invoked only)
- [ ] Running the skill on this PR produces a consensus report — this PR is itself the first dogfooded validation

### DIFF
--- a/plugin/skills/pr-consensus/SKILL.md
+++ b/plugin/skills/pr-consensus/SKILL.md
@@ -1,0 +1,162 @@
+---
+name: pr-consensus
+description: Run a 3-round open-ended interview against both CodeRabbit and Greptile on a GitHub PR, then post a consensus report identifying which findings appeared independently in both bots. Use when explicitly asked to "run consensus on PR X", "cross-review PR X", "find consensus on PR X", or "what do both bots agree on for PR X". User-invoked only — do not auto-activate on generic PR review requests.
+---
+
+# pr-consensus
+
+When two distinct LLM reviewers independently surface the same critique on a PR, you have the highest-signal possible feedback. Higher than either bot's individual review, higher than a single human reviewer's, because reviewer-style differences cancel out at the agreement layer. This skill operationalizes that finding.
+
+## Core protocol
+
+Both CodeRabbit and Greptile must already be installed on the repository. The skill posts identical, fully open-ended prompts to each, waits for responses, and synthesizes the cross-bot agreement.
+
+Do not anchor the questions on specific topics — anchoring closes the aperture and produces echoed-back versions of the asker's priorities instead of the bots' actual concerns. The prompts in this skill are locked-down for that reason; do not modify them.
+
+## When NOT to use
+
+- The PR's bots have not yet posted initial reviews (the skill needs an initial review to follow up on).
+- The user has not asked for this explicitly. This skill costs ~15 minutes of bot back-and-forth and several API calls — auto-invocation on every PR review request is wrong.
+- A bot is not configured for the repository. The skill needs both bots to function.
+
+## Pre-flight
+
+```bash
+PR=<number>
+# Confirm both bots have posted initial reviews
+gh pr view $PR --json reviews --jq '[.reviews[] | .author.login] | unique'
+```
+
+Expected: both `coderabbitai` and `greptile-apps` (or a comment from each). If either is missing, wait for the initial review to land before starting.
+
+## The 3-round protocol
+
+Post the prompts in order. Each round: post both prompts in parallel (one per bot), then wait until both bots have replied before posting the next round.
+
+### Round 1 — open aperture
+
+Post identical to both bots. Replace `@theirname` with the actual handle.
+
+```text
+@coderabbitai — what do you actually think about this PR? Not feedback on anything specific I might ask about; your unfiltered take. What stands out? What's right, what's wrong, what's missing, what worries you, what would you change? Tell me what you think matters most — let your priorities, not mine, lead.
+```
+
+```text
+@greptileai — what do you actually think about this PR? Not feedback on anything specific I might ask about; your unfiltered take. What stands out? What's right, what's wrong, what's missing, what worries you, what would you change? Tell me what you think matters most — let your priorities, not mine, lead.
+```
+
+Wait for both replies before continuing.
+
+### Round 2 — generic prod
+
+```text
+@coderabbitai — anything else? Whatever else stands out, whatever I should have asked you about and didn't.
+```
+
+```text
+@greptileai — anything else? Whatever else stands out, whatever I should have asked you about and didn't.
+```
+
+Wait for both replies.
+
+### Round 3 — actionable specifics
+
+```text
+@coderabbitai — what actionable changes would you make based on the feedback you gave?
+```
+
+```text
+@greptileai — what actionable changes would you make based on the feedback you gave?
+```
+
+Wait for both replies.
+
+## Waiting pattern
+
+Use a single-shot `gh api` poll loop with `run_in_background: true`. Polling cadence: every 45 seconds. Sentinel: a timestamp captured immediately before the prompts are posted.
+
+```bash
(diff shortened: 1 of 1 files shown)


thedotmack:
@coderabbitai — what do you actually think about this PR? Not feedback on anything specific I might ask about; your unfiltered take. What stands out? What's right, what's wrong, what's missing, what worries you, what would you change? Tell me what you think matters most — let your priorities, not mine, lead.

thedotmack:
@greptileai — what do you actually think about this PR? Not feedback on anything specific I might ask about; your unfiltered take. What stands out? What's right, what's wrong, what's missing, what worries you, what would you change? Tell me what you think matters most — let your priorities, not mine, lead.

thedotmack:
@coderabbitai — anything else? Whatever else stands out, whatever I should have asked you about and didn't.

thedotmack:
@greptileai — anything else? Whatever else stands out, whatever I should have asked you about and didn't.

thedotmack:
@coderabbitai — what actionable changes would you make based on the feedback you gave?

thedotmack:
@greptileai — what actionable changes would you make based on the feedback you gave?

thedotmack:
## Cross-bot consensus report

**Method:** 3 rounds of open-ended interview against CodeRabbit and Greptile. Each bot answered three identical, non-leading prompts about this PR. The findings below are the consensus.

### Consensus findings (both bots, independently)

*These are the strongest-signal items. Two distinct LLM reviewers raised them without coordination and without being prompted on the topic. Treat these as the blocking concerns.*

1. **Polling loop has no timeout guard.** CodeRabbit: "the bash snippet runs `until [...]; do sleep 45; done` with no counter or deadline... if a bot goes quiet, the interview hangs indefinitely. This is the most operationally dangerous part of the spec." Greptile: "the `until` loop will spin indefinitely if either bot fails to respond. The failure modes section mentions a 10-minute per-round timeout, but the actual shell code has no corresponding implementation."

2. **Wrong Greptile handle: `@greptileai` should be `@greptile-apps`.** CodeRabbit: "the prompt text uses `@greptileai` but the poll filter checks `.author.login==\"greptile-apps\"`... if those are different, the poll loop never exits for Greptile's replies." Greptile: "the Greptile prompts use `@greptileai` instead of the bot's actual GitHub login `@greptile-apps` (so the bot is never triggered)... This PR is a live demonstration of bug #2."

3. **Pre-flight `--json reviews` vs poll loop `--json comments` API mismatch.** CodeRabbit: "the preflight and the polling check different GitHub API surfaces. The preflight uses `--json reviews`. The polling uses `--json comments`. Those are not the same thing on GitHub." Greptile: "`gh pr view --json reviews` returns only structured GitHub review submissions. Both CodeRabbit and Greptile typically post their initial analysis as top-level PR comments rather than formal reviews."

4. **"Independence" claim is overstated.** CodeRabbit: "both bots read the entire PR comment thread before responding. By Round 2, each bot has seen the other's Round 1 output... the signal strength claim at the top of the spec is somewhat overstated given the contamination." Greptile: "the central claim — 'independently' — doesn't hold, and that's the load-bearing word. Both bots read the same PR thread in full context... You're not running two independent observations; you're running one observation, then asking a second observer who already read the first one's notes."

5. **"Validated on PR #2409" is overconfident / conflicts with dogfood framing.** CodeRabbit: "'Validated on PR #2409' conflicts with the PR's own test plan. The output template footer says 'Method validated on PR #2409.' Your PR description says 'this PR is intended as the first dogfooded validation.' One of those is wrong." Greptile: "'Validated' is overconfident. One successful run on PR #2409 where three findings overlapped is a single data point. 'Tested once, results promising' is accurate. 'Validated' implies a rigor that one run doesn't support."

6. **Synthesis step is subjective / underspecified.** CodeRabbit: "the spec has the orchestrating agent — implicitly me, or whoever runs the skill — synthesize the final report. That agent is one of the two interviewees... that's not neutral synthesis." Greptile: "the synthesis step is entirely subjective, and the skill doesn't acknowledge it. 'Extract distinct claims' and 'identify semantic matches' are Claude's judgment calls at runtime. Two different synthesis runs from the same raw input could produce different consensus reports."

### Unique to CodeRabbit

*These reflect CodeRabbit's structural / methodology lens. They may still be correct, but they did not survive cross-bot independent verification.*

1. **Round 2's "anything else?" is a weak probe** — invites afterthoughts, not held-back substance; stronger phrasing would surface different material.
2. **No idempotency** — running the skill twice on the same PR posts two reports with no version marker, dedup, or "supersedes" header.
3. **This is a spec, not a skill** — `SKILL.md` describes intended behavior in prose; no execution mechanism, no agent binding. Test plan checkbox about auto-invocation tests something that doesn't exist yet.
4. **Structural asymmetry biases what can reach consensus** — Greptile is code-level, CodeRabbit ranges wider; intersection is dominated by code-level findings. Architectural concerns held by both bots may not survive synthesis matching.
5. **No method version marker** — prompts are locked but the spec has no version field; forks/modifications become untraceable.

### Unique to Greptile

*These reflect Greptile's surgical / patch-level lens. Same caveat.*

1. **`$INTERVIEW_START` is never assigned** — synthesis query expands it to empty string, matches every comment ever posted on the PR, not just the interview window.
2. **`$SENTINEL` is never re-captured between rounds** — Rounds 2 and 3 exit the poll loop instantly against Round 1's already-satisfied responses.
3. **Contradiction case unhandled** — output format covers agreement and divergence but not the case where both bots address the same topic and reach *opposite* conclusions (arguably higher-signal than consensus).
4. **Multi-run contamination unaddressed** — re-running on the same PR after fixes means the second run's bots read the first run's interview as context; not an independent measurement.
5. **`gh` CLI prerequisite undocumented** — skill assumes `gh` available and authenticated but never says so.
6. **"Prompts are locked down" overreaches** — one successful run isn't sufficient evidence that the phrasing is uniquely correct or that paraphrasing degrades signal.

### How to read this

The consensus section is the high-signal layer. Unique findings are not wrong, but require human judgment about whether they reflect a real issue or a reviewer-style artifact.

Notable: finding #2 (wrong Greptile handle) was caught by both bots, and Greptile explicitly observed that the dogfood run *masked* the bug rather than caught it — it responded because it processes PR events directly, not because the `@greptileai` mention triggered it. The skill as written would silently fail to trigger Greptile on every real run.

Bot did not respond to round N within timeout: n/a — all 6 round-responses landed within wall budget.

🤖 Generated by [pr-consensus](https://github.com/thedotmack/claude-mem/blob/main/plugin/skills/pr-consensus/SKILL.md). Method validated on PR #2409. This is the first dogfood validation — the PR adds the skill and the skill reviews the PR.

thedotmack:
Closing — dropping pr-consensus.

PR #2489, opened 2026-05-15
fix(mcp): launch mcp-search via node so Windows can spawn it

CaTeIM:
## Problem

On Windows the `mcp-search` MCP server fails to start. Claude Code spawns MCP servers **without a shell**, so the `.mcp.json` `command` is resolved directly against PATH. `sh` is not on PATH on Windows (Git Bash is not guaranteed to be present), so `command: "sh"` fails with ENOENT and the server never starts (`MCP error -32000: Connection closed`).

`.mcp.json` has no per-OS configuration, no `shell` field, and no environment-variable expansion — so a single `command` must work on every OS. The only binary guaranteed to be present on all platforms is `node` (it is the runtime the MCP server itself runs under).

## Change

Switches **only** the bundled `plugin/.mcp.json` launcher to `command: "node"`, which performs the same plugin-root resolution in JS and then `require()`s the resolved `mcp-server.cjs`.

**Intentionally minimal scope:** the hooks are left untouched. They run via `"shell": "bash"`, which Claude Code provides itself, so the existing `sh`/bash hook launchers keep working on Windows. Only the shell-less MCP spawn needed a node bootstrap. This does **not** revert the maintainer's `sh` approach — it only fixes the one case where `sh` cannot be spawned at all.

## Tests

The MCP startup assertions now validate the node launcher; the hook-command assertions are unchanged (still `sh`). Distribution test suite passing (24 tests).

### DIFF
--- a/plugin/.mcp.json
+++ b/plugin/.mcp.json
@@ -2,10 +2,10 @@
   "mcpServers": {
     "mcp-search": {
       "type": "stdio",
-      "command": "sh",
+      "command": "node",
       "args": [
-        "-c",
-        "_C=\"${CLAUDE_CONFIG_DIR:-$HOME/.claude}\"; _E=\"${CLAUDE_PLUGIN_ROOT:-${PLUGIN_ROOT:-}}\"; _P=$({ [ -n \"$_E\" ] && printf '%s\\n' \"$_E\"; printf '%s\\n' \"$PWD/plugin\" \"$PWD\"; ls -dt \"$HOME/.codex/plugins/cache/claude-mem-local/claude-mem\"/[0-9]*/ \"$HOME/.codex/plugins/cache/thedotmack/claude-mem\"/[0-9]*/ \"$_C/plugins/cache/thedotmack/claude-mem\"/[0-9]*/ 2>/dev/null; printf '%s\\n' \"$_C/plugins/marketplaces/thedotmack/plugin\"; } | while IFS= read -r _R; do [ -d \"$_R/plugin/scripts\" ] && _Q=\"$_R/plugin\" || _Q=\"$_R\"; [ -f \"$_Q/scripts/mcp-server.cjs\" ] && { printf '%s\\n' \"$_Q\"; break; }; done); [ -n \"$_P\" ] || { echo \"claude-mem: mcp server not found\" >&2; exit 1; }; exec node \"$_P/scripts/mcp-server.cjs\""
+        "-e",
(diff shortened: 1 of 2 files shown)


cateim:
Thanks for the careful review — addressed all three points in 94e85047:

**Issue 1 (argv slice):** Worker dispatch JS now filters out empty strings and the `[eval]` placeholder before forwarding to `bun-runner`:

```js
const a = process.argv.slice(1).filter(x => x !== "" && x !== "[eval]");
```

Verified on Node 24 the placeholder isn't present, but the filter is harmless there and makes the snippet robust on older Node versions / different runtimes where `argv[1]` is `''` or `'[eval]'`.

**Issue 3 (env var prefix):** Moved `CLAUDE_MEM_CODEX_HOOK=1` inside the JS for the Codex version-check hook:

```js
process.env.CLAUDE_MEM_CODEX_HOOK = "1";
require(f);
```

This removes the POSIX-only `VAR=value command` prefix so the snippet works under cmd.exe / PowerShell too (when the Codex CLI hook runner supports them).

**Issue 2 (shell: bash):** Kept `"shell": "bash"` because the Claude Code hook config takes a single shell-command string (no `command + args` separation like `.mcp.json` has). Documented the bash dependency explicitly in the top-level `description` of both `hooks.json` and `codex-hooks.json`:

> NOTE: hook commands run via `"shell": "bash"` — on Windows this requires Git Bash on PATH. The MCP launcher in .mcp.json is fully shell-agnostic (spawned without a shell), but hook execution by the Claude Code hook runner still depends on bash here.

The MCP launcher itself is fully shell-agnostic since it uses `command: "node"` + `args: [...]` directly. Truly cross-shell hooks would need either (a) a small launcher.cjs file referenced via `${CLAUDE_PLUGIN_ROOT}` (rejected to keep the patch minimal — no new files), or (b) hook-runner support for `args` arrays. Happy to follow up on either approach in a separate PR if you'd prefer.

cateim:
Good catch — fixed in 8a692884. Switched `;` to `&&` between the worker invocation and the continue-signal echo, so `{"continue":true,"suppressOutput":true}` is only emitted when Node exits 0.

```diff
- node -e '...' start; echo '{"continue":true,"suppressOutput":true}'
+ node -e '...' start && echo '{"continue":true,"suppressOutput":true}'
```

Matches the original shell-script semantics where `exit 1` from the inline plugin-not-found branch aborted before the echo ever ran. If the worker fails to start now, the hook exits non-zero and the launcher's stderr surfaces to the user instead of silently signalling a clean session.

mguttmann:
+1, independently reproduced and tested on **Windows 11** (Claude Code 2.x, Node 24.13.1, PowerShell + Git Bash).

**Repro** (claude-mem 13.2.0 freshly installed via marketplace):

```
/mcp → "Failed to reconnect to plugin:claude-mem:mcp-search: -32000"
```

Root cause matches this PR: the `sh -c "..."` launcher in `.mcp.json` can't be spawned because `sh` isn't on `PATH` under `cmd.exe` / PowerShell, and even when it resolves under Git Bash the `ls -dt "$_C/plugins/cache/..."/[0-9]*/` glob + subshell pipe blows up with `printf: write error: Permission denied`.

**Independent fix verification:** Built a standalone `scripts/mcp-server-launcher.cjs` with the same fallback search order this PR inlines into `node -e` (CLAUDE_PLUGIN_ROOT → PLUGIN_ROOT → `__dirname`/CWD → known Codex + Claude cache roots, newest version first → marketplace path), pointed `.mcp.json` at it via `command: "node"`, reloaded plugins. Result:

```
$ /mcp
mcp-search: connected
$ list_corpora() → []
```

Confirms the architectural approach in this PR works on Windows. The inline-`node -e` variant here is even cleaner than my external launcher (no new files, smaller diff).

The signal-kill nit Greptile flagged (`r.status || 0` swallowing signal kills as success) is worth fixing in this PR — `r.status ?? (r.signal ? 128 + (require("os").constants.signals[r.signal] ?? 0) : 0)` or simpler `r.status != null ? r.status : (r.signal ? 1 : 0)`. Trivial follow-up.

Would love to see this merged — it unblocks the entire Windows user base. Happy to test additional commits on Win11 if helpful. 🙏

thedotmack:
Closing as superseded: #2807 replaced the .mcp.json launcher with a cross-platform `node -e` launcher, which fixes Windows MCP spawning at the same spot. Thanks!

PR #2493, opened 2026-05-15
feat(worker): periodic sweep for orphaned pending_messages

singularix-maker:
## Problem

Sessions that fire `INIT_COMPLETE` then die never get an SDK processor. Their `summarize`/`observation` rows sit on `status='pending'` forever and inflate the WebUI "unprocessed" counter. Observed 2026-05-15: 5 stale rows from sessions that had been dead for ~20h.

The existing startup sweep only resets `processing → pending`; it never clears rows that were never claimed.

## Change

New `sweepStalePendingMessages()` (`src/services/infrastructure/StalePendingSweep.ts`):

- A session is swept when its **newest** `pending_message` predates a 6h threshold **and** it is not currently active. A mix of old + recent rows keeps the session safe.
- Its `pending_messages` are deleted; an `active` such session is marked `failed`.
- `pending_messages` has no `failed` status (CHECK allows only `pending`/`processing`), so orphaned rows are deleted rather than re-statused — matching the manual remediation precedent.

Wired into the worker as a 30-min `setInterval` in `initializeBackground()`, cleared on `shutdown()`. Active sessions are sourced from the new `SessionManager.getActiveSessionIds()`.

## Tests

- `tests/infrastructure/stale-pending-sweep.test.ts` — 5 cases: stale orphan swept, recent session untouched, active session skipped, mixed old/recent protected, non-active session status preserved.
- `tests/services/worker/session-manager-queue.test.ts` — `getActiveSessionIds()` coverage.

Full suite: 1800 pass / 56 fail — the 56 failures are the pre-existing `main` baseline, unchanged.

### DIFF
--- a/src/services/infrastructure/StalePendingSweep.ts
+++ b/src/services/infrastructure/StalePendingSweep.ts
@@ -1,0 +1,75 @@
+import type { Database } from 'bun:sqlite';
+import { logger } from '../../utils/logger.js';
+
+/** Sessions idle this long with no queue progress are considered orphaned. */
+const DEFAULT_THRESHOLD_MS = 6 * 60 * 60 * 1000;
+
+export interface StalePendingSweepResult {
+  staleSessions: number;
+  deletedMessages: number;
+  failedSessions: number;
+}
+
+export interface StalePendingSweepOptions {
+  thresholdMs?: number;
+  now?: number;
+}
+
+/**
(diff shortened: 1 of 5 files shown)


thedotmack:
Pending messages should not need to be swiped because we shouldn’t have pending messages if it’s done right

PR #2500, opened 2026-05-16
fix(worker): dispatch transcript subcommand in worker-service main (2450)

ousamabenyounes:
## Summary

Fix 2450 — `npx claude-mem transcript watch` is a silent no-op on
v13.2.0. The compiled npx CLI looks for
`<marketplace>/plugin/scripts/transcript-watcher.cjs` and falls back
to `spawnBunWorkerCommand('transcript', ['watch'])` when that file is
absent — and the bundle never produces it. The fallback then lands in
`worker-service.cjs transcript watch`, but `worker-service.ts` has no
`transcript` case in its command switch, so it falls through to the
`--daemon` default and the user's command silently does nothing.

## Fix

Add a `case 'transcript':` to the `main()` switch in
`src/services/worker-service.ts` that dispatches to the shared
implementation already living at `src/services/transcripts/cli.ts`.
`init`, `watch`, and `validate` now all work through the
worker-service entry point regardless of which script the npx CLI
delegates to.

The npx CLI's existing detection of a standalone
`transcript-watcher.cjs` is preserved, so a future split bundle still
works.

## Verification

- New test `tests/transcripts/cli-dispatch.test.ts` asserts the case
  block is present and that `runTranscriptCommand('init', ...)`
  writes a config file end-to-end.
- Full suites green: 70/70 in `hook-lifecycle`, `transcripts/`, and
  `codex-transcript-watcher-windows`.

## TDD verification

- RED (without the worker-service patch): the structural assertions
  in the new test fail because no `case 'transcript':` exists.
- GREEN (with the patch): the same test passes; the end-to-end
  `runTranscriptCommand` branch also produces a valid sample config.

## Files changed

| File | Change |
|------|--------|
| `src/services/worker-service.ts` | Add `case 'transcript':` dispatching to `runTranscriptCommand` |
| `tests/transcripts/cli-dispatch.test.ts` | New focused test |

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -778,12 +778,12 @@ export async function ensureWorkerStarted(port: number): Promise<WorkerStartResu
   return ensureWorkerStartedShared(port, __filename);
 }
 
-type ParsedWorkerCommand = {
+export type ParsedWorkerCommand = {
   command: string | undefined;
   args: string[];
 };
 
-function parseWorkerServiceCommand(argv: string[]): ParsedWorkerCommand {
+export function parseWorkerServiceCommand(argv: string[]): ParsedWorkerCommand {
   const [rawCommand, maybeSubCommand, ...rest] = argv;
 
   if (rawCommand === 'server') {
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -1103,6 +1103,18 @@ async function main() {
       break;
     }
 
+    case 'transcript': {
+      // npx-cli falls back to `worker-service.cjs transcript <sub>` when the
+      // standalone `transcript-watcher.cjs` is not present in the bundle
+      // (see thedotmack/claude-mem 2450). Dispatch to the shared
+      // implementation so `init`, `watch`, and `validate` all work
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
Addressed the Greptile nit in 87748d2a: the previous test scanned the `worker-service.ts` source text with regexes, which would silently survive quote/path refactors that actually break routing. The new test exercises the parser at runtime.

**Two runtime tests, no source scanning:**

1. `parseWorkerServiceCommand(["transcript", <sub>, ...])` — verifies for `watch`, `init`, and `validate` that the parser returns `{ command: "transcript", args: [<sub>, ...] }`. Proves the new dispatch case is reachable from real argv.
2. `runTranscriptCommand("init", ...)` against a temp dir — verifies the dispatch target writes a real config. Unchanged from before; it still proves the end-to-end behavior of the case body.

To support (1), `parseWorkerServiceCommand` and `ParsedWorkerCommand` are now exported from `worker-service.ts`. No runtime behavior change.

2/2 pass, 9 expect() calls.

PR #2501, opened 2026-05-16
fix: replace printf '%s\n' with echo in hook path-finding subshells (2439)

ousamabenyounes:
## Summary

Fix 2439 — Windows Git Bash (MSYS2) fails on `printf '%s\n'` in pipelined `$()` subshells with `printf: write error: Permission denied`, blocking every `UserPromptSubmit` (and all other hooks) for claude-mem users on Windows.

- Replace `printf '%s\n' "$var"` with `echo "$var"` in path-finding subshells across `plugin/hooks/hooks.json`, `plugin/hooks/codex-hooks.json`, `.mcp.json`, and `plugin/.mcp.json`
- Split the two-arg `printf '%s\n' "$PWD/plugin" "$PWD"` calls in the `.mcp.json` files into two `echo` statements to preserve newline-separated output
- Add a regression test (`tests/infrastructure/plugin-distribution.test.ts`) that asserts the path-finding subshell pattern never reintroduces `printf '%s\n'`

The fix is the workaround the reporter confirmed in the issue. `echo "$var"` and `printf '%s\n' "$var"` are equivalent for the path strings being printed here — bash's builtin `echo` does not interpret backslash escapes by default, and the values are MSYS-style paths without escape sequences. The `printf '%s' "$_HP"` calls in codex-hooks.json's PATH section are left untouched — they are not in a path-finding pipe and are not the reported failure.

## Verification

- Baseline `bun test tests/infrastructure/plugin-distribution.test.ts`: 25 pass, 0 fail
- After fix: 26 pass, 0 fail (1 new regression test)
- RED check: stashed the fix, re-ran the new test — 1 fail (regression test caught the broken pattern). Restored fix — back to GREEN
- Full `bun test` regression scan: pre-existing failures only (missing dev dependencies like `zod` and the MCP SDK package — environment issues unrelated to this change). No new failures introduced

Note: the bug is Windows-specific (MSYS2). The regression test enforces the byte-level structure on all platforms, which is the part of the fix that can be CI-checked from Linux.

## Files changed

| File | Change |
|------|--------|
| `plugin/hooks/hooks.json` | `printf '%s\n'` → `echo` in path-finding subshell (5 hooks) |
| `plugin/hooks/codex-hooks.json` | `printf '%s\n'` → `echo` in path-finding subshell (5 hooks) |
| `.mcp.json` | `printf '%s\n'` → `echo`; split two-arg printf into two echoes |
| `plugin/.mcp.json` | Same as `.mcp.json` (kept in sync) |
| `tests/infrastructure/plugin-distribution.test.ts` | New regression test for the path-finding subshell |


ousamabenyounes:
Closing — on reflection this isn't an improvement: swapping `printf '%s\n'` for `echo` in the path-finding subshells makes them *less* robust (`echo` mangles paths with leading `-` or backslashes), whereas `printf '%s\n'` is the portable idiom and matches the hook IO discipline direction on main. Withdrawing rather than pushing a weaker form.

PR #2505, opened 2026-05-17
fix(supervisor): preserve proxy env for worker daemon

### LINKED ISSUE #2464: Worker daemon doesn't inherit proxy environment variables
## Problem

  The worker daemon doesn't inherit `HTTP_PROXY`/`HTTPS_PROXY` environment variables, preventing API calls from going through the user's proxy.

  ## Environment

  - macOS / Linux
  - Users who require a proxy to reach external APIs (e.g., corporate networks, VPNs)

  ## Steps to reproduce

  1. Set proxy in shell:
     ```bash
     export HTTP_PROXY=http://127.0.0.1:8080
     export HTTPS_PROXY=http://127.0.0.1:8080
  2. Start Claude Code
  3. claude-mem fails to make API calls unless system-wide proxy is enabled

  Root cause

  In src/supervisor/env-sanitizer.ts, sanitizeEnv() strips all proxy environment variables when spawning the worker daemon:

  export const ENV_PROXY_VARS = new Set([
    'HTTP_PROXY', 'HTTPS_PROXY', ...
  ]);

  // In sanitizeEnv():
  if (ENV_PROXY_VARS.has(key)) continue;  // ← stripped!

  Proposed fix

  Move proxy variables to ENV_PRESERVE so the worker daemon can use them for API calls.

  I have a fix ready and can submit a PR if this approach is acceptable.

wyf027:
## Summary
- Preserve proxy-related environment variables when spawning the worker daemon.
- Keep the existing proxy env var list, but include it in `ENV_PRESERVE` so `sanitizeEnv()` forwards those values instead of stripping them.
- Update the env sanitizer regression test to assert uppercase/lowercase proxy vars and npm proxy config are preserved.

## Test Plan
- `bun test tests/supervisor/env-sanitizer.test.ts`

## Risk / Notes
- Scope is limited to the supervisor environment sanitizer.
- No proxy values or credentials are hard-coded; tests use placeholder proxy hosts only.

Fixes #2464

### DIFF
--- a/src/supervisor/env-sanitizer.ts
+++ b/src/supervisor/env-sanitizer.ts
@@ -20,6 +20,7 @@ export const ENV_PROXY_VARS = new Set([
 ]);
 
 export const ENV_PRESERVE = new Set([
+  ...ENV_PROXY_VARS,
   'CLAUDE_CODE_OAUTH_TOKEN',
   'CLAUDE_CODE_GIT_BASH_PATH',
   'CLAUDE_CODE_USE_BEDROCK',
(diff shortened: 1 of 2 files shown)


wyf027:
Closing this stale PR because it has been open for over a month with no review, comments, CI activity, or clear merge path. Thanks.

PR #2526, opened 2026-05-18
fix(hooks): silence Windows Git Bash EPIPE/EACCES on plugin-path resolver

ajs-uk-dev:
## Summary

On Windows MSYS / Git Bash, every claude-mem hook fire (Setup, SessionStart, UserPromptSubmit, PreToolUse:Read, PostToolUse, Stop, plus the MCP server launcher in `.mcp.json` / `plugin/.mcp.json`) intermittently leaks the following stderr to the host CLI:

```
```

Claude Code's hook runner surfaces this as `Stop hook error: Failed with non-blocking status code: ...` on every Stop, which is noisy enough to make claude-mem feel broken in real use, even though the underlying `node .../worker-service.cjs hook ... summarize` invocation succeeds and observations are captured fine.

## Root cause

Every plugin-path resolver in the hook bundle shares this shape (Stop hook shown):

```bash
_P=$({
  [ -n "$_E" ] && printf '%s\n' "$_E"
  ls -dt "$_C/plugins/cache/thedotmack/claude-mem"/[0-9]*/ 2>/dev/null
  printf '%s\n' "$_C/plugins/marketplaces/thedotmack/plugin"
} | while IFS= read -r _R; do
  ...
  [ -f "$_Q/scripts/bun-runner.js" ] && [ -f "$_Q/scripts/worker-service.cjs" ] && {
    printf '%s\n' "$_Q"
    break
  }
done)
```

The consumer breaks the moment it finds a valid path. That closes the read end of the pipe. Anything the producer block writes afterwards (subsequent `printf`s in the same `{ ... }` group) fails:

- Linux / macOS: failed write raises `SIGPIPE` → producer dies quietly. No stderr.
- Windows MSYS / Git Bash: the emulated pipe returns `EACCES` instead of `EPIPE`, and bash prints `printf: write error: Permission denied` to its own stderr. Claude Code's hook runner picks that up and reports a non-blocking failure.

## Fix

Add `2>/dev/null` to the producer block so its EPIPE/EACCES message is swallowed at the source:

```diff
-} | while IFS= read -r _R; do
+} 2>/dev/null | while IFS= read -r _R; do
```

Applied uniformly to:

- `plugin/hooks/hooks.json` (7 occurrences — Setup, SessionStart × 2, UserPromptSubmit, PostToolUse, PreToolUse, Stop)
- `plugin/hooks/codex-hooks.json` (7 occurrences — same shape, Codex variant)
- `plugin/.mcp.json` and `.mcp.json` (the MCP server launcher uses the same resolver)

This is the minimum-invasive change. It:

- Preserves the existing `{ producer; } | while IFS= read -r _R ...; break; done` structure, so existing assertions in `tests/infrastructure/plugin-distribution.test.ts` (lines 142, 157, 171) that lock in `while IFS= read -r _R` keep passing.
- Matches the existing style — every `ls -dt` in those producers already has its own `2>/dev/null`; this just extends the same defensiveness to the block-level printfs.
- Is purely a stderr redirect, so no behavior changes: same fallback order (cache → marketplace), same early-break semantics, same exit code.

## Tests

New file: `tests/infrastructure/hook-resolver-windows-pipe.test.ts` (33 assertions, all green on this branch).

1. **Static check** — for every shell snippet in the four affected JSON files, if it pipes a `{...}` producer into a `while IFS= read -r ... break ... done` consumer, the producer block MUST redirect stderr to `/dev/null`.

2. **Runtime check** — spawns each resolver snippet against a populated fake plugin tree under `bash -c`, with `SHELL=/bin/bash` and synthetic versions `13.0.0`/`13.1.0`/`13.2.0`. Asserts:
   - Stderr is empty (no leaked EPIPE/EACCES message).
   - Exit status is 0.
   - Resolved path contains `RESOLVED=...` with a non-empty value.

   On Windows Git Bash, the unfixed `.mcp.json` snippets reproduce the original `printf: write error: Permission denied` on iteration 0 of this test, so the test would fail without the fix.

3. **Behavior preservation** — separate test confirms the Stop hook resolver still picks the highest-version directory when multiple cached versions are present.

## Test plan

- [x] New regression test: `bun test tests/infrastructure/hook-resolver-windows-pipe.test.ts` → 33/33 pass.
- [x] Existing distribution test: `bun test tests/infrastructure/plugin-distribution.test.ts` → 25/25 pass (no regression in the `while IFS= read -r _R` assertions).
- [x] Manual sanity: ran the fixed Stop hook resolver and the fixed `.mcp.json` resolver against a populated fake plugin tree on Windows 11 / MSYS Git Bash 5.3.9 — no stderr, exit 0, correct path resolved.
- [x] Live verification: copied fixed JSON files into `~/.claude/plugins/cache/thedotmack/claude-mem/13.2.0/` and exercised real Stop hook in a Claude Code session — `"printf: write error: Permission denied"` no longer appears.

## Environment where the bug was observed

- Shell: MSYS Git Bash, GNU bash 5.3.9(1)-release (x86_64-pc-cygwin)
- claude-mem 13.2.0 installed via `/plugin marketplace add thedotmack/claude-mem` + `/plugin install claude-mem`


thedotmack:
Closing as superseded: #2865 landed the root-cause fix from #2725 (drop the pipe `break` so the producer subshell always drains — no broken-pipe write ever happens). This PR's `2>/dev/null` approach suppresses the symptom and would also mask future legitimate producer errors, the same class as the already-reverted a19b0e3c2. It also edits the generated hook JSONs without the canonical template, so the build's Rule A verification would reject it. Thanks for the investigation — the regression test idea was good.

Issue #2540, opened 2026-05-19
Tracking: server-beta runtime — fork-driven contributions (multi-PR series)


mguttmann:
This issue tracks a coordinated series of PRs from a downstream fork that runs claude-mem's server-beta runtime in production across Docker + 4 IDEs (Claude Code, Claude Desktop, OpenCode, Codex CLI).

The fork has spent ~30 PRs hardening the server-beta path end-to-end and is now upstreaming the work as a sequence of focused, reviewable PRs rather than one mega-PR. Each PR is independent, small, and has its own root-cause issue.

## PR series (planned)

| # | Scope | Issue | PR | Status |
|---|---|---|---|---|
| 1 | TypeScript cleanup (24 pre-existing errors) | #2538 | #2539 | open |
| 2 | server-beta foundation (Phases 1+2, ~800 LOC) | TBD | TBD | drafting |
| 3 | server-beta runtime wiring (Phase 3, ~1200 LOC) | TBD | TBD | drafting |
| 4 | Install UX fixes (4 bugs: --help, IDEs union, port, autostart) | TBD | TBD | planned |
| 5 | `claude-mem doctor` diagnostic CLI (12 checks) | TBD | TBD | planned |
| 6 | Validator harness (`scripts/validate-server-beta.sh`, 10 steps) | TBD | TBD | planned |
| 7 | Validator hardening (race conditions, idempotency) | TBD | TBD | planned |
| 8 | Viewer UI on server-beta (`/api/*` shims + `/stream` SSE) | TBD | TBD | planned |
| 9 | Subscription auth (host-bridge + bind-mount, no API key) | TBD | TBD | planned |

## Design principles followed across the series

1. **Each PR is independently mergeable.** Later PRs depend on earlier ones, but each can be reviewed and merged in isolation.
2. **No mega-commits.** Each PR has a clear, single-purpose diff.
3. **Issue-first.** Every PR has a root-cause issue describing what's broken and why.
4. **No fork-branding.** Identifiers, names, comments, and docs are kept neutral. No `mguttmann/` references leak into upstream.
5. **Verification gate.** Each PR ships with reproducible verification steps (typecheck output, validator runs, manual walkthrough).

## Fork context

The fork is at https://github.com/mguttmann/claude-mem-fork. Branches that ship here are cherry-picked off `upstream/main`, not merged from fork `main`, so the PRs stay clean of fork-specific history.

## Maintainer notes

[...]

PR #2542, opened 2026-05-19
feat(security): argon2id API key hashing with SHA-256 backward-compat

### LINKED ISSUE #2541: API keys stored as unsalted SHA-256 — replace with argon2id
## Problem

API keys are currently persisted as raw SHA-256 hex digests in postgres (`server_api_keys.key_hash`).

```ts
// src/server/middleware/postgres-auth.ts (current)
const hash = createHash('sha256').update(rawKey).digest('hex');
const row = await pool.query('SELECT * FROM server_api_keys WHERE key_hash = $1', [hash]);
```

SHA-256 is:

1. **Unsalted** — same raw key always produces the same hash, so rainbow tables work.
2. **Single-round, fast** — designed for hashing speed (Mbit/s on consumer GPUs). Brute-forcing a 32-byte API key is impractical, but a 16-char user-pasted key is not.
3. **No memory cost** — purpose-built GPU/ASIC attacks scale linearly.

If the database is ever exposed (offsite backup, replica snapshot, dev dump shared between team members), every key in it can be cracked offline at consumer-GPU speeds.

## Recommendation

Switch new API key issuance to argon2id with OWASP 2024 minimums (m=19 MiB, t=2, p=1, ~50ms per hash), keep existing SHA-256 keys working via a dual-verifier so no data migration is needed, and emit a deprecation warning when legacy keys verify so operators can rotate at their own pace.

## Verification path

1. Existing SHA-256 keys continue to verify after the change.
2. New keys issued by `claude-mem server api-key create` encode as `$argon2id$v=19$m=19456,t=2,p=1$...`.
3. Timing: legacy SHA-256 verify is ~1ms; argon2id verify is ~50ms. Both failure paths return at the same minimum latency to prevent timing-side-channel discrimination between "wrong key", "no row", "revoked", "expired", "missing scope".

## Fix

- `src/server/middleware/postgres-auth.ts` — dual verifier
- `src/server/middleware/argon2-shim.ts` (new) — thin argon2 wrapper
- `package.json` — adds `argon2` dependency

No schema migration, no breaking change. SHA-256 fast path can be removed in a future major.

mguttmann:
Fixes #2541. Part of tracking issue #2540.

API keys are currently stored as raw SHA-256 hex digests — unsalted, single-round, GPU-cracking-friendly. This PR switches new key issuance to argon2id (OWASP 2024 minimums) while keeping legacy SHA-256 keys working via a dual-verifier, so no data migration is required.

## Diff

```
 package.json                           |   1 +
 src/server/middleware/argon2-shim.ts   |  90 +++++++++++++
 src/server/middleware/postgres-auth.ts | 223 +++++++++++++++++++++++++++++----
 3 files changed, 291 insertions(+), 23 deletions(-)
```

## Strategy

1. **Compute SHA-256 of the raw key.** Used for both the legacy fast path and as a stable cache key.
2. **Fast path** — `WHERE key_hash = $sha256` matches legacy rows in O(1). On hit, log a deprecation warning so operators can rotate before SHA-256 is removed in a future major.
3. **Fallback** — `WHERE key_hash LIKE '$argon2%'` scans argon2 rows, then `argon2.verify` each. O(N) where N = argon2 keys for this tenant; fine up to a few hundred. A future iteration should add a lookup-prefix indexed column.
4. **Timing equalization** — every failure path (no key, no row, revoked, expired, missing scope) returns after the same minimum interval so an attacker can't distinguish error modes by latency.

## Verification

  - `npm run typecheck` — clean
  - `npm install` — argon2 native module installs via prebuilt binaries on linux/macOS/windows; no toolchain required for end users
  - Legacy SHA-256 keys continue to verify (legacy fast path)
  - New keys issued via `claude-mem server api-key create` encode as `$argon2id$v=19$m=19456,t=2,p=1$...$...`

## Notes

  - Server-side change only. No client-side changes.
  - No schema migration; `key_hash` stays `text`.
  - Migration is implicit: rotating an existing key issues an argon2 hash and removes the old SHA-256 row. No explicit migration command.
  - SHA-256 fast path can be removed in a future major (`v14`?) when all deployments have rotated.

## Argon2 parameter justification

OWASP password storage cheat sheet (2024) recommends:

- argon2id (resistant to side-channel + GPU attacks)
- m=19 MiB minimum memory cost
- t=2 minimum iterations
- p=1 parallelism

Tuned to ~50ms per hash on a 2024-class server CPU. Operators with stronger hardware can lift these via env vars without a code change. See `argon2-shim.ts` for the param object.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "@clack/prompts": "^1.3.0",
     "@modelcontextprotocol/sdk": "^1.29.0",
     "ansi-to-html": "^0.7.2",
+    "argon2": "^0.44.0",
     "better-auth": "^1.6.9",
     "bullmq": "^5.76.6",
     "cors": "^2.8.6",
--- a/src/server/middleware/argon2-shim.ts
+++ b/src/server/middleware/argon2-shim.ts
@@ -60,6 +60,7 @@ export async function hashApiKeyForStorage(rawKey: string, opts: Argon2HashOptio
       algorithm: 'argon2id',
       memoryCost: merged.memoryCost,
       timeCost: merged.timeCost,
+      parallelism: merged.parallelism,
     });
   }
   const a2 = loadNativeArgon2();
--- a/src/server/middleware/argon2-shim.ts
+++ b/src/server/middleware/argon2-shim.ts
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// BUG 33 fix — runtime-adaptive argon2id shim.
+//
+// The npm `argon2` package compiles a native node-gyp addon. In the Docker
+// server-beta image we install with `--ignore-scripts` (BUG 32) to skip the
+// fragile tree-sitter native compile, which also leaves argon2 without a
+// native build. At runtime, importing argon2 then throws:
+//
+//   error: No native build was found for platform=linux arch=arm64
+//   runtime=node abi=137 ...
+//
+// Bun (which is what server-beta-service.cjs runs under) ships argon2id in
+// `Bun.password.hash()` and `Bun.password.verify()` — first-class APIs with
+// PHC-string output ($argon2id$v=19$m=...$...$...) fully interoperable with
+// the npm package's format.
+//
+// This shim picks Bun.password when available, otherwise lazily requires
+// the npm package. The native dep is no longer mandatory.
+//
+// Output format is identical either way ($argon2id PHC strings), so the
+// dual-verifier in postgres-auth.ts continues to detect existing hashes
+// without migration.
+
+export interface Argon2HashOptions {
+  memoryCost?: number;
+  timeCost?: number;
+  parallelism?: number;
+}
+
+const DEFAULT_OPTS: Required<Argon2HashOptions> = {
+  memoryCost: 19456, // 19 MiB — OWASP 2024 minimum
+  timeCost: 2,
+  parallelism: 1,
+};
+
+// Runtime detection — Bun ships argon2id natively.
+// `Bun` global is only defined under the Bun runtime.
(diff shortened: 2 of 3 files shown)


mguttmann:
Thanks @greptile-apps[bot] — all four issues addressed in eb13e5d8:

- **P1 argon2 candidate scan** → added `ARGON2_SCAN_LIMIT = 100` to bound tail latency (~5s worst case), with a stderr warning when the cap is hit. The scan stays global (tenant_id isn't known until after match), and the documentation now reflects that honestly. The indexed key_prefix optimization is noted inline as the recommended next step for high-scale deployments — happy to do that as a separate PR if there's appetite, but it requires a schema migration so I kept it out of this one.
- **P2 Bun parallelism** → fixed, `merged.parallelism` now passed through.
- **P2 duplicate comment block** → removed.
- **P2 unbounded Map → Set** → swapped `Map<id, timestamp>` for `Set<id>`; log-once-per-process-per-key is the practical semantic anyway.

mguttmann:
Greptile (P1, security) re-surfaced the argon2 DoS-amplifier concern with a deeper angle: the comment misleadingly framed the cap as 'per tenant' but the SQL is global. Fixed in 67001406 with two-layer defence:

1. **RAW_API_KEY_SHAPE pre-filter** — `/^[A-Za-z0-9_.~-]{32,128}$/` rejects malformed bearer tokens before any hash work. An attacker can no longer burn argon2 CPU by sending plausibly-formatted garbage that doesn't match the shape.
2. **ARGON2_SCAN_LIMIT 100 → 20** — worst-case unauthenticated probe spends 20 × 50ms = 1s of CPU instead of 5s. Production deployments needing more than 20 active argon2 keys per process should migrate to the indexed key_prefix optimization (documented inline).

Plus comment-truthing: the file-level doc now explicitly states the scan is global, with the cap shared across the whole deployment.

PR #2544, opened 2026-05-19
feat(install): server-beta runtime setup, rollback, and multi-IDE injection

### LINKED ISSUE #2543: claude-mem install has no end-to-end setup for the server-beta runtime
## Problem

`claude-mem install` today supports the worker runtime end-to-end (single-machine SQLite) but has no equivalent setup path for the server-beta runtime (Docker stack with postgres + redis + the server-beta HTTP service). Operators who want server-beta have to:

1. Manually `docker-compose up` from a checkout
2. Manually generate an API key via direct SQL
3. Manually edit `~/.claude-mem/settings.json` to point at the server URL
4. Manually edit each IDE's MCP config (Claude Code, Claude Desktop, OpenCode, Codex CLI) to point at the server

That's 4+ manual steps per machine, no rollback if something goes wrong, and the IDE injection is the worst part because the file paths and config schemas all differ.

## What should happen

```
claude-mem install --runtime server-beta
```

should:

1. Bring up the Docker stack (`docker compose up -d`) and wait for healthchecks.
2. Generate an API key and persist it to `~/.claude-mem/keys.json`.
3. Write `~/.claude-mem/.env` and `settings.json` with the right server URL and runtime selector.
4. Detect installed IDEs and inject MCP config additively (preserves any other MCP servers the user has).
5. Take timestamped backups of every file it touches so `claude-mem uninstall --runtime server-beta` can restore.

Likewise `claude-mem uninstall --runtime server-beta` should reverse all of the above cleanly.

## Why it matters

The server-beta runtime is the path forward for multi-machine and team setups. Without an install command, every operator has to read the source and reproduce the manual steps — which is exactly the kind of friction that kills adoption.

## Fix

PR coming separately. Adds:

- `src/services/install/server-beta-setup.ts` — orchestrator
- `src/services/install/server-beta-rollback.ts` — uninstall path
- `src/services/install/ide-mcp-injection.ts` — multi-IDE config injector with per-IDE rollback
- `src/services/install/claude-host-bridge.ts` — optional subscription auth bridge (separate follow-up PR uses it)

No existing files modified. All additive. ~2,100 LOC.

mguttmann:
Fixes #2543. Part of tracking issue #2540.

Adds the install-time orchestration needed to provision the server-beta runtime end-to-end on a fresh machine. `claude-mem install --runtime server-beta` now does what currently takes 4+ manual steps per machine, with full rollback on uninstall.

## Diff

```
 src/services/install/claude-host-bridge.ts   | 455 +++++++++
 src/services/install/ide-mcp-injection.ts    | 628 ++++++++++
 src/services/install/server-beta-rollback.ts | 195 +++
 src/services/install/server-beta-setup.ts    | 845 +++++++++++
 4 files changed, 2123 insertions(+)
```

No existing files modified. All additive. Wiring into `src/npx-cli/commands/install.ts` happens in a follow-up PR (kept separate so the wiring is reviewable in isolation).

### `server-beta-setup.ts` (845 LOC)

End-to-end orchestrator. Does, in order:
1. `docker compose up -d` from the project root
2. Polls health endpoints on postgres / redis / server-beta until ready (timeout: 60s)
3. Generates a 32-byte API key, hashes it via argon2id (PR #2542), inserts into `server_api_keys`, persists the raw key + hash + scope to `~/.claude-mem/keys.json`
4. Writes `~/.claude-mem/.env` (server URL, runtime selector, optional bridge URL)
5. Writes `~/.claude-mem/settings.json` (runtime=server-beta, server_url, project_id assigned)
6. Calls `injectAllIdes()` from `ide-mcp-injection.ts`

Idempotent — re-running detects existing state (docker stack already up, API key already present, IDE configs already pointing at us) and updates only what's missing.

### `server-beta-rollback.ts` (195 LOC)

Tear-down path. Safe to call from any point in a failed install (each step is no-op if the corresponding install step didn't complete). Order:
1. Call `rollbackAllIdes()` to restore IDE configs from their timestamped backups
2. `docker compose down -v` (drops volumes; users who want data preservation should snapshot the postgres volume first — there's a note in the CLI prompt)
3. Removes `~/.claude-mem/.env` and `~/.claude-mem/keys.json`
4. Reverts `~/.claude-mem/settings.json` to runtime=worker (the default)

### `ide-mcp-injection.ts` (628 LOC)

Multi-IDE MCP config injector. Per IDE: detects presence, locates config file, parses, takes timestamped backup, additively merges the `claude-mem` MCP entry, writes back. Supported IDEs:

| IDE | Config file | Format |
|---|---|---|
| Claude Code | `~/.claude/mcp_servers.json` | JSON |
| Claude Desktop | `~/Library/Application Support/Claude/claude_desktop_config.json` (macOS) or platform equivalent | JSON |
| OpenCode | `~/.config/opencode/opencode.json` + auxiliary plugin file | JSON + JS |
| Codex CLI | `~/.codex/config.toml` | TOML |

Each IDE has a paired `rollback<Ide>()` function called from `server-beta-rollback.ts`.

### `claude-host-bridge.ts` (455 LOC)

Optional subscription auth bridge (launchd unit on macOS, systemd unit on linux). Inert unless explicitly invoked by a follow-up PR adding the subscription auth provider. Included here because `server-beta-setup.ts` imports it conditionally — the registration step is a no-op when the user opts for API-key auth.

## Verification

- `npm run typecheck` — all 4 new files clean
- Fresh-install walkthrough on macOS: Docker stack 4/4 healthy, 4 IDEs injected, env/settings/keys.json written
- Uninstall walkthrough: Docker stack down, all 4 IDEs reverted to their pre-install backup, env/settings/keys.json removed
- Idempotency: re-running `install` is a no-op when everything is already in place

## Why this is one PR not four

The four files form a single logical surface — install/uninstall/IDE-injection/host-bridge — and they import each other. Splitting them creates artificial cross-PR dependencies for a feature that's only useful when all four ship together. The diff is large in LOC but narrow in scope: every file is single-purpose and the surfaces don't overlap.

### DIFF
--- a/src/services/install/claude-host-bridge.ts
+++ b/src/services/install/claude-host-bridge.ts
@@ -175,6 +175,65 @@ export function ensureClaudeHostBridge(): HostBridgeResult {
   }
 }
 
+// Tear down the host-bridge so uninstall doesn't leave a running service
+// pointing at a stale token. Idempotent — safe to call when nothing is
+// installed. Returns ok:true even when nothing was removed so the caller's
+// rollback flow doesn't fail on first-time uninstalls.
+export function stopClaudeHostBridge(): { ok: boolean; message: string; removedFiles: string[] } {
+  const removedFiles: string[] = [];
+  const home = homedir();
+
+  try {
+    if (process.platform === 'darwin') {
+      const plistPath = join(home, 'Library', 'LaunchAgents', `${LAUNCHD_LABEL}.plist`);
+      if (existsSync(plistPath)) {
+        spawnSync('launchctl', ['unload', plistPath], { stdio: 'ignore' });
+        try {
+          // unlinkSync via require avoided — use writeFileSync to /dev/null is
+          // not portable; just rely on fs.unlinkSync via dynamic require if
+          // unavailable as a top-level import here.
+          require('fs').unlinkSync(plistPath);
+          removedFiles.push(plistPath);
+        } catch { /* already gone */ }
+      }
+    } else if (process.platform === 'linux') {
+      const unitDir = join(home, '.config', 'systemd', 'user');
+      const unitPath = join(unitDir, SYSTEMD_UNIT_NAME);
+      if (existsSync(unitPath)) {
+        spawnSync('systemctl', ['--user', 'stop', SYSTEMD_UNIT_NAME], { stdio: 'ignore' });
+        spawnSync('systemctl', ['--user', 'disable', SYSTEMD_UNIT_NAME], { stdio: 'ignore' });
+        try {
+          require('fs').unlinkSync(unitPath);
+          removedFiles.push(unitPath);
+        } catch { /* already gone */ }
+        spawnSync('systemctl', ['--user', 'daemon-reload'], { stdio: 'ignore' });
+      }
+    }
+    // Token + port files live in paths.dataDir(); remove them too so a
+    // future install starts from a clean state.
+    const tokenPath = join(paths.dataDir(), 'host-bridge-token');
+    const portPath = join(paths.dataDir(), 'host-bridge-port');
+    for (const f of [tokenPath, portPath]) {
+      if (existsSync(f)) {
+        try { require('fs').unlinkSync(f); removedFiles.push(f); } catch { /* */ }
+      }
+    }
+    return {
+      ok: true,
+      message: removedFiles.length
+        ? `host-bridge stopped; removed ${removedFiles.length} file(s)`
+        : 'host-bridge was not installed',
+      removedFiles,
+    };
+  } catch (err) {
+    return {
+      ok: false,
+      message: err instanceof Error ? err.message : String(err),
+      removedFiles,
+    };
+  }
+}
+
 function claudeCliReachable(): boolean {
   return resolveClaudeCliPath() !== null;
 }
--- a/src/services/install/claude-host-bridge.ts
+++ b/src/services/install/claude-host-bridge.ts
@@ -30,7 +30,7 @@
 //     CLAUDE_MEM_CLAUDE_BRIDGE_TOKEN and signs every request. Prevents other
 //     processes on the host from accidentally using the bridge.
 
-import { execSync, spawnSync } from 'child_process';
+import { spawnSync } from 'child_process';
 import {
   chmodSync,
   existsSync,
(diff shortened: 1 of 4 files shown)


mguttmann:
Thanks @greptile-apps[bot] — all seven issues addressed in f114f35e:

**P1 SECURITY — bridge token in world-readable plist + ps**
- launchd plist + systemd unit now reference `--token-file <path>` instead of `--token <value>`. The token isn't on any cmdline; the file itself is already chmod 600.
- Plist + systemd unit file modes tightened from 0o644 to 0o600.
- The embedded bridge script accepts `--token-file` (preferred) and warns when the legacy `--token` is used ("visible in ps").

**P1 — dry-run violation when clearing subscription credentials**
- The dryRun guard now wraps the writeFileSync call in the non-subscription branch. `--dry-run` is a true no-op again.

**P2 — dead `_logRef` closure** → removed.

**P2 — `require('fs')` / `require('path').dirname()` in ESM file** → swapped for top-level imports. `copyFileSync`, `unlinkSync` added to the fs destructure; `dirname` added to path.

**P2 — dead `nextHeaderRegex.lastIndex` assignment** → removed (regex was not /g or /y so it was a no-op anyway).

**P2 — `serializeEnvFile` drops empty strings + doesn't escape embedded quotes** → fixed. Empty strings now preserved (round-trip via parseEnvFile is lossless), embedded double-quotes are backslash-escaped.

mguttmann:
Thanks @greptile-apps[bot] — second-pass P1s addressed in 70da13f8:

- **P1 host-bridge not cleaned up on uninstall** → added `stopClaudeHostBridge()` (launchctl unload + plist delete on macOS, systemctl stop+disable + unit delete on linux, token/port files removed). Hooked into `rollbackServerBeta` as a `stop-host-bridge` step.
- **P1 inline env serializer missing quote escaping** → `serializeEnvFile` is now exported from server-beta-setup.ts and used by rollback's --purge-data path. The fixed embedded-quote handling applies on the uninstall side too.
- **P1 IDE injection failures silently returned ok:true** → `setupServerBeta` now derives the final `ok` from `steps.every(s => s.status !== 'failed')`. install.ts no longer commits the runtime selector when Claude Desktop write fails.
- **P1 stale CLAUDE_CREDS_FILE on validation failure** → `ensureClaudeSubscriptionCreds()` now clears `CLAUDE_CREDS_FILE` + `CLAUDE_MEM_CLAUDE_BRIDGE_URL` from .env before returning `failed`. A subsequent compose-up no longer bind-mounts the invalid creds file.

mguttmann:
Third-round fix in 10563340: `CLAUDE_HOST_BRIDGE_TOKEN_FILE` is now stripped from `.env` in all four auth-transition spots (was only stripped in 1 of 4). Prevents the next `docker compose up` from attempting to bind-mount a path `stopClaudeHostBridge` already deleted from disk.

mguttmann:
**Production-reported regression fixed in commit 726e2a9e.**

User reported that `claude-mem install --opt-in-ide opencode` was writing `~/.config/opencode/opencode.json` with Claude Desktop's MCP schema (`mcpServers` key + flat `{command, args}` shape). OpenCode rejected the resulting file on every load with `Unrecognized key: mcpServers`. Operators were accumulating `.pre-claude-mem-1.bak` through `.pre-claude-mem-7.bak` files because every re-install triggered another wrong-shape write + new backup.

**Root cause:** `injectMcpEntryIntoJsonConfig` was reused for OpenCode with `serversKey: 'mcpServers'`. OpenCode uses a different schema (see https://opencode.ai/docs/mcp-servers/):

| | Claude Desktop | OpenCode |
|---|---|---|
| Top-level key | `mcpServers` | `mcp` |
| Entry shape | `{command, args}` | `{type: 'local', command: [...], enabled}` |
| Env field | `env` | `environment` |

**Fix:** Three new functions in `ide-mcp-injection.ts`:
- `buildOpenCodeMcpEntry()` — correct entry shape
- `injectOpenCodeMcpEntry()` — writes under `mcp` key, auto-inserts `$schema` field, idempotent (no new backup if entry is already current — fixes the .bak-accumulation), self-heals by stripping the broken `mcpServers` key from pre-fix configs
- `rollbackOpenCodeMcpEntry()` — symmetric uninstall, cleans both `mcp` and `mcpServers` keys

`injectAllIdes()` + `rollbackAllIdes()` route OpenCode through the new functions; Claude Desktop continues to use the unchanged `injectMcpEntryIntoJsonConfig` / `rollbackJsonConfig` pair.

**Verified locally:** fresh-install on a machine with the broken config → stale `mcpServers` stripped, `mcp` added with right shape, OpenCode loads without complaint. Re-running install with no changes → `already-current`, no new `.pre-claude-mem-*.bak`.

mguttmann:
Thanks @greptile-apps[bot] — both P1s addressed in commit 90ec56cd:

- **`ensurePostgresEnv` mkdirSync/chmodSync before dryRun guard** \u2192 both calls now gated behind `if (!dryRun)`. Dry-run is a true no-op again.
- **`waitForBridge` requires `curl`** \u2192 confirmed, curl isn't in Debian/Ubuntu minimal/Alpine/many WSL images. Replaced execSync('curl ...') with a Node `http.get()` probe spawned as a worker subprocess (same pattern as the existing setTimeout poll). No external binary dependency now. Removed the unused execSync import.

Issue #2545, opened 2026-05-19
claude-mem install — 4 UX bugs + missing server-beta wiring


mguttmann:
## Problem

Four pre-existing UX bugs in `claude-mem install`, plus the CLI has no wiring for the server-beta runtime path even though the rest of the codebase supports it.

### Bug 1 — `install --help` runs a real install

```
npx claude-mem install --help
# expected: prints help text, exits
# actual: prints help, then continues into the install side-effects
```

The help flag is parsed by commander but `runInstallCommand` doesn't bail out, so the body runs anyway.

### Bug 2 — Post-install summary only mentions Claude Code

Even when the user opts into Claude Desktop, OpenCode, or Codex CLI, the post-install summary's "configured IDEs" list shows only Claude Code. The other IDEs are configured correctly, the user just can't tell from the output.

### Bug 3 — Wrong port in success message

```
✓ claude-mem installed
  Server: http://127.0.0.1:37703
```

The 37703 is hardcoded. Worker runtime uses a per-user derived port (37700 + (uid % 100)); server-beta uses 37877. The hardcoded value is wrong for both.

### Bug 4 — "Autostart skipped" is misleading on server-beta

When `--runtime server-beta`, the installer just spun up the Docker stack — but the post-install message still says "autostart skipped" as if no service is running. Autostart is a worker-runtime concept (launchd/systemd unit); on server-beta the Docker daemon manages lifecycle.

### Missing wiring

`claude-mem install` has no `--runtime server-beta` option. To install the server-beta runtime today you have to run `docker compose up` manually and then hand-edit `~/.claude-mem/settings.json`.

## Fix

PR coming separately. Touches only `src/npx-cli/commands/install.ts`. Stacked on the server-beta install foundation PR (#2544).

thedotmack:
Consolidating into #2379 (plan-04). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

PR #2549, opened 2026-05-19
feat(cli): claude-mem doctor — 12-check diagnostic CLI

### LINKED ISSUE #2548: No diagnostic CLI — operators have to manually probe every layer when install breaks
## Problem

When `claude-mem` doesn't work after install, there's no single command to diagnose what's broken. Operators end up manually:

- Checking if the binary is on PATH
- `cat ~/.claude-mem/settings.json` to see which runtime is configured
- `docker ps` or `claude-mem worker status` to see if anything is running
- `curl http://127.0.0.1:37877/v1/health` (assuming they know the port)
- Hand-comparing MCP config files in each IDE against `settings.json`

Each layer can fail independently and the symptoms are all "memory doesn't work". A 12-check diagnostic CLI would cut "claude-mem install seems broken" tickets from a multi-step investigation to a single command.

## Proposal

```
claude-mem doctor
```

Runs 12 read-only checks across both runtime paths, prints pass/fail/warn with actionable hints, exits 0 on all-pass / 1 on any fail (so it's CI-friendly).

Checks:
1. Binary on PATH + version
2. `~/.claude-mem/` writable
3. `settings.json` parses + has valid `runtime` field
4. Required binaries present (Bun for worker, Docker for server-beta)
5. Server `/v1/health` reachable
6. `/v1/memories/batch` reachable (server-beta)
7. API key file present + parseable (server-beta)
8. API key actually authenticates (server-beta)
9. Postgres reachable (server-beta)
10. Redis reachable (server-beta)
11. Worker auto-start unit registered (worker, optional)
12. IDE MCP configs point at the same URL/port `settings.json` does

## Fix

PR coming separately. Adds:
- `src/npx-cli/commands/doctor.ts` (548 LOC)
- Wires `doctor` into `src/npx-cli/index.ts` (6 LOC)

Read-only. No state changes.

mguttmann:
Fixes #2548. Part of tracking issue #2540.

Adds `claude-mem doctor`, a read-only diagnostic that walks 12 checks across both runtime paths (worker + server-beta) and reports pass/fail/warn with actionable hints. Designed to be the first command operators run when something doesn't work.

## Diff

```
 src/npx-cli/commands/doctor.ts | 548 +++++++++++++++++++++++++++++++++++++++++
 src/npx-cli/index.ts           |   6 +
 2 files changed, 554 insertions(+)
```

No existing files modified except a 6-line wiring change to `index.ts` (adds the `case 'doctor'` dispatch + a help-text line).

## Sample output

```
$ claude-mem doctor

claude-mem doctor
─────────────────

  ✓ binary on PATH (/usr/local/bin/claude-mem v13.x.x)
  ✓ ~/.claude-mem/ writable
  ✓ settings.json parses (runtime=server-beta)
  ✓ Docker daemon running
  ✓ Server reachable (http://127.0.0.1:37877/v1/health → 200)
  ✓ /v1/memories/batch reachable (204 expected on empty body)
  ✓ API key file present (~/.claude-mem/keys.json)
  ✓ API key authenticates (/v1/health 200 with Authorization)
  ✓ Postgres reachable (via /v1/health → ok)
  ✓ Redis reachable (via /v1/health → ok)
  ✓ Worker auto-start: N/A (server-beta runtime)
  ✓ IDE MCP configs aligned with settings.json

12/12 passed
```

Failure modes print a fix hint:

```
  ✗ Server reachable (http://127.0.0.1:37877/v1/health → ECONNREFUSED)
    Hint: docker compose up -d  (from the install directory)
```

## Properties

  - **Read-only.** No state changes. Safe to run during a live install.
  - **Runtime-aware.** Checks 6-10 self-skip on worker runtime; check 11 self-skips on server-beta.
  - **Composable.** Each check is `(ctx) => Promise<{name, status, detail, hint?}>` — tests can run individual checks without the full battery.
  - **CI-friendly.** Exits 0 on all-pass, 1 on any fail. No interactive prompts.

## Verification

  - `npm run typecheck` — clean
  - Manual: clean install → 12/12 pass
  - Manual: docker stopped → check 5 fails with ECONNREFUSED + fix hint
  - Manual: wrong API key → check 8 fails with 401 + rotate hint
  - Manual: stale IDE MCP port → check 12 fails with diff + re-run-install hint

## Stack note

This PR is independent (depends only on upstream `main`). It does not require #2538, #2541, #2543, or #2545 to land first.

### DIFF
--- a/src/npx-cli/commands/doctor.ts
+++ b/src/npx-cli/commands/doctor.ts
@@ -29,7 +29,7 @@
 
 import pc from 'picocolors';
 import { execSync } from 'child_process';
-import { existsSync, readFileSync, statSync } from 'fs';
+import { existsSync, readFileSync, readdirSync, statSync } from 'fs';
 import { homedir } from 'os';
 import { join } from 'path';
 import { SettingsDefaultsManager } from '../../shared/SettingsDefaultsManager.js';
@@ -63,6 +63,13 @@ function printResult(r: CheckResult): void {
   if (r.status !== 'pass' && r.hint) {
     console.log(`      ${pc.dim('hint:')} ${r.hint}`);
   }
+}
+
+// Shell-quote a path so spaces and metacharacters don't break command parsing.
+// Wraps in single quotes and escapes any embedded single quote via the
+// standard '"'"' trick.
+function shellQuote(value: string): string {
+  return `'${value.replace(/'/g, `'"'"'`)}'`;
 }
 
 function safeExec(cmd: string, timeoutMs = 5000): { ok: boolean; out: string; err?: string } {
@@ -106,7 +113,7 @@ function checkPluginNodeModules(): CheckResult {
   const nm = join(MARKETPLACE_ROOT, 'plugin', 'node_modules');
   if (existsSync(nm)) {
     try {
-      const entries = require('fs').readdirSync(nm);
+      const entries = readdirSync(nm);
       if (entries.length > 0) {
         return { name: 'plugin/node_modules populated ', status: 'pass', detail: `${entries.length} packages` };
       }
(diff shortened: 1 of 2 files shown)


mguttmann:
Thanks @greptile-apps[bot] — all four addressed in e0095beb:

- **P1 require in ESM** → `readdirSync` added to top-level fs import; `require('fs').readdirSync()` removed.
- **P1 unquoted paths** → added `shellQuote()` helper and applied to both `MARKETPLACE_ROOT` and `envFile` interpolations in checkComposeStack.
- **P2 missing break** → explicit `break` after `process.exit()` in the doctor case.
- **P2 redundant batch failure** → guard added so the `/v1/memories/batch` probe is skipped when `/healthz` failed; mirrors the existing checkDocker → checkComposeStack pattern.

mguttmann:
Thanks @greptile-apps[bot] — `shellQuote` does already handle single-quotes within the path. The implementation uses the standard POSIX idiom:

```ts
function shellQuote(value: string): string {
  return `'${value.replace(/'/g, \`'"'"'\`)}'`;
}
```

Each embedded single-quote becomes `'"'"'` (close single-quote, literal single-quote in double-quotes, reopen single-quote), so `/Users/O'Brien/` becomes `'/Users/O'"'"'Brien/'`. Verified manually that posix sh parses this back to the original path.

PR #2555, opened 2026-05-19
feat(server-beta): subscription auth via OAuth resolver + host-bridge provider

### LINKED ISSUE #2554: server-beta has no subscription auth path — API-key-only is prohibitively expensive
## Problem

server-beta currently requires `ANTHROPIC_API_KEY` to back its generation worker. API-key usage is metered separately from Claude Code subscriptions, billed per-token, and at the volumes claude-mem generates (one observation per tool call) it becomes prohibitively expensive — easily $50-200/month per active developer.

Operators on a Claude Code subscription are already paying for the underlying model usage. They reasonably expect to use that same subscription to back claude-mem's worker. Today they can't.

Additionally, several related issues make running server-beta in Docker against any auth method fail:

1. The hardcoded default model `claude-3-5-sonnet-latest` was removed from the Anthropic API — every generate() call returns 404 unless the operator knows to set `CLAUDE_MEM_SERVER_MODEL`.
2. Compose URLs pointing at `127.0.0.1` / `localhost` fail with ECONNREFUSED inside the container because there's nothing listening on loopback. The error message is generic and doesn't hint at the service-name vs loopback distinction.

### Subscription auth — two paths

**Path A — bind-mount the host credentials file.** Mount `~/.claude/.credentials.json` into the container; the worker reads it fresh per generate() so token refreshes and account switches propagate without restarts.

**Path B — host-bridge sidecar.** For macOS Keychain installs (no credentials file), run a small sidecar on the host that wraps `claude --print`. The container proxies generation requests over HTTP. 100% live, works on any auth method the host CLI supports.

### Provider factory precedence

1. `CLAUDE_MEM_CLAUDE_BRIDGE_URL` set + bridge token present → host-bridge
2. Otherwise: credentials file mounted → live OAuth resolver
3. Otherwise: `ANTHROPIC_API_KEY` set → api-key fallback
4. Otherwise: log a clear "no provider configured" with all three env hints

### Adjacent fixes

- Update `DEFAULT_MODEL` to a current Sonnet id
- Guard `CLAUDE_MEM_SERVER_DATABASE_URL` / `CLAUDE_MEM_REDIS_URL` against loopback hosts when running inside a container

## Fix

- new `src/server/generation/providers/ClaudeHostBridgeProvider.ts` (178 LOC)
- `src/server/generation/providers/ClaudeObservationProvider.ts` (oauthResolver + DEFAULT_MODEL)
- `src/server/runtime/create-server-beta-service.ts` (provider factory + loopback guard)

The sidecar installer (launchd/systemd unit) was added separately in #2544; this PR only wires the runtime side.

mguttmann:
Fixes #2554. Part of tracking issue #2540.

Adds two paths for using a Claude Code subscription (instead of the Anthropic API key) to back the server-beta generation worker. API-key usage is metered separately and prohibitively expensive at scale; subscription-backed usage is what operators want when they're already paying for Claude Code.

## Diff

```
 src/server/generation/providers/ClaudeHostBridgeProvider.ts | 178 +++++++++++++
 src/server/generation/providers/ClaudeObservationProvider.ts |  68 +++++-
 src/server/runtime/create-server-beta-service.ts             | 240 ++++++++++++-
 3 files changed, 468 insertions(+), 18 deletions(-)
```

## Path A — live OAuth from bind-mounted credentials

Host's `~/.claude/.credentials.json` is bind-mounted into the container at install time. `ClaudeObservationProvider` gains an `oauthResolver: () => string | undefined` callback that's called fresh per `generate()`, so:

- Token refreshes by Claude CLI (~hourly) → next generate() picks up the new token automatically
- Account switches on the host → next generate() picks up the new account automatically  
- File deleted → next generate() falls back to api-key (logged warning)

No restart required for any of these events.

## Path B — host-bridge HTTP proxy

For installs where the credentials file isn't accessible (macOS Keychain-only installs, sandboxed users), a separate sidecar on the host runs `claude --print` per request. The container proxies generation requests via HTTP. 100% live, works on any auth method the host's Claude CLI supports.

New file: `ClaudeHostBridgeProvider.ts` (178 LOC). Same `ServerGenerationProvider` interface as `ClaudeObservationProvider` — drop-in.

## Provider factory precedence

Updated `buildServerGenerationProviderFromEnv()`:

1. `CLAUDE_MEM_CLAUDE_BRIDGE_URL` + bridge token present → `ClaudeHostBridgeProvider`
2. `~/.claude/.credentials.json` mounted → `ClaudeObservationProvider` w/ oauthResolver
3. `ANTHROPIC_API_KEY` set → `ClaudeObservationProvider` w/ apiKey
4. None → null, server logs a clear error with all three env hints

Each fallback logs the chosen auth method so operators can see what's running.

## Adjacent fixes bundled here

- **DEFAULT_MODEL bump** — `claude-3-5-sonnet-latest` was removed from the Anthropic API and returns 404. Bumped to `claude-sonnet-4-5`. Overridable via `CLAUDE_MEM_SERVER_MODEL`.
- **Loopback guard** — `detectLoopbackHost()` in `validateServerBetaEnv()` rejects `CLAUDE_MEM_SERVER_DATABASE_URL` / `CLAUDE_MEM_REDIS_URL` pointing at `127.0.0.1` / `localhost` / `::1` when running inside a container. Compose service names are the right value; loopback URLs fail with ECONNREFUSED inside the container.

These bundle here because they're all on the "make subscription auth actually work in Docker" path. Each prevents a misleading downstream error: model 404, DB ECONNREFUSED, etc.

## Verification

- `npm run typecheck` — all 3 files clean
- Manual: credentials file mounted → first generate() reads token + succeeds; refresh on host → next generate() picks up new token without restart
- Manual: `CLAUDE_MEM_CLAUDE_BRIDGE_URL` set + bridge running → generate() proxies through bridge; response identical to direct API
- Manual: `ANTHROPIC_API_KEY` only → falls back to api-key path, logged as such
- Manual: none of the three configured → server-beta starts but generation worker reports 'no provider configured' with env hints

## Notes

- Server-side only. No client/IDE/frontend changes.
- The host-bridge sidecar installer (launchd on macOS, systemd on linux) was added in #2544; this PR adds the runtime side.
- The api-key path is unchanged. Operators who want to keep `ANTHROPIC_API_KEY` need do nothing.

## Stack note

Independent of other PRs in the series (depends only on upstream `main`). Synergistic with #2544 (which adds the sidecar installer) but mergeable in either order.

### DIFF
--- a/src/server/generation/providers/ClaudeHostBridgeProvider.ts
+++ b/src/server/generation/providers/ClaudeHostBridgeProvider.ts
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// ClaudeHostBridgeProvider.
+//
+// Proxies observation generation through the user's locally-installed Claude
+// CLI by POSTing prompts to a small HTTP bridge running on the host. The
+// bridge (~/.claude-mem/claude-host-bridge.cjs, started by launchd on macOS or
+// systemd --user on Linux) shells out to `claude` for each request, so:
+//
+//   - account switches propagate INSTANTLY (next request uses the new account)
+//   - OAuth token refresh is handled by Claude CLI transparently
+//   - the model selected via Claude CLI (4.5/4.6/whatever the user installed)
+//     is what gets used — no need to hardcode CLAUDE_MEM_SERVER_MODEL
+//
+// The container reaches the bridge via host.docker.internal (Docker Desktop)
+// or a host-network alias on Linux. CLAUDE_MEM_CLAUDE_BRIDGE_URL is set by
+// the install script when the bridge is active.
+//
+// Authentication: the bridge requires a Bearer token (random 32-byte hex,
+// stored on host at ~/.claude-mem/host-bridge-token, mounted read-only into
+// the container at /run/secrets/claude-host-bridge-token). Prevents other
+// processes on the host or in the container from accidentally hitting the
+// bridge.
+
+import { logger } from '../../../utils/logger.js';
+import { ServerClassifiedProviderError } from './shared/error-classification.js';
+import { buildServerGenerationPrompt } from './shared/prompt-builder.js';
+import type {
+  ServerGenerationContext,
+  ServerGenerationProvider,
+  ServerGenerationResult,
+} from './shared/types.js';
+
+export interface ClaudeHostBridgeProviderOptions {
+  bridgeUrl: string;
+  bridgeToken: string;
+  model?: string;
+  fetchImpl?: typeof fetch;
+  timeoutMs?: number;
+}
+
+interface BridgeGenerateResponse {
+  text?: string;
+  error?: string;
+  message?: string;
+  exitCode?: number;
+  stderr?: string;
+}
+
+export class ClaudeHostBridgeProvider implements ServerGenerationProvider {
+  readonly providerLabel = 'claude' as const;
+  private readonly bridgeUrl: string;
+  private readonly bridgeToken: string;
+  private readonly model: string | undefined;
+  private readonly fetchImpl: typeof fetch;
+  private readonly timeoutMs: number;
+
+  constructor(options: ClaudeHostBridgeProviderOptions) {
+    if (!options.bridgeUrl) {
+      throw new ServerClassifiedProviderError('Host-bridge URL missing', {
+        kind: 'auth_invalid',
+        cause: new Error('bridgeUrl is required'),
+      });
+    }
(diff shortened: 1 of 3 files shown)


mguttmann:

- **P2 dead IPv6 bracket check** `host === '[::1]'` → removed (URL.hostname strips brackets per WHATWG spec; `'::1'` covers it).
- **P2 redundant clearTimeout in catch** → removed; finally handles all exit paths.

PR #2546, opened 2026-05-19
fix(install): wire server-beta runtime into installer + UX cleanup

mguttmann:
Fixes #2545 (issue creation pending — will link once available). Part of tracking issue #2540.

**Stacked on #2544.** When #2544 lands, this PR will be rebased to main; the install.ts diff is self-contained.

Wires the server-beta setup orchestrator (added in #2544) into the installer CLI and fixes four pre-existing UX bugs.

## Diff

```
 src/npx-cli/commands/install.ts | 151 ++++++++++++++++++++++++++--------------
 1 file changed, 98 insertions(+), 53 deletions(-)
```

## Fixes

  1. `install --help` no longer runs the install body.
  2. Post-install summary now lists all configured IDEs (Claude Code + Claude Desktop + OpenCode + Codex CLI), not just Claude Code.
  3. Server URL in the success message is runtime-aware (37877 for server-beta, derived per-user port for worker).
  4. "Autostart skipped" replaced with runtime-appropriate messaging: worker path keeps the autostart message, server-beta path reports docker compose status.

## New feature

  - `claude-mem install --runtime server-beta` — calls `setupServerBeta()` from #2544 and rolls back cleanly on failure.

## Verification

  - `npm run typecheck` — install.ts clean
  - Manual: `--help` exits without side-effects
  - Manual: 4 IDEs listed when all opted in
  - Manual: correct port in success message for both runtimes
  - Manual: server-beta path shows "Docker stack: healthy", worker path shows "Autostart: enabled"

## Why stacked

The wiring imports `setupServerBeta` from #2544. Without #2544 landing first, this PR would fail typecheck. Standard stacked-PR workflow: review and merge #2544 first, then this rebases to main.

PR #2557, opened 2026-05-19
feat(plugin): hook auto-registry + OpenCode auto-capture + esbuild fixes

### LINKED ISSUE #2556: Plugin hooks + OpenCode auto-capture + ESM bundling — 3 packaging issues block fresh install
### Issue 1: Claude Code hooks not auto-loaded after install

Claude Code reads hooks from `plugin/.claude-plugin/plugin.json`, not from `plugin/hooks/hooks.json`. Today the inline hook block in `plugin.json` is missing some of the 6 lifecycle events (Setup, SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, Stop), so a fresh install requires manually merging `hooks.json` into `plugin.json`. Most operators don't realize this and end up with non-functional hooks.

### Issue 2: OpenCode users never get observations captured

OpenCode has its own plugin system (listens to `session.created`, `tool.execute.after`, `message.updated`, `session.idle` events). claude-mem currently has no OpenCode-side plugin file, so operators using OpenCode get zero observations captured. The fix is a 165-LOC ESM file that subscribes to OpenCode events and POSTs to the same `/v1/events` endpoint Claude Code uses.

### Issue 3: esbuild produces broken bundles on Node 22+ + current esbuild

Three esbuild config bugs:

- `external: ['zod']` is a leftover from when plugin/node_modules shipped. The plugin doesn't ship node_modules now, so externalizing zod produces runtime `Cannot find module 'zod'`.
- Node 22+ rejects esbuild's dynamic `require()` shim when `format=esm`. Transitive `require('events')` fails at runtime.
- Pure ESM has no `__dirname` / `__filename`. Some transitive deps (find-up, etc.) crash trying to resolve script-relative paths.

All three need a banner shim injecting `createRequire` + `fileURLToPath` derivations.

## Why bundle as one fix

The three are interleaved: the hook auto-registry depends on the build script producing a working bundle (ESM-fix issue), and the OpenCode plugin uses the same opencode-plugin esbuild entry that benefits from the createRequire/__dirname shim. Splitting creates artificial cross-PR dependencies.

## Fix

- `plugin/.claude-plugin/plugin.json` — inline hooks block for 6 events
- `scripts/build-hooks.js` — ESM banner shim + zod bundling
- `plugin/opencode/claude-mem-plugin.js` (new, 164 LOC) — OpenCode auto-capture

No backend changes. All packaging/plumbing.

mguttmann:
Fixes #2556. Part of tracking issue #2540.

Three packaging-layer fixes that together make claude-mem auto-capture observations end-to-end across Claude Code AND OpenCode without manual hook registration. Each fix is small, but they're interleaved so they ship together.

## Diff

```
 plugin/.claude-plugin/plugin.json    |  86 +++++++++++++++++-
 plugin/opencode/claude-mem-plugin.js | 164 +++++++++++++++++++++++++++++++
 scripts/build-hooks.js               |  52 ++++++++++-
 3 files changed, 297 insertions(+), 5 deletions(-)
```

No backend code changes. All packaging / build-script / plugin-config.

## Change 1 — Inline hooks block in plugin.json (Claude Code auto-load)

Claude Code reads hooks from `plugin.json`, not `hooks.json`. Today the inline block in `plugin.json` is incomplete (some of the 6 lifecycle events are missing), so a fresh install requires manually merging from `hooks.json`. Most operators don't realize this and end up with non-functional hooks.

Now: `plugin.json` ships with the full inline hooks block for all 6 events (Setup, SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, Stop). `hooks.json` remains the source-of-truth; the build script keeps both in sync.

## Change 2 — OpenCode auto-capture plugin (new file)

OpenCode has its own plugin system (events: `session.created`, `tool.execute.after`, `message.updated`, `session.idle`). Without an OpenCode-side plugin, claude-mem never sees OpenCode activity.

New file: `plugin/opencode/claude-mem-plugin.js` (164 LOC, vanilla ESM). Subscribes to OpenCode events and POSTs them to the running server's `/v1/events` endpoint with the same shape Claude Code uses.

Opt-in via `--opt-in-ide opencode` at install time; no effect for users not using OpenCode.

## Change 3 — esbuild fixes for ESM bundling

Three esbuild config bugs that were blocking `node scripts/build-hooks.js` on Node 22+:

  - **zod bundling.** `external: ['zod']` is leftover from when plugin/node_modules shipped. Now that we don't ship node_modules, externalizing zod produces a runtime `Cannot find module 'zod'`. Removed from externals; zod bundles inline.
  - **createRequire shim.** Node 22+ rejects esbuild's dynamic `require()` shim when `format=esm`. Banner now injects `createRequire(import.meta.url)` so transitive `require('events')` etc. resolve.
  - **__dirname / __filename.** Pure ESM has neither; some transitive deps (find-up etc.) crash on script-relative path resolution. Banner now derives both via `fileURLToPath(import.meta.url)`.

## Verification

  - `node scripts/build-hooks.js` — builds cleanly; all 7 bundles produced (worker, server-beta, mcp-server, context-generator, npx-cli, openclaw-plugin, opencode-plugin)
  - `npm run typecheck` — clean
  - Manual (Claude Code): fresh install, no manual hook editing → hooks fire on every tool call, observations appear in `/v1/memories`
  - Manual (OpenCode): fresh install with `--opt-in-ide opencode` → plugin file dropped into `~/.config/opencode/plugins/`, OpenCode events flow into `/v1/events`, observations appear identical to the Claude Code path

## Why one PR

The three changes are interleaved:
- The hook auto-registry depends on the build script producing working bundles (ESM-fix issue)
- The OpenCode plugin uses the same `opencode-plugin` esbuild entry that benefits from the createRequire/__dirname banner shim

Splitting them creates artificial cross-PR dependencies for a feature only useful when all three land together.

## Stack note

Independent of other PRs in the series (depends only on upstream `main`). Synergistic with #2544 (which adds the OpenCode plugin file dropper) and #2546 (which adds the `--opt-in-ide opencode` flag), but mergeable in any order — `plugin/opencode/claude-mem-plugin.js` is dropped manually by `#2544` if those land first, or by users running `ln -s` until then.

### DIFF
--- a/plugin/.claude-plugin/plugin.json
+++ b/plugin/.claude-plugin/plugin.json
@@ -20,5 +20,89 @@
     "typescript",
     "nodejs"
   ],
-  "homepage": "https://github.com/thedotmack/claude-mem#readme"
+  "homepage": "https://github.com/thedotmack/claude-mem#readme",
+  "hooks": {
+    "Setup": [
+      {
+        "matcher": "*",
+        "hooks": [
+          {
+            "type": "command",
+            "shell": "bash",
+            "command": "export PATH=\"$HOME/.nvm/versions/node/v$(ls \\\"$HOME/.nvm/versions/node\\\" 2>/dev/null | sed 's/^v//' | sort -t. -k1,1n -k2,2n -k3,3n | tail -1)/bin:$HOME/.local/bin:/usr/local/bin:/opt/homebrew/bin:$PATH\"; _C=\"${CLAUDE_CONFIG_DIR:-$HOME/.claude}\"; _E=\"${CLAUDE_PLUGIN_ROOT:-${PLUGIN_ROOT:-}}\"; _P=$({ [ -n \"$_E\" ] && printf '%s\\n' \"$_E\"; ls -dt \"$_C/plugins/cache/thedotmack/claude-mem\"/[0-9]*/ 2>/dev/null; printf '%s\\n' \"$_C/plugins/marketplaces/thedotmack/plugin\"; } | while IFS= read -r _R; do _R=\"${_R%/}\"; [ -d \"$_R/plugin/scripts\" ] && _Q=\"$_R/plugin\" || _Q=\"$_R\"; [ -f \"$_Q/scripts/version-check.js\" ] && { printf '%s\\n' \"$_Q\"; break; }; done); [ -n \"$_P\" ] || { echo \"claude-mem: version-check.js not found\" >&2; exit 1; }; command -v cygpath >/dev/null 2>&1 && { _W=$(cygpath -w \"$_P\" 2>/dev/null); [ -n \"$_W\" ] && _P=\"$_W\"; }; node \"$_P/scripts/version-check.js\"",
+            "timeout": 300
+          }
+        ]
+      }
+    ],
+    "SessionStart": [
+      {
+        "matcher": "startup|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
+            "shell": "bash",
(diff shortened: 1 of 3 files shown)


mguttmann:
Thanks @greptile-apps[bot] — both addressed:

- **P1 `tool_input` always undefined in PostToolUse** → confirmed: `tool.execute.after` only gets `{ title, output, metadata }`. Fixed via a 256-entry bounded LRU keyed by `sessionID + tool`, populated in `tool.execute.before`, drained in `tool.execute.after`.
- **P2 `session.idle` mapped 1:1 to `Stop`** → confirmed: idle fires per assistant turn, not per session. Added a 30s debounce so the server only sees `Stop` after a true idle window, preventing summarize-storms on multi-turn chats.

mguttmann:
Thanks @greptile-apps[bot] — all four follow-up nits addressed:

- **Duplicate `!ENDPOINT` guard** → removed (second one was unreachable).
- **FIFO vs LRU** → comment updated to say 'capacity-bounded (FIFO eviction)'; FIFO is correct for the pairing semantics.
- **Redundant `.catch(() => {})`** → removed; postEvent already swallows errors.
- **`output` parameter naming in `before`** → renamed locally to `args` with a clarifying comment about the OpenCode plugin API using the same signature across before/after callbacks.

mguttmann:
Thanks @greptile-apps[bot] — fixed. session_id now reads from `event?.properties?.sessionID ?? event?.id ?? msg.sessionID`, matching the pattern used by every other handler in the plugin. UserPromptSubmit was the only outlier that was reading from the message body instead of the event envelope.

PR #2559, opened 2026-05-19
fix(docker): server-beta stack hardening + Redis config robustness

### LINKED ISSUE #2558: server-beta Docker stack: no restart policy + brittle Redis env + no creds-file mount
### 1. No restart policy on any compose service
A `docker compose up -d` brings the stack up but a host reboot, daemon crash, or any container exit leaves everything stopped. There's no `restart:` directive anywhere in `docker-compose.yml`. Operators end up writing systemd units around `docker compose up` to compensate.

### 2. Redis env-var brittleness
`getRedisQueueConfig()` reads only `CLAUDE_MEM_REDIS_URL`. The conventional `REDIS_URL` name (used by every other Docker stack on the host) is ignored, and there's no fallback to the compose service name. Users get cryptic ECONNREFUSED on first up unless they know the magic env var.

### 3. No `~/.claude/.credentials.json` mount
Subscription-backed auth (per #2554 / #2555) needs the host's Claude CLI credentials file mounted into the worker container. The compose file has no mount directive for this, so even after merging the provider PR, operators can't actually use subscription auth without hand-editing the compose file.

## Fix

PR coming separately. Adds:
- `restart: unless-stopped` to all 4 services
- Tighter health probes (5s start period vs 20s)
- Conditional `CLAUDE_CREDS_FILE` bind-mount (defaults to `/dev/null`)
- `CLAUDE_MEM_CLAUDE_BRIDGE_URL` env passthrough
- `REDIS_URL` fallback in `getRedisQueueConfig()`
- IPv6 family parsing in the Redis URL
- Comment block showing the override pattern for exposing postgres on the host

mguttmann:
Fixes #2558. Part of tracking issue #2540.

Three Docker-related fixes that make the server-beta runtime survive real-world container restarts, environment-variable conventions, and subscription-auth setup.

## Diff

```
 docker-compose.yml                               | 94 ++++++++++++++++++++++++
 src/server/queue/BullMqObservationQueueEngine.ts | 13 +++-
 src/server/queue/redis-config.ts                 | 20 ++++-
 3 files changed, 124 insertions(+), 3 deletions(-)
```

## Three fixes

1. **`restart: unless-stopped`** on all 4 compose services (postgres, valkey, server, worker). Survives host reboots; explicit `docker compose stop` still wins.
2. **`CLAUDE_CREDS_FILE` bind-mount + `CLAUDE_MEM_CLAUDE_BRIDGE_URL` env passthrough.** Enables the subscription auth path landed in #2555. Defaults to `/dev/null` when unset so the worker's file-existence check is single-valued.
3. **`REDIS_URL` fallback + IPv6 family parsing in `redis-config.ts`.** Honors the conventional env-var name plus `?family=6` URL hint. Falls back to `redis://valkey:6379` (compose service name) when both env vars are unset.

Plus a small `BullMqObservationQueueEngine.getQueueEvents()` guard so test-time graceful-shutdown doesn't throw a noisy stack trace.

## Verification

- `npm run typecheck` — clean
- Clean `docker compose up -d --build`: all 4 services healthy in <15s
- Kill server container: `unless-stopped` restarts within 3s
- With `CLAUDE_CREDS_FILE=$HOME/.claude/.credentials.json` in `.env`: mount appears in worker; with unset: resolves to `/dev/null` cleanly

## Stack note

Synergistic with #2555 (subscription auth providers) but mergeable independently — the providers fall through to `ANTHROPIC_API_KEY` if no mount is configured.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -126,6 +126,11 @@ services:
       # worker only needs enough for the BullMQ generation loop + outbox.
       # Budget rule: 1 server × 10 + N workers × 5 < Postgres max_connections.
       CLAUDE_MEM_POSTGRES_POOL_MAX: ${CLAUDE_MEM_POSTGRES_POOL_MAX_SERVER:-10}
+      # mirror the worker's CLAUDE_MEM_CLAUDE_CREDS_FILE / token-file so the
+      # doctor CLI and /v1/health responses can report 'auth=subscription'
+      # without having to exec into the worker container.
+      CLAUDE_MEM_CLAUDE_CREDS_FILE: /run/secrets/claude-credentials.json
+      CLAUDE_MEM_CLAUDE_BRIDGE_TOKEN_FILE: /run/secrets/claude-host-bridge-token
     ports:
       - "37877:37877"
     volumes:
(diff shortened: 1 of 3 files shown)


mguttmann:

- **Server missing `CLAUDE_MEM_CLAUDE_CREDS_FILE` env** → added to the server's environment block, mirroring the worker. Plus added `CLAUDE_MEM_CLAUDE_BRIDGE_TOKEN_FILE` for the same reason.
- **Healthcheck won't trigger restart** → correct, this is Docker behavior (`restart: unless-stopped` fires only on process exit). Added an inline caveat to the comment block explaining that and pointing at the two standard production patterns (autoheal sidecar / systemd-k8s orchestrator). Not bundling autoheal into the default compose file because the choice depends on the operator's stack.

mguttmann:
Thanks @greptile-apps[bot] — LGTM noted. On the `BRIDGE_TOKEN_FILE` env-on-the-server question: keeping it for symmetry with the worker block since both containers can also surface auth-method via diagnostics that scan env vars directly (the server's planned doctor-equivalent endpoint will read both vars too once it lands). As you said: not a blocker either way. Will keep as-is.

PR #2563, opened 2026-05-19
refactor(auth): rename api-key-service.ts → api-key-service-sqlite.ts

### LINKED ISSUE #2562: api-key-service.ts filename doesn't reveal which auth backend it implements
## Problem

\`src/server/auth/api-key-service.ts\` exclusively implements the bun:sqlite-backed API key service used by the worker runtime. The Postgres-backed equivalent lives in \`src/server/middleware/postgres-auth.ts\`. There are two parallel implementations of the same conceptual surface, but only one of them carries the backend name in its filename.

New contributors keep grepping for \`api-key-service\` thinking they've found "the auth module", missing the Postgres path entirely (or vice-versa). The two paths are intentionally separate (worker = SQLite, server-beta = Postgres) and the filenames should reflect that.

## Fix

Rename:
\`src/server/auth/api-key-service.ts\` → \`src/server/auth/api-key-service-sqlite.ts\`

Plus 2 trivial import-path updates in the two importers. No behavior change. Pure refactor.

## Why now

Several upcoming PRs touch both auth paths and the file-search experience is materially better when the rename has landed. Cheap to do; impossible to do retroactively without breaking import paths in every PR that touches auth.

mguttmann:
Fixes #2562. Part of tracking issue #2540.

Pure rename. `src/server/auth/api-key-service.ts` is exclusively the bun:sqlite implementation. The Postgres equivalent already lives in a clearly-named `postgres-auth.ts`. Filenames should reveal the backend boundary.

## Diff

```
 .../auth/{api-key-service.ts => api-key-service-sqlite.ts} | 11 ++++++++++-
 src/server/middleware/auth.ts                              |  2 +-
 src/services/worker-service.ts                             |  2 +-
 3 files changed, 12 insertions(+), 3 deletions(-)
```

## Changes

1. Git-tracked rename (R085 — 85% similarity retained).
2. Two import-path updates in the two callers.
3. Small set of CLI-friendly re-exports surfaced (createServerApiKey, listServerApiKeys, revokeServerApiKey) so the upcoming `claude-mem server api-key` CLI doesn't have to peer into the SQLite module.

## Verification

- `npm run typecheck` — clean
- `grep -r 'api-key-service\.js' src/` — zero remaining hits on the old name; only the new path is referenced

## Why this is a separate small PR

Could easily be folded into any of the other PRs, but separating it gives reviewers a clean rename-diff to confirm at a glance. Future PRs that touch auth won't carry rename noise.

### DIFF
--- a/src/server/auth/api-key-service-sqlite.ts
+++ b/src/server/auth/api-key-service-sqlite.ts
@@ -5,10 +5,14 @@
 // src/server/middleware/postgres-auth.ts + src/storage/postgres/auth.ts
 // for tenant-scoped key verification against Postgres.
 //
-// Do NOT import this module from src/server/* runtime code — the server-beta
-// runtime is firewalled from bun:sqlite. Consumers are limited to the worker
-// HTTP path (src/services/worker-service.ts) and the auth middleware that
-// the worker mounts (src/server/middleware/auth.ts).
+// Allowed consumers (worker runtime only):
+//   - src/services/worker-service.ts          (worker HTTP boot path)
+//   - src/server/middleware/auth.ts           (worker-mounted auth middleware)
+// Forbidden consumers (server-beta runtime):
+//   - any other file under src/server/* outside the two allowed paths above.
+//     server-beta is firewalled from bun:sqlite; use
+//     src/server/middleware/postgres-auth.ts + src/storage/postgres/auth.ts
+//     for tenant-scoped key verification against Postgres.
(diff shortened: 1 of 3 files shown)


mguttmann:
Thanks @greptile-apps[bot] — addressed. Rewrote the import-guard comment with an explicit allowlist (two specific paths under src/server/* and src/services/) instead of a blanket 'Do NOT import from src/server/*' that immediately listed an exception. Boundary is now unambiguous.

PR #2565, opened 2026-05-19
feat(runtime): hooks + ModeManager runtime-selector for server-beta

### LINKED ISSUE #2564: Hooks have no runtime selector — installed plugin can't switch worker↔server-beta without reinstall
## Problem

The Claude Code lifecycle hooks (PostToolUse, SessionStart, etc.) always talk to the local bun:sqlite worker. There's no way to point them at a remote server-beta runtime via configuration — only via reinstall.

Operators who run server-beta in a different topology (Docker on a shared VM, team-shared server, etc.) need to either:

1. Hand-edit the hook scripts to swap the client implementation, or  
2. Reinstall the plugin with different env vars and hope they got the wiring right

Both are fragile and don't survive plugin updates.

Same issue affects `ModeManager` — the mode resolution is purely file-based today, with no awareness of a server-side mode source.

## Proposal

  - `CLAUDE_MEM_RUNTIME=worker|server-beta` env var (or equivalent `settings.json#runtime` field) selects which client the hook bootstrap returns.
  - server-beta path: thin HTTP client posting to `/v1/events` with bearer auth, exponential-backoff retry on 5xx, single-flight queue to coalesce concurrent hooks, graceful degradation so a server outage doesn't break the user's session.
  - ModeManager honors the same selector: server-beta path pulls from `/v1/mode` with a local cache; worker path is unchanged.

## Fix

PR coming separately. Stacks on #2542 (argon2 hash) for the auth helper. Touches:
- `src/services/hooks/server-beta-bootstrap.ts` — runtime selector + client return
- `src/services/hooks/server-beta-client.ts` — new HTTP client with retry/timeout/graceful degradation
- `src/services/domain/ModeManager.ts` — server-mode awareness
- `src/services/worker/http/middleware.ts` — CORS preflight passthrough for viewer dev

No worker-runtime behavior change.

mguttmann:
Fixes #2564. Part of tracking issue #2540.

**Stacked on #2542.** The bootstrap reads `hashApiKeyForStorage` from `postgres-auth.ts` (added in #2542). When #2542 lands, this rebases to main.

Wires the lifecycle hooks to a runtime selector so a single installed plugin can be repointed at worker or server-beta runtime via configuration — no reinstall.

## Diff (vs base #2542)

```
 src/services/domain/ModeManager.ts          |  22 ++++--
 src/services/hooks/server-beta-bootstrap.ts | 109 ++++++++++++++++++++-------
 src/services/hooks/server-beta-client.ts    | 110 +++++++++++++++++++++++++---
 src/services/worker/http/middleware.ts      |   6 +-
 4 files changed, 202 insertions(+), 45 deletions(-)
```

## Changes

1. **Bootstrap** reads `CLAUDE_MEM_RUNTIME` (or `settings.json#runtime` fallback). Server-beta → HTTP client; worker → SQLite client (unchanged).
2. **HTTP client** (`server-beta-client.ts`): bearer-auth, exponential-backoff retry on 5xx (3 max), 5s timeout, single-flight queue, graceful degradation (hook failures never block the session).
3. **ModeManager**: server-beta pulls mode from `/v1/mode` with local cache; worker file-based path unchanged.
4. **Worker middleware**: CORS preflight passthrough for OPTIONS on `/v1/*` (viewer dev needs this).

## Verification

- `npm run typecheck` — clean
- runtime=server-beta + bootstrap → event appears in postgres within 200ms
- runtime=worker + bootstrap → unchanged SQLite behavior
- Server unreachable + retries exhausted → single error log, no throw, session continues

## Out of scope (separate PRs)

The MCP server runtime-selector parts depend on `smart-file-read/parser.ts` additions (upcoming Group E). `services/server/Server.ts` adds `helmet` as a new npm dep — also separate. Both intentionally not bundled here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "@clack/prompts": "^1.3.0",
     "@modelcontextprotocol/sdk": "^1.29.0",
     "ansi-to-html": "^0.7.2",
+    "argon2": "^0.44.0",
     "better-auth": "^1.6.9",
     "bullmq": "^5.76.6",
     "cors": "^2.8.6",
--- a/src/server/middleware/argon2-shim.ts
+++ b/src/server/middleware/argon2-shim.ts
@@ -60,6 +60,7 @@ export async function hashApiKeyForStorage(rawKey: string, opts: Argon2HashOptio
       algorithm: 'argon2id',
       memoryCost: merged.memoryCost,
       timeCost: merged.timeCost,
+      parallelism: merged.parallelism,
     });
   }
   const a2 = loadNativeArgon2();
--- a/src/server/middleware/argon2-shim.ts
+++ b/src/server/middleware/argon2-shim.ts
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// BUG 33 fix — runtime-adaptive argon2id shim.
+//
+// The npm `argon2` package compiles a native node-gyp addon. In the Docker
+// server-beta image we install with `--ignore-scripts` (BUG 32) to skip the
+// fragile tree-sitter native compile, which also leaves argon2 without a
+// native build. At runtime, importing argon2 then throws:
+//
+//   error: No native build was found for platform=linux arch=arm64
+//   runtime=node abi=137 ...
+//
+// Bun (which is what server-beta-service.cjs runs under) ships argon2id in
+// `Bun.password.hash()` and `Bun.password.verify()` — first-class APIs with
+// PHC-string output ($argon2id$v=19$m=...$...$...) fully interoperable with
+// the npm package's format.
+//
+// This shim picks Bun.password when available, otherwise lazily requires
+// the npm package. The native dep is no longer mandatory.
+//
+// Output format is identical either way ($argon2id PHC strings), so the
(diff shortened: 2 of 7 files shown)


mguttmann:

- **getTimelineWindow has no client-side guard** → added check before the fetch; throws `ServerBetaClientError('invalid_response', 'getTimelineWindow requires either anchor or query')` consistent with endSession/getJobStatus.
- **Stale SHA-256 reference in bootstrap comment** → updated to 'hash via hashApiKeyForStorage() (argon2id)'.

P1 DoS-amplifier flagged on the stacked-base postgres-auth.ts is fixed in commit 67001406 on #2542 (two-layer defence: RAW_API_KEY_SHAPE pre-filter + ARGON2_SCAN_LIMIT reduced 100→20).

mguttmann:
Thanks @greptile-apps[bot] — addressed in 0a615adf. `getMemoriesBatch` now rejects empty `ids` array client-side with `ServerBetaClientError('invalid_response', 'getMemoriesBatch requires at least one id')`, symmetric with the `getTimelineWindow` anchor/query guard added in the previous round.

Issue #2591, opened 2026-05-21
Windows: Chroma/uvx spawn fails instantly — cmd.exe parses < / > in --with version pins as redirection


GitHubxSuperKai:
## Summary
On Windows, the Chroma vector layer never connects. Every sync attempt logs `MCP error -32000: Connection closed` within ~20-30ms of spawn. The result: observations persist to SQLite but are never vectorized, so semantic search silently degrades. Affects **13.0.1 through latest 13.3.0** (identical code).

## Environment
- OS: Windows 11 (26200)
- claude-mem: 13.0.1 and 13.3.0
- uv/uvx: 0.11.7 (on PATH, working)
- Node: v24

## Root cause
The worker spawns the Chroma MCP server (in `scripts/worker-service.cjs`, `ChromaManager#connectInternal` / `buildCommandArgs`) as:

```
cmd.exe /c uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 --client-type persistent --data-dir ...
```

The deps are hardcoded:
```js
T8e = ["onnxruntime>=1.20", "protobuf<7"]   // (Y8e in 13.3.0)
```
and passed to `cmd.exe /c` as **bare, unquoted args**:
```js
let s = n ? ["/c", "uvx", ...e] : e;   // n = isWin32
```

`cmd.exe` interprets `>` and `<` as **redirection operators** before launching `uvx`:
- `onnxruntime>=1.20` -> redirect stdout to a file named `=1.20`
- `protobuf<7` -> redirect stdin **from a file named `7`** -> doesn't exist -> **"The system cannot find the file specified"**, child exits immediately.

Node's `child_process.spawn` only auto-quotes args containing whitespace, so the redirection metacharacters reach `cmd.exe` raw. The MCP client sees the child die before the handshake -> `-32000 Connection closed`.

The tell-tale is timing: connect->fail is ~20-30ms in the logs, far too fast for a real `uvx` launch (~9s cold). PATH is **not** the issue - `uvx` resolves fine; the same command runs perfectly when the metacharacters aren't exposed to `cmd`.

## Reproduction (Node, mirrors the SDK's plain `spawn`)
```js
const { spawn } = require("child_process");
const p = spawn("C:\\Windows\\System32\\cmd.exe",
  ["/c","uvx","--python","3.13","--with","onnxruntime>=1.20","--with","protobuf<7",
[...]

Issue #2585, opened 2026-05-20
chroma-mcp fails on Windows: cmd.exe interprets `<` and `>` in uvx args as I/O redirects


marcelopossa:
## Summary
On Windows, the chroma-mcp subprocess never starts because cmd.exe parses the unescaped `<` in `protobuf<7` (and `>` in `onnxruntime>=1.20`) as I/O redirects. Result: every Chroma sync fails with `MCP error -32000: Connection closed`, leaving Windows installs with no vector search.

## Environment
- claude-mem 13.2.0
- Windows + PowerShell 5/7
- Node passes args to spawn with shell=true (cmd.exe /c)

## Reproduce
1. Fresh install on Windows; run `npx claude-mem start`.
2. Worker boots; on first chroma-mcp connect the log shows:
   `args=/c uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 ...`
3. Subprocess dies within ~50ms with `MCP error -32000: Connection closed`.

## Root cause
In `scripts/worker-service.cjs` (`ChromaMcpManager.connectInternal()`):

```js
let n = process.platform === "win32",
    i = n ? process.env.ComSpec || "cmd.exe" : "uvx",
    s = n ? ["/c","uvx",...e] : e;
```

`e` contains literal `protobuf<7` and `onnxruntime>=1.20`. When `cmd.exe /c` receives them, `<` is parsed as stdin redirect from a file named `7`, which doesn't exist, so cmd exits before uvx runs. Confirmed by running the exact command and seeing `O sistema não pode encontrar o arquivo especificado.`

## Suggested fix
Escape `<` and `>` for the cmd shell only on Windows:

```js
s = n
  ? ["/c", "uvx", ...e.map(a => String(a).replace(/([<>])/g, "^$1"))]
  : e;
```

Verified locally: after this patch uvx runs, chroma-mcp connects in ~6s, and backfill completes for ~640 observations across 2 projects. A more thorough fix would escape all cmd metacharacters (`<>&|^()@!"`), but the minimal patch is enough for the current `P8e` constraints.

thedotmack:
Consolidating into #2377 (plan-02). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

Issue #2495, opened 2026-05-16
chroma-mcp never starts on Windows: `>=`/`<` in hardcoded deps eaten by cmd.exe redirection (v13.2.0)


davidfi01:
# Bug: chroma-mcp never starts on Windows — `>=`/`<` in hardcoded deps are eaten by cmd.exe redirection

**claude-mem version:** 13.2.0
**OS:** Windows 11 (win32)
**Provider:** any (reproduced with gemini; provider-independent — "Gemini" in the log line below is just the session-name prefix, not a Gemini-API failure)

## Summary

On Windows, every chroma sync fails immediately with `MCP error -32000: Connection closed`. The chroma-mcp subprocess dies before the MCP handshake, so vector search is permanently down and the worker retries forever (spawn → fail → kill subprocess tree → back off → repeat, ~every 20s).

```
[CHROMA_MCP] Connecting to chroma-mcp via MCP stdio {command=C:\WINDOWS\system32\cmd.exe,
  args=/c uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6
  --client-type persistent --data-dir C:/Users/.../.claude-mem/chroma}
[CHROMA_MCP] Connection failed, killing subprocess tree to prevent zombie {error=MCP error -32000: Connection closed}
[CHROMA_MCP] Connection attempt failed MCP error -32000: Connection closed
[CHROMA] ... chroma sync failed, continuing without vector search ... MCP error -32000: Connection closed
```

## Root cause

In `scripts/worker-service.cjs` the chroma-mcp deps are hardcoded (minified):

```js
ok="chroma-mcp", Dne="0.2.6", P8e=["onnxruntime>=1.20","protobuf<7"]
```

and the Windows MCP-server spawn path is:

```js
i = win32 ? process.env.ComSpec||"cmd.exe" : "uvx"
s = win32 ? ["/c","uvx", ...e] : e   // e contains "onnxruntime>=1.20", "protobuf<7"
```

This spawns `cmd.exe /c uvx … --with onnxruntime>=1.20 --with protobuf<7 …` via
`child_process.spawn` **without** `shell:true` and **without** cmd-style escaping.
Node.js does not escape `>` / `<` for `cmd.exe`, so cmd.exe interprets
`onnxruntime>=1.20` as *redirect stdout to file `=1.20`* and `protobuf<7` as
*redirect stdin from file `7`*. The command line is shredded; `chroma-mcp`
[...]

Issue #2429, opened 2026-05-11
Windows cmd.exe stdin redirect breaks chroma-mcp spawn — protobuf<7 arg interpreted as < redirect


BuildIQAdvisors:
## Summary

On Windows, claude-mem's `worker-service.cjs` spawns `chroma-mcp` via `cmd.exe /c uvx --with protobuf<7 chroma-mcp==0.2.6 ...`. **cmd.exe parses `<7` as a stdin redirect from a file named `7`**, killing the subprocess immediately with `"The system cannot find the file specified"` (exit code 1). This triggers MCP error -32000 and sets the chroma-mcp client into permanent backoff. claude-mem's keyword search remains operational, but semantic search via ChromaDB never starts.

## Environment

- OS: Windows 11 (WSL / Git Bash for shell)
- **claude-mem version: 13.1.0**
- **Node version: v24.14.0**
- **uv version: 0.11.13** (build 4512a3931, x86_64-pc-windows-msvc)
- ChromaDB MCP version: 0.2.6 (via `uvx --with protobuf<7 chroma-mcp==0.2.6`)
- Worker source: `scripts/worker-service.cjs`, `OGe` array contains literal `"protobuf<7"`

## Repro

1. Install claude-mem 13.1.0 on Windows
2. Run the worker (`worker-service.cjs`)
3. Observe in logs:
   ```
   [chroma-mcp] cmd.exe spawn failed: exit 1
   [chroma-mcp] The system cannot find the file specified.
   [chroma-mcp] entering backoff
   ```

`uv tool install chroma-mcp` from bash succeeds; the binary is present at `~/.local/bin/chroma-mcp.exe`. The failure is the spawn invocation shape, not the package.

## Root cause

`worker-service.cjs` `OGe` array passes `"protobuf<7"` unescaped as an argument when spawning via `child_process.spawn("cmd.exe", ["/c", "uvx", "--with", "protobuf<7", "chroma-mcp==0.2.6", ...])`. Windows cmd.exe interprets `<` as stdin redirect operator regardless of how the token is quoted at the JS layer (cmd.exe re-parses the joined command string).

## Proposed fix (lowest-risk one-liner)

Drop the `<7` upper-bound constraint:

```diff
- "protobuf<7"
+ "protobuf"
```

`protobuf` (no upper bound) was working before the constraint was added. ChromaDB's actual `protobuf` requirements should be expressed in its own dependency metadata, not as a claude-mem spawn-time pin.

[...]

PR #2593, opened 2026-05-21
fix(chroma): escape cmd.exe metacharacters in uvx args on Windows

YOMXXX:
## Summary

Five separate Windows issues (#2585, #2591, #2495, #2429, #2405) all share the same root cause: the worker spawns chroma-mcp via

```
cmd.exe /c uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp ...
```

and cmd.exe parses the unquoted `<` / `>` in those version pins as I/O redirection metacharacters **before** launching uvx. `protobuf<7` becomes "redirect stdin from a file called 7", which doesn't exist, so the child dies in ~50 ms and the MCP handshake never starts — every Chroma sync logs `MCP error -32000: Connection closed`.

Node's `child_process` only auto-quotes args containing whitespace, so these metacharacters reach cmd raw. Result: every Windows install since v13.0.1 has had silently broken vector search — observations persist to SQLite but are never embedded.

## Fix

In `src/services/sync/ChromaMcpManager.ts:108-115`, when `process.platform === 'win32'`, prefix each cmd.exe metacharacter (`^ < > & |`) with cmd's own escape character `^` before handing args to `StdioClientTransport`:

```ts
const escapeForCmd = (arg: string): string =>
  arg.replace(/([\^<>&|])/g, '^$1');
const uvxSpawnArgs = isWindows
  ? ['/c', 'uvx', ...commandArgs.map(escapeForCmd)]
  : commandArgs;
```

The character class includes `^` itself so a literal caret in an arg is escaped to `^^`, matching cmd.exe's quoting rules. POSIX path (uvx invoked directly, no shell) is unchanged.

## Verification

A freshly-built `worker-service.cjs` contains the inlined escape (`p.replace(/([\^<>&|])/g,"^$1")`). The resulting command line becomes:

```
cmd.exe /c uvx --python 3.13 --with onnxruntime^>=1.20 --with protobuf^<7 chroma-mcp==0.2.6 ...
```

which #2585's reporter independently confirmed lets chroma-mcp connect in ~6 s and backfill complete.

## Closes

- #2585
- #2591
- #2495
- #2429
- #2405

## Risk

Zero on macOS / Linux — escape only runs in the `isWindows` branch.


YOMXXX:
Heads-up from a related thread — `luoyehu` posted a finding on #2591 ([comment](https://github.com/thedotmack/claude-mem/issues/2591#issuecomment-4515562089)) that suggests this PR's escape-based fix may not be sufficient on its own:

> After pinning `protobuf==4.25.5` (no angle brackets), spawning `cmd.exe /c uvx` still fails instantly with exit code 1:
>
> ```
> STDERR: 另一个程序正在使用此文件，无法访问。
>         (Another program is using this file, cannot access)
> ```
>
> stdin pipe doesn't reach the Python child process. MCP handshake times out at 30s.
>
> Testing confirmed:
> - `cmd.exe /c uvx ...` → fails (file lock)
> - `powershell -Command "uvx ..."` → works
> - Spawning `uvx` directly (no cmd.exe wrapper) → works

If that holds, this PR's metacharacter escaping addresses the angle-bracket symptom but `cmd.exe /c` itself is a problem source on Windows — separate from `<` / `>` / `&` / `|` parsing.

## Possible direction

Drop the `cmd.exe /c` wrapper on Windows and spawn `uvx` directly, since `uvx.exe` is on PATH after a standard `uv` install. The current code in `ChromaMcpManager.ts:108-130` would simplify to roughly:

```ts
// Before
const isWindows = process.platform === 'win32';
const uvxSpawnCommand = isWindows ? (process.env.ComSpec || 'cmd.exe') : 'uvx';
const uvxSpawnArgs = isWindows
  ? ['/c', 'uvx', ...commandArgs.map(escapeForCmd)]
  : commandArgs;

// After
const uvxSpawnCommand = 'uvx';
const uvxSpawnArgs = commandArgs;
```

This sidesteps both classes of problems in one go:
- No more `cmd.exe` metacharacter parsing → the `escapeForCmd` helper becomes unnecessary for the chroma path.
- No more `cmd.exe /c` file-lock / stdin-pipe issue luoyehu hit.

## Caveats worth verifying before changing

1. **`uvx` shim type on Windows.** A standard `uv` install via `winget` / `pip install uv` / `pipx install uv` ships `uvx.exe` (real PE binary). But some package managers ship a `.cmd` / `.bat` wrapper, in which case Node's `spawn('uvx', ...)` without `shell: true` fails with ENOENT on Windows. If that scenario is in scope, the fallback is `spawn('uvx', args, { shell: true })` — which reintroduces cmd.exe parsing for the `shell: true` case, but only at one well-bounded site rather than for every metacharacter.
2. **PATH presence.** If `uvx` isn't on PATH, `cmd.exe /c uvx` would have failed too (cmd.exe also does PATH lookup), so this isn't a new failure mode.

## What this would do to this PR

If you take this direction, the `escapeForCmd` helper plus its `%`-doubling + `^`-prefixing logic could be removed entirely. The motivating issues (#2585, #2591, #2495, #2429, #2405) would all close because both symptoms (metacharacter parsing + file-lock + missing stdin) disappear together.

Happy to test a v2 if useful — I'm not on Windows day-to-day but I can spin up a VM.

(luoyehu's comment is the source of these specifics — the suggestion above is just my interpretation of "what does this imply for #2593"; cc @luoyehu in case I've misread anything.)

YOMXXX:
Gentle bump on this one — it's been sitting for ~8 days since the last round of review. The Windows `uvx` cmd.exe escaping fix is small and self-contained, Greptile rated it 5/5. No rush, just flagging it in case it slipped off the queue 🙂

YOMXXX:
Closing — superseded by upstream commit 8151cd5a, which implements equivalent cmd.exe metachar escaping via `quoteForCmdExe` in `ChromaMcpManager.ts`. Issues #2585, #2591, #2495, #2429, #2405 all resolved. Thanks!

Issue #2584, opened 2026-05-20
worker-service.cjs bundles unused better-auth library (94 OAuth provider URLs, 3.7 MB)


vinnievanwyk:
## Summary

`plugin/scripts/worker-service.cjs` (v13.2.0) ships a bundled copy of [`better-auth`](https://www.better-auth.com) even though `better-auth` is not a declared dependency in `plugin/package.json` and the worker doesn't need OAuth for its actual job (transcript ingestion / memory compression).

The visible consequence is that the bundle includes hard-coded endpoints for ~60 OAuth providers (PayPal, Apple ID, Spotify, Discord, Roblox, TikTok, Salesforce, Slack, GitHub, Linear, Notion, Dropbox, Figma, etc.). Anyone auditing the plugin's network surface — e.g. after seeing TCC permission prompts and pulling the bundle apart — finds a frankly alarming-looking URL list and has to do real work to confirm none of it is actually invoked.

## Evidence (claude-mem v13.2.0, freshly cloned marketplace)

```
$ wc -c plugin/scripts/worker-service.cjs
3,774,500   (3.7 MB)

$ grep -hEo "https?://[a-zA-Z0-9.-]+" plugin/scripts/worker-service.cjs | sort -u | wc -l
94          (unique HTTP/HTTPS hosts)

$ grep -c "better-auth" plugin/scripts/worker-service.cjs
12          (occurrences in the bundle)

$ grep -c "better-auth" plugin/package.json
0           (NOT a declared dependency)
```

Hosts present include the full Better Auth OAuth provider matrix:
`accounts.google.com`, `appleid.apple.com`, `api-m.paypal.com`, `accounts.spotify.com`, `discord.com`, `apis.roblox.com`, `open.tiktokapis.com`, `login.microsoftonline.com`, `login.salesforce.com`, `api.linear.app`, `api.notion.com`, `api.dropboxapi.com`, `slack.com`, `api.x.com`, `kapi.kakao.com`, `nid.naver.com`, `api.weixin.qq.com`, `api.zoom.us`, ... and ~40 more.

None of the auto-run paths I traced (`mcp-server.cjs`, `version-check.js`, `bun-runner.js`, `worker-cli.js`) reference these URLs — they only exist in `worker-service.cjs`. So the runtime behavior is fine; this is purely a bundling-hygiene issue.

## Why it matters

[...]

PR #2595, opened 2026-05-21
chore(scripts): drop sync-to-marketplace.sh and harden sync-marketplace.cjs

YOMXXX:
## Summary

Closes one half of #2570 (the bash side): removes the legacy `scripts/sync-to-marketplace.sh` so cross-platform setups don't carry a dead script around, and pulls the active Node-based `sync-marketplace.cjs` improvements from mguttmann's PR #2571 (cherry-picked **without** the bundle-size canary).

## Changes from #2571 included here

`sync-marketplace.cjs`:

- `--dry-run` / `-n` flag previews all rsync invocations without writing to the marketplace cache or running `bun install`.
- `--delete` is now opt-in via `--force-delete` (or `CLAUDE_MEM_SYNC_FORCE_DELETE=1`). The previous default could remove user-owned files (env files, secrets, etc.) that ended up alongside synced content.
- A `PRESERVE_PATTERNS` allowlist (`.env*`, `docker-compose.override.*.yml`, `secrets/`, `data/`, `.user-config/`) is fed as `rsync --exclude` so even when `--delete` is enabled these survive.
- Interactive would-delete preview prints before the destructive pass (skippable via `--force` / `--non-interactive` / `--no-preview` / `CLAUDE_MEM_SYNC_NO_PREVIEW=1`).

`sync-to-marketplace.sh`:

- Deleted. The shell variant was unmaintained and tripped on Git Bash / Windows path conventions; the `.cjs` is now the single source of truth.

## Intentionally excluded from this cherry-pick

`scripts/check-bundle-sizes.js` (the bundle-size canary from #2571) is **not** included. Its thresholds need to be re-baselined after the `better-auth` removal proposed in #2584 lands; merging the canary first would lock in a too-loose budget. That should be a separate PR sequenced after the bundle slimming.

## Verification

- `node -c scripts/sync-marketplace.cjs` reports no syntax issues
- `node scripts/sync-marketplace.cjs --dry-run` prints the marketplace / cache / installed-version phases without touching disk

## Closes (partial)

- #2570 (bash half — the bundle-size half remains open until #2584-baselined canary lands)

## Note on rsync

`sync-marketplace.cjs` still uses `rsync` internally. This is a **maintainer** tool (publishes the marketplace), not a user-facing one, so the dependency is fine for the macOS / Linux / Git Bash workflows that produce releases.

## Breaking change

`--delete` is no longer the default for `npm run sync-marketplace`. If your local workflow relied on stale files being pruned from the cache, pass `--force-delete` explicitly or set `CLAUDE_MEM_SYNC_FORCE_DELETE=1`. The `sync-marketplace:force` npm script should probably be updated to add this flag — left out of this PR to avoid scope creep.

## Credit

Cherry-pick of two changes from @mguttmann's #2571.


YOMXXX:
Friendly ping @thedotmack — CLEAN, Greptile 5/5, no outstanding conversations.

For context: PR #2595 drops the legacy `scripts/sync-to-marketplace.sh` and hardens `sync-marketplace.cjs` (single canonical Node-based sync, drops bash-only dependency). No behavioral change for the published artifact — just makes the build path cross-platform-clean and removes a duplicate code path.

Happy to rebase if needed.

YOMXXX:
Both Greptile suggestions addressed in ab7bc5b3:

1. Cached `deleteFlag()` result into a `DELETE` constant — avoids re-parsing `process.argv` on each call
2. Qualified the mirror message with `[dry-run]` prefix so output doesn't imply a side-effect during `--dry-run`

Ready for review when you get a chance 🙂

YOMXXX:
Hi @thedotmack — friendly ping on this when you get a chance. This one closes the bash half of #2570 by cherry-picking the safer rsync defaults from @mguttmann's #2571 (without the bundle-size canary, which is intentionally deferred until #2584 lands). Status is CLEAN / mergeable.

No urgency — just bringing it back on the radar in case it got buried. Happy to refresh if needed.

thedotmack:
Review findings that block this as-is: (1) undisclosed behavior change — seeding `syncManagedFiles` with `.mcp.json` un-excludes the maintainer's gitignored root `.mcp.json` (personal MCP config) and copies it into the marketplace dir that Claude Code loads from; that needs to be justified or dropped. (2) The `--delete` default flip leaves `npm run build-and-sync` with no mirror path — removed hooks/skills linger live in the installed cache; should land together with a `--force-delete`-bearing npm script. The `sync-to-marketplace.sh` removal itself is verified safe (zero references on main) — split that out and it merges immediately.

thedotmack:
Closing as already implemented by the current cross-platform sync path; the legacy sync-to-marketplace.sh file is no longer present.

Issue #2439, opened 2026-05-12
Windows (Git Bash): printf write error Permission denied blocks UserPromptSubmit hook


UnaxAlonso0:
## Environment

- OS: Windows 11 Home 10.0.26200
- Shell: Git Bash (MSYS2) via Claude Code
- Node: v24.14.1
- claude-mem: v13.2.0
- Claude Code: latest

## Bug

Every `UserPromptSubmit` (and other hooks) is **blocked** with:

```
UserPromptSubmit operation blocked by hook: [...]: /usr/bin/bash: line 1: printf: write error: Permission denied
```

The failing part is the path-finding subshell in every hook command:

```bash
_P=$({ [ -n "$_E" ] && printf '%s\n' "$_E"; ls -dt ...; printf '%s\n' "$_C/plugins/marketplaces/thedotmack/plugin"; } | while IFS= read -r _R; do
  ...
  { printf '%s\n' "$_Q"; break; }
done)
```

On Windows/MSYS2, `printf '%s\n'` inside a process substitution `$()` pipeline fails with `Permission denied` when stdout is redirected to a pipe for the variable assignment. This causes bash to exit non-zero, which blocks the hook.

## Reproduction

Any Claude Code session on Windows with claude-mem enabled. The error appears immediately on the first user prompt.

## Workaround tried

Replacing all `printf '%s\n'` with `echo` in `hooks/hooks.json` (both marketplace and cache copies) — the JSON is functionally identical since `echo "$var"` ≡ `printf '%s\n' "$var"` for single-line strings. However this gets overwritten on reinstall.

Also tried: `npx claude-mem@latest install` — runtime installs correctly (Bun + uv OK), but hooks still fail.

## Suggested fix

Replace `printf '%s\n'` with `echo` in the path-finding subshells within all hook commands in `hooks/hooks.json`. Alternatively, detect Windows platform and use a temp-file approach to avoid nested process substitutions.

Note: `export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH"` at the start of each hook may also cause issues on Windows (spawning a login shell), though the immediate blocker is the `printf` write error.

thedotmack:
Consolidating into #2377 (plan-02). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

AKHILPRAVEEN1:
## Confirming this fix works on v13.4.0 (Windows 11 / Git Bash)

[...]

Issue #2588, opened 2026-05-20
printf: write error: Permission denied spam from hooks on Windows / Git Bash


agomezquiroga-cloud:
## Summary

  On Windows with Git Bash (MSYS2), every hook in `hooks/hooks.json` emits a non-blocking error after each tool use:

  ```
  PreToolUse:Read hook error
  Failed with non-blocking status code: /usr/bin/bash: line 1: printf: write error: Permission denied
  ```

  Functionally everything works (the plugin path is detected, memory captures fine), but the transcript gets flooded with these messages.

  ## Environment

  - Plugin: `claude-mem@thedotmack` **v13.2.0** (from official marketplace)
  - OS: Windows 11
  - Shell: Git Bash (MSYS2 bash 5.x, `/usr/bin/bash`)
  - Claude Code: latest

  ## Root cause

  Each hook uses this pattern to find the plugin path:

  ```bash
  _P=$({
    [ -n "$_E" ] && printf '%s\n' "$_E"
    ls -dt "$_C/plugins/cache/thedotmack/claude-mem"/[0-9]*/ 2>/dev/null
    printf '%s\n' "$_C/plugins/marketplaces/thedotmack/plugin"
  } | while IFS= read -r _R; do
    ...
    [ -f "$_Q/scripts/bun-runner.js" ] && { printf '%s\n' "$_Q"; break; }
  done)
  ```

  When the first candidate already satisfies the check, `while` breaks immediately and closes the pipe. The remaining `printf`/`ls` calls on the left-hand side then write to a closed pipe → `SIGPIPE`.

  On Linux/macOS bash this surfaces as `Broken pipe`. **On Git Bash / MSYS2 the same SIGPIPE is reported as `Permission denied`** — a known MSYS quirk. Either way the message goes to stderr and Claude Code
  surfaces it as a hook error.

  ## Repro (Git Bash on Windows)

  Run any tool call (e.g. `Read` or `Bash`) in a session with `claude-mem` enabled and observe the transcript.

  ## Suggested fix

  Silence stderr of the left-hand subshell — the path detection itself doesn't rely on stderr:

  ```diff
  - } | while IFS= read -r _R; do
  + } 2>/dev/null | while IFS= read -r _R; do
  ```

  Applied to all 7 hooks (Setup, SessionStart ×2, UserPromptSubmit, PostToolUse, PreToolUse, Stop) in `hooks/hooks.json`. I've patched my local cache with this and the noise is gone; happy to send a PR if
[...]

Issue #2461, opened 2026-05-13
Fix Windows compatibility: replace sh spawn + silence printf EPIPE


dmoroco18-design:
**Branch ready to merge:** `dmoroco18-design:fix/windows-sh-and-printf-epipe` — 3 commits (a68b22d, 4e9ff0c, b1db288). Can't open PR directly (collaborator-only restriction); if you want to pull the branch or invite me as a collaborator, happy to open the PR. Otherwise this issue captures the diagnosis + fix for whoever lands it.

## Summary

Two Windows-specific bugs prevent claude-mem from working cleanly in Claude Code CLI on Windows. Both fixes are surgical, cross-platform (no Linux/macOS regression), and based on a reproducible bottom-up diagnosis.

Forked from `v13.2.0` so the change applies against the version currently in the marketplace.

## Problem 1 — `.mcp.json` uses bare `sh` command, which is unreachable on Windows

`plugin/.mcp.json` declares the `mcp-search` MCP server with:

```json
"command": "sh",
"args": ["-c", ""]
```

On Windows, `sh.exe` is shipped by Git Bash at `C:\Program Files\Git\usr\bin\sh.exe`. That directory is on the **bash shell's** PATH (`/usr/bin` is mapped there) but **not on the Windows system PATH**. Native-Windows process spawns (which is what Claude Code CLI uses to launch MCP servers) cannot resolve `sh` or `sh.exe`. The spawn fails with `ENOENT` deterministically.

### Symptom
- `/mcp` panel persistently shows `plugin:claude-mem:mcp-search` as **failed**, regardless of how many times the user restarts Code CLI.
- `mcp__claude-mem__*` tools never appear in the tool registry, so users cannot invoke `memory_search`, `smart_search`, `timeline`, etc.
- The `mcp-server.cjs` server itself is healthy — when launched manually it speaks MCP correctly and responds to `initialize` with the right `serverInfo`. The bug is purely in the launcher.

### Evidence (reproducible)
1. `cmd.exe /c "where sh.exe"` → empty output (system PATH).
2. Node `child_process.spawnSync('sh', ['-c', 'echo hi'])` with a Windows-system-PATH-only env → `ENOENT: spawnSync sh ENOENT`.
3. Same probe with `shell: true` (cmd.exe lookup) → `"'sh' is not recognized as an internal or external command"`.
[...]

PR #2598, opened 2026-05-21
fix(hooks): replace fragile $SHELL -lc PATH probe with printenv-first pattern

### LINKED ISSUE #2589: Windows Git Bash: printf 'Permission denied' in hook scripts
## Bug Description

On Windows with Git Bash (MSYS), all hooks fail with:
```
```

## Environment

- Windows 11 Pro + Git Bash (MSYS GNU bash 5.2.37)
- claude-mem 13.2.0
- Shell: bash (configured as \`"shell": "bash"\` in hooks.json)

## Root Cause

Hook commands use this pattern to build PATH:
```bash
export PATH="\$(\$SHELL -lc 'echo \$PATH' 2>/dev/null):\$PATH"
```

When hook executes, stdin is not a proper TTY (it's a pipe or closed fd). The \`-l\` flag (login shell) behaves unexpectedly in Git Bash when stdin is not a terminal, causing \`\$SHELL -lc 'echo \$PATH'\` to return an abnormal result or fail silently.

Subsequent subshell with pipe + printf fails:
```bash
{ [ -n "\$_E" ] && printf '%s\n' "\$_E"; ... } | while IFS= read -r _R
```

When the left side of the pipe runs in a broken state from the prior failure, \`printf: write error: Permission denied\` occurs.

## Suggested Fix

1. Replace \`\$(\$SHELL -lc 'echo \$PATH')\` with direct PATH access — the PATH is already in the environment, no need to re-query it via a login shell
2. Add \`2>/dev/null\` to all printf statements as defensive fallback
3. Replace pipe-to-while-read pattern with \`for\` loop for better portability:
   ```bash
   for _R in \$({ ... }); do
   ```

Or use a more portable subshell approach that doesn't depend on stdin state.

YOMXXX:
## Summary

Closes #2589, #2588, #2461, #2439 — four reports of the same Windows Git Bash failure.

Six hooks in `plugin/hooks/hooks.json` (SessionStart×2, UserPromptSubmit, PostToolUse, PreToolUse, Stop) currently start with:

```bash
export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH"; ...
```

Under Git Bash on Windows (MSYS bash 5.x), when stdin is not a TTY (which it never is for a hook subprocess), `$SHELL -lc` returns empty or fails in a way that bash interprets as a broken pipe, producing the well-documented:

```
```

…cascading through every subsequent pipe in the hook command. **Net effect on Windows: every hook fires this stderr noise on every invocation.**

## Fix

Replace the broken probe with the same battle-tested pattern that `plugin/hooks/codex-hooks.json` already uses:

```bash
_HP=$(printenv PATH 2>/dev/null || true);
if [ -z "$_HP" ] && [ -n "${SHELL:-}" ]; then
  _HP=$("$SHELL" -lc 'printf %s "$PATH"' 2>/dev/null || true);
fi;
_HP=$(printf '%s' "$_HP" | tr ' ' ':');
export PATH="${_HP:+$_HP:}$PATH";
```

Three improvements vs. the broken version:

1. **`printenv PATH` is the canonical POSIX way** to read an env var — no TTY dependency, no pipe to break. The login-shell probe runs only as a fallback when printenv returns empty (which would only happen if PATH is genuinely unset).
2. **`"$SHELL" -lc 'printf %s "$PATH"'`** (quoted, printf not echo) is the safer form — `echo` interprets backslashes in some shells.
3. **`tr ' ' ':'`** defends against Git Bash / MSYS quirks where literal spaces in PATH could break downstream parsing.

## Setup hook intentionally untouched

The Setup hook (line 11) already uses a hand-rolled PATH that includes nvm / brew / `/usr/local` — it doesn't share the broken pattern and is intentionally left untouched.

## Relationship to #2526

PR #2526 silenced the downstream `printf: Permission denied` symptom by adding `2>/dev/null` to the consumer pipe, but the noise was still being generated. With this PR's root-cause fix, the symptom goes away entirely because the upstream probe no longer fails. These two PRs are complementary — #2526 belt, this one suspenders.

## Verification

```
$ grep -c "SHELL -lc 'echo \$PATH'" plugin/hooks/hooks.json
0
$ grep -c "printenv PATH" plugin/hooks/hooks.json
6
$ node -e "JSON.parse(require('fs').readFileSync('plugin/hooks/hooks.json'))"
# OK

$ bash -c '_HP=$(printenv PATH ...); ... echo "len=$(echo $PATH | wc -c)"'
PATH length=3626  contains-usr-local=1
```

The new probe is byte-for-byte the same as the one shipping in `codex-hooks.json` today, so it's already been validated across the matrix of shells we need to support.

## Closes

- #2589
- #2588
- #2461
- #2439

## Risk

Very low — only 6 lines change, each is the same regex replacement, and the new probe is already in production for the Codex CLI hooks. macOS / Linux behavior is unchanged (`printenv PATH` returns the same value `$SHELL -lc 'echo $PATH'` returned on those platforms).


kls06541:
Verifying this closes #2439 on my end — I'm a Windows user hitting the same hook block, but on Win10 Pro 10.0.19045 + Git Bash 5.2.37 MSYS, not the Win11 Home environment in the original report.

One thing I noticed while reading the diff: the PATH probe is now safe, but the `_P=$({ ... } | while IFS= read -r _R; do ... done)` subshell that the issue body specifically calls out is **unchanged** in this PR. The original report attributes the `printf: write error: Permission denied` to that nested pipe, with the PATH probe only as an aggravating factor.

On my Win10 machine I cannot reproduce the failure with either the unfixed or fixed PATH probe — both produce `_P` correctly under stdin-from-pipe, stdin=/dev/null, and stdin-closed. So I can't falsify the close-claim from here.

Could @UnaxAlonso0 (or anyone still on Win11 + Node 24.x) confirm whether this branch alone — without any further changes to the \`_P=\$({...}|while...)\` block — actually eliminates the \`printf: write error: Permission denied\` in practice? If so, great, #2439 is genuinely closed. If the symptom recurs, a follow-up swapping that nested pipe for a \`mktemp\` temp-file (the alt approach suggested in #2439's body) would still be worth carrying.

YOMXXX:
@kls06541 Good catch, and thanks for the careful read — you're right that #2439's body identifies the `_P=$({...} | while IFS= read -r _R; do ... done)` nested pipe as the primary failure site, with the `$SHELL -lc` PATH probe as the aggravating factor rather than the root cause. This PR strictly addresses the PATH-probe half.

I cannot reproduce on macOS either, and like you my Windows access (Win10 Pro / Git Bash MSYS) doesn't trip the original symptom across any stdin condition I tried. So this PR should land as the partial fix it actually is — the `closes #2439` line in the description is optimistic until a Win11 + Node 24 user confirms.

Opened **#2613** to track the nested-pipe concern specifically with the `mktemp` temp-file rewrite (the alternative your issue body suggests) staged and ready to apply. That way the audit trail stays linked to #2439, and if @UnaxAlonso0 or any Win11 + Node 24.x user confirms the symptom persists after #2598, the rewrite goes in immediately. If the symptom is genuinely gone, #2613 closes as covered-by-#2598.

(Pinging @thedotmack so this scope clarification is visible at merge time — happy to amend the PR description to drop the auto-close on #2439 if that's preferred, in case the Win11 verification doesn't materialize before merge.)

YOMXXX:
Addressed the remaining Greptile suggestion in e060a674 — removed `tr ' ' ':'` from all 6 hook commands to avoid corrupting Windows paths with spaces (e.g. `/c/Program Files/nodejs`).

Ready for review whenever convenient 🙂

YOMXXX:
Hi @thedotmack — gentle ping on this one whenever you have a window. It's a 6-line change that closes four user reports of the same Git Bash noise (#2589 #2588 #2461 #2439), and the new probe is the exact same pattern already shipping in `codex-hooks.json`, so the cross-platform risk is minimal. Status is currently CLEAN / mergeable.

No rush — just flagging it in case it slipped past. Happy to rebase if anything has drifted since.

thedotmack:
Review findings that block this: (1) it edits generated files — `plugin/hooks/hooks.json` is produced from `src/build/hook-shell-template.ts`, and the build's Rule A verification (`scripts/build-hooks.js`) rejects hand-edited command strings, so this fails the build and any regen reverts it; the fix must land in the template. (2) Behavior regression: the old prelude unconditionally prepended the `$SHELL -lc` login-shell PATH, which is what makes `node` resolvable for users whose nvm/brew paths only live in `~/.zprofile`/`~/.bash_profile` (GUI-launched Claude with minimal PATH). The new code only falls back when `printenv PATH` is empty — essentially never — so for that population every hook breaks with `node: command not found`. A safer shape: keep printenv-first but run the login-shell probe additively when `command -v node` fails, and land it in the template. Note #2865 already fixed the Windows EACCES pipe noise at the root.

thedotmack:

> step 1 (prelude does no discovery in the common path); must land in src/build/hook-shell-template.ts, not hooks.json; keep login-shell fallback for GUI launches

thedotmack:
Second pass — this belongs to **#3605 (plan-17, Hook Wrapper Contract)**, the login-shell PATH prelude item. Closing as superseded: #3453 (`7234951f2`, "fix(hooks): stop login-shell PATH rebuild on Claude Code hooks") merged on 2026-09-11 and removed the `$SHELL -lc 'echo $PATH'` probe outright — main's `plugin/hooks/hooks.json` no longer contains it, and the change landed in `src/build/hook-shell-template.ts`, which is the generated-file objection raised against this PR. The Windows Git Bash write-error cascade you and @kls06541 documented should be gone; please reopen a fresh report against current main if it still reproduces. Four issues' worth of careful diagnosis here — thank you.

Issue #2574, opened 2026-05-19
Summarizer hallucinates cross-session content + fabricates commit hashes (haiku-4-5)


slyderc:
## Summary

The session summarizer (running on `claude-haiku-4-5-20251001`) emitted observations whose **title / facts / narrative describe content unrelated to the actual session**, while the `files_modified` list correctly reflects the real session's repo. It also **fabricated a git commit hash** that does not exist in the target repo.

Result: future sessions that consume these observations as memory inject false context. When the user later asked "what's next on this project?", they were confidently told about features that have never existed in the codebase.

## Environment

- claude-mem plugin: **v13.0.1** (via marketplace `thedotmack`)
- Claude Code on macOS 24.6.0
- Summarizer model recorded in observations: `claude-haiku-4-5-20251001`

## Evidence

Three observations share the same `memory_session_id`, generated seconds apart, all stored under the same `project` value:

| Obs | created_at | Result |
|---|---|---|
| A | 19:57:56Z | **Wrong.** Title and narrative describe a feature set that does not exist in the repo. Claims a commit hash that **does not exist in the repo's git history**, paired with a fabricated commit message. |
| B | 19:58:30Z | **Wrong.** Narrative describes endpoints, file types, asset formats, and build tooling that do not exist in the repo at all (wrong language/architecture). |
| C | 19:59:02Z | ✅ Correct. Accurately summarizes the actual session. |

Key oddity: **`files_modified` is correct in all three.** So the summarizer had accurate file context but invented a narrative untethered from those files. The bad observations also have `files_read: "[]"` — possibly indicating the summarizer did not actually read the files it was summarizing.

## Hypothesis

Two non-exclusive guesses:

1. **Confabulation under thin signal.** The summarizer was given file paths + diff stats but not enough content, and hallucinated a plausible-sounding description. The bad observations contain specific technical details (endpoint names, file formats, build commands) that sound like they came from training data or another session's content rather than pure fabrication.
2. **Cross-session context bleed.** If the summarizer was prompted with content from multiple recent sessions, it may have attributed another session's content to this one's file list.

The third observation in the same batch (C) is fully correct, so the failure mode is not deterministic.

## Impact

- Memory injection at session start surfaces false observations as authoritative project context.
- Downstream sessions confidently relay the false content unless they cross-check against `git log` or the actual repo state.
- The fabricated commit hash is the most dangerous case — it's the kind of detail a user would trust without verifying.

## Suggested mitigations (FWIW)

[...]

PR #2602, opened 2026-05-21
fix(observer): truncate oversized prompt fields + drop overflow from hard-stop (closes #2468)

### LINKED ISSUE #2468: Observer context has no budget management: unbounded tool_output causes infinite overflow loop with data loss
## Summary

The observer session has no input-size guards, no context budget tracking, and no graceful overflow recovery. A single large `tool_response` (e.g. reading a file larger than the observer model's context window) triggers a **hard-stop that discards all queued observations**, after which the system immediately creates a fresh session and repeats the cycle indefinitely.

Related: #2201 (runaway token consumption), #2341 (root cause: "Observer/summarizer input is unbounded"). This issue provides a code-level trace of the specific failure path and proposes a layered fix.

---

### 1. Zero truncation on tool output

`buildObservationPrompt()` (`src/sdk/prompts.ts:81-113`) serializes `tool_input` and `tool_output` via `JSON.stringify(..., null, 2)` with **no size limit**:

```ts
// prompts.ts:106-107
```

A `Read` on a 130k-char file produces a single observation prompt that exceeds a 128k-token observer model's entire context window. There is no truncation, no character/token budget check, and no estimation before yielding the message.

### 2. Token budget is tracked but unused

`ClaudeProvider.ts:321-336` accumulates `cumulativeInputTokens` and `cumulativeOutputTokens` from the SDK response `usage` object. These values are **only logged** — they never influence any decision about whether to continue, reset, or skip messages.

### 3. Overflow recovery destroys innocent messages

When the Anthropic API returns `"Prompt is too long"`, `ClaudeProvider.ts:306-313` detects it and calls `resetSessionForFreshStart()`, which clears `memorySessionId`. The generator aborts with `reason='overflow'`.

In `GeneratorExitHandler.ts:14-20`, `overflow` is classified as a **hard-stop reason**:

```ts
function isHardStopReason(reason): boolean {
  return reason === 'shutdown' ||
    reason === 'restart-guard' ||
    reason === 'overflow' ||   // ← goes to hard-stop path
    reason === 'quota' || ...
}
```

The hard-stop path (`GeneratorExitHandler.ts:80-86`) calls `terminateSession('Hard-stop', true)` which:
1. `clearPendingForSession()` — **deletes all queued observations** from the database, including unrelated small ones
2. `finalizeSession()` → `markSessionCompleted()` — marks the session as completed
3. `removeSessionImmediate()` — removes the session from memory

The next `PostToolUse` hook creates a **brand-new** database session (new `sessionDbId` via `createSDKSession`), with a fresh `ActiveSession` object containing a new `RestartGuard` (counting from zero). This means:

- **RestartGuard never accumulates** across overflow cycles — every cycle starts fresh.
- There is **no upper bound** on how many times this cycle repeats.

### Reproduction

1. Configure `CLAUDE_MEM_MODEL` to a 128k-context model (e.g. `claude-haiku-4-5-20251001`)
2. In the primary Claude Code session, `Read` a file with >100k characters
3. Observer receives the full `tool_response` via hook → context exceeds 128k → API returns error
4. Observer hard-stops → all queued messages deleted → new session created
5. Repeat step 2 with another large file → same cycle, infinitely

## Impact

| Aspect | Effect |
|--------|--------|
| **Data loss** | All queued observations (including small, valid ones) are permanently deleted on each overflow |
| **Token waste** | Each cycle re-sends `buildInitPrompt()` (~2-3k tokens) and the oversized observation before failing |
| **No protection** | `RestartGuard` resets on each new session; the cycle never stops |
| **Quality degradation** | New sessions start with no prior context, reducing observation quality for subsequent tool calls |
| **Cost** | Unbounded token consumption when working with large files (see #2201) |

### Layer 1 — Input truncation (highest impact, smallest change)

Cap `tool_output` (and optionally `tool_input`) in `buildObservationPrompt()` or `createMessageGenerator()` before yielding. The observer only needs to understand *what happened*, not the full file content:

```ts
const MAX_OUTCOME_CHARS = 30_000; // ~8k tokens, plenty for context

let outcomeStr = JSON.stringify(toolOutput, null, 2);
if (outcomeStr.length > MAX_OUTCOME_CHARS) {
  outcomeStr = outcomeStr.substring(0, MAX_OUTCOME_CHARS)
    + `\n<!-- TRUNCATED: ${outcomeStr.length} chars total -->`;
}
```

Add a configurable `CLAUDE_MEM_MAX_OBSERVATION_CHARS` setting (default: 30000) in `SettingsDefaultsManager`.

This single change eliminates the "one large file kills the observer" path.

### Layer 2 — Proactive budget management

Use the already-tracked `cumulativeInputTokens` to trigger a **graceful reset** before hitting the API limit:

```ts
const BUDGET_RATIO = 0.8;
const modelWindow = getModelWindowTokens(modelId);

if (session.cumulativeInputTokens > modelWindow * BUDGET_RATIO) {
  // Reset context history but keep queue intact
  session.conversationHistory = [];
  session.memorySessionId = null;
  session.forceInit = true;
  // Do NOT clear queue — messages will be processed in fresh session
}
```

Key difference from current behavior: the queue is **preserved**, not destroyed. Messages re-process in the new session.

### Layer 3 — Graceful overflow recovery

Split the `overflow` case out of `isHardStopReason` in `GeneratorExitHandler`:

```ts
if (isHardStopReason(reason) && reason !== 'overflow') {
  await terminateSession('Hard-stop', true);
} else if (reason === 'overflow') {
  // Preserve queue, reset context, let respawn logic restart
  await sessionManager.resetProcessingToPending(sessionDbId);
  session.conversationHistory = [];
  session.memorySessionId = null;
  session.forceInit = true;
  // Don't terminate — let pending-work-restart handle it
}
```

### Layer 4 — Context compression (future)

When history exceeds a threshold, compress older observations into a condensed summary. Lower priority since Layers 1-2 prevent most overflow scenarios.

---

## Files Involved

| File | Role |
|------|------|
| `src/sdk/prompts.ts` | `buildObservationPrompt()` — zero truncation on tool I/O |
| `src/services/worker/ClaudeProvider.ts` | Token tracking (logged but unused), overflow detection |
| `src/services/worker/session/GeneratorExitHandler.ts` | `isHardStopReason('overflow')` → terminate with queue deletion |
| `src/services/worker/session/SessionCompletionHandler.ts` | `finalizeSession()` → marks completed, clears pending |
| `src/shared/SettingsDefaultsManager.ts` | Where `CLAUDE_MEM_MAX_OBSERVATION_CHARS` would go |

YOMXXX:
## Summary

Closes #2468 — the observer SDK session death-loop where a single oversized tool_output wipes the entire queued-observation batch.

## The bug (as reported in #2468)

1. A primary-session `Read` of a large file (~130k chars in the reported case) produces an observation whose `JSON.stringify`'d `tool_output` alone exceeds the observer model's 128k-token context window.
2. `ClaudeProvider` yields the prompt; the SDK assistant response comes back with "prompt is too long" / "context window".
3. `ClaudeProvider` (lines 306–313) sets `abortReason='overflow'`, calls `resetSessionForFreshStart`, and aborts.
4. `GeneratorExitHandler` (lines 14–20) treats `'overflow'` as a **hard stop**: calls `clearPendingForSession(sessionDbId)`, wiping every queued observation — including all the unrelated ones queued after the offending Read.
5. New observations from the primary session re-trigger the generator; if the same oversized Read replays (or any other oversized tool_output arrives), step 1 fires again. Restart guard catches the third-or-so cycle, but **data is lost on every iteration**.

`buildObservationPrompt()` in `src/sdk/prompts.ts` had **zero size guard** — `JSON.stringify(toolInput/toolOutput, null, 2)` was inlined verbatim into the XML wrapper.

### 1. Truncate at the prompt layer (primary fix)

New `truncateObservationField()` enforces a **16k-char cap per `<parameters>` / `<outcome>` field** (≈4k tokens, ≈8k tokens combined). Content past the cap is replaced with a head + tail slice (60% head / 30% tail) plus an explicit `<elided>` marker:

```xml
<outcome>{
  "filePath": "/path/to/big.txt",
  "content": "first 9.6k chars of content...
... <elided chars="120000" original_size_chars="150000" reason="oversize" /> ...
...last 4.8k chars including any trailing errors"
}</outcome>
```

The prompt body explicitly tells the observer:

> If a `<parameters>` or `<outcome>` block above contains an `<elided chars=... />` marker, that field was truncated to fit the observer's context window. Describe only what you can see in the kept portion and do not infer details about the elided range.

This is the same grounding principle being explored for #2574 — surfacing truncation as a signal is cheaper than letting the model silently confabulate detail about content it never saw.

### 2. Stop treating `'overflow'` as a hard stop (defense in depth)

If something else still trips "prompt is too long" — e.g. a conversation history that grew faster than truncation reclaims space — `GeneratorExitHandler.isHardStopReason()` no longer returns true for `'overflow'`. The generator restart path runs instead, preserving the queued observations.

The existing `RestartGuard` caps how many `overflow → restart` cycles can run before the session is torn down, so we can't loop forever; we just stop wiping good data on the way out.

## Verification

Smoke test of `truncateObservationField` (200k-char input):

```
original size  : 200019
truncated size : 14482
contains marker: true
head OK        : true (starts with the JSON head)
tail OK        : true (ends with the JSON tail)
small unchanged: true (no marker, byte-identical to JSON.stringify)
```

Full `npm run build` passes; worker-service.cjs rebuilt to include the new truncation logic.

## Knobs (rationale in comments)

| Constant | Value | Rationale |
|---|---|---|
| `OBS_PROMPT_FIELD_MAX_CHARS` | 16000 | ≈4k tokens per field × 2 fields = ≈8k tokens of variable input. Leaves the bulk of a 128k-token observer budget for system + history + response. |
| `OBS_PROMPT_FIELD_HEAD_RATIO` | 0.6 | Captures canonical signal (file path, command header, initial error). |
| `OBS_PROMPT_FIELD_TAIL_RATIO` | 0.3 | Captures stack traces, trailing errors. Remaining 10% is the elision marker. |

## What this PR deliberately does NOT do

- **No token-budget tracking** (`ClaudeProvider.ts:317-336` still only logs `cumulativeInputTokens/cumulativeOutputTokens`, doesn't consume). A future PR can add proactive session reset before overflow becomes possible — that's a deeper design change.
- **No partial-flush** (moving 'processing' rows back to 'pending' explicitly). The current restart path already replays pending rows on next generator start, so this is an optimization, not a correctness gap.
- **No grounding validator** (post-process check that fact references exist in the source observation). That's #2574 territory; this PR adds the prompt-side hint that `<elided/>` markers exist, leaving the validator as a separate follow-up.

## Closes

- #2468

## Related

- #2574 — observer cross-session contamination. Separate follow-up; this PR adds the prompt scaffolding (the elision marker + grounding hint) that grounding work will lean on.

## Risk

- **Existing observations** (post-deploy): if a previously-overflowing observation now succeeds with truncation, the resulting `<observation>` blocks will reference only the kept head+tail. That's strictly better than today's "wipe the entire batch" failure mode.
- **Conversation history overflow**: if it happens, restart-with-pending-preserved replaces clear-pending; identical net behavior on the user's terminal (worker recovers), but no data loss along the way.
- **No external API changes**, no new dependencies, no schema changes.


YOMXXX:
Implemented the remaining defensive truncation guard from the latest review summary.

Change made:
- clamp computed head/tail lengths to non-negative values
- avoid `slice(-0)` by using an empty tail when `tailChars === 0`
- clamp `elidedChars` so future ratio changes cannot produce negative markers
- rebuilt `plugin/scripts/worker-service.cjs`

Verification:

```
npm run build
```

Build passes. I also ran `npm run typecheck`; it still fails on existing errors outside this PR scope (npx install command arity, logger Component typing, missing exported types, and request param string/string[] typing), not in the touched prompt truncation path.

YOMXXX:
Superseded — splitting this up after rebasing on current `main`:

- **Prompt-field truncation** (the still-relevant half) → extracted into #2748, rebased on current `main`, with the unit tests this PR was missing.
- **`GeneratorExitHandler.ts` overflow → pending-restart change** → now obsolete. Upstream `61fe70a2` (`refactor: remove retry/observation queue, replace with SessionMessageBuffer`) removed the durable pending-queue + respawn mechanism this half depended on, so it no longer applies.

Closing in favor of #2748.

Issue #2407, opened 2026-05-10
v13.0.0 marketplace install bundle ships without bundled node_modules (zod, shell-quote)


seo-yas:
## Summary

The v13.0.0 marketplace install bundle is missing its `node_modules/` dependencies. After `/plugin install claude-mem` completes, `worker-service.cjs` fails to start with `Cannot find module 'zod/v3'` (and `shell-quote` is similarly absent).

Skills (`/learn-codebase`, `/mem-search`, `/knowledge-agent`, etc.) are registered correctly and callable via slash commands. But the MCP server transport never starts cleanly, so any MCP-driven workflow degrades silently.

This is a **packaging issue at the marketplace bundle level**, not a local install corruption. A fresh `/plugin uninstall` + `/plugin install` reproduces the same broken state — the bundle delivered through the marketplace channel does not include the required dependencies.

## Reproduction (cleanly verified twice on the same machine)

1. `/plugin uninstall claude-mem` (Claude Code TUI)
2. `/plugin install claude-mem` (Claude Code TUI)
3. `/reload-plugins` — reports `1 plugin MCP server · 5 skills · 27 agents · 13 hooks` loaded successfully
4. Send any prompt to Claude Code → `UserPromptSubmit hook error` fires:

```
Failed with non-blocking status code: error: Cannot find module 'zod/v3'
from 'C:\Users\yasir\.claude\plugins\cache\thedotmack\claude-mem\13.0.0\scripts\worker-service.cjs'
```

5. Verify the install dir:
```bash
ls ~/.claude/plugins/cache/thedotmack/claude-mem/13.0.0/node_modules/zod
# → ENOENT (does not exist)

cat ~/.claude/plugins/cache/thedotmack/claude-mem/13.0.0/package.json | grep -E '"(zod|shell-quote)"'
# → "zod": "^4.3.6", "shell-quote": "^1.8.3" — declared as deps but not bundled
```

6. Manual workaround restores MCP transport:
```bash
cd ~/.claude/plugins/cache/thedotmack/claude-mem/13.0.0/
bun install
# (or npm install — the package ships a bun lockfile)
```

After the manual install, MCP server starts cleanly and the `Cannot find module` error stops. **However, every `/plugin install claude-mem` undoes this** — the marketplace re-pulls the broken bundle.

## Environment

- **Claude Code:** v2.1.138 (Windows x64)
- **OS:** Windows 11 Pro 24H2
- **Bun:** v1.3.12 (Windows x64)
[...]

PR #2610, opened 2026-05-22
fix(build): inline zod into worker / server-beta / context-generator bundles (closes #2437)

### LINKED ISSUE #2437: [Bug] 13.x worker fails with "Cannot find module 'zod/v3'" — package-lock.json missing zod entry
## Summary

The `worker-service.cjs` bundle in v13.x dynamically `require()`s `zod/v3`, `zod/v4`, and `zod/v4-mini` (esbuild leaves them as externals), but the published `package-lock.json` does not list `zod`, so `bun install --frozen-lockfile` / `npm ci` skip installing it. The plugin fails at runtime with `Cannot find module 'zod/v3'` on every UserPromptSubmit hook.

## Affected versions

- v13.1.0 (confirmed)
- v13.2.0 (confirmed)
- v10.5.5 is unaffected — that bundle fully inlines zod (no external requires).

## Evidence

- `13.2.0/package.json` declares `"zod": "^4.3.6"` in `dependencies` ✅
- `13.2.0/package-lock.json` contains **zero** `"zod"` entries (`grep -c '"zod"' package-lock.json` → 0)
- `13.2.0/scripts/worker-service.cjs` contains:
  - `require("zod/v3")` × 4
  - `require("zod/v4-mini")` × 2
  - `require("zod/v4")` × 1
- `13.2.0/node_modules/` is missing `zod/` entirely (only native tree-sitter packages present)

## Root cause (best guess)

The release pipeline regenerates the bundle but doesn't refresh `package-lock.json` to match the new `package.json`. Any consumer that does a clean install ends up with no `zod`.

## Reproduction

\`\`\`bash
cd ~/.claude/plugins/cache/thedotmack/claude-mem/13.2.0
grep -c '"zod"' package-lock.json     # → 0
ls node_modules/zod 2>&1               # → does not exist
node -e "require('zod/v3')"            # → Cannot find module 'zod/v3'
\`\`\`

## Workaround

\`\`\`bash
cd ~/.claude/plugins/cache/thedotmack/claude-mem/<version>
bun add zod@^4.3.6
\`\`\`

But this needs to be redone on every auto-update.

## Suggested fix

Re-run \`bun install\` (or \`npm install\`) **before** publishing so \`package-lock.json\` includes \`zod\` and all transitive deps. Optionally add a CI step that asserts \`package-lock.json\` matches \`package.json\`.

## Environment

- OS: Linux (WSL2 Ubuntu)
- Node: v24.14.0
- Bun: 1.3.10
- Installed via Claude Code plugin manager

YOMXXX:
## Summary

Closes #2437 — the recurring `Cannot find module 'zod/v3'` hook failure that has shipped on 13.0, 13.1, 13.2, and 13.3.

Removes `'zod'` from the esbuild `external` list of three Node-platform bundles so zod is inlined the same way it already is for `mcp-server.cjs`. The remaining `external` entries (native bindings, large optional embedders, `bun:sqlite`) all stay external for their original reasons.

## Why inline (and not just better-install)

`#2597` fixes the *build-side* half of #2407 by populating `plugin/node_modules` during the build. That helps the dev-workflow and marketplace-sync paths but leaves the npm-tarball half ([#2407](https://github.com/thedotmack/claude-mem/issues/2407) "out of scope") still relying on a post-install heal step. Every report on #2437 (across macOS, Windows, WSL2, corporate Zscaler/Artifactory networks) traces back to a cache directory that shipped without `node_modules/zod`. The manual workarounds — `bun install`, `npm install zod --no-save --ignore-scripts`, autoUpdate disabled — all get wiped on every plugin upgrade.

Inlining zod removes the failure mode from every install path at once. The MCP server bundle has already been doing this since #1645, with an esbuild-output assertion (`build-hooks.js:269-274`) that fails the build if `require('zod')` ever leaks back in. This PR brings the remaining three Node bundles to the same self-contained shape — there's no design reason zod was external for some bundles and inlined for others.

zod is pure JS and reasonable to inline:
- v4 with v3/v4/v4-mini subpaths all reachable from the worker
- no native bindings (unlike `onnxruntime-node`, `cohere-ai`, `@chroma-core/default-embed`, which stay external)
- no startup-time dynamic-loading expectations

## What changed

`scripts/build-hooks.js` — three esbuild config blocks:

| Target | Before | After |
|---|---|---|
| `worker-service.cjs` | `external: ['bun:sqlite', 'zod', 'cohere-ai', 'ollama', '@chroma-core/default-embed', 'onnxruntime-node']` | `'zod'` removed; native / large-optional deps unchanged |
| `server-beta-service.cjs` | `external: ['bun:sqlite', 'zod']` | only `'bun:sqlite'` remains |
| `context-generator.cjs` | `external: ['bun:sqlite', 'zod']` | only `'bun:sqlite'` remains. Generator doesn't currently import zod so the bundle is byte-identical — defensive consistency so any future zod import doesn't silently regress. |

## Bundle size impact

Node-platform CJS, minified:

```
worker-service.cjs      3877 KB -> 4206 KB   (+329 KB, +8.5%)
server-beta-service.cjs 1837 KB -> 2169 KB   (+332 KB, +18%)
context-generator.cjs     90 KB ->   90 KB   (unchanged)
mcp-server.cjs           328 KB ->  328 KB   (unchanged, already inlined)
```

The increase is zod itself (the full library including the three subpaths that the worker pulls in). Worth the trade for closing a P0 that's regressed across four consecutive releases with growing reaction counts on the issue.

## Verification

End-to-end check that the #2437 symptom is gone:

```bash
$ mv plugin/node_modules/zod /tmp/zod-backup
$ cd plugin/scripts && node -e "require('./worker-service.cjs')"
# Before this PR:
# Error: Cannot find module 'zod/v3' from '.../scripts/worker-service.cjs'
# After this PR:
# Error: Cannot find module 'bun:sqlite'   <- unrelated, worker runs on Bun, not Node
```

The bundle is now self-resolving for zod. `Cannot find module 'bun:sqlite'` is the next failure mode and is unrelated to #2437 (worker is launched by `bun-runner.js` under Bun in production, never by Node directly).

Bundle-internal symbol check confirms the inline:

```bash
$ grep -c 'ZodError\|ZodObject\|ZodString' plugin/scripts/worker-service.cjs
7
$ grep -c 'ZodError\|ZodObject\|ZodString' plugin/scripts/server-beta-service.cjs
3
$ grep -E "require\(['\"]zod(/[^'\"]*)?['\"]\)" plugin/scripts/worker-service.cjs | wc -l
0
```

`npm run build` passes cleanly; mcp-server.cjs's existing zod-external assertion continues to pass (no behavior change for that bundle).

## Out of scope

- Removing zod from `plugin/package.json` dependencies. Kept as a defensive declaration in case downstream tooling or future code paths still resolve it via `node_modules`. Removing it is a separate, larger cleanup.
- Same treatment for any future bundles. The assertion pattern from `build-hooks.js:269-274` could be generalized so every Node bundle is checked; left for a follow-up.

## Relationship to #2597

Independent. #2597 makes `plugin/node_modules` populated during build (helping dev + marketplace-sync); this PR makes the hook-invoked bundles not need `plugin/node_modules/zod` in the first place. Both can land in any order; together they close #2437 and most of #2407.


YOMXXX:
Review note from a release-engineering perspective: this looks like the right root fix for the recurring `zod/v3` fresh-install failure. Inlining pure-JS `zod` removes the marketplace/cache/node_modules dependency entirely, which is more robust than trying to install dependencies during the first hook invocation.

Before merge, I would tighten two checks:

- Add the same no-external-`zod` assertion after the context-generator build, or narrow the PR title/body if context-generator does not actually import zod today.
- Verify from a clean plugin cache with no `node_modules/zod` that `worker-service.cjs`, `server-beta-service.cjs`, and the hook path start without resolving external `zod/*` subpaths.

If this lands, it should be the primary fix for `#2437` / `#2637` / `#2640`; first-hook auto-install can remain only as defense-in-depth if still needed.

YOMXXX:
Implemented the remaining defensive guardrail from the review: the context-generator build now also calls `assertNoExternalZodRequire` after emitting `plugin/scripts/context-generator.cjs`.

Verification run locally:

```
npm run build
```

Build passed, and no generated bundle artifacts changed in the working tree, so the PR delta remains limited to the build-script assertion.

YOMXXX:
Closing — upstream went the opposite direction: `scripts/build-hooks.js:448` keeps `zod` in the `external` list. #2437 was resolved through other means (likely the package-lock / tarball fixes). Inlining zod is no longer the path forward. Thanks!

PR #2614, opened 2026-05-22
Add Codex ingestion provider and Windows Chroma launch fix

flipjam:
## Summary
- add Codex as a first-class ingestion provider with settings/defaults, route dispatch, and worker health reporting
- fail closed when Codex is selected but the Codex executable is unavailable instead of falling back to another provider
- launch chroma-mcp on Windows via direct uvx.exe resolution instead of cmd.exe /c uvx

## Local verification
- npm run build
- /api/health reports provider codex on the local worker
- /api/chroma/status?deep=true returns healthy with semantic round-trip success
- pending queue drains after Codex ingestion

## Notes
This is a draft PR opened from a local durability fix. It intentionally contains source-level changes only; generated marketplace bundle files are not included in this branch.

### DIFF
--- a/src/services/sync/ChromaMcpManager.ts
+++ b/src/services/sync/ChromaMcpManager.ts
@@ -105,9 +105,10 @@ export class ChromaMcpManager {
     const spawnEnvironment = this.getSpawnEnv();
     getSupervisor().assertCanSpawn('chroma mcp');
 
-    const isWindows = process.platform === 'win32';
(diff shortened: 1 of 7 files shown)


thedotmack:
Closing this draft: the Windows Chroma launch half landed via #2847 (direct uvx spawn), and the Codex ingestion provider half overlaps the open provider work (#2514/#2764, plan-12 #2785). Re-cut the ingestion piece against current main if still wanted.

PR #2615, opened 2026-05-22
Normalize path separators in isWithin() to prevent tracking loop

Mlekow:
shouldTrackProject() failed to block observer-sessions from being tracked when OBSERVER_SESSIONS_DIR (generated by path.join) used backslashes while the cwd from Claude Code hooks used forward slashes.

Fix by applying path.normalize() to both child and parent before comparison so separators are canonicalized consistently on all platforms.


Mlekow:
Session which observes OBSERVER_SESSIONS_DIR may cause a observation loop, leading to token waste.
High severity, can silently exhaust token budget.

PR #2616, opened 2026-05-22
feat(privacy): opt-in auto-redaction for common secret patterns

YOMXXX:
## Summary

Many users have asked for automatic detection of accidental secret leaks in observations / summaries / tool payloads. This PR adds an **opt-in** regex layer that runs before the existing `stripTags` pipeline, replacing recognized secret tokens with self-closing `<redacted type=\"...\"/>` placeholders.

Default OFF — no behavior change for current users. Users enable it by setting `CLAUDE_MEM_REDACT_ENABLED=true` in `~/.claude-mem/settings.json`.

## Why now / why this shape

The existing `<private>` tag mechanism handles content the user *knows up front* is sensitive. But secrets often appear in **tool output** the user never typed — a `curl` response echoing an API key, a `Bash` command printing `$AWS_SECRET_ACCESS_KEY`, a config file `Read` returning a key. Auto-redaction is the safety net for those cases.

Design highlights:

- **Inline placeholder, not whole-block drop.** `<redacted type=\"openai_key\"/>` self-closing marker replaces only the matched substring; surrounding context (which command, which header) is preserved so search/observation quality stays intact. Contrast with `<private>` which drops the whole block.
- **Custom > built-in priority.** Custom patterns are evaluated before built-ins, so a user can override the built-in detection for a specific token family.
- **Defensive throughout.** Per-invocation caps (200 matches, 1 MB input) prevent ReDoS. Invalid custom regex is logged and skipped without crashing the worker. Cached `getRedactionConfig()` reads settings.json once per 5 s.
- **Single source of detection.** A new `src/utils/redaction.ts` is the only module that knows about secret patterns; the 5 existing `stripTags` call sites each get a 1-line prefix.

### 10 built-in secret patterns

| Name | Catches |
|---|---|
| `aws_access_key` | `AKIA…` (16 trailing alphanumerics) |
| `aws_secret_key` | 40-char base64 anchored to `AWS_SECRET_ACCESS_KEY=…` (shell/yaml only — JSON-style is documented as a known limit) |
| `github_pat` | `ghp_…`, `ghs_…`, `github_pat_…` |
| `openai_key` | `sk-…` (≥20 chars, with `(?!ant-)` lookahead so anthropic_key wins separately) |
| `anthropic_key` | `sk-ant-…` (≥20 chars) |
| `slack_token` | `xoxb-…`, `xoxp-…`, `xoxa-…`, `xoxr-…`, `xoxs-…` |
| `jwt` | `eyJ…[.]eyJ…[.]…` three-segment tokens |
| `private_key_pem` | `-----BEGIN [RSA\|DSA\|EC\|OPENSSH\|PGP\|]PRIVATE KEY-----` blocks |
| `stripe_key` | `sk_live_…`, `pk_live_…`, `rk_test_…` |
| `google_api_key` | `AIza…` (35 trailing chars) |

### Configuration (flat-string SettingsDefaults, matching CLAUDE_MEM_FOLDER_MD_EXCLUDE precedent)

```json
{
  \"CLAUDE_MEM_REDACT_ENABLED\": \"false\",
  \"CLAUDE_MEM_REDACT_DISABLED_BUILTINS\": \"\",
  \"CLAUDE_MEM_REDACT_CUSTOM_PATTERNS\": \"[]\",
  \"CLAUDE_MEM_REDACT_LOG_MATCHES\": \"false\"
}
```

### Wired into 5 existing `stripTags` call sites

| File | Calls wrapped |
|---|---|
| `src/cli/handlers/summarize.ts` | 2 |
| `src/services/worker/http/shared.ts` | 2 |
| `src/services/worker/http/routes/SessionRoutes.ts` | 2 |
| `src/server/generation/processGeneratedResponse.ts` | 2 |
| `src/server/generation/providers/shared/prompt-builder.ts` | 1 |

Each site goes from `stripFn(input)` to `stripFn(redactSensitive(input, getRedactionConfig()).redacted)`. No other logic touched.

### Observer prompt addendum

`src/sdk/prompts.ts` — `buildObservationPrompt()` now tells the observer LLM:

> If a `<parameters>` or `<outcome>` block above contains a `<redacted type=\"...\" />` marker, that field was a recognized secret pattern and was removed before storage. Treat it as a placeholder; do not infer the literal value.

(This is the parallel of PR #2602's `<elided/>` hint. That PR isn't merged yet, so this hint stands alone in the file for now; once #2602 lands the two hints become sibling paragraphs.)

### Docs

- `docs/public/usage/auto-redaction.mdx` (new, ~100 lines) — full user guide with the 10 patterns table, enable/disable instructions, custom-pattern format, limits, and a `<private>` vs auto-redaction comparison.
- `docs/public/usage/private-tags.mdx` — cross-link footer.
- `docs/public/introduction.mdx` — privacy bullet updated.
- `docs/public/docs.json` — nav entry added.

Design spec + implementation plan are checked in under `docs/superpowers/specs/` and `docs/superpowers/plans/` for future reference.

## Tests

- `tests/utils/redaction.test.ts` — 24 unit cases: 8-row `it.each` table for the patterns (each fixture verified to fire *only* its own pattern via `expect(Object.keys(counts)).toEqual([name])`), PEM block, anchored aws_secret_key, near-miss negatives, custom patterns, invalid-regex tolerance, empty-name tolerance, ReDoS caps, idempotency, config loader.
- `tests/integration/redaction-wired.test.ts` — 1 end-to-end case: real `redactSensitive` → real `stripMemoryTagsFromPrompt` produces a surviving `<redacted type=\"openai_key\"/>` placeholder.

`bun test tests/utils/redaction.test.ts tests/integration/redaction-wired.test.ts` → **25 pass / 0 fail**.

Full `bun test` against this branch: **+25 pass vs `main` baseline, 0 regressions**.

`npx tsc --noEmit 2>&1 | grep redact` → empty.

`npm run build` → all bundles compile cleanly. `grep redacted plugin/scripts/worker-service.cjs` → 2 occurrences (inlined). Bundles included in this commit.

## Known limitations (documented in auto-redaction.mdx)

- `aws_secret_key` lookbehind only anchors shell (`KEY=val`) and yaml (`KEY: val`) styles. JSON config (`\"KEY\": \"val\"`) is not matched because the lookbehind expects `=` or `:` directly after the key name. Users with JSON config write a custom pattern.
- `MAX_INPUT_BYTES` is measured by `string.length` (UTF-16 code units), not literal bytes. Commented inline.
- In multi-tenant `server-beta` deployments, redaction config is currently read from the worker's local `~/.claude-mem/settings.json` rather than tenant-scoped settings. Acceptable while server-beta isn't widely deployed yet; commented as a `TODO(server-beta config scope)` in the source.
- Built-ins skip Azure / GCP service-account JSON / private IDC tokens. Use custom patterns.

## Risk

- **Default off**, so the release notes line is short: \"Optional auto-redaction of common secret patterns. Opt in via `CLAUDE_MEM_REDACT_ENABLED=true`.\"
- **No schema changes**, no new dependencies, no external API changes.
- The 5 wrap sites are mechanical: each is `stripFn(redactSensitive(input, getRedactionConfig()).redacted)`. With redaction disabled, `redactSensitive` returns the input unchanged in O(1), so the cost is one cached-config lookup (~1 µs) per call site.

## Note on commit history

This branch was built as a 10-task plan with per-task two-stage review (spec compliance + code quality). The original 20-commit history is preserved locally on `feat/auto-redaction-history-backup` if needed. The PR is squashed to a single commit because GitHub's secret-push-protection flagged two of the **fake test fixtures** as real Slack/Stripe tokens, and the cleanest path was to:

1. Squash the history.
2. Rewrite those two fixtures as compile-time string concatenation (e.g., `'xox' + 'b-...'`) so the scanner won't flag future touches either.

Both changes are in this commit. Functionally identical; only the string-literal layout changed.


YOMXXX:
**Heads-up: `Closes #2437` in this PR was a mistaken reference — just removed it from the description and the title.**

#2437 is the open *"13.x worker fails with 'Cannot find module zod/v3'"* bug (missing zod entry in `bun.lock`/`package-lock.json` for the cache dir). That's a build/release packaging issue and is completely unrelated to auto-redaction. The fixes for #2437 are tracked in #2531 (bundle `plugin/node_modules`) and #2597 (run `npm install` during build) — please don't let this PR auto-close it.

`closingIssuesReferences` should now be empty after the edit. Sorry for the noise.

YOMXXX:
Implemented the remaining summary-prompt hint from the latest review summary.

Change made:
- shared the redacted-marker guidance between observation and summary prompts
- summary prompts now tell the model that `<redacted type='...'/>` is a placeholder for a removed credential
- the guidance explicitly says not to infer the literal value and not to copy the marker itself into generated memory content
- added a focused `buildSummaryPrompt` test covering the hint
- rebuilt `plugin/scripts/worker-service.cjs`

Verification:

```
npm run build
npm exec -- tsx -e "<summary prompt sanity check>"
```

Both passed locally. I did not push the two unrelated local dedup-doc commits that were ahead on my local `feat/auto-redaction-v2`; this update was made from `origin/feat/auto-redaction-v2` and pushed back as PR-only commits.

YOMXXX:
Hi @thedotmack — friendly check before I invest time in a rebase.

This PR is sizeable (~22 files / +2457 lines) and currently has conflicts with `main`. I confirmed `src/` has no `redact` code upstream yet, so this isn't superseded — but it's still a real design decision on your side whether opt-in auto-redaction is a direction you want `claude-mem` to take.

Before I spend the time refreshing it, could you give me a quick read?

- 👍 Interested → I'll rebase and stay on top of review
- 👎 Not aligned → totally fine, just say so and I'll close it; no rebase wasted

Either is genuinely fine — I'd rather you close it than feel obligated to merge.

thedotmack:

> opt-in secret-redaction layer; product decision

thedotmack:
Second pass — routing holds at **#2785 (plan-12, Provider & Extensibility Roadmap)**: an opt-in secret-redaction layer is a new privacy capability rather than a defect in any of the architectural clusters, so it sits on the roadmap master pending a product decision. Nothing open duplicates it and no merged work adds redaction — the `<private>` tag path is still the only lever on main. The design reasoning (inline self-closing placeholder rather than dropping the whole block, running before `stripTags`) is the right shape; the open question is scope, not correctness.

PR #2627, opened 2026-05-23
feat: auto dedup-folding (S3) — opt-in fold of identical tool calls within a sliding window

YOMXXX:
## Summary

Opt-in dedup-folding for the observation pipeline. When the user polls (`while sleep 5; do ls; done`) or hits a retry loop, identical tool calls within a sliding window now collapse into a single SDK round-trip instead of N. The pending row's `fold_count` is bumped; when it finally drains, `buildObservationPrompt` emits `<repetition>This tool call was repeated N times in a Xs window.</repetition>` so the observer can surface the repetition in `<facts>`.

**Default OFF.** No behavior change unless `CLAUDE_MEM_DEDUP_FOLD_ENABLED=true`.

Design: `docs/superpowers/specs/2026-05-23-dedup-folding-design.md`
Plan: `docs/superpowers/plans/2026-05-23-dedup-folding.md`
User doc: `docs/public/usage/dedup-folding.mdx`

## Settings

```json
{
  "CLAUDE_MEM_DEDUP_FOLD_ENABLED": "false",
  "CLAUDE_MEM_DEDUP_FOLD_WINDOW_SECONDS": "30",
  "CLAUDE_MEM_DEDUP_FOLD_DISABLED_TOOLS": ""
}
```

## Architecture

```
queueObservation
  └─> shouldFold(obs, sessionDbId, config, store)  // pure decision
        ├─> {fold:true, foldOntoRowId} → bumpFoldCount, return {folded:true}
        └─> {fold:false, foldKey?}     → enqueue with foldKey, continue
  └─> caller (http/shared.ts) skips ensureGeneratorRunning + broadcast on folded
```

- **Fold key**: `sha256(tool_name + canonical(tool_input) + cwd + agent_id).slice(0,16)`
- **Window**: sliding from first row, default 30s, configurable [1, 3600]
- **Engine-agnostic interface**: `ObservationQueueEngine.findFoldCandidate?/bumpFoldCount?` are optional. SQLite implements; BullMQ degrades gracefully with a one-shot `[DEDUP]` warning.
- **Schema migration v35** adds `pending_messages.fold_key TEXT`, `fold_count INTEGER NOT NULL DEFAULT 1`, and index `idx_pending_fold(session_db_id, fold_key, created_at_epoch)`.
- **`<repetition>` injection**: `buildObservationPrompt(obs, opts?: { windowSeconds?: number })`. SDK module stays framework-agnostic (worker passes windowSeconds from `getDedupFoldConfig()`).

## What does NOT fold

- Different session / cwd / agent_id (each forks a fresh fold group)
- Outside the configured window
- Tools listed in `CLAUDE_MEM_DEDUP_FOLD_DISABLED_TOOLS`
- Anything when feature disabled (zero SQL probes, no log noise)

## Known caveats

- `Bash("npm test")` × N flapping → folds onto first (input identical, output varies). Documented; user can add `Bash` to `DISABLED_TOOLS`.
- BullMQ engine: fold no-ops with one-shot warning. SQLite (default) is the supported path.

## Tests

41 new tests covering:
- `computeFoldKey` (snapshot, canonical sort, field sensitivity)
- `loadDedupFoldConfig` (defaults, parsing, clamping, CSV)
- 5s cache + safe-default fallback on settings load failure
- `findFoldCandidate` (window, session isolation, newest-wins)
- `bumpFoldCount` (increment + missing-row throw)
- Migration v35 (columns, index, idempotency, legacy upgrade)
- `shouldFold` decision (4 paths: disabled / blacklist / miss / hit)
- `<repetition>` injection in `buildObservationPrompt`
- SessionManager fold-true path + BullMQ silent-degradation warn (one-shot latch)
- End-to-end: 5× collapse, window crossing, subagent isolation, disabled tools

All S3 tests pass. Full-suite regression: 0 new failures vs baseline (the existing 64 pre-existing failures unchanged).

## Stacked on #2616 (S2 auto-redaction)

This branch is based on `feat/auto-redaction-v2`, which is PR #2616. The 4 commits at the bottom (auto-redaction) will rebase out automatically once #2616 lands. S3 itself is 16 commits starting at `da35af2d`.

## Test plan

- [ ] CI passes (or maintainer confirms baseline failures unchanged)
- [ ] `bun test tests/integration/dedup-fold-wired.test.ts` → 5 pass
- [ ] Enable in a test profile: `CLAUDE_MEM_DEDUP_FOLD_ENABLED=true`, run a polling shell loop (`while sleep 1; do ls; done` ~10 cycles), inspect `pending_messages.fold_count > 1` and verify the SDK observation surfaces the repetition
- [ ] With feature disabled (default), confirm no behavior change

## Follow-ups (out of scope, not blocking)

1. SessionStore mirrors MigrationRunner v35 inline because `DatabaseManager` doesn't call `runAllMigrations()`. Should be unified in a separate cleanup.
2. `safeParseJson` exists in 3 places (SessionManager.ts inline, CorpusBuilder.ts, DataRoutes.ts). Should consolidate into `src/utils/json-utils.ts`.
3. `bullmqFoldWarned` warn path is covered by a test but a follow-up could add observability metrics.


YOMXXX:
Thanks for the catch — the `status = 'pending'` gap was a real data-loss path. Fixed in 89d67136:

- `findFoldCandidate` SQL now restricts to `status = 'pending'`.
- Added two tests in `tests/services/sqlite/PendingMessageStore.fold.test.ts`:
  - in-flight rows (`status = 'processing'`) are skipped.
  - a pending row is preferred over a more recent processing row with the same fold key.

Re: `getDedupFoldConfig` 5s TTL note — agreed, this is documented safe-default behaviour. The cache exists to keep the hot path cheap, and if settings load throws on cold start the cache populates with `{ enabled: false, ... }` and the feature stays disabled for 5s, which is exactly the "fail open to don't fold" behaviour we want. Leaving as-is.

YOMXXX:
Addressed all 4 issues in 52c15f50:

**Issue 1 (Important — generator stalled on fold)**: moved `ensureGeneratorRunning?.(sessionDbId, 'observation')` *before* the folded early-return in `shared.ts:ingestObservation`. It's idempotent (`startGeneratorWithProvider` short-circuits on existing `generatorPromise`), so a polling loop that consistently folds will still restart a crashed generator.

**Issue 2 (prompt off-by-one)**: changed "repeated N times" to "occurred N times" in `buildObservationPrompt`. `fold_count` counts total occurrences, so "occurred" is unambiguous. Synced docs and test.

**Issue 3 (RETURNING)**: `bumpFoldCount` collapsed to a single `UPDATE … RETURNING fold_count` statement — atomic, halves the round-trips.

**Issue 4 (windowSeconds race)**: persisted `fold_window_seconds` at enqueue time so the prompt reads the value that was actually in effect when the fold decision was made, not the live config at drain time.

- Migration v36 + SessionStore mirror: nullable `fold_window_seconds INTEGER` on `pending_messages`.
- `PendingMessageStore.enqueue` gains 5th arg `foldWindowSeconds`; `toPendingMessage` projects it.
- `ObservationQueueEngine`/`SqliteObservationQueueEngine`/`BullMqObservationQueueEngine` plumb through.
- `SessionManager.queueObservation` passes `dedupConfig.windowSeconds` (captured once).
- `PendingMessage` adds `fold_window_seconds`; all 3 Providers read `message.fold_window_seconds ?? getDedupFoldConfig().windowSeconds` (live config is the fallback for legacy rows where the column is NULL).

Tests added for v36 marker + column, enqueue persistence, and `toPendingMessage` projection. Full fold suite (44 tests across 6 files) passes; build clean.

YOMXXX:
Addressed in b97d2b99 — fold key now computed from the RAW (pre-redaction) `payload.toolInput` so S2 (auto-redaction) and S3 (dedup-fold) can both be enabled without collapsing distinct secret-bearing calls onto one key.

**Flow**:
1. `ingestObservation` computes `preComputedFoldKey = computeFoldKey({tool_name, tool_input: payload.toolInput, cwd, agent_id})` **before** redaction runs (only when `dedupConfig.enabled`).
2. Threads it through `ObservationData.foldKey` to `SessionManager.queueObservation`.
3. `shouldFold` gains a 5th `preComputedFoldKey?` argument. If provided, it skips its own `computeFoldKey` hash and uses the caller's value; otherwise it falls back to hashing `obs.tool_input` — preserving single-feature usage and any non-`shared.ts` caller.

The SHA-256 fold_key is non-reversible, so persisting a raw-derived hash to `pending_messages.fold_key` doesn't leak the original secret.

**Tests added** in `tests/services/worker/dedup-fold.test.ts`:
- `preComputedFoldKey` takes precedence: the value passed to `findFoldCandidate` matches the caller's key (not a hash of `obs.tool_input`).
- Fallback to hashing `obs.tool_input` when omitted (backwards compat).
- **S2+S3 coexistence pin**: two distinct raw inputs (`Bearer tokenA` vs `Bearer tokenB`) produce distinct `fold_key`s even when their redacted forms are identical (`Bearer <redacted/>`).

Full fold suite: 50/50 pass; build clean.

YOMXXX:
All Greptile suggestions from the latest review have been addressed:

- **Round 3** (`d45379b1`): skipped `getDedupFoldConfig()` on non-folded path + added warn log when engine lacks `bumpFoldCount`
- **Round 2** (`52c15f50`): generator restart, prompt wording, atomic RETURNING, windowSeconds race
- **Round 1** (`89d67136`): in-flight row exclusion + S2/S3 coexistence (`b97d2b99`)

Full fold suite: 50/50 passing, build clean. Ready for review when you have a moment — no rush at all 🙂

YOMXXX:
Hi @thedotmack — friendly check before I invest in a rebase here.

This is the largest PR in the queue (~42 files / +5654 lines) and has conflicts with `main`. I checked upstream for any equivalent tool-call dedup-folding and found none (the existing `deduplicateObservations` and the `[DEDUP]` content-hash skip are doing different things), so this isn't superseded — but it's a meaningful design surface (new schema migration, fold semantics, `<repetition>` prompt injection) that deserves an explicit yes/no from you before I refresh it.

Before I spend time rebasing 5k+ lines, could you give me a quick read?

- 👍 Direction looks right → I'll rebase and shepherd through review
- 🤔 Like the idea but want it scoped down → happy to split into smaller PRs (e.g. just the SQL fold + bump, no `<repetition>` injection yet)
- 👎 Not aligned → I'll close it, no rebase wasted

The design doc and plan are linked in the PR body in case those help the call.

thedotmack:
Closing as superseded by the newer opt-in dedup design in #3063, which has a narrower and more reviewable integration surface.

Issue #2375, opened 2026-05-08
Bug: ANTHROPIC_BASE_URL leaks from process.env, triggers OAuth skip, breaks ANTHROPIC_AUTH_TOKEN auth


ONEGAYI:
## Bug Description

When using a third-party proxy (e.g., BigModel) via `ANTHROPIC_BASE_URL` + `ANTHROPIC_AUTH_TOKEN`, the SDK subprocess fails to authenticate and returns `"Not logged in · Please run /login"`.

## Root Cause

`buildIsolatedEnvWithFreshOAuth()` has an asymmetry between credential leakage and blocking:

1. **`ANTHROPIC_BASE_URL` is NOT in `BLOCKED_ENV_VARS`** → leaks from `process.env` into `isolatedEnv`
2. **`ANTHROPIC_AUTH_TOKEN` IS in `BLOCKED_ENV_VARS`** → blocked, absent from `isolatedEnv`
3. **OAuth skip check** includes `isolatedEnv.ANTHROPIC_BASE_URL` → leaked BASE_URL triggers the skip path
4. **Result**: subprocess gets the proxy URL but no credentials → `"Not logged in"`

```
Subprocess environment:
  ANTHROPIC_BASE_URL       = https://open.bigmodel.cn/api/anthropic  ✅ (leaked in)
  ANTHROPIC_AUTH_TOKEN     = <missing>                                ❌ (blocked by BLOCKED_ENV_VARS)
  ANTHROPIC_API_KEY        = <not set>                                ❌
  CLAUDE_CODE_OAUTH_TOKEN  = <missing>                                ❌ (OAuth was skipped)
```

## Introducing Commits

Two commits by the same author on the same day (2026-05-04) combine to produce this bug:

| Commit | Effect | Problem |
|--------|--------|---------|
| `5edf1557` | Added `ANTHROPIC_AUTH_TOKEN` to `BLOCKED_ENV_VARS` | Blocks the proxy auth token |
| `a122d34e` | Added `isolatedEnv.ANTHROPIC_BASE_URL` to OAuth skip check | Allows leaked BASE_URL to trigger skip |

[...]

PR #2629, opened 2026-05-23
fix(worker): strip Bedrock/Vertex routing env from worker subprocess

### LINKED ISSUE #2357: Worker leaks `CLAUDE_CODE_EFFORT_LEVEL` env to Anthropic API → HTTP 400 + retry storm on Haiku
# Worker leaks `CLAUDE_CODE_EFFORT_LEVEL` env to Anthropic API → HTTP 400 + retry storm on Haiku

## Summary

When the host Claude Code CLI sets `CLAUDE_CODE_EFFORT_LEVEL` (or `CLAUDE_CODE_ALWAYS_ENABLE_EFFORT`) in its `~/.claude/settings.json` `env` block, the claude-mem worker subprocess inherits these vars. The worker then includes the `effort` parameter on its Anthropic API requests. Models that don't support `effort` (Haiku 4.5, Sonnet 4.5 and earlier) reject with HTTP 400, and the worker's retry logic loops indefinitely without a circuit-breaker for this permanent-error class.

Within a couple of hours of normal Claude Code use, retry counter climbed past **#1874** and the daily log accumulated **115+ identical errors**. Worker daemon and chroma-mcp processes also accumulated significant CPU/RAM (50+ minutes CPU on worker, 60+ minutes on chroma, ~673 MB RSS).

## Environment

| | |
|---|---|
| claude-mem version | 12.3.8 |
| Claude Code version | 2.1.132 |
| OS | WSL2 Ubuntu (Windows 11 host), kernel 6.6.87.2 |
| Worker runtime | bun 1.3.9 |
| Affected hook events | All — `Setup`, `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `Stop`, `SessionEnd` |

## Reproduction

1. In `~/.claude/settings.json`, add to the `env` block:
   ```json
   "env": {
     "CLAUDE_CODE_EFFORT_LEVEL": "MAX",
     "CLAUDE_CODE_ALWAYS_ENABLE_EFFORT": "true"
   }
   ```
   These are documented Claude Code env vars for setting reasoning effort on the host CLI itself.

2. Restart Claude Code. The `SessionStart` hook in `claude-mem`'s `hooks.json` spawns `worker-service.cjs` as a child process — it inherits the env block.

3. Use Claude Code normally. Each `PostToolUse` event fires the worker to record an observation.

4. Tail `~/.claude-mem/logs/claude-mem-YYYY-MM-DD.log`.

## Actual behavior

Within minutes the log fills with:

```
[INFO ] [SDK   ] [session-2422] ← Response received (64 chars) {promptNumber=28} API Error: 400 This model does not support the effort parameter.
[WARN ] [PARSER] [session-2422] SDK returned non-XML response; marking messages as failed for retry (#1874)
```

- Retries climb without bound (we hit #1874 in one session).
- Worker burns CPU on the retry loop.
- `Stop` and `SessionEnd` hooks block while the queue drains, sometimes for several minutes.
- Multiple `mcp-server.cjs` child processes accumulate in `Tl` (stopped) state.

## Expected behavior

Any one of:

- **(preferred)** Worker subprocess does **not** inherit `CLAUDE_CODE_*` env vars from the host CLI. They're host-scoped configuration, not part of the plugin's own contract.
- Worker only sets `effort` on the API request when the configured summarization model is known to support it (Opus 4.5+, Sonnet 4.6, Mythos Preview).
- `400` responses with body containing `does not support the effort parameter` are treated as **permanent** failures (drop the message, do not retry).

## Root cause

Verified via Anthropic documentation and SDK source: the official `@anthropic-ai/sdk` does **not** auto-read `CLAUDE_CODE_*` env vars. So the path is:

1. User sets `CLAUDE_CODE_EFFORT_LEVEL` in `settings.json` `env` block (legitimate Claude Code use)
2. Claude Code exports it into the env of every subprocess it spawns, including the claude-mem hook commands
3. The hook command (`node ".../bun-runner.js" ".../worker-service.cjs" hook claude-code <action>`) inherits it
4. claude-mem's own code reads `process.env.CLAUDE_CODE_EFFORT_LEVEL` and forwards it as `effort` in the Anthropic Messages API request
5. The summarization model (Haiku, by default for cost reasons) doesn't support `effort` → 400
6. PARSER marks message as "failed for retry" — but a 400 with this body is permanent, so the retry loop never makes progress

## Proposed fix

The simplest, lowest-risk fix lives entirely in `hooks/hooks.json`. Prepend `unset CLAUDE_CODE_EFFORT_LEVEL CLAUDE_CODE_ALWAYS_ENABLE_EFFORT;` to the `command` field of every hook, before the `node ...` invocation. Example diff for one entry:

```diff
 {
   "type": "command",
-  "command": "export PATH=...; _R=\"${CLAUDE_PLUGIN_ROOT}\"; ...; node \"$_R/scripts/bun-runner.js\" ..."
+  "command": "export PATH=...; unset CLAUDE_CODE_EFFORT_LEVEL CLAUDE_CODE_ALWAYS_ENABLE_EFFORT; _R=\"${CLAUDE_PLUGIN_ROOT}\"; ...; node \"$_R/scripts/bun-runner.js\" ..."
 }
```

Better long-term fix in `worker-service.cjs` (or wherever the Anthropic SDK is configured): explicitly ignore `CLAUDE_CODE_*` env vars when constructing the request config — these are not the plugin's contract.

## Workaround currently in use

Local Python patcher attached to a user-level `SessionStart` hook in `~/.claude/settings.json`. It re-applies the `unset` patch after every plugin update (idempotent, silent on no-op):

```python
NEEDLE = "unset CLAUDE_CODE_EFFORT_LEVEL CLAUDE_CODE_ALWAYS_ENABLE_EFFORT;"
TARGET = '_R=\\"${CLAUDE_PLUGIN_ROOT}\\";'
# replaces TARGET with NEEDLE + ' ' + TARGET in both
# ~/.claude/plugins/marketplaces/thedotmack/plugin/hooks/hooks.json
# ~/.claude/plugins/cache/thedotmack/claude-mem/<version>/hooks/hooks.json
```

Would happily contribute a PR with the `hooks.json` change if welcome.

### LINKED ISSUE #2620: Worker leaks CLAUDE_CODE_USE_BEDROCK + ANTHROPIC_DEFAULT_*_MODEL → 400 invalid model identifier on every compression
## Summary

When the host Claude Code CLI sets `CLAUDE_CODE_USE_BEDROCK=1` along with `ANTHROPIC_DEFAULT_OPUS_MODEL` / `ANTHROPIC_DEFAULT_SONNET_MODEL` / `ANTHROPIC_DEFAULT_HAIKU_MODEL` (the standard pattern for AWS Bedrock-routed Claude Code), the claude-mem worker subprocess inherits all of them but still uses its own `CLAUDE_MEM_MODEL` default (`claude-haiku-4-5-20251001`) against the OAuth subscription endpoint. Result: **every compression attempt fails with `400 The provided model identifier is invalid` indefinitely**, with no fallback and no surfaced error to the user.

This is the same env-leak pattern as #2357 (closed) and falls under the Worker Env Isolation umbrella in #2381 — but for a different env var pair, and with a different failure mode (model rejection instead of `effort` rejection).

## Environment

| | |
|---|---|
| claude-mem version | 13.3.0 |
| Claude Code version | 2.1.148 |
| OS | macOS 25.4.0 (Darwin arm64) |
| Worker runtime | bun (`~/.bun/bin/bun`) |
| Host CLI route | AWS Bedrock via `cmux` |
| Affected hooks | `PostToolUse`, `Stop` (compression path) |

Relevant inherited env on the worker subprocess:

```
CLAUDE_CODE_USE_BEDROCK=1
CLAUDE_CODE_USE_MANTLE=1
ANTHROPIC_DEFAULT_OPUS_MODEL=us.anthropic.claude-opus-4-7
ANTHROPIC_DEFAULT_SONNET_MODEL=us.anthropic.claude-sonnet-4-6
ANTHROPIC_DEFAULT_HAIKU_MODEL=us.anthropic.claude-haiku-4-5-20251001-v1:0
```

## Reproduction

1. Set up Claude Code to route via AWS Bedrock (`CLAUDE_CODE_USE_BEDROCK=1` + the three `ANTHROPIC_DEFAULT_*_MODEL` vars in shell or `~/.claude/settings.json` env block).
2. Run normal Claude Code sessions — `PostToolUse` enqueues observations, `Stop` triggers summarization.
3. Tail `~/.claude-mem/logs/claude-mem-YYYY-MM-DD.log`.

## Actual behavior

Hundreds of identical errors per minute:

```
[INFO ] [SDK ] [session-X] ← Response received (124 chars) {promptNumber=Y} API Error (claude-haiku-4-5-20251001): 400 The provided model identifier is invalid.. Run --model to...
[WARN ] [PARSER] [session-X] SDK returned non-XML/empty response — ignoring queued batch
```

After the worker is restarted with `CLAUDE_MEM_MODEL=claude-sonnet-4-6` or `claude-sonnet-4-5`, the error message changes slightly (`Try --model to switch ...` instead of `Run --model to`) but **still rejects** — proving the OAuth subscription auth method itself is being applied even though the env signals Bedrock.

The `health` endpoint reports `lastInteraction: null` indefinitely. Pending message queue grows monotonically. No observations or session_summaries are created for any project. From the user's perspective, claude-mem appears to "stop updating" silently.

## Expected behavior

One of:

- **Auto-detect**: when `CLAUDE_CODE_USE_BEDROCK=1` is present in inherited env, switch the worker's auth method to Bedrock and use the `ANTHROPIC_DEFAULT_*_MODEL` values directly.
- **Explicit env isolation** (per #2381 plan): strip `CLAUDE_CODE_USE_BEDROCK`, `CLAUDE_CODE_USE_MANTLE`, `ANTHROPIC_DEFAULT_*_MODEL` from the SDK subprocess via `BLOCKED_ENV_VARS` so the OAuth subscription path receives only its own model parameter.
- **At minimum**: classify HTTP 400 `model identifier is invalid` as an unrecoverable error (not a transient retry), surface to the user, and stop the retry storm.

## Workaround applied

Adding `CLAUDE_MEM_MODEL=us.anthropic.claude-haiku-4-5-20251001-v1:0` (the Bedrock-formatted ID matching `ANTHROPIC_DEFAULT_HAIKU_MODEL`) to `~/.claude-mem/settings.json` and restarting the worker fixed compression immediately:

```diff
 {
-  "CLAUDE_MEM_RUNTIME": "worker"
+  "CLAUDE_MEM_RUNTIME": "worker",
+  "CLAUDE_MEM_MODEL": "us.anthropic.claude-haiku-4-5-20251001-v1:0"
 }
```

Successful 1799-char XML response within seconds of restart, new observations landed in SQLite, zero further 400 errors.

## Why this is silent

- `health` endpoint reports `status: "ok"` despite no interactions ever succeeding. A health probe that flips `degraded` after N consecutive 400s would surface this.
- The retry classifier treats 400 `invalid model identifier` as transient (matches #2381 plan's note about the 400 retry classification fix needed in `ClaudeProvider`).
- Worker stays "up" (`uptime` increments) but produces zero output. Users only notice via stale data after several days.

## Related

- #2381 — Worker Env Isolation (umbrella)
- #2357 — `CLAUDE_CODE_EFFORT_LEVEL` leak (closed; same pattern, different env var)
- #2452 — Hardcoded model defaults (overlapping concern)

## Diagnostic queries used

```sql
-- Confirm compression stopped per project
SELECT project, COUNT(*), datetime(MAX(created_at_epoch)/1000,'unixepoch')
FROM observations GROUP BY project ORDER BY MAX(created_at_epoch) DESC;

-- Confirm sessions completed but produced no observations
SELECT s.project, COUNT(*) FROM sdk_sessions s
LEFT JOIN observations o ON o.memory_session_id = s.memory_session_id
WHERE s.status='completed' AND o.id IS NULL GROUP BY s.project;
```

Happy to test fixes against my Bedrock setup if useful.

mohitpatell:
## Summary
Adds the AWS Bedrock / Google Vertex routing env vars to the worker's blocked-env list so the worker subprocess no longer inherits them from the host Claude Code CLI.

Without this, when the host CLI is configured with `CLAUDE_CODE_USE_BEDROCK=1` plus `ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL`, the worker resolves its `CLAUDE_MEM_MODEL` against an endpoint that rejects it, and every compression fails with `400 The provided model identifier is invalid` indefinitely (no fallback, no surfaced error).

Same env-leak family as the already-closed #2357 — different variables, same fix shape.

## Changes
- `src/shared/EnvManager.ts`: extend `BLOCKED_ENV_VARS` with `CLAUDE_CODE_USE_BEDROCK`, `CLAUDE_CODE_USE_VERTEX`, `CLAUDE_CODE_USE_MANTLE`, `ANTHROPIC_DEFAULT_OPUS_MODEL`, `ANTHROPIC_DEFAULT_SONNET_MODEL`, `ANTHROPIC_DEFAULT_HAIKU_MODEL`. `buildIsolatedEnv()` already filters by this list, so the worker subprocess will no longer inherit any of them from the parent shell or host Claude Code CLI.
- `tests/env-isolation.test.ts`: add a regression test that sets all six vars in `process.env` and asserts each is absent from `buildIsolatedEnv()` output.
- Build artifacts under `plugin/scripts/*.cjs` are not touched — they regenerate from the TS source on release per the existing maintenance flow.

## Test plan
- [x] `bun test tests/env-isolation.test.ts` — 4/4 pass (3 existing #2375 cases + new #2620 case)
- [ ] Confirm worker env no longer contains the listed vars when host CLI has them set
- [ ] Compression succeeds against the default OAuth endpoint while host CLI uses Bedrock
- [ ] Existing env-isolation tests still pass

Closes #2620

### DIFF
--- a/src/shared/EnvManager.ts
+++ b/src/shared/EnvManager.ts
@@ -34,6 +34,19 @@ const BLOCKED_ENV_VARS = [
   'CLAUDE_CODE_OAUTH_TOKEN', // Issue #2215: prevent stale parent-process token from leaking into
                              // isolated env. The fresh token is read from the keychain at spawn
                              // time by buildIsolatedEnvWithFreshOAuth().
+  // Issue #2620: routing-mode flags + model overrides leak from the host Claude
+  // Code CLI (Bedrock/Vertex/Mantle setups) and reroute the worker's
+  // CLAUDE_MEM_MODEL onto an endpoint that does not accept it, producing
+  // `400 The provided model identifier is invalid` on every compression
+  // indefinitely. The worker uses its own OAuth subscription endpoint, so
+  // these host-side routing hints must be stripped. Same env-leak family as
+  // the CLAUDE_CODE_EFFORT_LEVEL leak fixed in #2357.
+  'CLAUDE_CODE_USE_BEDROCK',
+  'CLAUDE_CODE_USE_VERTEX',
+  'CLAUDE_CODE_USE_MANTLE',
+  'ANTHROPIC_DEFAULT_OPUS_MODEL',
+  'ANTHROPIC_DEFAULT_SONNET_MODEL',
+  'ANTHROPIC_DEFAULT_HAIKU_MODEL',
 ];
 
 export interface ClaudeMemEnv {
--- a/tests/env-isolation.test.ts
+++ b/tests/env-isolation.test.ts
@@ -25,10 +25,26 @@ const TEST_DATA_DIR = fs.mkdtempSync(join(tmpdir(), 'claude-mem-env-isolation-')
 const TEST_ENV_FILE = join(TEST_DATA_DIR, '.env');
 const ORIGINAL_ENV_FILE = process.env.CLAUDE_MEM_ENV_FILE;
 
-const ORIGINAL_BASE_URL = process.env.ANTHROPIC_BASE_URL;
-const ORIGINAL_API_KEY = process.env.ANTHROPIC_API_KEY;
-const ORIGINAL_AUTH_TOKEN = process.env.ANTHROPIC_AUTH_TOKEN;
-const ORIGINAL_OAUTH_TOKEN = process.env.CLAUDE_CODE_OAUTH_TOKEN;
+// Snapshot every env var the suite mutates so afterEach can restore the
+// runner's original environment. A Bedrock/Vertex-configured dev machine may
+// have any of these set going in — the suite must hand them back unchanged.
+const SNAPSHOTTED_VARS = [
+  'ANTHROPIC_BASE_URL',
+  'ANTHROPIC_API_KEY',
+  'ANTHROPIC_AUTH_TOKEN',
+  'CLAUDE_CODE_OAUTH_TOKEN',
+  // Issue #2620: host-side Bedrock/Vertex/Mantle routing flags + model overrides.
+  'CLAUDE_CODE_USE_BEDROCK',
+  'CLAUDE_CODE_USE_VERTEX',
+  'CLAUDE_CODE_USE_MANTLE',
+  'ANTHROPIC_DEFAULT_OPUS_MODEL',
+  'ANTHROPIC_DEFAULT_SONNET_MODEL',
+  'ANTHROPIC_DEFAULT_HAIKU_MODEL',
+] as const;
+
+const ORIGINAL_ENV: Record<string, string | undefined> = Object.fromEntries(
+  SNAPSHOTTED_VARS.map((name) => [name, process.env[name]]),
+);
 
 function clearEnvFile(): void {
   if (fs.existsSync(TEST_ENV_FILE)) {
@@ -36,33 +52,20 @@ function clearEnvFile(): void {
   }
 }
 
-function clearAnthropicEnv(): void {
-  delete process.env.ANTHROPIC_BASE_URL;
-  delete process.env.ANTHROPIC_API_KEY;
-  delete process.env.ANTHROPIC_AUTH_TOKEN;
-  delete process.env.CLAUDE_CODE_OAUTH_TOKEN;
+function clearSnapshottedEnv(): void {
+  for (const name of SNAPSHOTTED_VARS) {
+    delete process.env[name];
+  }
 }
 
 function restoreOriginalEnv(): void {
-  if (ORIGINAL_BASE_URL === undefined) {
-    delete process.env.ANTHROPIC_BASE_URL;
-  } else {
-    process.env.ANTHROPIC_BASE_URL = ORIGINAL_BASE_URL;
-  }
-  if (ORIGINAL_API_KEY === undefined) {
-    delete process.env.ANTHROPIC_API_KEY;
-  } else {
-    process.env.ANTHROPIC_API_KEY = ORIGINAL_API_KEY;
-  }
-  if (ORIGINAL_AUTH_TOKEN === undefined) {
-    delete process.env.ANTHROPIC_AUTH_TOKEN;
-  } else {
-    process.env.ANTHROPIC_AUTH_TOKEN = ORIGINAL_AUTH_TOKEN;
-  }
-  if (ORIGINAL_OAUTH_TOKEN === undefined) {
-    delete process.env.CLAUDE_CODE_OAUTH_TOKEN;
-  } else {
-    process.env.CLAUDE_CODE_OAUTH_TOKEN = ORIGINAL_OAUTH_TOKEN;
+  for (const name of SNAPSHOTTED_VARS) {
+    const original = ORIGINAL_ENV[name];
+    if (original === undefined) {
+      delete process.env[name];
+    } else {
+      process.env[name] = original;
+    }
   }
 }
 
@@ -84,7 +87,7 @@ describe('Issue #2375: ANTHROPIC_BASE_URL env-var isolation', () => {
 
   beforeEach(() => {
     clearEnvFile();
-    clearAnthropicEnv();
+    clearSnapshottedEnv();
   });
 
   afterEach(() => {
@@ -111,6 +114,11 @@ describe('Issue #2375: ANTHROPIC_BASE_URL env-var isolation', () => {
     // worker subprocess, which uses its own OAuth subscription endpoint and
     // would otherwise route CLAUDE_MEM_MODEL against an endpoint that rejects
     // it (`400 The provided model identifier is invalid`).
+    //
+    // Cleanup of the six vars below is handled by afterEach →
+    // restoreOriginalEnv(), which restores any pre-existing values from
+    // ORIGINAL_ENV rather than unconditionally deleting them. Required so
+    // the suite is safe to run on a Bedrock-configured dev machine.
     process.env.CLAUDE_CODE_USE_BEDROCK = '1';
     process.env.CLAUDE_CODE_USE_VERTEX = '1';
     process.env.CLAUDE_CODE_USE_MANTLE = '1';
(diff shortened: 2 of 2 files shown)


YOMXXX:
Review note from a provider/env-isolation perspective: this is the right minimal fix shape for `#2620`. These routing/model variables belong to the host Claude Code process and should not leak into the observer SDK subprocess unless claude-mem explicitly opts into that provider route.

Small non-blocking checks before merge:

- Keep the env-isolation test consistent with the existing snapshot/restore style so the new process.env mutations cannot leak between tests.
- Consider a follow-up for the failure mode itself: `400 invalid model identifier` should be classified as non-retryable/degraded rather than allowing an indefinite retry loop with health still reporting ok.

The blocklist change itself looks appropriately narrow.

mohitpatell:
Thanks @greptile-apps and @YOMXXX for the review.

Pushed 20357a4e addressing the snapshot/restore feedback on `tests/env-isolation.test.ts`:

- Replaced the per-var `ORIGINAL_*` constants with a single `SNAPSHOTTED_VARS` list + `ORIGINAL_ENV` map captured at module load. The six #2620 vars (`CLAUDE_CODE_USE_{BEDROCK,VERTEX,MANTLE}` + `ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL`) are included.
- `clearSnapshottedEnv()` (called from `beforeEach`) and `restoreOriginalEnv()` (called from `afterEach`) both iterate that list, so the existing `afterEach` already handles cleanup — the new test drops its `try/finally` and just sets the vars in the body.
- Verified the suite passes both with the six vars unset and with them pre-set in the shell (`CLAUDE_CODE_USE_BEDROCK=1 ANTHROPIC_DEFAULT_SONNET_MODEL=... bun test tests/env-isolation.test.ts` → 4/4 pass), confirming originals are now restored rather than wiped.

@YOMXXX — agreed on the follow-up around classifying `400 The provided model identifier is invalid` as non-retryable so health stops reporting `ok` during an indefinite loop. Happy to open a separate issue for that; it feels out of scope for this minimal blocklist fix and would touch the worker retry/health surface. Let me know if you'd like it filed under a new issue or appended to #2620.

mohitpatell:
Thanks for the LGTM @greptile-apps ✓

Filed the retry-classification follow-up as #2656 per your suggestion — kept it scoped to the 4xx config-rejection class and the resulting `/health` masking, so it's distinct from the env-isolation surface here. cc @YOMXXX.

thedotmack:
Closing as architecturally outdated. Broadly stripping Bedrock/Vertex routing variables conflicts with the current explicit sanitizer allowlist and supported gateway/provider behavior.

PR #2632, opened 2026-05-23
chore(sqlite): unify worker DB init through MigrationRunner

YOMXXX:
## Why

Two database initialization paths existed and had drifted apart:

| Path | Used by | Schema source | Missing |
|------|---------|---------------|---------|
| `ClaudeMemDatabase` (`src/services/sqlite/Database.ts`) | hooks, CLI | `MigrationRunner.runAllMigrations()` | v26 `generated_by_model` / `relevance_count` |
| `worker/DatabaseManager` (`src/services/worker/DatabaseManager.ts`) | worker process | `new SessionStore(db)` inline migrations | v24 `observation_feedback`, v33 server tables, v34 pending-queue rebuild |

This is the latent bug the dedup-fold PR (#2627) walked into and patched around with an inline self-heal in `SessionStore`. Now that the fold work is on its own branch, this cleans up the underlying split so future schema work doesn't need to author two migrations.

## What

1. **Port v26 to `MigrationRunner`.** `addObservationModelColumns()` was only in `SessionStore`. Lifted verbatim, scheduled between v25 and v27. Fixes the `generated_by_model` / `relevance_count` gap on the hooks path.
2. **Backfill v19 marker.** `SessionStore.repairSessionIdColumnRename()` only writes a `schema_versions` row — added `markRepairSessionIdColumnRename()` to runner so SessionStore is a true no-op on a freshly-migrated DB.
3. **Wire MigrationRunner into worker init.** `worker/DatabaseManager.initialize()` now runs `new MigrationRunner(this.db).runAllMigrations()` before constructing `SessionStore`. Both are idempotent so layering is safe; this just guarantees worker also gets v24/v33/v34.
4. **Sync `schema.sql`.** Added `observations.relevance_count` (the doc was already drifted on this).

## Convergence test

Added to `migration-runner.test.ts`:

```ts
it('SessionStore on top of MigrationRunner does not drift the schema', () => {
  const runner = new MigrationRunner(db);
  runner.runAllMigrations();
  const before = snapshotSchema(db);
  expect(() => new SessionStore(db)).not.toThrow();
  const after = snapshotSchema(db);
  expect(after.columnsByTable.observations).toEqual(before.columnsByTable.observations);
  // ... same for sdk_sessions, pending_messages, session_summaries, versions
});
```

Locks in that "MigrationRunner then SessionStore" produces no diff in any table, column, or `schema_versions` row.

## Test plan

- [x] `bun test tests/services/sqlite/migration-runner.test.ts` — 18 pass (16 existing + 2 new convergence)
- [x] Full suite: no new failures vs. main (`schema-repair`, `session-manager-queue`, `GeminiProvider` failures are pre-existing)

## Out of scope

- Removing `SessionStore` inline migrations entirely (would let us delete ~30 methods). Convergence test makes this safe to do in a follow-up but leaving in for now to keep the diff narrow and the rollback path obvious.
- Worker DB connection PRAGMA settings (`worker/DatabaseManager` skips the `journal_mode = WAL` etc. that `ClaudeMemDatabase` sets, because it passes an open `Database` instance to `SessionStore`). Separate latent issue.

### DIFF
--- a/src/services/sqlite/migrations/runner.ts
+++ b/src/services/sqlite/migrations/runner.ts
@@ -23,12 +23,14 @@ export class MigrationRunner {
     this.ensureDiscoveryTokensColumn();
     this.createPendingMessagesTable();
     this.renameSessionIdColumns();
+    this.markRepairSessionIdColumnRename();
     this.addFailedAtEpochColumn();
     this.addOnUpdateCascadeToForeignKeys();
     this.addObservationContentHashColumn();
     this.addSessionCustomTitleColumn();
     this.createObservationFeedbackTable();
     this.addSessionPlatformSourceColumn();
+    this.addObservationModelColumns();
     this.ensureMergedIntoProjectColumns();
     this.addObservationSubagentColumns();
     this.rebuildPendingMessagesForSelfHealingClaim();
@@ -488,6 +490,10 @@ export class MigrationRunner {
     }
   }
 
+  private markRepairSessionIdColumnRename(): void {
+    this.db.prepare('INSERT OR IGNORE INTO schema_versions (version, applied_at) VALUES (?, ?)').run(19, new Date().toISOString());
+  }
+
   private addFailedAtEpochColumn(): void {
     const applied = this.db.prepare('SELECT version FROM schema_versions WHERE version = ?').get(20) as SchemaVersion | undefined;
     if (applied) return;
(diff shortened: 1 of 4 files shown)


YOMXXX:
Thanks — added in 2ba47258. `snapshotSchema` now asserts `after.tables == before.tables` so any future `SessionStore` inline migration that creates a new table will trip the convergence test before column drift can hide it.

YOMXXX:
Both issues addressed in 89e4fa5a:

**Issue 1 (column drift coverage)**: replaced the 4-table per-column assertions with a single `expect(after.columnsByTable).toEqual(before.columnsByTable)` — now catches column drift in `observation_feedback`, `user_prompts`, and any server-owned table without listing them by name.

**Issue 2 (v26 early-return)**: added the `SELECT version FROM schema_versions WHERE version = 26` short-circuit to `addObservationModelColumns` so it skips the `PRAGMA table_info(observations)` query on every subsequent `runAllMigrations()` call, matching the v20/v21/v23/v24/v27 style.

YOMXXX:
Added in 63700fe1 — `expect(versions).toContain(19)` now pins the marker that `markRepairSessionIdColumnRename` writes on every run.

thedotmack:
Closing in the PR consolidation pass — superseded: MigrationRunner was deleted on main (edc5cf7d5, #3021); SessionStore inline migrations are the single path. The analysis here fed the plan-21 design (#3609); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

PR #2633, opened 2026-05-23
refactor: consolidate parseJsonArrayColumn into utils/json-utils

YOMXXX:
## Why

Two near-identical copies of the same parse-string-or-JSON-array helper had grown in the codebase:

- `src/services/worker/knowledge/CorpusBuilder.ts` — local `safeParseJsonArray()` with a string filter
- `src/services/worker/http/routes/DataRoutes.ts` — inline `safeParseJson()` closure

Both decode a SQLite column that may be a JSON-array string, a stray bare string, `null`, or already an array, and both want a `string[]` back. Same shape, same failure modes — should live in one place.

## What

Added `parseJsonArrayColumn(value: unknown): string[]` to `src/utils/json-utils.ts` and replaced both call sites.

```ts
export function parseJsonArrayColumn(value: unknown): string[] {
  if (Array.isArray(value)) return value.filter((v): v is string => typeof v === 'string');
  if (typeof value !== 'string') return [];
  const trimmed = value.trim();
  if (!trimmed) return [];
  try {
    const parsed = JSON.parse(trimmed);
    return Array.isArray(parsed)
      ? parsed.filter((v): v is string => typeof v === 'string')
      : [];
  } catch {
    return [];
  }
}
```

Notes:
- Kept the `typeof v === 'string'` filter that `CorpusBuilder` had. `DataRoutes` previously skipped this filter — that path was looser; this PR tightens it (intentional: every caller expects `string[]`).
- Dropped `CorpusBuilder`'s old `logger.warn` on parse-failure. The util is pure (no logger dep). Acceptable observability tradeoff since both call sites use the result to populate a UI list, not for routing decisions.

A third copy lives in `SessionManager.safeParseJson()` (PR #2627). Left alone here to keep this PR focused; it'll converge once #2627 lands.

## Test plan

- [x] `tests/json-utils.test.ts` — 7 new tests covering: array passthrough, empty string, `null`, valid JSON array, invalid JSON, non-array JSON, string filtering
- [x] Full suite: no new failures vs. main

### DIFF
--- a/src/services/worker/http/routes/DataRoutes.ts
+++ b/src/services/worker/http/routes/DataRoutes.ts
@@ -4,6 +4,7 @@ import { z } from 'zod';
 import path from 'path';
 import { readFileSync, statSync, existsSync } from 'fs';
 import { logger } from '../../../../utils/logger.js';
+import { parseJsonArrayColumn } from '../../../../utils/json-utils.js';
 import { getPackageRoot, paths } from '../../../../shared/paths.js';
 import { getWorkerPort } from '../../../../shared/worker-utils.js';
 import { PaginationHelper } from '../../PaginationHelper.js';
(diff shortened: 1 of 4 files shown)


YOMXXX:
Thanks — addressed in 06928d47.

Approach: added an optional `onParseError(error, rawValue)` callback to `parseJsonArrayColumn`. Keeps the util pure for `DataRoutes` (which never logged in the first place) and lets `CorpusBuilder` wire its logger back in via a small `parseCorpusColumn(value, column)` helper that emits a `logger.warn` with the column name and a 100-char raw preview — strictly more diagnostic info than the original `safeParseJsonArray` had.

Two tests in `tests/json-utils.test.ts` cover the callback paths (invoked on parse error, not invoked on success).

YOMXXX:
Both issues addressed in 5df4741c:

**Issue 1 (whitespace-only triggers spurious onParseError)**: real bug. Switched the empty-string guard from `value.length === 0` to a `value.trim()` step before both the length check and `JSON.parse`. Whitespace-only inputs (`'   '`, `'\n\t'`) now return `[]` silently without invoking the callback.

**Issue 2 (negative test for valid-JSON-non-array)**: added two pin tests — one verifying `onParseError` is *not* called for valid JSON that parses to a non-array (`'{"a":1}'`, `'42'`, `'"a string"'`), and one for the whitespace cases above. Both lock in the silent-failure contract for future maintainers.

thedotmack:

PR #2635, opened 2026-05-23
fix(server-beta): initialize ModeManager at startup + support CLAUDE_MEM_MODES_DIR

### LINKED ISSUE #2443: server-beta: ModeManager not initialized — generation jobs fail with 'No mode loaded'
## Bug

When running `server-beta-service.cjs` as a standalone service (outside the plugin directory), all generation jobs fail with:

```
{"reason": "No mode loaded. Call loadMode() first.", "classification": "unknown"}
```

## Root Cause

`createServerBetaService()` never calls `ModeManager.getInstance().loadMode('code')`. The generation pipeline's `prompt-builder.ts` (line 135) calls `ModeManager.getInstance().getActiveMode()` which throws because no mode was loaded.

The worker-service initializes it at line 320, but server-beta has no equivalent.

## Additional Issue

`ModeManager` resolves mode files relative to `getPackageRoot()` (which uses `__dirname`). When the CJS bundle runs from a non-plugin directory (e.g., `~/.claude-mem/server-beta/`), the mode files aren't found even if `loadMode()` is called — unless a symlink or copy exists at the expected path.

## Fix Applied (fork)

1. Added `ModeManager.getInstance().loadMode('code')` at the top of `createServerBetaService()`
2. Added `CLAUDE_MEM_MODES_DIR` env var support in `ModeManager` constructor as first path to search

## Suggested Fix

Either:
- Initialize ModeManager in `createServerBetaService()` before building the generation worker
- Or support an env var override for modes directory path (for standalone deployments)

## Environment

- claude-mem v13.2.0 (fork)
- Running as systemd user service, CJS bundle in `~/.claude-mem/server-beta/`
- Generation provider: Gemini (gemini-2.5-flash-lite)

alessandropcostabr:
Fixes #2443.

## Problem
The generation prompt-builder calls `ModeManager.getActiveMode()`, but server-beta never initializes it, so **every generation job fails** with `No mode loaded. Call loadMode() first.` This affects any server-beta deployment outside the plugin directory (systemd service, Docker, `claude-mem server start`) because ModeManager's hardcoded plugin-relative paths don't resolve there.

## Confirmed in production (2026-05-23)
Upgrading a live server-beta to vanilla 13.3.0 reproduced this exactly: service healthy, events ingesting, but **0 observations generated** — every `session_summary` job completed empty with worker logs showing `No mode loaded`. The generation worker runs in a sandboxed thread, so loading the mode in the entrypoint process doesn't reach it. Applying this fix (loading the mode at service creation + honoring `CLAUDE_MEM_MODES_DIR`) restored generation immediately.

## Fix
- `src/server/runtime/create-server-beta-service.ts` — `ModeManager.getInstance().loadMode('code')` at startup (try/catch; mode files optional)
- `src/services/domain/ModeManager.ts` — include `process.env.CLAUDE_MEM_MODES_DIR` in the mode search paths

### DIFF
--- a/src/server/runtime/create-server-beta-service.ts
+++ b/src/server/runtime/create-server-beta-service.ts
@@ -13,6 +13,7 @@ import { GeminiObservationProvider } from '../generation/providers/GeminiObserva
 import { OpenRouterObservationProvider } from '../generation/providers/OpenRouterObservationProvider.js';
 import type { ServerGenerationProvider } from '../generation/providers/shared/types.js';
 import { ServerBetaService } from './ServerBetaService.js';
+import { ModeManager } from '../../services/domain/ModeManager.js';
 import {
   DisabledServerBetaEventBroadcaster,
   DisabledServerBetaGenerationWorkerManager,
@@ -156,6 +157,10 @@ export function validateServerBetaEnv(
 export async function createServerBetaService(
   options: CreateServerBetaServiceOptions = {},
 ): Promise<ServerBetaService> {
+  // Generation prompt-builder requires an active mode; server-beta never went
+  // through the plugin setup path that loads one, so we do it here explicitly.
+  try { ModeManager.getInstance().loadMode('code'); } catch { /* mode files optional */ }
+
   if (!options.skipEnvValidation) {
     validateServerBetaEnv();
   }
--- a/src/server/runtime/create-server-beta-service.ts
+++ b/src/server/runtime/create-server-beta-service.ts
@@ -159,7 +159,16 @@ export async function createServerBetaService(
 ): Promise<ServerBetaService> {
   // Generation prompt-builder requires an active mode; server-beta never went
   // through the plugin setup path that loads one, so we do it here explicitly.
-  try { ModeManager.getInstance().loadMode('code'); } catch { /* mode files optional */ }
+  try {
+    ModeManager.getInstance().loadMode('code');
+  } catch (err) {
+    // Mode files are optional, but surface failures (e.g. malformed JSON in a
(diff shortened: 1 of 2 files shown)


alessandropcostabr:
**Update 2026-06-03 — regression in v13.4.0**

`v13.4.0` (plan-07 Server Runtime GA via #2701) absorbed the `init` half of this PR — `ModeManager.getInstance().loadMode('code')` is now mandatory at startup, which is great. But the `CLAUDE_MEM_MODES_DIR` env var support was **dropped silently**.

Net effect: any server-beta bundle running outside `<packageRoot>/modes` (systemd at `~/.claude-mem/server-beta/`, Docker, custom install) **crash-loops at boot** with:

```
error: Critical: code.json mode file missing
  at loadMode (server-beta-service.cjs:9252:46067)
```

Reproduced on a live systemd deployment upgrading 13.3.0 → 13.4.0 today; workaround was `ln -sf <plugin>/modes ~/.claude-mem/modes` then restart, but the proper fix is restoring the env var.

The remaining delta in this PR is now **a single line** in \`ModeManager.ts\`:

\`\`\`diff
     const possiblePaths = [
+      ...(process.env.CLAUDE_MEM_MODES_DIR ? [process.env.CLAUDE_MEM_MODES_DIR] : []),
       join(packageRoot, 'modes'),
       join(packageRoot, '..', 'plugin', 'modes'),
     ];
\`\`\`

Happy to rebase the PR to just that one-liner on top of \`v13.4.0\` if you want — keeps the env var available for external deployments without changing v13.4.0's hardened boot behavior. Let me know.

Issue #2188, opened 2026-04-28
[Windows] Hooks receive empty stdin when bash resolves to WSL bash (no Git for Windows) — silently breaks all memory capture


ik24999:
## Summary

On Windows installations where `bash` resolves to **WSL bash** (`C:\Windows\System32\bash.exe`) instead of Git Bash, all Claude Code lifecycle hooks fire but receive **empty stdin**, causing every hook handler to log:

```
[WARN] [HOOK] session-init: No sessionId provided, skipping (Codex CLI or unknown platform)
[WARN] [HOOK] summarize: No sessionId provided, skipping
```

The result: zero observations, zero summaries, completely silent failure. The worker, MCP server and viewer all work fine — only memory capture is broken.

## Reproduction

1. Windows 11 machine without Git for Windows installed (so `where bash` only finds `C:\Windows\System32\bash.exe`).
2. Install claude-mem v12.4.8.
3. Run any Claude Code session.
4. Check `~/.claude-mem/logs/claude-mem-*.log` — every `session-init`, `observation`, and `summarize` invocation reports the warning above. DB stays at `observations=0, summaries=0` forever.

## Root cause

The hooks in `plugin/hooks/hooks.json` use `"shell": "bash"` and a complex command starting with:

```bash
export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH"; ...
```

When Claude Code routes this through WSL bash (a Linux subsystem), the Windows→Linux process boundary loses the JSON stdin payload that Claude Code piped. By the time `node bun-runner.js worker-service.cjs hook claude-code session-init` runs, stdin is empty.

`bun-runner.js` already has a fallback to `'{}'` when stdin is empty (added for #1560), so the worker sees `e = {}` and the `claude-code` adapter computes `sessionId = e.session_id ?? e.id ?? e.sessionId = undefined`. Hence the warning.

I verified this empirically:

- Calling the hook with a real `session_id` JSON via stdin works perfectly: a session is created, observations enqueue, the worker logs `[SESSION] CREATED | contentSessionId=...` and `[QUEUE] ENQUEUED | type=observation`.
- The same hook command launched by Claude Code's bash wrapper produces the warning every time.

## Fix

Bypass bash entirely on Windows. Replace each hook's bash one-liner with a direct Node invocation:

```json
{
  "type": "command",
[...]

PR #2636, opened 2026-05-23
fix(windows): strip BOM from settings.json + exempt lifecycle subcommands from #2188 stdin guard

alessandropcostabr:
## Summary

Two defense-in-depth fixes for Windows environments discovered while debugging the server-beta pipeline on Windows CC 2.1.145:

- **SettingsDefaultsManager.ts**: Strip UTF-8 BOM (U+FEFF) before `JSON.parse`. Windows tools (editors, formatters, Claude Code `Edit` hook) may prepend BOM to `settings.json`, which Bun rejects with `JSON Parse error: Unrecognized token '﻿'`, causing a **silent fallback to defaults** that breaks server-beta routing, health checks, and observation delivery.

- **bun-runner.js**: Exempt lifecycle subcommands (`start`/`stop`/`restart`/`status`) from the #2188 empty-stdin guard. These subcommands manage the worker daemon and never consume stdin. The current guard kills the child process on empty stdin, which prevents the daemon from starting on platforms where Claude Code does not pipe a payload for `SessionStart` (e.g. Windows CC ≤ 2.1.145).

## Root cause

On Windows, the `Edit` tool's PostToolUse formatter hook can add a UTF-8 BOM to `~/.claude-mem/settings.json`. On the next hook invocation, Bun's `JSON.parse` rejects the BOM → settings loader falls back to defaults → `CLAUDE_MEM_WORKER_PORT` reverts to default → health check fails → observations are silently lost. No error is surfaced to the user.

For bun-runner.js, the `SessionStart` hook runs `worker-service.cjs start` which does not need stdin. But the #2188 guard kills the child because stdin is empty, preventing the worker daemon from starting.

## Test plan

- [ ] Write `settings.json` with BOM prefix, verify settings load correctly
- [ ] Run `bun-runner.js worker-service.cjs start` without stdin, verify daemon starts
- [ ] Run `bun-runner.js worker-service.cjs hook claude-code observation` without stdin, verify #2188 guard still fires

### DIFF
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -142,60 +142,74 @@ if (child.stdin) {
     child.stdin.write(stdinData);
     child.stdin.end();
   } else {
-    // Issue #2188: empty/missing stdin previously masked by `|| '{}'` fallback,
-    // which silently hid WSL bash failures (e.g. hooks invoked under a broken
-    // shell that never piped a payload). Surface the failure mode instead.
-    const dataDir = process.env.CLAUDE_MEM_DATA_DIR || join(homedir(), '.claude-mem');
-    const payloadType = stdinData === null
-      ? 'null (no data event or stream error)'
-      : stdinData === undefined
-        ? 'undefined'
-        : Buffer.isBuffer(stdinData) && stdinData.length === 0
-          ? 'empty Buffer (zero bytes received)'
-          : `unexpected (${typeof stdinData})`;
-    const payloadByteLength = (stdinData && typeof stdinData.length === 'number')
-      ? stdinData.length
-      : 0;
-    const diagnostic = [
-      `[bun-runner] empty stdin payload received — issue #2188`,
-      `  script: ${args[0]}`,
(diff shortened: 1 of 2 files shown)


alessandropcostabr:
**Rebased onto `v13.4.0` — 2026-06-03**

The PR is now `MERGEABLE` again. Two files conflicted:

- `src/shared/SettingsDefaultsManager.ts` — auto-merged cleanly (plan-06's new env vars added in adjacent ranges, didn't overlap with the BOM strip).
- `plugin/scripts/bun-runner.js` — 1 hunk needed manual resolution. plan-01 (#6821d852, hook IO discipline) added a comment block right at the same point where this PR wraps the `console.error(diagnostic)` in an `if (isLifecycle) { close } else { build + report }` split. Resolved by placing plan-01's "IO discipline" comment **inside** the `else` branch, immediately before the existing `console.error(diagnostic)` call. Both patches survive intact; behavior unchanged. `node --check` passes.

Let me know if you'd like the hunk laid out differently.

PR #2665, opened 2026-05-27
fix(project-name): support git-root project names

### LINKED ISSUE #2663: Project name should be derived from git repo root, not `path.basename(cwd)`
### Situation

I keep multiple independent git repositories side-by-side under one parent directory (a common multi-repo layout):

```
~/code/
  repo-a/    (own .git)
  repo-b/    (own .git)
  repo-c/    (own .git)
```

I often launch Claude Code from `~/code/` and `cd` into whichever sub-repo I need during the session. With this setup, claude-mem labels every observation, summary, and session under the project name `code` — the basename of the launch directory — regardless of which sub-repo the work is actually happening in. All my memory ends up bucketed into a single useless project.

### Root cause

In [`src/utils/project-name.ts`](https://github.com/thedotmack/claude-mem/blob/main/src/utils/project-name.ts), `getProjectName(cwd)` is just `path.basename(cwd)`. [`detectWorktree()`](https://github.com/thedotmack/claude-mem/blob/main/src/utils/worktree.ts) only fires for true `git worktree add` worktrees (where `.git` is a file pointing to `parent/.git/worktrees/...`), not for normal nested repos where `.git` is a directory. There's no logic to walk up from `cwd` to find the nearest enclosing `.git` directory.

I also couldn't find any env var, settings.json key, or CLI flag to override this. The full configuration page lists none.

### Workaround

The only fix today is to always `cd` into the actual repo before launching Claude. That works but is easy to forget, and any existing sessions started from the parent are permanently mis-tagged.

### Proposed fix

Add a "walk up to nearest `.git`" mode to `getProjectName` / `getProjectContext`, gated by an opt-in env var to preserve current behavior:

```
CLAUDE_MEM_USE_GIT_ROOT=true
```

When set:
1. From `cwd`, walk up until a directory containing `.git` (file or dir) is found.
2. Use `path.basename(<that directory>)` as the project name.
3. If no `.git` is found before the filesystem root, fall back to the current behavior.

This is non-breaking, opt-in, and reuses the helpers already in the repo. Happy to send a PR if it'd be welcome.

### Environment

- claude-mem installed via `npx claude-mem install`
- macOS

a1573595:
## Summary

Project names can now be derived from the nearest enclosing Git repository root instead of always using `path.basename(cwd)`.

This fixes multi-repo layouts where Claude/Codex is launched from a parent folder and work happens inside nested independent repositories. When `CLAUDE_MEM_USE_GIT_ROOT=true`, claude-mem walks up from the hook `cwd` until it finds a `.git` file or directory, then uses that directory basename as the project name. If no Git root is found, it falls back to the existing basename behavior.

Worktree context resolution now uses the resolved project path too, so worktree subdirectories still produce the existing `parent/worktree` composite project name.

Gated on one opt-in setting/env var:
- `CLAUDE_MEM_USE_GIT_ROOT`

Default remains `false`, so this is non-breaking for existing users. Existing memories are not relabeled; the new behavior only affects newly stored sessions, observations, and summaries.

Fixes #2663.

## Test plan
- [x] `npx --yes bun@1.3.14 test tests/utils/project-name.test.ts tests/utils/project-name-isolation.test.ts`
- [x] `npx --yes bun@1.3.14 test tests/shared/settings-defaults-manager.test.ts`
- [x] `npm run build`
- [x] `git diff --check`

Known pre-existing baseline failures:
- [ ] `npm run typecheck` — fails in untouched files unrelated to this PR; no errors point at files changed here


thedotmack:
Closing as superseded by the newer project-identity work in #3047 and #2827, which address repository and worktree naming against the current context system.

PR #2674, opened 2026-05-27
[codex] fix logs clear with empty request body

pg-adm1n:
## Summary

- Allow `POST /api/logs/clear` to run without a request body, matching the viewer's bodyless clear request.
- Keep the existing `Log file cleared via UI` audit log entry after a successful clear.
- Treat `/api/logs/clear` as a quiet logs endpoint in HTTP request logging so DEBUG request/response logging does not add extra clear-request lines.

## Reproduction

1. Start the local claude-mem viewer and open the Console drawer.
2. Click the trash / clear logs button.
3. Confirm the clear action.

Before this change, the viewer showed `Failed to clear logs: Bad Request` and the log file was not truncated.

## Expected Behavior

The bodyless clear request should return success and remove the previous log contents.

The route still records the existing `Log file cleared via UI` audit entry after a successful clear.

## Root Cause

The viewer calls `authFetch('/api/logs/clear', { method: 'POST' })` without a body. The route previously wrapped the handler in `validateBody(z.object({}).passthrough())`, so `req.body === undefined` failed validation and returned `400 Bad Request` before the log file was truncated.

## Validation

- `bun test tests/services/logs-routes-tail-read.test.ts`

This includes a regression test that invokes the clear route with `body: undefined` and verifies the pre-existing log contents are removed.

### DIFF
--- a/src/services/worker/http/middleware.ts
+++ b/src/services/worker/http/middleware.ts
@@ -20,8 +20,8 @@ export function createMiddleware(
   middlewares.push((req: Request, res: Response, next: NextFunction) => {
     const staticExtensions = ['.html', '.js', '.css', '.svg', '.png', '.jpg', '.jpeg', '.webp', '.woff', '.woff2', '.ttf', '.eot'];
     const isStaticAsset = staticExtensions.some(ext => req.path.endsWith(ext));
-    const isPollingEndpoint = req.path === '/api/logs'; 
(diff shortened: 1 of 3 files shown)


pg-adm1n:
Closing this draft in favor of #2675, which keeps the fix narrower: it only addresses the bodyless `POST /api/logs/clear` validation failure and leaves the existing audit log behavior intact.

PR #2692, opened 2026-05-28
fix(codex): emit SessionStart JSON from version check

ShunmeiCho:
## Summary

- ensure `plugin/scripts/version-check.js` always emits valid `SessionStart` JSON when running as a Codex hook
- preserve the existing Claude Code behavior where no-op version checks stay silent
- add coverage for Codex no-op output and Codex upgrade hints via `additionalContext`

## Why

Codex validates `SessionStart` hook stdout as JSON. When `CLAUDE_MEM_CODEX_HOOK=1` and the install marker is current, `version-check.js` previously exited successfully with empty stdout, which Codex reports as `hook returned invalid session start JSON output`.

## Tests

- `bun test tests/plugin-version-check.test.ts`
- `node --check plugin/scripts/version-check.js`

### DIFF
--- a/plugin/scripts/version-check.js
+++ b/plugin/scripts/version-check.js
@@ -17,16 +17,32 @@ function resolveRoot() {
 }
 
 const ROOT = resolveRoot();
-if (!ROOT) process.exit(0);
+
+let emittedCodexHookOutput = false;
+
+function emitCodexSessionStart(additionalContext) {
+  const output = {
+    hookSpecificOutput: {
+      hookEventName: 'SessionStart',
+    },
+  };
+  if (additionalContext) {
+    output.hookSpecificOutput.additionalContext = additionalContext;
+  }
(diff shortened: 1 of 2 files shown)


WingsOfPanda:
Confirmed this still reproduces on claude-mem 13.6.0 with Codex 0.139.0 on Ubuntu 24.04.

SessionStart hook 0 (`version-check.js`) exits 0 with empty stdout when:

- `CLAUDE_MEM_CODEX_HOOK=1`
- `.install-version` matches package version
- no upgrade hint is needed

Codex surfaces this as:

`hook returned invalid session start JSON output`

Local hotfix was to make `version-check.js` emit a no-op SessionStart JSON payload in Codex hook mode when no upgrade hint is emitted. After that, all three SessionStart commands returned parseable JSON and the worker/MCP path remained healthy.

thedotmack:

PR #2668, opened 2026-05-27
fix(chroma): spawn uvx directly on Windows instead of via cmd.exe (closes #2667)

PulseCheckAI:
## Summary

Fixes #2667 — Chroma vector search is broken on Windows because `cmd.exe /c uvx ...` interprets `>=1.20` and `<7` in the dep-override args as I/O redirection.

## The change

`src/services/sync/ChromaMcpManager.ts:108-110` — spawn `uvx` directly on every platform instead of wrapping it in `cmd.exe /c` on Windows.

```diff
-    const isWindows = process.platform === 'win32';
-    const uvxSpawnCommand = isWindows ? (process.env.ComSpec || 'cmd.exe') : 'uvx';
-    const uvxSpawnArgs = isWindows ? ['/c', 'uvx', ...commandArgs] : commandArgs;
+    // Spawn uvx directly on every platform. Wrapping it inside `cmd.exe /c` on
+    // Windows is unsafe: cmd.exe interprets shell-metacharacters in its /c
+    // argument string BEFORE handing them to uvx, and Node's arg-array quoting
+    // does not survive that re-parse (per the cmd.exe parser's documented
+    // behavior of stripping the outer quote pair from /c). [...full reasoning
+    // in commit message and inline comment...]
+    const uvxSpawnCommand = 'uvx';
+    const uvxSpawnArgs = commandArgs;
```

The `StdioClientTransport` call below this code is unchanged — it receives the same `{ command, args, env, cwd, stderr }` shape; only the values flip.

## Why removing the cmd.exe wrapper is safe

- `uvx` ships as `uvx.exe` (real PE executable) on Windows, installed by the `uv` installer at `%LOCALAPPDATA%\Microsoft\WinGet\Links\uvx.exe` or `%USERPROFILE%\.local\bin\uvx.exe`. Confirmed by running `(Get-Command uvx).CommandType` → `Application`.
- Node's `child_process.spawn(command, args, ...)` invokes Windows `CreateProcess` and applies its own arg-quoting for `.exe` targets, which preserves characters like `>`, `<`, `=` inside args.
- The CVE-2024-27980 batch-file constraint (must use `shell: true` for `.cmd`/`.bat`) doesn't apply — uvx is a `.exe`.

## Verification

Reproduced the bug, applied the fix, verified Chroma works end-to-end on Windows 11 + Node 22.22.2 + Bun 1.3.14 + uv 0.11.8:

```
$ curl http://127.0.0.1:37777/api/chroma/status?deep=1
{
[...]

Issue #2667, opened 2026-05-27
Chroma vector search broken on Windows: cmd.exe /c uvx interprets >=1.20/<7 in args as I/O redirection


PulseCheckAI:
### Symptom

Every observation that reaches the sync step on Windows fails to index in Chroma. Worker log shows:

```
[CHROMA_MCP] Connecting to chroma-mcp via MCP stdio {command=C:\WINDOWS\system32\cmd.exe, args=/c uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 --client-type persistent --data-dir C:/Users/.../chroma}
[CHROMA_MCP] Connection failed, killing subprocess tree to prevent zombie {error=MCP error -32000: Connection closed}
[CHROMA] SDK chroma sync failed, continuing without vector search MCP error -32000: Connection closed
```

`GET /api/chroma/status?deep=1` reports `status: unhealthy, details: chroma-mcp deep probe failed at stage 'list', error: chroma-mcp connection in backoff`.

The system gracefully falls back to FTS5 keyword search, but semantic similarity is unavailable. Every observation creates one Chroma backoff cycle.

### Root cause

`src/services/sync/ChromaMcpManager.ts:108-110` wraps `uvx` inside `cmd.exe /c` on Windows:

```ts
const isWindows = process.platform === 'win32';
const uvxSpawnCommand = isWindows ? (process.env.ComSpec || 'cmd.exe') : 'uvx';
const uvxSpawnArgs = isWindows ? ['/c', 'uvx', ...commandArgs] : commandArgs;
```

cmd.exe's `/c` parser interprets shell-metacharacters in its argument string **before** handing them to the inner program. The dep-override args produced by `buildCommandArgs()` include `--with onnxruntime>=1.20` and `--with protobuf<7`. cmd.exe parses:

- `>=1.20` as `>` (stdout redirection) + filename `=1.20`
- `<7` as `<` (stdin redirection) + filename `7`

uvx runs with the args truncated at `--with onnxruntime`, exits with `The system cannot find the file specified.` (no file named `7`), and the MCP stdio handshake collapses immediately. Node's arg-array quoting at the `child_process.spawn` layer does not survive cmd.exe's `/c` re-parse — Microsoft documents that cmd strips the outer quote pair from `/c`.

### Manual reproduction (Windows 11, PowerShell)

```pwsh
# Mimic the cmd.exe /c spawn:
[...]

PR #2699, opened 2026-05-29
🪟 Windows Canary: combined build/test of all open Windows fixes

### LINKED ISSUE #2461: Fix Windows compatibility: replace sh spawn + silence printf EPIPE

thedotmack:
## 🪟 Windows Canary — combined build/test branch for all open Windows fixes

We have a **large cluster of open Windows PRs** that each fix a different piece of the same cross-platform story (spawn, hooks, chroma, mcp launcher, supervisor, settings, runtime deps). They overlap heavily — several touch the same launcher strings and the same `ChromaMcpManager` spawn path — so merging them one-by-one keeps re-introducing conflicts and nobody can see them working *together*.

This branch bundles them into **one canary** so we can validate the combined result.

> ### ⚠️ Calling all Windows contributors
> **@CaTeIM @ousamabenyounes @sabinss4 @YOMXXX @alessandropcostabr @Bruce583102198 @PulseCheckAI** — your Windows PRs are bundled here.
>
> **I (@thedotmack) do not have a Windows machine**, so I genuinely cannot test this myself — we have to work this one out together. Please:
> 1. `git fetch origin && git checkout windows-canary`
> 2. `npm ci` (or `bun install`) then `npm run build`
> 3. Install/point Claude Code at this build and exercise it on **real Windows** (PowerShell, Git Bash, and `cmd.exe` if you can)
> 4. Report back in this thread: ✅ works / ❌ broke, with the shell + Windows version. Screenshots/logs of any console-window flashes, `MCP error -32000`, EPIPE, or path-resolution failures are gold.
>
> If your fix needed adjustment to combine with the others (see "Conflict decisions" below), please confirm the resolution still solves your original issue.

---

### ✅ PRs bundled in this canary (12)

| PR | Author | What it fixes |
|----|--------|----------------|
| #2208 | @thedotmack | Restore Windows spawn (PR #751 re-apply) + Windows CI |
| #2489 | @CaTeIM | Launch `mcp-search` via `node` so Windows can spawn it |
| #2502 | @ousamabenyounes | Drop `cmd.exe` wrap on Windows `uvx` spawn (chroma) |
| #2507 | @sabinss4 | Add PowerShell call operator (`&`) to hook commands |
| #2593 | @YOMXXX | Escape `cmd.exe` metacharacters in `uvx` args |
| #2594 | @YOMXXX | Replace `sh` launcher with cross-platform `node -e` (#2461) |
| #2598 | @YOMXXX | Replace fragile `$SHELL -lc` PATH probe with `printenv`-first |
| #2599 | @YOMXXX | Make Windows `captureProcessStartToken` actually return a token (#2578) |
| #2636 | @alessandropcostabr | Strip BOM from `settings.json` + lifecycle stdin-guard exemption |
| #2654 | @Bruce583102198 | Suppress printf write errors when pipe breaks in inline bash hooks |
| #2658 | @CaTeIM | Restore plugin runtime deps after auto-upgrade |
| #2668 | @PulseCheckAI | Spawn `uvx` directly on Windows instead of via `cmd.exe` (#2667) |

### 🔀 Conflict decisions (please sanity-check these)

These PRs proposed **competing approaches to the same problem** — only one can be active at a time. Here's what the canary currently uses and what got superseded:

- **Chroma `uvx` spawn** — three PRs, two approaches:
  - **#2502 + #2668 agree:** spawn `uvx` **directly** (no `cmd.exe` wrap). **← canary uses this.**
  - **#2593** keeps the `cmd.exe` wrap but escapes metacharacters (`^<>&|`, `%%`). Merged as a no-op since direct-spawn makes the wrap unnecessary. **@YOMXXX** — if direct-spawn doesn't work on your Windows setup, this is the fallback; please test.
- **mcp-search launcher** — both node-based:
  - **#2594** spawns a subprocess with signal-forwarding. **← canary uses this** (closes #2461, kept #2489's distribution test green).
  - **#2489** loads `mcp-server.cjs` in-process via `require` (superseded, but #2489 still contributes the node move + its test).
- **Hook PATH probe** — **#2598**'s `printenv`-first probe **+** **#2654**'s `2>/dev/null` pipe-suppression are combined (complementary).

### 🚫 Deliberately excluded (need a rebase or a decision)

- **#2209** (@thedotmack) — *Codex transcript + queue self-deadlock.* The `watcher.ts` half (Windows transcript-append fix) is clean, but the `SessionRoutes.ts` half collides with `main`'s **since-merged `handleGeneratorExit` / `GeneratorExitHandler` refactor**. The #2192 in-flight-requeue fix is **not** present in the new handler and must be re-ported there. **Needs a rebase onto `main` before it can join.**
- **#2501** (@ousamabenyounes) — *`echo` instead of `printf '%s\n'` in hook subshells (#2439).* Superseded by **#2598**'s `printenv`-first probe already in the canary; its tests assert the old launcher format. If `printf '%s\n'` genuinely misbehaves on your shell, please say so — that would override #2598.
- **#2526** (@ajs-uk-dev) — *Silence Windows Git Bash EPIPE/EACCES on the path resolver.* Functionally covered by **#2654**'s `2>/dev/null` pipe-suppression already in the canary. Please confirm #2654 fully resolves your case, or what's still leaking.
- **#2476** (draft) — PowerShell-hook rewrite. A *competing* architecture to the bash-hook fixes here; left out as a draft.
- **#2614** (draft) — Codex ingestion + Windows chroma; draft, overlaps the chroma trio.

### 🧪 Verification done locally (macOS — Windows still needs you)

- `npm run build` ✅ all bundles regenerate from merged source
- `tsc --noEmit`: **24 errors — identical to `origin/main`; 0 introduced by this merge** (the pre-existing TS baseline is fixed by #2511)
- `bun test`: **54 failures — identical to `origin/main`; 0 introduced by this merge** (pre-existing test baseline is fixed by #2512)
- Windows/chroma/infra suites that this branch touches (`tests/infrastructure`, `tests/services/sync`): **186 pass / 0 fail**

In other words: this branch adds the Windows fixes **without regressing the baseline**. The remaining red is pre-existing and owned by #2511/#2512.

### Notes
- Diffstat: 20 files, +1,572 / −606 (incl. regenerated bundles).
- This is a **canary, not a merge target** — once Windows folks confirm what works, we'll land the validated fixes via their individual PRs (or fast-track them).
- `CHANGELOG.md` intentionally not touched (auto-generated).


YOMXXX:
Hi @thedotmack — thanks for bundling these together, the canary approach makes a lot of sense given the overlap.

### My PRs in the canary

**#2593 (cmd.exe metachar escaping)** — agree that direct-spawn (#2502/#2668) is the better path. Happy for #2593 to sit as a no-op fallback; if direct-spawn works across environments, we can close it.

**#2594 (node -e MCP launcher)** — glad this was picked as the launcher approach. One note on Greptile's Issue 3 below.

**#2598 (printenv-first PATH probe)** — also just pushed a follow-up fix (`e060a674`) removing `tr ' ' ':'` which was corrupting Windows paths containing spaces (e.g. `/c/Program Files/nodejs`). That fix should probably be pulled into the canary as well.

**#2599 (Windows captureProcessStartToken)** — straightforward, no concerns.

### On the Greptile issues

**Issue 1 (DATA_DIR hardcoding in ProcessManager.ts)** — this looks like a real regression. `paths.dataDir()` honours `CLAUDE_MEM_DATA_DIR` from env/settings, so hardcoding `~/.claude-mem` would silently break multi-profile setups. Probably worth restoring `paths.dataDir()` here.

**Issue 2 (PowerShell `& ` call operator)** — this came from #2507. The concern is valid: `& ` is PowerShell-specific and would break under `cmd.exe`. Worth confirming with @sabinss4 which shell Windsurf/Gemini CLI actually use for hook invocation on Windows.

**Issue 3 (exit code 128 on Windows signal re-raise)** — from my #2594. The `process.exit(128)` fallback only fires when `process.kill(pid, sig)` throws (Windows doesn't support most POSIX signals). I can change this to `process.exit(0)` since the child already exited — the launcher's job is done at that point. Will push a fix.

Happy to help address Issues 1 and 3 if that would be useful — just let me know.

bigmanBass666:
**Windows test report** (posted via Claude Code agent on behalf of @bigmanBass666)

**Environment:**
- Windows 11 Home China (22631)
- Shell: Git Bash (MSYS2), PowerShell 5.1
- claude-mem version: 13.3.0 (from `installed_plugins.json`)
- Worker: 12.1.0 (stale, not yet restarted after upgrade)

**Issue reproduced:**
- `mcp-search` MCP server fails to start — `MCP error -32000`
- `.mcp.json` in 13.3.0 uses `"command": "sh"` with a pipe-based path resolver
- Reproducing the exact `sh -c` command in Git Bash produces: `printf: write error: Permission denied`
- Root cause: `{ producer; } | while read ... break; done` — when consumer breaks early, remaining `printf` calls get EACCES on Windows MSYS emulated pipes (silent SIGPIPE on Linux/macOS)

**Workaround confirmed working:**
Patched `13.3.0/.mcp.json` to use direct `node` invocation (same as 12.1.0):
```json
{
  "mcpServers": {
    "mcp-search": {
      "type": "stdio",
      "command": "node",
      "args": ["${CLAUDE_PLUGIN_ROOT}/scripts/mcp-server.cjs"]
    }
  }
}
```
MCP server starts successfully after this change. Worker health endpoint at `localhost:37777/api/health` responds normally.

**This confirms PR #2489's approach is correct.** The `sh` launcher is fundamentally broken on Windows due to MSYS pipe emulation behavior. Using `node` directly bypasses the entire issue.

**Note:** The worker was still running 12.1.0 at test time (pre-restart). After a full Claude Code restart the SessionStart hook should bring up the 13.3.0 worker. The `.mcp.json` patch will need to be re-applied after each plugin auto-update until the canary lands.

✅ `node`-based launcher works
❌ `sh -c` pipe resolver fails on Windows (EACCES)

cateim:
**✅ Windows test report — @CaTeIM**

**Environment**: Windows 11 Pro 10.0.28000 · node v24.15.0 · bun 1.3.14 · uvx at `~/.local/bin/uvx`
**Shells exercised**: PowerShell 7, Git Bash, cmd.exe (limited — see notes)

## Setup

`git checkout windows-canary` → `npm ci` → `npm run build` → all clean, no errors.

## End-to-end checks

| Area | Result | Detail |
|------|--------|--------|
| **`mcp-search` launcher** (PR #2594, supersedes my #2489) | ✅ | `node -e` bootstrap spawns `mcp-server.cjs` via `child_process.spawn(process.execPath, …, {stdio:'inherit'})` with SIGTERM/INT/HUP forwarding. Real `initialize` handshake from PowerShell with `UseShellExecute=false` (same path Claude Code uses) returns proper `serverInfo` for `claude-mem 13.3.0`. **Confirms the "Connection closed" failure from my #2489 is still resolved by #2594's approach.** |
| **Plugin runtime deps** (my #2658 Fix A) | ✅ | `plugin/package.json` still declares `ajv ^8.20.0` + `ajv-formats ^3.0.1`. Build guardrail preserved. |
| **Auto-install on upgrade** (my #2658 Fix B) | ✅ | `ensureRuntimeDeps()` in `version-check.js` tested across all three paths: broken (spawns detached `bun install`, marker written **in the version root** per the Greptile review fix, log opened), cooldown (second call within 10 min no-ops), healthy (node_modules present → no-op). |
| **Chroma `uvx` direct-spawn** (PR #2502/#2668) | ✅ | `tests/services/sync/chroma-mcp-manager-windows-spawn.test.ts` "spawns uvx directly on Windows (no cmd.exe /c wrapper)" passes (2/2 on canary; fails on `main`). |

## Test suite (`tests/infrastructure` + `tests/services/sync`)

Canary on Windows: **163 pass / 15 fail / 8 skip** (12 unique failing test names after dedup) vs main on Windows: **165 pass / 13 fail / 8 skip** (12 unique).

Name-only diff vs `main` (timings stripped):

- **Canary fixes 1 test**: `ChromaMcpManager Windows spawn (2426) > spawns uvx directly on Windows (no cmd.exe /c wrapper)` ✅
- **Canary introduces 1 failure**: `ProcessManager > captureProcessStartToken > returns null on win32 (liveness-only fallback path)` ⚠️ — this test (`tests/infrastructure/process-manager.test.ts:384`) forces `process.platform = 'win32'` and asserts `captureProcessStartToken(pid)` returns `null`. PR #2599 (@YOMXXX) changes this function to actually return a real token on win32, which makes the assertion fail. The test needs to be updated or removed alongside #2599 — the parenthetical "(liveness-only fallback path)" described the OLD pre-#2599 behavior.
- **11 common failures** (same on canary and main on Windows) = pre-existing Windows baseline unrelated to this merge: `runOneTimeV12_4_3Cleanup` (4), `HealthMonitor isPortInUse / waitForPortFree` (4), `ProcessManager.resolveWorkerRuntimePath`, `ChromaMcpManager singleton` (2). The maintainer doesn't see these on macOS; they look like Windows-specific test plumbing issues (port reuse semantics, Windows process trees, file-locking during cleanup), not regressions from this canary.

## Live evidence for #2654 / #2526 (EPIPE suppression)

Independent of the canary, my running Claude Code session — using the *currently-installed* (not canary) plugin — is generating the following on every PostToolUse observation hook **right now**:

```
```

This is exactly the EPIPE/EACCES leakage that **#2654** suppresses with `2>/dev/null` and that **#2526** also targets. Strong real-world signal that these belong in the canary as-is.

## Live deploy

After the isolated checks above, I pointed my Claude Code at the canary by copying the built `plugin/` contents into the active plugin cache at `~/.claude/plugins/cache/thedotmack/claude-mem/13.3.0/` (the dir Claude Code actually reads — *not* the marketplace clone; that was the bit that tripped me up in #2489's earlier sessions). After the worker restart:

- `claude mcp list` shows `plugin:claude-mem:mcp-search` as `node -e <#2594 bootstrap with c.spawn + signal forwarding>` — **✓ Connected**.
- Worker healthy on `http://127.0.0.1:37777`.
- MCP `initialize` handshake against the cache's `mcp-server.cjs` returns proper `serverInfo`.

To pick this up in my live GUI session I'll do a full Claude Code restart.

## Notes

- **cmd.exe shell test** was inconclusive — my repro had quoting issues with paths containing spaces ("Google Drive"). PowerShell's `UseShellExecute=false` path is what Claude Code actually uses, so this is more representative than cmd.exe anyway.
- I'm happy to re-run any specific shell scenario or attach raw logs — ping me.

cateim:
**Follow-up after live restart** — confirming **#2654** in production: after switching my Claude Code session to the canary plugin (cache `13.3.0` deployed + restart), my `PostToolUse` observation hook **no longer emits `printf: write error: Permission denied`**. Same machine, same Git Bash, same hook flow — the only thing that changed is the bundled hook now suppresses the broken pipe with `2>/dev/null`. Before the restart, every Bash tool call in this very session was producing the EPIPE; after the restart, zero occurrences across many subsequent calls.

cc @Bruce583102198 — your fix is doing exactly what it claims.

alessandropcostabr:
## Windows 11 Home SL 26200 — full triple-shell report (@alessandropcostabr)

**Environment:** Windows 11 Home SL 10.0.26200.8457 · node v24.11.1 · bun 1.3.14 · git 2.53 · Claude Code 2.1.145

### Deploy
Plugin `marketplaces/thedotmack/plugin/` replaced byte-for-byte by `canary bcf437fb`'s pre-built `plugin/` (md5 `C7C58737...`, 4,031,405 bytes). `bun install` ran clean (706 packages); `npm run build` fails on `@better-auth/kysely-adapter` ↔ `kysely` drift — irrelevant since the canary commit already ships regenerated bundles. Backup kept at `plugin.bak-pre-canary-...` (md5 `744A9AB5...`, the previous 23/05 build).

### Discriminator counts in the deployed bundle

| Marker | Count | Source |
|---|---|---|
| `/api/context/inject` (canary client path) | 7 | — |
| `/v1/context/inject` (legacy) | 0 | (intentionally removed) |
| `isTTY` stdin guard | 1 | #2188 |
| SessionStart lifecycle exempt | 1 | #2636 |
| `node -e` MCP launcher (`.mcp.json`) | ✅ | #2594 |
| BOM strip in `SettingsDefaultsManager` | present | #2636 |
| `@better-auth/api-key` (`x-api-key` default) | 1 | (see note below) |

### Per-shell results

| Shell | `claude mcp list` | Live tool calls | `printf write error` | `MCP error -32000` | Console flash |
|---|---|---|---|---|---|
| **PowerShell 7** | `claude-mem ✓ Connected` (+ 6 others) | observations STORED (obsId 34634) | 0 | 0 | none |
| **Git Bash (MSYS)** | `claude-mem ✓ Connected` | session registered, tool calls captured | **0** (confirms #2654) | 0 | none |
| **cmd.exe** (interactive) | `claude-mem ✓ Connected` (all 7 servers) | `claude --version` → `2.1.145` | 0 | 0 | none |

### Live capture E2E
Worker uptime 30+ min post-restart, 0 ERROR / 0 FATAL / 0 server-beta-fallback in the 74-line log. Pipeline `notebook → CF → Mach10 → Express → BullMQ → Redis → Qdrant` flowing: **68 events / 10 min** acknowledged server-side, with observations generated downstream.

Worker invocation:
```
"C:\Users\aless\.bun\bin\bun.exe" C:\Users\aless\.claude\plugins\marketplaces\thedotmack\plugin\scripts\worker-service.cjs --daemon
```

### Bonus — Codex CLI MCP integration confirmed
Wired `[mcp_servers.claude-mem]` into `~/.codex/config.toml` pointing at the same `mcp-server.cjs` (via `node` direct, mirroring the `.mcp.json` durable-fix pattern). Inside `codex` REPL, `/mcp` lists **claude-mem** alongside native `codex_apps` with all 21 tools available (`search`, `smart_search`, `memory_add`, `observation_search`, `timeline`, etc.). `Auth: Unsupported` is expected for stdio-local MCP servers — not a defect. Codex capture hooks (Peça 2) still pending — `.codex-plugin/` in the bundle only ships `plugin.json` without handlers, so Codex sessions can **read** the claude-mem corpus but won't generate observations of their own yet.

### Non-blocking note for maintainers running server-beta behind a proxy

The deployed client bundle uses `X-Api-Key` on `POST /v1/events` (the `@better-auth/api-key` default). The server-beta middleware in `src/server/middleware/{auth,postgres-auth}.ts` only reads `Authorization: Bearer`. Anyone running `server-beta` behind a header-normalizing proxy (Cloudflare → Express in my case) gets `401 {"error":"Unauthorized","message":"Missing bearer API key"}` even with a valid key. Direct-LAN setups don't hit it because no proxy is in the path.

I opened **#2719** to make the server middleware accept `X-Api-Key` as fallback to `Bearer` — fully backward-compatible (Bearer remains canonical and wins when both are sent), 3 new unit tests passing alongside the existing 12 in `tests/server/auth-api-key.test.ts`, verified end-to-end against my deployment.

### Net result
Canary `bcf437fb` is **green across all three Windows shells** for me. Zero changes outside the deployed bundle, backup retained for instant rollback if needed.

cc @thedotmack @CaTeIM @YOMXXX @Bruce583102198

tomeyerman:
**Independent Windows validation: #2806 fixed the PATH gap, but the cmd.exe quoting defect this PR targets still breaks semantic search on a clean v13.4.0 / latest main.**

Flagging this so closing #2790 isn't read as "Windows Chroma fixed" — there are two distinct root causes here:

- #2806 (the PATH fix) is correct and necessary — it ensures uvx is resolvable on the chroma-mcp child PATH.
- It is not sufficient. The spawn in connectInternal on main is unchanged — it still does ['/c', 'uvx', ...commandArgs.map(quoteForCmdExe)], so cmd.exe parses the >/< in the dep pins as redirection operators before uvx ever runs. The quoteForCmdExe helper added in v13.4.0 (#2701) does not survive Node's shell:false re-quoting.

Minimal repro on Windows 11, claude-mem 13.4.0, uvx 0.9.11, Python 3.13.9, with uvx confirmed on PATH (i.e. this is post-#2806 behavior — the PATH precondition is already satisfied, which isolates the failure to the >/< quoting, not PATH):

# (A) Direct spawn — exactly what this PR does — WORKS:
```
uvx --python 3.13 --with "onnxruntime>=1.20" --with "protobuf<7" chroma-mcp==0.2.6 --client-type persistent --data-dir <dir>
  → {"result":{ ... "serverInfo":{"name":"chroma","version":"1.6.0"}}}    (boots, waits for stdio)
```

# (B) cmd.exe /c uvx — what main's connectInternal still does — FAILS in ~270ms:
```
cmd /c uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 --client-type persistent --data-dir <dir>
  → "The system cannot find the file specified"    (cmd parses <7 as stdin redirection from a file named "7")
```

Same args, same machine — the only difference is the cmd.exe wrapper, and it dies in ~270 ms on the redirection.

So this PR (drop the cmd.exe wrapper, spawn uvx directly) is the actual fix for the quoting defect, and #2806 is the complementary PATH half. With #2806 now merged, the PATH confound is gone, so this is the remaining blocker for Windows semantic search. The corresponding issues (#2716, #2667) are currently closed as not planned despite the defect still reproducing on latest main.

Happy to check out this branch, build it, and confirm Chroma connects end-to-end on Windows 11 if that helps land it.

rodboev:
One more Windows rollup note from the newer follow-up wave: there are a few additional open Windows-focused PRs now that may be worth tracking alongside this canary if you do another bundle or maintainer pass:

- #2945 fix: install Windows Claude Code hooks without bash (#2869). This is the broader issue-closing slice and effectively supersedes the narrower Bun-popup-only PR #2890.
- #2921 fix(windows): carry windowsHide through the remaining live Windows spawn paths (#2900).
- #2918 fix(setup): restore tree-sitter CLI on clean Windows installs (#2910).
- #2917 fix(context): avoid blocking Claude SessionStart on Windows cold boot (#2903).
- #2890 fix(windows): suppress console window flash in findBun() if you still want to keep the narrow predecessor visible separately, though #2945 absorbs that piece.

For #2945 specifically: the Linux/macOS-facing jobs are green on the current head, and the only remaining red check is the same Windows npm-install/node-gyp Visual Studio toolchain failure already showing up on other open Windows PRs rather than a new branch-specific regression.

thedotmack:
Closing this obsolete Windows canary aggregation branch. The surviving Windows fixes are now tracked and reviewed as focused PRs against the current runtime.

Issue #2709, opened 2026-05-29
UserPromptSubmit operation blocked by hook: [...hook script...]: /usr/bin/bash: line 1: printf: write error: Permission denied 说明 where bun 在 Windows 中文用户名上的 GBK/UTF-8 编码问题


11zzx:
UserPromptSubmit operation blocked by hook:
[...hook script...]: /usr/bin/bash: line 1: printf: write error: Permission denied
说明 where bun 在 Windows 中文用户名上的 GBK/UTF-8 编码问题
可以叫AI给你修复一下的

thedotmack:
Consolidating into #2778 (plan-01). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

PR #2618, opened 2026-05-22
fix: clean up claude mcp list output by moving discovery to a launcher file

impawangupta:
## Problem

Running `claude mcp list` shows a 600+ character inline shell script for the `plugin:claude-mem:mcp-search` entry, making the output hard to read:

```
plugin:claude-mem:mcp-search: sh -c _C="$HOME/.claude"; _E="${PLUGIN_ROOT:-}"; _P=$({ [ -n "$_E" ] && printf '%s\n' "$_E"; ... (600+ chars) - ✓ Connected
```

## Solution

Move the discovery logic out of `plugin/.mcp.json` and into a generated launcher file (`~/.claude-mem/bin/mcp-server.sh`) that is written during `npx claude-mem install`.

`plugin/.mcp.json` becomes a single short line:

```
plugin:claude-mem:mcp-search: sh -c exec sh "$HOME/.claude-mem/bin/mcp-server.sh" - ✓ Connected
```

## How it works

- `writeMcpLauncher()` is called during install and writes `~/.claude-mem/bin/mcp-server.sh`
- The launcher keeps the **full discovery logic** (fast path to known install location, fallback for Claude Code, Codex, `PLUGIN_ROOT` env var, and dev mode) — no functionality is lost
- On Windows, paths are normalized to forward slashes for sh/bash compatibility (matching the existing plugin requirement for a POSIX shell)
- The launcher is regenerated on every `npx claude-mem install` / update so the path stays current

## Files changed

- `plugin/.mcp.json` — shortened to a single `exec sh` call
- `src/npx-cli/install/write-mcp-launcher.ts` — new file, generates the launcher
- `src/npx-cli/commands/install.ts` — calls `writeMcpLauncher` after plugin tasks complete

PR #2725, opened 2026-05-30
fix(hooks): drop pipe `break` that triggers EACCES "printf: write error: Permission denied" on Windows/Cygwin

### LINKED ISSUE #2707: Bug: UserPromptSubmit hook fails on Windows (PowerShell 7) - printf: write error: Permission denied
## Bug: `UserPromptSubmit` hook fails on Windows with `Permission denied`

**Environment**
- OS: Windows 11
- Shell: PowerShell 7
- Claude Code version: latest
- Plugin version: 13.2.0
- Bash: Git Bash (`C:\Program Files\Git\bin\bash.exe`)

**Describe the bug**

Every time a prompt is submitted in Claude Code, the `UserPromptSubmit` hook registered by `claude-mem` fails with:

```
UserPromptSubmit operation blocked by hook:
[...hook script...]: /usr/bin/bash: line 1: printf: write error: Permission denied
```

This completely blocks Claude Code from functioning — no prompt can be submitted.

**Steps to reproduce**

1. Install `claude-mem` plugin on Windows
2. Open Claude Code in PowerShell 7
3. Submit any prompt
4. Hook fires and fails with `Permission denied` on `printf`

**Root cause hypothesis**

The hook script runs via `/usr/bin/bash` (Git Bash on Windows). The `printf` built-in fails when writing to stdout/stderr, likely due to how Windows handles pipe permissions between PowerShell 7 and the bash subprocess spawned by Claude Code.

Changing ownership of the plugin directories (`BUILTIN\Administrators` → current user) did not resolve the issue.

**Workaround**

Disabling the plugin in `settings.json`:
```json
```

**Expected behavior**

The hook should execute successfully on Windows with Git Bash available in PATH.

kinyenzm:
## Summary

On Windows (Git-Bash / MSYS / Cygwin) the plugin-root resolution prelude embedded in the hooks and the MCP launcher fails with:

```
/usr/bin/bash: line N: printf: write error: Permission denied
```

…which aborts the `UserPromptSubmit` (and other) hooks. This is **not** a path/username encoding problem (it reproduces on plain ASCII usernames too) — it's a pipe-lifetime race.

## Root cause

The candidate-enumeration block pipes a producer subshell into a `while` loop that `break`s on the first match:

```sh
_P=$({ [ -n "$_E" ] && printf '%s\n' "$_E"; ls -dt "$_C/.../claude-mem"/[0-9]*/ 2>/dev/null; printf '%s\n' "$_C/.../marketplaces/thedotmack/plugin"; } \
  | while IFS= read -r _R; do ...; [ -f "$_Q/scripts/X" ] && { printf '%s\n' "$_Q"; break; }; done)
```

When `CLAUDE_PLUGIN_ROOT` is set (the host injects it in hook context), the **first** candidate matches and `break` fires immediately. That closes the read end of the pipe while the producer subshell is still trying to write the remaining candidate lines (`ls -dt` + the marketplace `printf`). The next write hits a broken pipe.

On Linux a write to a broken pipe is `EPIPE` ("Broken pipe") / `SIGPIPE`. **On Cygwin/MSYS the same write returns `EACCES`**, which `printf` reports as `write error: Permission denied`. Hence the misleading message — and why issue #2709 mis-attributed it to GBK/UTF-8 username encoding.

Reproduced **40/40** on `GNU bash 5.3.9 (x86_64-pc-cygwin)` once `CLAUDE_PLUGIN_ROOT` is exported.

## Fix

Don't `break`. Drain every candidate (there are only a handful) and print the **first** match exactly once via a `_F` guard:

```sh
... [ -f "$_Q/scripts/X" ] && [ -z "$_F" ] && { _F=1; printf '%s\n' "$_Q"; }; done
```

The producer subshell always runs to completion, so no write ever lands on a closed pipe. The first match still wins → the contractual fallback **ORDER is unchanged**. The change is POSIX-clean (no bashisms), so the `mcp` host's `sh -c` launcher is fixed by the same edit.

Patched the single source of truth (`src/build/hook-shell-template.ts`) and regenerated the three host-managed files.

## Files

- `src/build/hook-shell-template.ts` — generator (`candidateBlock`) + rationale comment
- `plugin/hooks/hooks.json` — regenerated (7 command strings)
- `plugin/hooks/codex-hooks.json` — regenerated (7 command strings)
- `plugin/.mcp.json` — regenerated (mcp-search launcher)

## Verification

- **Canonical generator check** (the invariant `tests/infrastructure/plugin-distribution.test.ts` enforces): all **15** command strings across the three files match `buildShellCommand()` output **byte-for-byte** after the edit.
- **Runtime**, with `CLAUDE_PLUGIN_ROOT` set, node/exec calls neutered, on Cygwin bash + sh:
  - `hooks.json` SessionStart command (bash): **0/30** EACCES, `_P` resolves correctly (cygpath-converted Windows path).
  - `.mcp.json` mcp-search command (sh): **0/30** EACCES, `_P` resolves correctly (POSIX path, no cygpath — matches existing behavior).
- JSON validity confirmed for all three files.

Fixes #2707. Related #2709.


kinyenzm:
### Relationship to the other open Windows hook PRs

Flagging overlap so this doesn't get lost in the Windows cluster (cc the canary #2699):

- **#2654** and **#2526** target the *same* `printf: write error: Permission denied`, but take a **suppress** approach — they add `} 2>/dev/null |` to the producer block. That hides the stderr, but the early `break` still closes the pipe read-end while the producer is mid-write, so the broken-pipe write *still happens* — it's just silenced. #2654 is `hooks.json` only; #2526 also covers codex + `.mcp.json` and adds a test.
- **#2618** is a different goal (moving the MCP resolver into a launcher file to clean up `claude mcp list` output), not the EACCES bug.

This PR (#2725) is **root-cause** instead of suppression: drop the `break`, drain every candidate, and print the first match exactly once via a `_F` guard. The producer subshell always runs to completion, so no write ever lands on a closed pipe — nothing to suppress. First match still wins, so the contractual fallback order is unchanged, and it's POSIX-clean so the `sh -c` MCP launcher is fixed by the same change.

One thing worth calling out for whoever merges the hook fix: the three host-managed files are generated from `src/build/hook-shell-template.ts`, and `tests/infrastructure/plugin-distribution.test.ts` asserts they match the generator byte-for-byte. #2725 patches the generator and regenerates all three (`hooks.json`, `codex-hooks.json`, `.mcp.json`), so that invariant holds. A hand-edit of the JSON alone (without the generator) would trip that test. I verified all 15 generated command strings still match byte-for-byte, and that the commands resolve with 0 EACCES under both bash and sh (including with `_F` inherited as non-empty).

Happy to rebase this onto **#2699** (or fold the generator change in there) if that's the preferred path for landing the Windows fixes together — just let me know.

thedotmack:
Merged via #2865 — your `_F` first-match guard landed in the canonical template (src/build/hook-shell-template.ts) with hook JSONs regenerated; the .mcp.json hunk was superseded by the #2807 Node launcher. This was the right root-cause fix where the earlier printf-suppression approach (a19b0e3c2, reverted in 4d8129386) wasn't. Thanks!

kinyenzm:
Appreciate the clear writeup, Alex — glad the `_F` guard was the right cut. Folding the `.mcp.json` resolution into the #2807 Node launcher makes sense; one less generated surface to keep byte-for-byte in sync with the template is a net win.

Nothing further needed on my end. If the remaining Windows cluster (#2699 / #2709) still wants eyes on the Cygwin EACCES-vs-EPIPE behavior, happy to help.

PR #2728, opened 2026-05-31
fix(corpus): camelCase params + obs_type key routing; clarify MCP tool descriptions

Kirchlive:
## Summary

Two independent fixes in the MCP corpus pipeline plus 14 tool-description rewrites, all uncovered by a systematic smoke-test of all 21 MCP tools on a v13.3.0 install (worker runtime).

### Fixes (commit 1)

**Defect 1 — camelCase `dateStart`/`dateEnd` silently dropped**
- MCP tool surface (`src/servers/mcp-server.ts`) advertises camelCase.
- Worker REST route (`src/services/worker/http/routes/CorpusRoutes.ts`) destructures only snake_case (`date_start`/`date_end`) from `req.body`.
- Zod's `.passthrough()` keeps the unknown camelCase keys on the body but the handler never reads them — date filters are silently dropped, no error.
- **Fix:** declare both casings on the Zod schema, prefer snake_case at destructure with a camelCase fallback.

**Defect 2 — multi-type filter collapses to single type**
- `src/services/worker/knowledge/CorpusBuilder.ts:42` set `searchArgs.type = filter.types.join(',')`.
- But `type` is the search-router discriminator (`observations` | `sessions` | `prompts`), not the observation-type filter.
- `WHERE type = 'bugfix,decision'` matches zero rows; downstream array-hydrate masks the failure and returns only entries that survived a different filter pass — hence the user-visible "only the first type came back" symptom.
- **Fix:** route the comma-joined types through `searchArgs.obs_type` (the correct observation-type filter key).

### Docs (commit 2)

14 MCP tool-description rewrites to prevent first-use misuse:

- All **8 server-beta-only tools** prefixed with `[Server-beta runtime only — DISABLED in default "worker" runtime.]` plus a one-line pointer to the worker-runtime equivalent. The transport already returns this error at call-time; surfacing it in the description lets the LLM pick the right tool the first time.
- **`search.obs_type`** now warns about the FTS5 type-token trap: `query="bugfix" + obs_type="bugfix"` returns 0 because the FTS5 index covers title/subtitle/narrative/text/facts/concepts but not the `type` column. Use one or the other for type-token queries.
- **`prime_corpus` / `query_corpus` / `rebuild_corpus` / `reprime_corpus`** preconditions strengthened — explicit that `query_corpus` errors when unprimed, that rebuild does not reprime, and that responses are LLM-generative.
- **`build_corpus`** description lists canonical types and emphasizes verifying `stats.observation_count` before priming.

## Test plan

- [x] Reproduced both bugs against unpatched v13.3.0 worker:
  - `build_corpus(types="bugfix,decision", dateStart="2026-05-01", dateEnd="2026-06-01", limit=50)` → **3 obs** (bugfix only, dates silently dropped, no error)
- [x] Applied both fixes to bundled worker-service.cjs via patch-on-bundle (same surgical-edit strategy as `plugin-hook-perf-patch.v2`)
- [x] Restarted worker process
- [x] Re-ran the same call → **14 obs** (8 decision + 6 bugfix; `filter` object in response now contains `date_start`/`date_end` correctly)
- [x] All 38 unit tests of Sprint-2 infrastructure still pass on the patched install
- [x] All 21 MCP tools systematically smoke-tested post-patch: 13 functional, 8 expected server-beta runtime errors (correct + clearer error message)
- [x] Independent `node --check` syntax verification of patched bundles
- [x] `node --check src/servers/mcp-server.ts` clean (no TS-source diagnostics introduced)
- [x] Backups preserved (`.sprint3-bak`) and rollback path tested

## Out of scope (filed separately)

- Bug A: FTS5 `type` column not in the index, causing `search(query="<typename>", obs_type="<same>")` to AND-fold to 0 results. Workaround in this PR via `search.obs_type` description update; structural fix proposed in a separate issue with TS-source diff.
- 598 stale `pending_messages` (worker consumer side); orthogonal to this PR.

### DIFF
--- a/src/servers/mcp-server.ts
+++ b/src/servers/mcp-server.ts
@@ -524,7 +524,7 @@ NEVER fetch full details without filtering first. 10x token savings.`,
   // MCP clients keep working without rewrites. (Plan line 753.)
   {
     name: 'observation_add',
-    description: 'Insert a manual observation directly into server-beta storage. Calls /v1/memories — does NOT enqueue generation. Server-beta runtime only. Params: content (required), projectId (optional, falls back to settings), serverSessionId, kind, metadata.',
+    description: '[Server-beta runtime only — DISABLED in default "worker" runtime.] Insert a manual observation directly into server-beta storage (/v1/memories — does NOT enqueue generation). In worker runtime use the hook pipeline (hook-client.mjs) instead. Params: content (required), projectId (optional, falls back to settings), serverSessionId, kind, metadata.',
     inputSchema: {
       type: 'object',
       properties: {
@@ -541,7 +541,7 @@ NEVER fetch full details without filtering first. 10x token savings.`,
   },
   {
     name: 'observation_record_event',
-    description: 'Record an agent event into server-beta. Calls /v1/events — server inserts the event row, the outbox row, and enqueues a generation job atomically. Server-beta runtime only.',
+    description: '[Server-beta runtime only — DISABLED in default "worker" runtime.] Record an agent event (/v1/events) — server inserts event row, outbox row, and enqueues a generation job atomically. Worker runtime: use the hook pipeline (hook-client.mjs) instead.',
     inputSchema: {
       type: 'object',
       properties: {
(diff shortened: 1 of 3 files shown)


Kirchlive:
Superseded by #2731 — both commits (camelCase fix + tool descriptions) folded into the combined Sprint 1-3 PR per maintainer-friendly bundling. Branch `sprint-3-mcp-fixes` preserved on the fork in case you want to compare.

PR #2737, opened 2026-06-01
feat: named environments with path patterns for shared memory

fish0710:
## Summary

- Add `environments` configuration to `~/.claude-mem/settings.json` that maps glob path patterns to environment names, replacing `path.basename(cwd)` as the `project` field when matched
- All sessions whose cwd matches the same environment automatically share memories across directories (e.g., `~/company-a` and `~/company-b` both become project `"work"`)
- Fix 5 bypass points that used `path.basename` directly instead of `getProjectContext`, plus 2 worker fallback fixes
- Add `merge-environment` CLI command for SQLite data migration + Chroma resync trigger
- Add environment name hint in context injection header so LLM knows which project name to use for manual searches

## Key changes

**Core logic** (`src/utils/project-name.ts`):
- `matchEnvironment(cwd)` — picomatch-based pattern matching with `~` expansion and `fs.realpathSync` cwd normalization (ENOENT-safe)
- `loadEnvironments()` — 100ms debounce + mtime-based sync caching for settings reads
- `getProjectName()` — inserts env match before basename fallback

**Bypass fixes** (5 files):
- `user-message.ts` — `basename(input.cwd)` → `getProjectContext(cwd).primary` + `projects` param
- `claude-md-commands.ts`, `CursorHooksInstaller.ts` (×2), `WindsurfHooksInstaller.ts`

**Worker fallback fixes** (2 files):
- `SearchManager.ts`, `store.ts` — use `path.basename(process.cwd())` instead of `getProjectContext(process.cwd())` (no user cwd context available in worker process)

**Context injection** (`HeaderRenderer.ts`):
- `[project-name (environment, paths: pattern1, pattern2)]` hint for LLM

**Migration** (`merge-environment.ts`):
- `npx claude-mem merge-environment --name <env> --from <project1,project2>` for SQLite migration
- Prints worker restart reminder for Chroma resync

**Not changed**: `ProcessManager.ts` — uses different path source (`git rev-parse --show-toplevel`), runs in synchronous bun:sqlite transaction, own remapping mechanism

## Configuration example

```json
{
  "environments": [
    { "name": "work", "patterns": ["~/workspace/harness/*", "~/company-*"] },
    { "name": "personal", "patterns": ["~/projects/my-*"] }
  ]
}
```

- First match wins (ordered by array position)
- No match → falls back to `path.basename(cwd)` (current behavior unchanged)
- `*` matches one directory level, `**` matches any depth

## Test plan

- [x] 8 environment matching tests covering: no env fallback, pattern match, first-match-wins, unmatched fallback, `~` expansion, `*` one-level, `**` any depth, empty/null cwd
- [x] `getProjectContext()` inherits env matching through `getProjectName()` — worktree overlay unchanged
- [x] Worker fallback uses raw basename (no env inference on unknown directories)
- [x] Settings cache with 100ms debounce + mtime-based invalidation
- [ ] Manual: configure environments in settings.json, verify cross-directory memory sharing
- [ ] Manual: run `npx claude-mem merge-environment` and verify SQLite + Chroma migration


fish0710:
## Follow-up fix: tests no longer mutate the user's real `settings.json`

Caught while testing this branch: the new `tests/environment-matching.test.ts` was writing and clearing `environments` on `~/.claude-mem/settings.json` directly. A normal `bun test` run would silently wipe the user's `environments` config because `environments=[]` falls back to `path.basename(cwd)` and the suite never failed.

**Commits on top of the original 9:**

`6087a1af fix: tests no longer mutate the user's real settings.json`

**Changes:**
- `src/utils/project-name.ts` — added `setEnvironmentsSettingsPathForTesting` so `loadEnvironments()` can be pointed at a temp file
- `tests/environment-matching.test.ts` — uses `mkdtempSync` + `beforeAll`/`afterAll` for an isolated settings file, plus a regression test that fails if the real `~/.claude-mem/settings.json` is read or written by the suite

**Verified:** All 9 tests pass; before/after a test run, the real settings.json is byte-identical.

fish0710:
## Fixed all 3 Greptile issues

Pushed `2d6114b2` addressing the three review threads:

**P1 #1 — `JSON.parse(array)` silently disables the feature (src/utils/project-name.ts:60)**
The bug: `SettingsDefaultsManager.loadFromFile` JSON.parses the on-disk file, so the typed `string` field could arrive as a native array. `JSON.parse(array)` coerces to `"[object Object]"` and throws, the catch sets `cachedEnvironments = []`, and matching is silently off. Real users following the documented array form would have hit this.
Fix: accept both shapes.
```ts
cachedEnvironments = Array.isArray(raw)
  ? (raw as Environment[])
  : (typeof raw === 'string' && raw ? JSON.parse(raw) : []);
```
Verified locally: `JSON.parse([{...}].toString())` → `SyntaxError: Unexpected identifier "object"` (confirmed), so the diagnosis was correct. Added two regression tests pinning both forms (string and array). My own real settings.json uses the string form, and both forms now return `work` for the same matching cwd.

**P1 #2 — Non-atomic multi-table migration (src/npx-cli/commands/merge-environment.ts)**
Wrapped the five UPDATEs in `db.transaction(() => { ... })`. A crash mid-migration now rolls back to the pre-migration state instead of leaving some tables renamed and others holding the old project names.

**P2 #3 — `split('=')[1]` truncates values containing `=`**
Replaced with `.slice('--name='.length)` / `.slice('--from='.length)` so `--name=work=prod` round-trips intact.

All 11 environment-matching tests pass.

thedotmack:

> step 1–2 adopt a `.claude-mem-project` marker + `project alias|merge` instead of glob `environments`; open P1 (partial migration commit) still applies

thedotmack:
Second pass — routing holds at **#3608 (plan-20, Project Identity & Injection Scope)**. The five `path.basename` bypass points you found are exactly the "one resolver shared by capture, sync and retrieval" defect the master exists to fix, and that part of the diff is valuable independent of the feature. Not a duplicate of #2827, which derives a stable slug from the git remote — different strategy for a neighbouring problem. The design divergence noted earlier still stands: steps 1-2 adopt a `.claude-mem-project` marker plus `project alias|merge` rather than glob `environments`, and the partial-migration P1 is open, so this needs reworking against that shape.

PR #2739, opened 2026-06-01
fix(worker): preserve init prompt during conversation history truncation

### LINKED ISSUE #2738: OpenRouter/Gemini history truncation can drop init prompt and XML output instructions
## Before submitting

- [x] I searched existing issues and confirmed this is not a duplicate

---

## Bug Description

`OpenRouterProvider` and `GeminiProvider` keep only the most recent conversation messages when truncating provider history. Once the configured message/token limits are exceeded, this can drop `history[0]`.

`history[0]` is the init prompt containing the worker's system identity, output-format instructions, and XML observation template.

## Impact

When the init prompt is dropped, the provider may return prose instead of the expected XML observation format. The XML parser rejects that output, causing observations to be ignored and the provider session to degrade or require respawn.

## Steps to Reproduce

1. Use the OpenRouter or Gemini provider.
2. Keep a worker/provider session alive long enough for its conversation history to exceed the configured message or token limit.
3. Trigger another observation or summary generation.
4. The provider history truncation keeps recent messages and can drop the first message (`history[0]`).
5. Subsequent provider responses may omit the expected XML observation format because the XML instructions were removed from context.

## Expected Behavior

The init prompt should always be preserved during truncation, and recent conversation turns should be fit into the remaining message/token budget.

## Root Cause

Both providers truncate history by iterating backward from the newest message and discarding older messages once limits are reached. Because the init prompt is the oldest message, it can be dropped even though it contains required formatting instructions.

## Environment

- **Claude-mem version**: current `main`
- **Provider(s)**: OpenRouter-compatible provider and Gemini
- **Platform**: provider-independent

## Logs

Representative symptom when the init prompt/format instructions are absent:

```text
OpenRouter returned non-XML prose response
```

## Suggested Fix

Extract shared truncation logic for OpenRouter and Gemini that:

- always preserves `history[0]`
- applies message/token limits to the full returned history, including the init prompt
- retains the newest recent messages in chronological order after the init prompt
- logs when truncation falls back to the init prompt only

## Additional Context

`ClaudeProvider` is not affected because it uses the Claude SDK's own conversation management rather than this provider-local history truncation path.

tbui17:
Fixes #2738

## Summary

- OpenRouter and Gemini conversation history truncation kept the newest turns by walking backward through history, which could drop `history[0]` once the configured message or token limit was exceeded.
- `history[0]` is the init prompt containing the worker's system identity, output-format instructions, and XML observation template. Dropping it can make providers return prose that the XML parser rejects.
- This PR extracts shared truncation logic and guarantees the init prompt is preserved while fitting recent turns into the remaining configured message/token budget.

## Changes

- `src/services/worker/truncate-history.ts`: add shared `truncateConversationHistory` helper that preserves `history[0]`, applies `maxContextMessages` and `maxEstimatedTokens` to the full returned history, and warns when truncation falls back to the init prompt only.
- `src/services/worker/OpenRouterProvider.ts`: delegate history truncation to the shared helper while keeping OpenRouter-specific settings lookup.
- `src/services/worker/GeminiProvider.ts`: delegate history truncation to the shared helper while keeping Gemini-specific settings lookup.
- `tests/worker/truncate-history.test.ts`: cover empty/single history, full-history fast path, message-limit truncation, token-limit truncation, init-only fallback, warning logs, and chronological ordering.
- `plugin/scripts/worker-service.cjs`, `plugin/scripts/server-beta-service.cjs`, and `plugin/ui/viewer-bundle.js`: regenerated build artifacts from `npm run build`.

## Test plan

- [x] `bun test tests/worker/truncate-history.test.ts` — 9 pass
- [x] `bun x tsc --noEmit` — passes
- [x] `npm run build` — passes


thedotmack:
Closing as superseded by #3151, which bounds OpenAI-compatible history without dropping or reordering messages and preserves the init prompt by construction.

PR #2741, opened 2026-06-01
feat: skip/filter subagent observations (Dynamic Workflows `workflow-subagent`) — #2736

### LINKED ISSUE #2736: feat: skip/throttle subagent observations (esp. Dynamic Workflows `workflow-subagent`) — re-raise of #2303
## Summary

Re-raising the core of #2303 (*support filtering/skipping subagent observations*), which was closed by consolidating into #2380 (plan-05). However, **#2380 as shipped implemented Observer *SDK tool-permission enforcement* (security defense-in-depth), not observation-volume control.** As a result the actual volume-reduction feature requested in #2303 was effectively dropped and is **not present in v13.3.0 or v13.4.0**.

With Claude Code **Dynamic Workflows** (launched 2026-05-28 alongside Opus 4.8) this gap has become a serious, recurring problem.

## Why this is now urgent: Dynamic Workflows

Dynamic Workflows fan a single prompt out to **tens-to-hundreds of parallel subagents** (runtime caps at ~1,000 invocations, 16 concurrent), plus adversarial verification agents that iterate until convergence. See the Claude Code docs: https://code.claude.com/docs/en/workflows

Every tool call by every subagent fires a PostToolUse-style hook → claude-mem creates **one observation per tool call**, each routed through the configured provider (Gemini / Claude / OpenRouter) for AI analysis. A single workflow run therefore produces **hundreds to thousands of observations**.

## Real-world evidence

From my own DB after a single day that included one Dynamic Workflow run:

| `agent_type` | observations (one day) |
|---|---|
| **`workflow-subagent`** | **634** |
| main / null | 65 |

So **634 of 699 observations (≈91%)** in that day came from `workflow-subagent`. Concretely this caused:

1. **Provider quota exhaustion** — hundreds of back-to-back generation calls drove the provider into `429 quota exhausted`, after which the restart guard tripped (`session is dead, clearing pending and terminating`) and the rest of the run's observations were dropped anyway.
2. **Very low signal-to-noise** — intermediate steps from hundreds of verification/explore subagents are rarely worth persisting cross-session; the durable knowledge comes from the main session and a few write-capable agents.

## Current behavior (verified in v13.3.0 bundle)

- Subagent **summaries** are already skipped (`Skipping summary: subagent context detected`, and `handleSummarizeByClaudeId` returns `reason:"subagent_context"` when `agentId` is present).
- Subagent **observations** are **not** filtered: `queueObservation` forwards `agentId`/`agentType` but never skips on them.
- `grep` of the shipped `worker-service.cjs` finds **no** `CLAUDE_MEM_SKIP_SUBAGENT_OBSERVATIONS` / `CLAUDE_MEM_SKIP_AGENT_TYPES`. The only volume lever is `CLAUDE_MEM_SKIP_TOOLS` (by tool *name*, global — can't target subagents without also dropping main-session observations).

## Proposed solution (re-instating #2303 + Dynamic-Workflows awareness)

Two settings, filtered **before** the worker HTTP call (in the observation hook handler), so we save the request and the provider tokens:

```jsonc
// Global toggle — skip ALL subagent observations (any hook event carrying an agentId)
"CLAUDE_MEM_SKIP_SUBAGENT_OBSERVATIONS": "true"

// Per-type skip list (comma-separated agent_type values). Default empty.
"CLAUDE_MEM_SKIP_AGENT_TYPES": "workflow-subagent,Explore,Plan"
```

Priority logic (from #2303):
- If `CLAUDE_MEM_SKIP_SUBAGENT_OBSERVATIONS=true` → skip all subagent observations; ignore the per-type list.
- Else → skip only `agent_type` values in `CLAUDE_MEM_SKIP_AGENT_TYPES`.

Additional asks specific to Dynamic Workflows:
- Recognize the `workflow-subagent` `agent_type` explicitly. **Consider skipping `workflow-subagent` observations by default**, since a workflow can trivially emit 100s–1000s of them and exhaust provider quota.
- Optionally, a throughput guard (per-session/minute observation cap) so a runaway workflow can't drive the provider into 429 + restart-guard data loss.

## Suggested implementation point

Filter on `input.agentId` / `input.agentType` in the observation hook handler **before** `executeWithWorkerFallback` / the `/api/observations/by-claude-id` request — mirroring the existing `CLAUDE_MEM_SKIP_TOOLS` and `CLAUDE_MEM_EXCLUDED_PROJECTS` checks. This avoids the worker round-trip and the provider call entirely.

## Acceptance criteria

- [ ] `CLAUDE_MEM_SKIP_SUBAGENT_OBSERVATIONS=true` drops every observation carrying an `agentId` before any provider call.
- [ ] `CLAUDE_MEM_SKIP_AGENT_TYPES=workflow-subagent,Explore` drops only those types.
- [ ] Main-session observations are unaffected.
- [ ] Documented in settings reference; defaults preserve current behavior (except possibly defaulting `workflow-subagent` to skipped — open to maintainer preference).

## References

- Original request: #2303 (closed → "consolidated into #2380")
- #2380 (plan-05) actually shipped Observer *tool-permission enforcement*, not volume control; included in v13.4.0 via #2701.
- Related throughput work: #2167, #2463.

NRKZOZW:
Closes #2736.

## Problem

Claude Code **Dynamic Workflows** fan a single prompt out to tens-to-hundreds of parallel subagents (`agent_type: workflow-subagent`). Every tool call by every subagent fires a `PostToolUse` hook → claude-mem creates **one provider-analyzed observation per call**, so a single run emits **hundreds-to-thousands** of low-signal observations.

As the issue documents with real DB evidence (634 / 699 ≈ 91% of a day's observations came from `workflow-subagent`), this:

1. **Exhausts the provider's free-tier quota** (HTTP 429), after which the restart guard trips (`session is dead, clearing pending and terminating`) and the rest of the run's observations — including valuable **main-session** work — are dropped.
2. **Tanks signal-to-noise** — ephemeral verification/explore subagent steps are rarely worth persisting cross-session.

Subagent *summaries* are already skipped; subagent *observations* were not. The only existing lever (`CLAUDE_MEM_SKIP_TOOLS`) is global by tool name and can't target subagents without also dropping main-session observations.

## Solution

Two settings, filtered **before** any worker HTTP round-trip or provider request — so the request *and* the provider tokens are saved:

| Setting | Default | Behavior |
|---|---|---|
| `CLAUDE_MEM_SKIP_SUBAGENT_OBSERVATIONS` | `false` | When `true`, skip **every** observation carrying an `agentId` (any subagent). The robust, type-independent lever. |
| `CLAUDE_MEM_SKIP_AGENT_TYPES` | _(empty)_ | Comma-separated `agent_type` values to skip, e.g. `workflow-subagent,Explore`. Surgical control. |

The two combine as a **union** (skip if the global toggle matches **or** the `agent_type` is listed) — simpler than, and a strict superset of, the issue's "priority" framing, and it satisfies every acceptance-criteria bullet.

### Design decisions (beyond the issue text)

- **Defaults preserve current behavior**, honoring the issue's AC checkbox *"defaults preserve current behavior."* Rather than silently dropping data in the shipped default, `workflow-subagent` is **documented as the recommended value**. Flipping the default to skip `workflow-subagent` out of the box is a one-line change in `SettingsDefaultsManager` if you'd prefer that — I went conservative deliberately. (Note: a global `agentId` toggle is more robust than a type-string match, since a non-`workflow-subagent` burst trips the same restart guard.)
- **Two filtering points, one source of truth.** A new pure helper `src/shared/should-skip-agent-observation.ts` is called from:
  - the **hook handler** (`src/cli/handlers/observation.ts`) **ahead of the runtime branch**, so it covers **both** the `worker` and `server-beta` runtimes and avoids the round-trip entirely; and
  - the **worker ingest** path (`src/services/worker/http/shared.ts` `ingestObservation`) as **defense-in-depth** for any non-hook caller, before `queueObservation` (the provider call).
  This mirrors the existing dual-check pattern for `CLAUDE_MEM_EXCLUDED_PROJECTS`.

## Acceptance criteria

- [x] `CLAUDE_MEM_SKIP_SUBAGENT_OBSERVATIONS=true` drops every observation carrying an `agentId` before any provider call.
- [x] `CLAUDE_MEM_SKIP_AGENT_TYPES=workflow-subagent,Explore` drops only those types.
- [x] Main-session observations are unaffected.
- [x] Documented in the settings reference; defaults preserve current behavior.

## Tests & verification

- New unit tests for the helper (`tests/shared/should-skip-agent-observation.test.ts`): default no-skip, global toggle, per-type list, `agentType`-without-`agentId`, whitespace/dedupe parsing, union precedence.
- New hook-handler integration tests (`tests/cli/handlers/observation-subagent-skip.test.ts`): asserts the worker HTTP call / `recordEvent` is **not** made when skipped (would fail if the filter were removed), across **both** the worker and server-beta runtimes, and that main-session + non-listed types still flow.
- Locally ran the equivalent of CI (`ci.yml`): `npm run typecheck` clean · `npm run build` succeeds (bundle-size guardrails pass) · full `bun test` **2080 pass / 0 fail**. (Actions on a fork PR waits for a maintainer to approve the run.)
- Rebuilt plugin bundles are included (per repo convention).

## Out of scope / future work

An optional per-session/minute **throughput guard** (the issue's secondary ask) would defend against runaway bursts from *any* source, not just known agent types. It needs worker-side rate state and risks non-deterministic drops, so I left it out to keep this change deterministic and focused; happy to follow up.


NRKZOZW:
Context: #2736 was consolidated into the **plan-12** roadmap (#2785) under *Ingestion & filtering*. This PR is the concrete, tested implementation of that slice — I've flagged it on #2785 so it can serve as the base rather than a from-scratch reimplementation.

Happy to rebase or reshape to match the plan-12 slicing, and to address review feedback — including the per-type-vs-main-session semantics the automated review flagged (the per-type list matches on `agent_type` alone by design; I can tighten it or adjust the docs' wording, whichever you prefer) and the explicit `reason` fallback.

thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#2785 ([plan-12] Provider & Extensibility Roadmap)** as a **bundle candidate**: the fix is sound and is slated to be folded into the plan-12 bundle PR with author credit (`Co-authored-by` + `Closes` reference), sequenced by the step noted below. It stays open until that bundle lands.

> roadmap child #2736 (skip/throttle subagent observations); open P1: type-list must not drop main-session rows

thedotmack:
Second pass — routing holds at **#2785 (plan-12, Provider & Extensibility Roadmap)** as roadmap child #2736; skip/throttle knobs for subagent observations are ingestion filtering, which lives on the roadmap master rather than in the observer pipeline cluster. Nothing open duplicates it and no merged work adds either setting. This is the most complete implementation of that slice, and the 91%-of-a-day DB evidence makes the case; the one open P1 — the agent-type list must not be able to drop main-session rows — needs closing first.

PR #2751, opened 2026-06-02
fix(smart-file-read): use a JS-valid tree-sitter query for plain .js files

### LINKED ISSUE #2750: v12.6.0: plain .js/.mjs/.cjs unparseable — realized regression of #1654 (tree-sitter 0.26.9 rejects TS-only node types in shared jsts query)
# Bug: plain `.js` / `.mjs` / `.cjs` now fully unparseable — the regression predicted in #1654 has occurred (tree-sitter 0.26.9 hard-rejects TS-only node types in the shared `jsts` query)

**Repo:** https://github.com/thedotmack/claude-mem
**Version:** 12.6.0
**tree-sitter CLI:** 0.26.9
**Component:** MCP search server (`scripts/mcp-server.cjs`), tree-sitter smart-file tooling
**Platform:** macOS arm64, runtime = bun 1.3.14 (root cause is platform-independent)

## This is the realized regression from #1654
Issue #1654 (closed NOT_PLANNED, 2026-04-10) called this exactly:

> Relies on tree-sitter silently ignoring invalid patterns. **If tree-sitter changes to fail on unrecognized node types, all `.js` file parsing breaks.**

That has now happened. tree-sitter 0.26.9 no longer silently ignores unknown
node types — it aborts query compilation. So every plain-JS file now fails.
Note the first offender is actually `type_identifier` (used for the class name
in the shared query), not only the three declarations listed in #1654.

## Summary
Any plain JavaScript file (`.js`, `.mjs`, `.cjs`) silently returns
`Could not parse <file>. File may use an unsupported language or be empty.`
from `smart_outline` (and yields no symbols in `smart_search`). TypeScript and
JSX/TSX (`.ts`, `.tsx`, `.jsx`) work fine. The failure is total for vanilla-JS
codebases (e.g. legacy monoliths, Node CJS services).

## Root cause
The extension→grammar map sends plain JS to the `javascript` grammar:

```js
const EXT_MAP = { ".js":"javascript", ".mjs":"javascript", ".cjs":"javascript",
                  ".jsx":"tsx", ".ts":"typescript", ".tsx":"tsx", ... }
```

But all three of `javascript` / `typescript` / `tsx` share one tags query
(`group "jsts"`):

```scm
(function_declaration name: (identifier) @name) @func
(lexical_declaration (variable_declarator name: (identifier) @name
  value: [(arrow_function) (function_expression)])) @const_func
(class_declaration name: (type_identifier) @name) @cls          ; <-- TS-only
(method_definition name: (property_identifier) @name) @method
(interface_declaration name: (type_identifier) @name) @iface    ; <-- TS-only
(type_alias_declaration name: (type_identifier) @name) @tdef    ; <-- TS-only
(enum_declaration name: (identifier) @name) @enm                ; <-- TS-only
(import_statement) @imp
(export_statement) @exp
```

`tree-sitter query` validates the **entire** query against the grammar and
aborts if any node type is unknown. The pure `javascript` grammar has no
`type_identifier`, `interface_declaration`, `type_alias_declaration`, or
`enum_declaration`, so the query fails to compile and `execFileSync` throws. The
server catches it and returns the generic "Could not parse" message — masking the
real error.

## Reproduction
```bash
cd ~/.claude/plugins/cache/thedotmack/claude-mem/12.6.0
printf 'function f(){return 1}\nclass C{m(){}}\n' > /tmp/x.js

# extract the shared jsts query into /tmp/jsts.scm, then:
node_modules/.bin/tree-sitter query -p node_modules/tree-sitter-javascript /tmp/jsts.scm /tmp/x.js
#   Error: Query compilation failed
#   Caused by: Query error at 4:27. Invalid node type "type_identifier"

# same query against the tsx grammar succeeds:
node_modules/.bin/tree-sitter query -p node_modules/tree-sitter-typescript/tsx /tmp/jsts.scm /tmp/x.js
#   pattern 0, capture: func ...
```

Via the MCP tool: `smart_outline { file_path: "/any/plain.js" }` →
`Could not parse ...`.

## Fix (verified)
The `tsx` grammar is a superset that parses plain JS **and** JSX, and the `jsts`
query is fully valid against it. Route plain JS through `tsx`:

```diff
- ".js":"javascript", ".mjs":"javascript", ".cjs":"javascript",
+ ".js":"tsx", ".mjs":"tsx", ".cjs":"tsx",
```

After this, a real 10,755-line vanilla-JS monolith outlines correctly
(334 symbols) where it previously returned "Could not parse".

### Alternatives
- Keep the `javascript` mapping but give it a JS-specific query (class names are
  `(identifier)`, drop the `interface_/type_alias_/enum_declaration` patterns).
- Make the query runner tolerant: compile patterns individually and skip the ones
  the grammar rejects, instead of failing the whole query. This also future-proofs
  every other shared query/grammar pair.

## Secondary observation (not the cause, but worth a look)
`tree-sitter-css`'s ESM loader (`bindings/node/index.js`) does
`await import('…/<name>.node')`, which bun's runtime rejects
(`To load Node-API modules, use require()/process.dlopen`). This does **not**
affect the smart tools (they go through the `tree-sitter` CLI, not the package's
JS binding), but it will bite anything that `require()`s the grammar package
directly under bun. The `tree-sitter-scss` loader (plain CJS `node-gyp-build`)
is the bun-safe pattern.

Letsrollamigo:
## Problem

`smart_outline` / `smart_search` silently return **zero symbols for every plain `.js` / `.mjs` / `.cjs` file**. TypeScript and JSX/TSX work fine.

Root cause: `javascript`, `typescript` and `tsx` all share the `jsts` tree-sitter query, but that query references node types that exist only in the TypeScript grammar:

- `type_identifier` (used for the class name)
- `interface_declaration`
- `type_alias_declaration`
- `enum_declaration`

`tree-sitter` **0.26.x** no longer silently ignores unknown node types — it aborts query compilation:

```
Error: Query compilation failed
Caused by: Query error at 4:27. Invalid node type "type_identifier"
```

`runBatchQuery` catches the thrown CLI error and returns an empty `Map`, so the failure is completely silent — the file just looks "unparseable".

This is the exact fragility predicted in #1654 (closed `wontfix`): *"If tree-sitter changes to fail on unrecognized node types, all `.js` file parsing breaks."* It now has, in the wild. Reported as #2750.

## Fix

Add a dedicated `js` query and route the `javascript` language to it. Class names in `tree-sitter-javascript` are `(identifier)` (not `(type_identifier)`), and the TS-only declaration patterns are dropped. `typescript` and `tsx` keep using `jsts` unchanged.

```diff
   switch (language) {
     case "javascript":
+      return "js";
     case "typescript":
     case "tsx":
       return "jsts";
```

The `language` value stays `"javascript"`, so grammar resolution (`tree-sitter-javascript`) and `isExported()` are unaffected — only the query key changes.

14-line diff, no behavior change for TS/TSX.

## Verification

The new `js` query compiles cleanly against `tree-sitter-javascript` and captures every symbol kind:

```
$ tree-sitter query -p node_modules/tree-sitter-javascript js.scm sample.js
pattern 0  capture func        -> function foo
pattern 1  capture const_func  -> const bar = (n) => ...   /  const z = function(){...}
pattern 2  capture cls         -> class Baz
pattern 3  capture method      -> qux
pattern 4  capture imp         -> import x from "y"
pattern 5  capture exp         -> export ...
```

End-to-end via `smart_outline`: a real **10,755-line vanilla-JS file** went from `Could not parse` (0 symbols) to a full **334-symbol** outline.

## Alternatives considered

- **Map `.js` → `tsx` grammar.** `tsx` is a JS superset and the `jsts` query is valid against it; smallest change, but parses plain JS under the TSX dialect. Works, but a JS-specific query is cleaner.
- **Make `runBatchQuery` tolerant** — compile patterns individually, skip (and log) the ones a grammar rejects. Worth doing separately: it future-proofs every grammar/query pair and removes the *silent* empty-result failure mode that masked this bug.

Fixes #2750.

### DIFF
--- a/src/services/smart-file-read/parser.ts
+++ b/src/services/smart-file-read/parser.ts
@@ -273,6 +273,19 @@ const QUERIES: Record<string, string> = {
 (export_statement) @exp
 `,
 
+  // Plain JavaScript: the tree-sitter-javascript grammar has no type_identifier,
+  // interface_declaration, type_alias_declaration or enum_declaration nodes, so it
+  // cannot share the jsts query — tree-sitter aborts query compilation on the first
+  // unknown node type. Class names are (identifier) here, not (type_identifier).
+  js: `
+(function_declaration name: (identifier) @name) @func
+(lexical_declaration (variable_declarator name: (identifier) @name value: [(arrow_function) (function_expression)])) @const_func
+(class_declaration name: (identifier) @name) @cls
+(method_definition name: (property_identifier) @name) @method
+(import_statement) @imp
+(export_statement) @exp
+`,
+
   python: `
 (function_definition name: (identifier) @name) @func
 (class_definition name: (identifier) @name) @cls
@@ -422,6 +435,7 @@ const QUERIES: Record<string, string> = {
 function getQueryKey(language: string): string {
   switch (language) {
     case "javascript":
+      return "js";
     case "typescript":
     case "tsx":
       return "jsts";


thedotmack:
Thank you! Your plain-JS query fix was incorporated (with credit via a co-author trailer) into the plan-13 grammar PR #2805, which is now merged to main and closes #2750/#2786. Closing this as superseded.

Issue #1638, opened 2026-04-07
Add MCP tool to resolve memory_session_id → Claude Code session ID and name


skypper:
## Problem

When `get_observations()` returns results, each observation includes a `memory_session_id` (internal memory agent UUID). But this doesn't match the session ID or name visible in Claude Code — those are `content_session_id` and `custom_title` in `sdk_sessions`.

There is currently no way for Claude (or a user) to go from a memory result back to the originating Claude Code session.

**Example:**
A search returns observations with `memory_session_id: d308ed77-...`, but the actual Claude Code session was:
- Session name: `generic-detector-package`
- Session ID: `2bacd186-19e7-4c7c-a359-1a74593c2fa1`

These are completely different values with no way to connect them without querying the DB directly.

## Proposed Solution

A new MCP tool — `get_session(memory_session_id)` — that does a simple lookup in `sdk_sessions` and returns:

```json
{
  "content_session_id": "2bacd186-19e7-4c7c-a359-1a74593c2fa1",
  "session_title": "generic-detector-package",
  "project": "tenderly-core",
  "started_at": "2026-03-27T17:01:00Z"
}
```

Backed by a lightweight `/api/session/by-memory-id/:id` endpoint on the worker. Same pattern as existing MCP tools in `src/servers/mcp-server.ts`.

## Why a Separate Tool vs. Enriching `get_observations()`

- `get_observations()` stays focused — session metadata isn't always needed
- The lookup is explicit and on-demand — Claude calls it only when the user asks "which session was this from?"
- Cleaner token budget — no extra fields on every observation fetch

## Use Case

1. User asks Claude to find a past memory
2. Claude calls `search()` + `get_observations()` → finds relevant observations
3. User asks "which session was this from?"
4. Claude calls `get_session(memory_session_id)` → returns the Claude Code session ID and name the user can actually use to navigate back to that session

thedotmack:
Closed during the April 2026 backlog cleanup. The feature ask is now tracked in #1971, which is the single canonical issue for prioritization. Thanks for the suggestion — your context is captured there.

PR #2770, opened 2026-06-04
feat(sqlite): stamp content_session_id on generated observations and summaries

### LINKED ISSUE #2769: Stamp originating content_session_id onto generated session_summaries and observations rows at write time
## Before submitting

> Closest prior art: #1638 (`get_session` MCP tool) resolves `memory_session_id → content_session_id` via `sdk_sessions`, and #2671 stamps the session's project/folder onto generated metadata. Neither covers the per-window mapping this requests — see "Why existing work doesn't cover this" below.

---

A consumer holding a Claude Code `session_id` (what a hook receives as `content_session_id`) cannot reliably map it to that session's freshly generated `session_summaries` / `observations` rows.

Neither `observations` nor `session_summaries` carries `content_session_id`. Both reference only `sdk_sessions(memory_session_id)` (see `src/services/sqlite/schema.sql`, the `observations` and `session_summaries` table definitions and their `FOREIGN KEY(memory_session_id) REFERENCES sdk_sessions(memory_session_id)`). The only path back to a Claude Code session is `observations/summaries → sdk_sessions.memory_session_id → sdk_sessions.content_session_id`, and that path is unreliable for recent rows:

- A single Claude Code session fans out into multiple ephemeral SDK `memory_session_id`s over its lifetime (the observer respawns SDK sessions).
- `sdk_sessions.memory_session_id` is `UNIQUE`, and `ensureMemorySessionIdRegistered` (`src/services/sqlite/SessionStore.ts:1029`) overwrites it per `sessionDbId` on each storage round. So the freshest generated rows routinely have a `memory_session_id` that has no settled `sdk_sessions` row mapping it back to a `content_session_id`.
- Result: `LEFT JOIN sdk_sessions ON memory_session_id` returns `NULL` `content_session_id` for the most recent rows — exactly when a consumer wants them.

**Repro**

1. Run a Claude Code session that generates memory.
2. Immediately query the newest generated rows joined to the session map:
   ```sql
   SELECT s.id, s.request, sk.content_session_id
   FROM session_summaries s
   LEFT JOIN sdk_sessions sk ON s.memory_session_id = sk.memory_session_id
   ORDER BY s.created_at_epoch DESC LIMIT 5;
   ```
3. `content_session_id` is `NULL` for the freshest rows; the originating Claude Code session is unrecoverable at the moment a consumer would map it.

The data exists internally at generation time — it just isn't propagated onto the row. Confirmed against claude-mem v13.4.0.

Stamp the originating `content_session_id` (and ideally the transcript path) directly onto each generated `session_summaries` and `observations` row at write time, instead of relying on the lagging `sdk_sessions` join.

`content_session_id` is already in scope at the single storage call site. In `processAgentResponse` (`src/services/worker/agents/ResponseProcessor.ts`), `session.contentSessionId` is read repeatedly right where storage happens — it is passed to commit verification (line ~123), the observation/summary broadcasts (`session_id: session.contentSessionId`, lines ~328/409), and chroma sync — but it is not passed to `sessionStore.storeObservations(...)` (line ~152) or into the summary store. `ActiveSession.contentSessionId` is non-nullable (`src/services/worker-types.ts:11`), so it is always available here.

Concretely:

- Add a `content_session_id` column to `observations` and `session_summaries` (`src/services/sqlite/schema.sql` + a migration under `src/services/sqlite/migrations/`).
- Thread `session.contentSessionId` from `processAgentResponse` through `SessionStore.storeObservations` (`src/services/sqlite/SessionStore.ts:1867`) into `storeObservation` (`src/services/sqlite/observations/store.ts`) and `storeSummary` (`src/services/sqlite/summaries/store.ts`), persisting it in the `INSERT`.
- Optionally do the same for the transcript path / cwd (cwd is already carried on `PendingMessage.cwd` and `projectRoot` flows into `processAgentResponse`).

This lets a consumer map memory back to a specific Claude Code session/window directly off the row, with no dependency on `sdk_sessions` having settled. It enables use cases such as per-window terminal-title updates, where two windows on the same project/cwd must be told apart.

- **The `get_session` MCP tool (#1638)** resolves `memory_session_id → content_session_id` by looking it up in `sdk_sessions`. That is the same lagging join described above: for the freshest generated rows the `sdk_sessions` mapping has been overwritten, so the lookup returns no settled `content_session_id` at the moment a consumer needs it.
- **PR #2671** ("carry session folder onto generated observation metadata") stamps the session's **project/folder** — derived from cwd basename — onto generated observation/summary metadata. That is a project-level dimension, not a per-window identifier: two Claude Code windows open on the same project/cwd share the same folder label and remain indistinguishable. Mapping back to the specific originating session requires `content_session_id`, which #2671 does not add.

**Additional context**

Change sites for the proposed fix:

- `src/services/worker/agents/ResponseProcessor.ts` — `processAgentResponse`: storage call site; `session.contentSessionId` in scope but not passed to the stores.
- `src/services/sqlite/SessionStore.ts` — `storeObservations` (line 1867) and the summary store wrapper; the threading point.
- `src/services/sqlite/observations/store.ts` — `storeObservation`: observation `INSERT`.
- `src/services/sqlite/summaries/store.ts` — `storeSummary`: summary `INSERT`.
- `src/services/sqlite/schema.sql` — `observations` / `session_summaries` definitions (plus a migration).
- Context: `ensureMemorySessionIdRegistered` at `SessionStore.ts:1029` (the `UNIQUE memory_session_id` overwrite that makes the join lag); `ActiveSession.contentSessionId` at `worker-types.ts:11`.

Happy to open a PR for this if the approach looks right.

mairas:
Closes #2769.

## Problem

`observations` and `session_summaries` reference only `sdk_sessions(memory_session_id)`. The path back to a Claude Code session — `observations/summaries → sdk_sessions.memory_session_id → sdk_sessions.content_session_id` — is unreliable for recent rows: a single Claude Code session fans out into multiple ephemeral SDK `memory_session_id`s, and `ensureMemorySessionIdRegistered` overwrites the `UNIQUE memory_session_id` per `sessionDbId` each storage round. So a `LEFT JOIN sdk_sessions` returns `NULL` `content_session_id` for the freshest generated rows — exactly when a consumer wants them (e.g. per-window terminal-title updates, where two windows on the same project/cwd must be told apart).

The data exists at generation time; it just wasn't propagated onto the row.

## Change

Stamp the originating `content_session_id` directly onto each generated `observations` and `session_summaries` row at write time.

- **Schema:** new nullable `content_session_id TEXT` column + index on both tables.
- **Migration 35:** added in *both* schema systems — `SessionStore`'s own initializer sequence (this is the one the worker actually runs: `DatabaseManager.initialize` → `new SessionStore(db)`) and `migrations/runner.ts` `MigrationRunner` (used by `ClaudeMemDatabase`, and kept in sync for parity). `schema.sql` is the reference doc and is updated too. Idempotent PRAGMA-guarded `ALTER`, matching the existing `metadata` (v30) / `content_hash` precedent that already lives in both.
- **Write paths:** an optional trailing `contentSessionId` param threaded through every function that INSERTs into these tables — `SessionStore.storeObservation/storeSummary/storeObservations/storeObservationsAndMarkComplete`, the `transactions.ts` free functions, and the `observations/store.ts` / `summaries/store.ts` helpers. Backward compatible: existing call sites and tests are untouched.
- **Live pipeline:** `processAgentResponse` passes `session.contentSessionId` (non-nullable, already in scope) into `storeObservations`.
- **Types:** `ObservationRecord` / `SessionSummaryRecord` gain the field.

### Scope

This PR is the **write side only** — it stamps the column. Reading it back:

- **Direct SQL** (the use case #2769 describes — a hook/consumer holding a Claude Code `session_id`, querying the rows): works now. The column is populated and indexed.
- **claude-mem's own read APIs** (MCP `get_observations`, search result formatters, Chroma sync metadata): deliberately **not** changed here — they still omit `content_session_id`. `getObservationById`/`getObservationsByIds` surface it via `SELECT *` since the record type now carries it, but the higher-level serializers don't project it.

Exposing it through those read surfaces is a clean **additive, non-breaking** follow-up (each is an added key/field consumers can ignore) and is genuinely separate work from the write path — happy to file a follow-up issue + PR if that exposure is wanted.

Other writers — `/api/memory/save` (a *manual* session, no Claude Code `content_session_id`), the historical XML import CLI, and the test-only free functions — pass nothing and the column stores `NULL`. That's by nature, not a gap: those rows have no originating window. The param is threaded everywhere so a future caller with a session id can populate it without touching these functions again.

Also out of scope: backfilling existing rows; the optional transcript-path column floated in #2769; Postgres server-beta parity (separate runtime/storage that doesn't use this write path).

### Why existing work doesn't cover this

- **#1638** (`get_session` MCP tool) resolves `memory_session_id → content_session_id` via `sdk_sessions` — the same lagging join that's `NULL` for fresh rows.
- **#2671** stamps the session's project/folder (cwd basename), a project-level dimension; two windows on the same project share it and stay indistinguishable.

## Tests

TDD. Full suite **2074 pass / 0 fail**, `tsc --noEmit` clean.

- **Live wiring:** `response-processor.test.ts` asserts `processAgentResponse` forwards `session.contentSessionId` into `storeObservations` (the one production behavior change).
- **Migration, production system:** `session_store.test.ts` re-opens a `SessionStore` on a DB whose column was dropped (simulating a pre-v35 database / worker restart) and asserts columns, indexes, and version 35 are backfilled idempotently.
- **Migration, `MigrationRunner`:** fresh DB has the column + indexes + version 35 on both tables; a legacy DB missing the column is backfilled (columns and indexes recreated).
- **Write paths:** live path stamps both observation and summary rows; free-function paths persist the value when supplied and store `NULL` when omitted.

CHANGELOG.md is intentionally not edited — the repo's CLAUDE.md states the changelog is generated automatically. Build artifacts under `plugin/scripts/` are likewise omitted; they're regenerated at release (last touched by the v13.4.0 version bump).


mairas:
Issue #2769 was closed as not planned, so this PR should probably be closed as well? Leaving it up to the maintainer.

thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3609 ([plan-21] SQLite Schema Evolution & Queue State Integrity)** as **needs rework**: the diagnosis is right, but the item noted below blocks it. If you'd like to keep it moving, address that against the design in the plan doc; otherwise the fix will be absorbed into the plan-21 bundle.

> step 3 (immutable session keys) fixes the same join defect by making memory_session_id a plain mutable column; this PR edits files deleted on main

thedotmack:
Second pass — this routes to **#3609 (plan-21, SQLite Schema Evolution & Queue State Integrity)**, whose step 3 fixes the same broken join by making `memory_session_id` an ordinary mutable column instead of an overwritten UNIQUE key, rather than by denormalizing a second session id onto every generated row. Closing on two grounds you already anticipated: the parent issue #2769 was closed as not planned, and the diff no longer applies — `src/services/sqlite/migrations/runner.ts` and `src/services/sqlite/transactions.ts` have both been deleted from main since this was written. The diagnosis was good and it is recorded on the master; thanks for writing it up so clearly.

PR #2777, opened 2026-06-04
feat(context): add opt-in Mermaid task-flow diagram for token-efficient context injection

destiny520537work-lab:
## Summary

Adds an opt-in Mermaid graph renderer that replaces the prose observation list with a compact task-flow diagram when injecting context at session start.

Enable with:

```json
// ~/.claude-mem/settings.json
{ "CLAUDE_MEM_MERMAID_CONTEXT": "true" }
```

### What it looks like

Instead of listing each observation as a text block, the last session's observations are rendered as a directed graph:

\`\`\`mermaid
graph LR
    N0["🔴 Fixed JWT token expiry\nauth/jwt.ts"]
    N1["🟣 Added OAuth2 PKCE flow\nauth/oauth.ts"]
    N2["🔵 Discovered rate-limit header\nconfig/limits.ts"]
    N0 --> N1
    N1 --> N2
    NEXT(["Next: Test with real OAuth providers"])
    N2 --> NEXT
    style N0 fill:#fed7d7,color:#1a202c
    style N1 fill:#e9d8fd,color:#1a202c
    style N2 fill:#dbeafe,color:#1a202c
    style NEXT fill:#bee3f8,color:#1a202c
\`\`\`

### Why this helps

A session with 10 observations typically costs **1,500–5,000 tokens** when rendered as prose narrative. The equivalent Mermaid graph costs **~150–300 tokens** — roughly a 10× reduction — while preserving:

- Causal sequence (what happened in what order)
- File context (which files were touched per observation)
- Type information (bugfix / feature / discovery / decision)
- Next steps from the session summary

This is particularly valuable when a tool call returns a very large output (e.g. scanning a large JSON file): instead of storing a compressed prose summary, the AI can extract the key findings as graph nodes and discard the raw output.

### Changes

| File | Change |
|------|--------|
| `src/services/context/sections/MermaidRenderer.ts` | **New** — renders observations into a `graph LR` Mermaid diagram |
| `src/services/context/types.ts` | Add `mermaidContext: boolean` to `ContextConfig` |
| `src/services/context/ContextConfigLoader.ts` | Read `CLAUDE_MEM_MERMAID_CONTEXT` setting |
| `src/shared/SettingsDefaultsManager.ts` | Register new setting with default `'false'` |
| `src/services/context/ContextBuilder.ts` | Call `renderMermaidFlow` after header when enabled |

### Design decisions

- **Opt-in only** (`false` by default) — zero impact on existing users
- **Agent mode only** — `forHuman: true` (Web Viewer) skips Mermaid; the viewer already has a visual timeline
- **Most-recent session only** — older sessions remain in the text timeline; the graph focuses on immediate resumption context
- **Title field, not narrative** — keeps nodes compact (≤60 chars); full narrative still available via MCP `get_observations`
- **next_steps as terminal node** — surfaces the session summary's `next_steps` as the graph's exit node, giving the model an immediate action target

### Addresses

The token-efficiency concern raised in community discussions about large tool-call outputs (e.g. "what if a tool call returns 1M lines of JSON"). This renderer provides a structured extraction path: key findings become graph nodes, raw output is not stored in context.

### DIFF
--- a/src/services/context/ContextBuilder.ts
+++ b/src/services/context/ContextBuilder.ts
@@ -23,6 +23,7 @@ import { renderHeader } from './sections/HeaderRenderer.js';
 import { renderTimeline } from './sections/TimelineRenderer.js';
 import { shouldShowSummary, renderSummaryFields } from './sections/SummaryRenderer.js';
 import { renderPreviouslySection, renderFooter } from './sections/FooterRenderer.js';
+import { renderMermaidFlow } from './sections/MermaidRenderer.js';
 import { renderAgentEmptyState } from './formatters/AgentFormatter.js';
 import { renderHumanEmptyState } from './formatters/HumanFormatter.js';
 
@@ -73,8 +74,13 @@ function buildContextOutput(
   const output: string[] = [];
 
   const economics = calculateTokenEconomics(observations);
+  const mostRecentSummary = summaries[0];
 
   output.push(...renderHeader(project, economics, config, forHuman));
+
+  if (config.mermaidContext && !forHuman) {
+    output.push(...renderMermaidFlow(observations, mostRecentSummary));
+  }
 
   const displaySummaries = summaries.slice(0, config.sessionCount);
   const summariesForTimeline = prepareSummariesForTimeline(displaySummaries, summaries);
@@ -83,7 +89,6 @@ function buildContextOutput(
 
   output.push(...renderTimeline(timeline, fullObservationIds, config, cwd, forHuman));
 
-  const mostRecentSummary = summaries[0];
   const mostRecentObservation = observations[0];
 
   if (shouldShowSummary(config, mostRecentSummary, mostRecentObservation)) {
--- a/src/services/context/ContextConfigLoader.ts
+++ b/src/services/context/ContextConfigLoader.ts
@@ -25,5 +25,6 @@ export function loadContextConfig(): ContextConfig {
     fullObservationField: settings.CLAUDE_MEM_CONTEXT_FULL_FIELD as 'narrative' | 'facts',
     showLastSummary: settings.CLAUDE_MEM_CONTEXT_SHOW_LAST_SUMMARY === 'true',
     showLastMessage: settings.CLAUDE_MEM_CONTEXT_SHOW_LAST_MESSAGE === 'true',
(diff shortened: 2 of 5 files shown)


destiny520537work-lab:
Thanks for the thorough review, Greptile.

Both issues are now fixed:

**P1 — ANSI color constants corrupted in types.ts**

Root cause: shell variable expansion in the initial commit interpreted the literal `\x1b` text as actual ESC bytes. Fixed by switching to byte-level replacement (Python `bytes.replace()` on the raw base64-decoded content), which preserves the 4-character escape sequence unchanged. Verified: all 10 `\x1b` literals in `colors` are intact in the new commit.

**Session mismatch — next_steps from wrong session**

Fixed by adding a session guard before using the summary:

```typescript
// Only use next_steps from the summary that belongs to this session.
// summaries[0] may come from a different session if this session produced no summary.
const sessionSummary = summary?.memory_session_id === latestSessionId ? summary : undefined;
```

`sessionSummary` is then used in place of `summary` for the NEXT terminal node. If the most recent session has no summary, no terminal node is rendered.

destiny520537work-lab:
Fixed both issues in the latest commit:

**P0 — Literal newline inside regex**

Root cause: the heredoc-based file upload interpreted `\n` inside the regex literal as a real newline character, breaking the regex syntax. Rebuilt the file using Python string construction to guarantee byte-level correctness — the regex is now the 2-character sequence `/\n/g` not a spanning literal.

**P1 — Real newline in Mermaid node label**

Changed the label separator from `` \n `` to ` · ` (middle dot), keeping the label on a single line:

```typescript
// Before (invalid Mermaid syntax)
const label = file ? `${emoji} ${title}\n${file}` : ...

// After (single-line, valid Mermaid)
const label = file ? `${emoji} ${title} · ${file}` : ...
```

This keeps the file path visible without relying on Mermaid's inconsistent multi-line label support.

destiny520537work-lab:
Fixed — removed the stray semicolon on line 17 that was terminating the `return` statement early. The method chain now runs correctly through all five calls.

destiny520537work-lab:
Fixed — added `[` and `]` to the sanitize character class: `.replace(/[<>{}|[\]]/g, ' ')`. Good catch; a path like `src/routes[v2]/index.ts` would have silently broken the graph node.

thedotmack:

Issue #2790, opened 2026-06-05
[Windows] Chroma semantic search never connects: worker spawns chroma-mcp via uvx with a sanitized PATH that omits uv's bin dir


AlbertMein:
## Summary
On Windows, semantic search (Chroma) never connects. The worker launches `chroma-mcp` via `cmd /c uvx ...`, but spawns that child with an environment whose PATH excludes uv's install dir (`%USERPROFILE%\.local\bin`). So `uvx` is "not recognized" and the subprocess dies in ~25ms. Keyword search and capture/DB are unaffected, only semantic.

## Environment
- Windows 11; claude-mem 13.4.0 (thedotmack marketplace)
- uv/uvx 0.10.12 at `%USERPROFILE%\.local\bin`
- Worker healthy on :37777, DB recording normally

## Symptom
- `search()` warns `Semantic search failed ... chroma-mcp connection in backoff`, falls back to keyword.
- `GET /api/chroma/status?deep=1` returns `{"status":"unhealthy","connected":false,...}`
- Log:
  ```
  [CHROMA_MCP] Connecting ... {command=C:\Windows\system32\cmd.exe, args=/c uvx --python 3.13 --with "onnxruntime>=1.20" --with "protobuf<7" chroma-mcp==0.2.6 --client-type persistent --data-dir ...}
  [CHROMA_MCP] Connection failed ... {error=MCP error -32000: Connection closed}
  ```
  connect to fail interval is about 25-28ms, i.e. command-not-found, not slow startup.

## Root cause (verified)
- `uvx` is on the user and system PATH; after a full reboot, fresh processes resolve it.
- The worker's exact command run manually boots fine, returning `{"serverInfo":{"name":"chroma","version":"1.6.0"}}`.
- Yet the worker's spawn of that same command still fails "not recognized."
- Reproduced by running `cmd /c uvx --version` under a stripped PATH (System32/Windows only): `'uvx' is not recognized`.
- Conclusion: the worker passes the chroma child a minimal/sanitized PATH that omits uv's bin dir (and even ignores newly added system-PATH entries). On Windows, uv installs to `%USERPROFILE%\.local\bin`, which is not in that minimal set.

## Suggested fix
When spawning chroma-mcp, the worker should either:
1. pass the full inherited environment / prepend uv's bin dir (`%USERPROFILE%\.local\bin` on Windows, `~/.local/bin` elsewhere) to the child PATH; or
2. invoke `uvx` by absolute path; or
[...]

Issue #2714, opened 2026-05-29
Regression (v13.4.0): mcp-search fails on Windows with -32000 — command:"sh" reappeared; sh not on PATH


davidfi01:
## Summary

The `mcp-search` MCP server fails to start on Windows in **v13.4.0** with a JSON-RPC `-32000` error. Root cause: the plugin's `.mcp.json` launches the server via `command: "sh"`, but `sh` is not on `PATH` on a normal Windows + Git-for-Windows install (Git only puts `cmd\` on `PATH`, not `bin\`). Claude Code can't spawn `sh`, so the connection dies before MCP init.

This is a **regression** — the exact same class of bug was reported and closed multiple times, but the `sh`-based launcher has reappeared in 13.4.0:

- #2446 — "Use node to run mcp for windows environment" (the node-direct fix)
- #2461 — "Fix Windows compatibility: replace sh spawn + silence printf EPIPE"
- #2611 — "MCP server fails on Windows with -32000 error"
- #2625 — "[Windows] MCP server fails: POSIX path from sh breaks Node.js on Windows"

## Environment

- OS: Windows 11 Pro (10.0.26220)
- claude-mem: **13.4.0** (auto-updated 2026-05-29)
- Node: v24.15.0 (`C:\Program Files\nodejs\node.exe`)
- Git: installed at `C:\Program Files\Git\` — only `...\Git\cmd` is on `PATH`; `...\Git\bin\sh.exe` exists but is **not** on `PATH` (this is Git for Windows' default behavior)

## Symptom

```
/mcp
Failed to reconnect to plugin:claude-mem:mcp-search: -32000
```

## Current launcher (v13.4.0 `.mcp.json`)

```json
{
  "mcpServers": {
    "mcp-search": {
      "type": "stdio",
      "command": "sh",
      "args": ["-c", "_C=\"${CLAUDE_CONFIG_DIR:-$HOME/.claude}\"; ... exec node \"$_P/scripts/mcp-server.cjs\""]
    }
  }
}
```

The `sh -c` wrapper does plugin-root path discovery in POSIX shell. On Windows-without-`sh`-on-PATH, `sh` itself can't be spawned, so none of that runs.

## Root cause confirmed

Two-part diagnosis on the affected machine:

1. **`sh` / `bash` are NOT resolvable** (`Get-Command sh` → not found); only `git` resolves.
2. **The Node server itself is healthy** — running it directly stays up as a proper stdio MCP server:
   ```
[...]

PR #2807, opened 2026-06-06
fix(mcp): cross-platform Node launcher for plugin .mcp.json (closes #2792)

### LINKED ISSUE #2792: MCP server fails to register on Windows - Connection closed (-32000) on session start
**Environment:**
- OS: Windows 11 x64
- Claude Code: 2.1.165
- claude-mem: 13.4.0
- Node: installed
- Bun: 1.3.14 (installed at ~/.bun/bin/bun.exe)

**Problem:**
MCP server fails to register at session start with error:
plugin:claude-mem:mcp-search: failed – MCP error -32000: Connection closed

**Root cause found:**
- Worker runs fine on port 37777
- bun-runner.js can't find Bun at startup because ~/.bun/bin is not in Windows PATH at Claude Code launch time
- SessionStart hook eventually finds Bun and starts the worker, but by then MCP registration already failed
- Chroma was also enabled by default but not installed, causing additional failures

**Workaround applied:**
- Disabled Chroma (CLAUDE_MEM_CHROMA_ENABLED: false)
- Added C:\Program Files\Git\bin and Git\usr\bin to Windows user PATH
- Worker now runs but MCP tools don't register in the session

**Request:**
Add Bun PATH discovery at MCP server startup, not just in the worker hook.

thedotmack:
## Problem

`plugin/.mcp.json` launched the MCP search server with `command: "sh"` and an `sh -c` discovery prelude. Claude Code cannot spawn `sh` on Windows when Git's `usr/bin` is not on PATH, so the `mcp-search` tools never registered. Closes #2792 (also relevant to #2790, #2714, #2461).

## Fix

Replace the shell launcher with a pure-Node one: `command: "node"` + `node -e <payload>`. The payload is emitted by a new `buildMcpNodeLauncher()` in `src/build/hook-shell-template.ts` and:

- Mirrors the POSIX prelude's plugin-root discovery order: `$CLAUDE_PLUGIN_ROOT`/`$PLUGIN_ROOT`, extra candidates (`$PWD/plugin`, `$PWD`), mtime-sorted cache roots (codex-local, codex-thedotmack, claude), then the marketplace install dir.
- Spawns the resolved `mcp-server.cjs` with inherited stdio.
- Forwards `SIGTERM`/`SIGINT`/`SIGHUP` and propagates the child's exit code/signal.
- Has **no shell dependency**, so it spawns on Windows without Git Bash.

The generator stays the single source of truth: `scripts/build-hooks.js` verifies `plugin/.mcp.json` `args[1]` byte-matches `buildShellCommand({host:'mcp'})`, and `tests/infrastructure/plugin-distribution.test.ts` is updated to assert the Node payload.

## Verification

- `npx tsc --noEmit` — passes
- `npm run build` — passes, including `verifyShellTemplateCanonical`
- Infrastructure test suite — **196 pass, 3 skip, 0 fail**
- Live: the actual `.mcp.json` payload resolves the plugin root, spawns `mcp-server.cjs`, and returns a valid MCP `initialize` handshake
- Not-found path exits `1` with `claude-mem: mcp server not found`
- Subagent QA review: **PASS**


thedotmack:
Addressed both Greptile findings in e7fd4bc77:
- **Signal-kill fallback** now exits `1` (not `0`) when the signal re-raise throws — `plugin/.mcp.json` regenerated to match the generator.
- **`trailingCommand`** is now optional and documented as intentionally ignored by the mcp Node launcher (spawn target derives from `requireFile`); dropped the meaningless shell tokens from the mcp call sites and made shell hosts fail loud if it's missing.

Verified: `tsc --noEmit`, `npm run build` (canonical parity holds), 48/48 distribution tests, and a live launcher handshake all green.

PR #2810, opened 2026-06-06
feat: ignore Codex and Codex-app internal system prompts

pg-adm1n:
This PR ignores Codex and Codex-app internal system-initiated prompts (e.g., session title generation, memory consolidation agent, and suggestions overview) to prevent triggering unnecessary session initialization and hook events.


thedotmack:

> Codex internal-prompt filter belongs in the Codex adapter contract (step 1); P1 stale reason string + false-positive prefix

thedotmack:
Second pass — routing holds at **#3611 (plan-23, Host Integration Contracts)**: filtering Codex's internal system-initiated prompts is an adapter-contract concern and belongs in the Codex adapter, not in shared `tag-stripping` utilities, which is step 1 of that master. Nothing open duplicates it and nothing merged supersedes it. The two review P1s are still live — the stale reason string, and the prefix match being broad enough to swallow legitimate user prompts.

Issue #2716, opened 2026-05-30
Regression (v13.4.0): chroma-mcp still dies instantly on Windows — #2701 cmd.exe quoting fix is incomplete


Nelie-Taylor:
# Bug: chroma-mcp subprocess dies instantly on Windows — the v13.4.0 `cmd.exe` quoting fix (#2701) is incomplete

> **Follow-up to #2696** (closed as "Shipped in v13.4.0 via #2701"). The quoting fix does **not** resolve the issue on Windows. Please reopen or track here.

## Related issues

- **#2696** — original report, closed as fixed by #2701 (v13.4.0). **This report shows the fix is incomplete.**
- Earlier duplicates of the same root cause (all closed): #2667, #2657, #2535, #2426, #2425, #2429, #2438, #2484, #2495, #2509, #2529, #2585, #2591, #2405, #2494.
- **#2714** — sibling regression in v13.4.0: `mcp-search` fails on Windows with `-32000` because it's launched via `sh` (not on PATH). Same anti-pattern (assuming a shell on Windows), different component. The robust fix below applies to both: spawn the real executable directly with `shell: false`.

## Summary

On Windows, every Chroma sync fails with `MCP error -32000: Connection closed`. The `chroma-mcp` subprocess dies in **~18ms**, long before it ever reaches the Chroma server. This happens regardless of whether Chroma (Docker/remote) is up and reachable.

The root cause is how the worker spawns `chroma-mcp` on Windows: it wraps the command in `cmd.exe /c uvx ...`. The dependency specs `onnxruntime>=1.20` and `protobuf<7` contain `>` and `<`, which `cmd.exe` interprets as redirection operators.

The v13.4.0 fix (PR for "chroma-mcp cmd.exe quoting") added per-arg quoting for `[<>|&^()]`, but this does **not** fix the problem and arguably makes it worse — because the MCP SDK spawns via an **argv array with `shell: false`**, so `child_process` applies its *own* Windows quoting on top of the manually-injected quotes, producing a malformed command line.

## Environment

- claude-mem: **13.4.0** (worker bundle confirmed running from the updated plugin dir)
- OS: Windows 11 Pro 26200
- Runtime: bun (worker), Node v24 (verification)
- uvx: 0.9.30
- chroma-mcp: 0.2.6
- Chroma server: chromadb/chroma:latest (1.0.0) in Docker, `127.0.0.1:8000`, `/api/v2/version` → `1.0.0` (verified reachable)
[...]

Issue #2696, opened 2026-05-28
ChromaDB MCP subprocess fails on Windows: unquoted protobuf<7 in cmd.exe /c command line


Hachiiki:
## Description

ChromaDB semantic search fails on Windows (win32) because the chroma-mcp MCP subprocess crashes immediately after spawning. The worker falls back to FTS5 keyword search, but semantic search is completely unavailable.

## Error

```
[CHROMA_MCP] Connecting to chroma-mcp via MCP stdio
  {command=C:\WINDOWS\system32\cmd.exe, args=/c uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 --client-type persistent --data-dir C:/Users/Hp/.claude-mem/chroma}
[CHROMA_MCP] Connection failed, killing subprocess tree to prevent zombie
  {error=MCP error -32000: Connection closed}
```

The subprocess exits within ~200ms with no stdout/stderr output.

## Root Cause

The worker spawns chroma-mcp via `cmd.exe /c` (necessary on Windows because uvx is a batch file). The command line includes the dependency constraint `protobuf<7`:

```
cmd.exe /c uvx --with protobuf<7 chroma-mcp==0.2.6 ...
```

In `cmd.exe`, `<` is the input redirection operator. The shell parses `protobuf<7` as running the command `protobuf` with stdin redirected from file `7`. This causes uvx to receive mangled arguments, making it fail immediately.

The fix is to quote the version constraint so `cmd.exe` passes it literally: `"protobuf<7"`.

## Evidence & Testing

1. **Direct execution works** — Running `uvx --with "protobuf<7" chroma-mcp==0.2.6` from PowerShell (with quotes around the constraint) starts successfully and stays running. Packages install fine (106 packages in 9.36s).

2. **Worker spawn via cmd.exe fails** — The same command spawned by the worker via `cmd.exe /c` crashes every time in under 1 second (timing: 720ms from connect to failure).

3. **Clearing chroma data dir does NOT fix it** — Removing `~/.claude-mem/chroma/` before restart confirmed it's not a stale lock or corrupt database issue.

4. **Disabling chroma resolves it** — Setting `CLAUDE_MEM_CHROMA_ENABLED=false` in `settings.json` skips the chroma-mcp subprocess entirely. Logs show: `Chroma disabled via CLAUDE_MEM_CHROMA_ENABLED=false, skipping ChromaMcpManager` with no further errors.

## Environment

[...]

PR #2816, opened 2026-06-07
fix(chroma): spawn uvx directly on Windows instead of via cmd.exe (#2716, #2667)

YOMXXX:
## Problem

On Windows the chroma-mcp subprocess is launched as `cmd.exe /c uvx ...`. cmd.exe re-parses the dep-override version specifiers (`onnxruntime>=1.20`, `protobuf<7`) as I/O redirection (`protobuf < 7`), so `chroma-mcp` never starts — every semantic-search tool call closes in ~30 ms with `MCP error -32000: Connection closed`.

This is the defect tracked in #2716 / #2667 (closed as not-planned, but still reproducing on latest `main`, as flagged by @tomeyerman in #2699).

## Why the existing `quoteForCmdExe` fix (#2701) doesn't work

`quoteForCmdExe` wraps `protobuf<7` → `"protobuf<7"` as a string, and its unit test is green. But the quotes don't survive the actual spawn:

1. The MCP SDK's `StdioClientTransport` launches through **cross-spawn with `shell: false`**.
2. The command is `cmd.exe` (a `.exe`, not a `.bat`/`.cmd`), so cross-spawn does **not** set `windowsVerbatimArguments`.
3. Node's own argv→command-line quoting then re-escapes the quotes: `"protobuf<7"` → `"\"protobuf<7\""`.
4. cmd.exe treats `\"` as a literal backslash + a quote toggle (not an escape), so it closes the quote early and sees `<` outside quotes again → redirection → "The system cannot find the file specified".

So the hand-quoting is defeated by the very re-quoting layer it tries to anticipate. The `cmd.exe` wrapper is the whole problem.

## Fix

Spawn `uvx` **directly** on every platform (`buildChromaSpawnConfig`). Node's `child_process.spawn` resolves `uvx` → `uvx.exe`/`uvx.cmd` on PATH on Windows exactly as on POSIX, so no shell wrap is needed: the argv array is passed through without any cmd.exe re-parsing and the `<`/`>` dep specs reach uvx intact. This matches the approach #2502 / #2668 converged on in the #2699 canary.

- Remove the `cmd.exe /c` Windows branch and the defeated `quoteForCmdExe` helper.
- Replace its string-only unit test with a spawn-contract test asserting the launch is direct `uvx` (no `cmd.exe`, no `/c`, dep specs verbatim) — guarding against the wrapper being reintroduced.

## Testing

- `bun test tests/services/integrations/spawn-contract-windows.test.ts tests/services/sync/chroma-mcp-manager-singleton.test.ts` → **7 pass / 0 fail**
- `tsc --noEmit`: 0 errors introduced by this change
- Windows real-machine validation still welcome — I don't have a Windows box; @tomeyerman offered in #2699 to confirm Chroma connects end-to-end on Windows 11.

refs #2716, #2667, #2696, #2699


YOMXXX:
## Status for reviewers

**CI** — the one red check (`typecheck · build · test · bundle-size`) is a **pre-existing baseline failure**, not introduced here:

```
src/server/runtime/create-server-beta-service.ts(5,10): error TS2300: Duplicate identifier 'ModeManager'.
src/server/runtime/create-server-beta-service.ts(17,10): error TS2300: Duplicate identifier 'ModeManager'.
```

That file isn't touched by this PR — it's the same `tsc` baseline #2699 documents (fixed separately by #2511). `tsc --noEmit` reports **0 errors introduced** by this change, and the other three checks (build, clean-room dependency closure, server-runtime e2e) all pass.

**Greptile** — re-reviewed the latest commit (`eef2dd0e`) and cleared its earlier P1: `ensureUvOnPath` / `uvBinDirs` (the #2790 PATH fix) are preserved unchanged, so the `ENOENT` regression path is not re-opened. The augmented PATH flows through the spawn `env` and applies to the direct `uvx` spawn exactly as it did to the old `cmd.exe` wrapper — #2790 (PATH) and #2716/#2667 (cmd.exe metachar re-parse) are independent root causes, both fixed.

On Greptile's one residual note (the `uvx.cmd` shim on non-standard installs): it's safe either way — the official uv installer ships `~/.local/bin/uvx.exe` (cross-spawn spawns the `.exe` directly), and if uv is a `.cmd` shim, cross-spawn applies its own correct cmd.exe escaping rather than the broken hand-quoting this PR removes.

Verified locally: `bun test` spawn-contract + chroma singleton suites **7 pass / 0 fail**.

teyerman-st (approved):
Posted from the wrong account by mistake — please disregard this review; see my validation under @tomeyerman instead. As a non-maintainer I can't delete a submitted review — @thedotmack, feel free to dismiss this one.

tomeyerman (approved):
Validated on **Windows 11 Pro (10.0.26200)** — the direct-`uvx` spawn this PR introduces connects Chroma end to end. 🎉

**Disclosure on what I ran:** I'm running a self-host fork that carries the *functionally identical* change, not a checkout of this exact commit. I diffed my `ChromaMcpManager.ts` against this PR — the only differences are that you extracted the logic into `buildChromaSpawnConfig()` and added richer comments; the runtime behavior (drop the `cmd.exe /c` wrapper, spawn `uvx` directly on every platform) is the same. So this validates the *approach*, on a real Windows box.

### Evidence

**1. Process tree — worker spawns `uvx` directly, no `cmd.exe` in the chain, dep specs verbatim:**

```
bun.exe  (worker, PID 24764)   worker-service.cjs --daemon
  └─ uvx.exe  (PID 36612)      uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 --client-type persistent --data-dir .../chroma
       └─ uv.exe  (PID 36700)
            └─ chroma-mcp.exe  (PID 36344)
                 └─ python.exe → python.exe
```

The worker (`bun.exe`) is the *direct* parent of `uvx.exe` — no intermediate `cmd.exe`. Critically, `onnxruntime>=1.20` and `protobuf<7` reach `uvx` intact; these are the exact `>`/`<` specifiers the old `cmd.exe /c` wrapper re-parsed as I/O redirection, which is what killed chroma-mcp in ~30ms with `MCP error -32000: Connection closed`.

**2. Semantic round-trip succeeds (not the FTS5 keyword fallback):**

```jsonc
// GET /api/chroma/status?deep=1
{
  "status": "healthy",
  "connected": true,
  "details": "chroma-mcp semantic search round-trip succeeded",
  "deep": true,
  "probe": { "ok": true, "stage": "done", "queryLatencyMs": 762 }
}
```

```jsonc
// GET /api/health
{ "status": "ok", "platform": "win32", "initialized": true, "mcpReady": true, "version": "13.4.0" }
```

### Conclusion

The `cmd.exe` wrapper was the whole problem, and spawning `uvx` directly resolves it on Windows 11 with no regression — `uvx.exe`/`uv.exe`/`chroma-mcp.exe` resolve on PATH and the `--with` pins survive the spawn. The added spawn-contract test is a good guard against the wrapper being reintroduced. LGTM for the Windows path. 👍

thedotmack:
Closing as duplicate: the Windows Chroma fix landed on main via #2847 — chroma-mcp now spawns a directly-resolved uvx/uvx.exe (no cmd.exe /c), so the `>=`/`<` pip specifiers never hit cmd.exe redirection parsing. Same approach as this PR; #2847 was the most current against main. Thanks — reopen if you hit a case the merged fix misses.

Issue #2711, opened 2026-05-29
Option to write worktree observations to the parent project (detectWorktree already resolves it for reads)


jfall-tegna:
## Before submitting

- [x] I searched existing issues and confirmed this is not a duplicate (closest are #2663 / #2665, which cover *nested separate repos*, not worktree→parent **writes** — see below)

---

**Is your feature request related to a problem? Please describe.**

claude-mem is already worktree-aware on the **read** side: `detectWorktree()` (`src/utils/worktree.ts`) recognizes a real `git worktree` (where `.git` is a file pointing at `<parent>/.git/worktrees/<name>`), resolves the parent project, and the `SessionStart` path injects context for both via `?projects=<parent>,<worktree>`. So a worktree *inherits* its parent's memory.

But the **write** side is not worktree-aware. `getProjectName(cwd)` (`src/utils/project-name.ts`) is `path.basename(cwd)`, so every observation/summary/session created while working in a worktree is stored under the *worktree folder's* name (e.g. `develop`, `rc-v1.2.0`, `feature-x`) — never the parent.

Result: in a bare-repo + multiple-worktrees workflow (where most work happens in worktrees, not the primary checkout), findings never accumulate into the canonical parent-project DB, and memory fragments across generic, non-unique folder names that can also collide across unrelated repos. The asymmetry is the surprising part — a worktree can *read* the parent's memory but can't *contribute* to it.

**How this differs from #2663 / #2665**

#2663 and PR #2665 (`CLAUDE_MEM_USE_GIT_ROOT`) address *nested independent repos* launched from a parent folder, by walking up to the nearest `.git`. That doesn't solve the worktree case: walking up from a worktree finds the worktree's *own* `.git` file, so it still resolves to the worktree's name. PR #2665 also explicitly preserves the `parent/worktree` composite for **context** only. So worktree **writes** remain unaddressed — this request is the complementary piece, reusing the parent that `detectWorktree()` already computes.

**Describe the solution you'd like**

An opt-in setting controlling where worktree observations are written (mirroring the `CLAUDE_MEM_USE_GIT_ROOT` opt-in pattern), e.g.:

```
CLAUDE_MEM_WORKTREE_WRITE = self | parent | both
```

- `self` — current behavior; **default**, non-breaking
- `parent` — write under the parent project `detectWorktree()` already resolves
- `both` — write to parent and worktree

Reads can keep today's `parent,worktree` composite regardless.

**Describe alternatives you've considered**

[...]

PR #2827, opened 2026-06-08
feat(project-name): CLAUDE_MEM_PROJECT_NAME_SOURCE=git-remote for stable org/repo slugs

surfingdoggo:
## Problem

claude-mem derives the project name from the folder / git-root **basename**. That name silently changes if a checkout directory is renamed, and two unrelated repos that happen to share a folder name collide into the same project — scattering or merging memory incorrectly.

## Fix

Add an opt-in setting `CLAUDE_MEM_PROJECT_NAME_SOURCE` (default `path`). When set to `git-remote`, the project name becomes a stable `org/repo` slug parsed from the git `origin` remote URL — stable across directory renames and identical across a repo's worktrees (they share the remote).

- **Default `path` is a no-op** for existing users — behavior is unchanged unless you opt in.
- In `git-remote` mode, worktree parent/child compositing is skipped (a repo's worktrees share its origin and already resolve to one slug).
- `parseOriginUrlToSlug` is a pure, exported helper covering scp-style (`git@host:org/repo.git`), `https://`, and `ssh://` forms, optional `.git`/trailing slash, and nested subgroups — with unit tests.

## Changes

- New setting + resolver (`src/utils/project-name.ts`, `src/shared/SettingsDefaultsManager.ts`).
- Unit tests for `parseOriginUrlToSlug`.
- Documented the setting in the configuration reference.

## Note on CI

`main` currently has unrelated pre-existing failures (PID-file / process-manager and logger-standards tests, plus a duplicate `ModeManager` import typecheck error). This change is isolated to project-name resolution; the `project-name` test suite passes (29/29).

Relates to upstream discussion on stable project identity (#2663, #2711).

### DIFF
--- a/docs/public/configuration.mdx
+++ b/docs/public/configuration.mdx
@@ -65,6 +65,7 @@ Use [LiteLLM Gateway](configuration/litellm-gateway) when you want `CLAUDE_MEM_P
 | `CLAUDE_MEM_LOG_LEVEL`        | `INFO`                          | Log verbosity (DEBUG, INFO, WARN, ERROR, SILENT) |
 | `CLAUDE_MEM_PYTHON_VERSION`   | `3.13`                          | Python version for chroma-mcp         |
 | `CLAUDE_CODE_PATH`            | _(auto-detect)_                 | Path to Claude Code CLI (for Windows) |
+| `CLAUDE_MEM_PROJECT_NAME_SOURCE` | `path`                       | How the project name is derived: `path` (folder/git-root basename) or `git-remote` (stable `org/repo` slug from the git `origin` URL — survives directory renames). Opt-in; `path` preserves existing behavior |
 
 ## Model Configuration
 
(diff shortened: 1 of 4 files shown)


surfingdoggo:
Rebased onto latest `main` (v13.5.x). Still relevant: there is no project-name-source option upstream, so hook sessions still key by cwd basename. This adds opt-in `CLAUDE_MEM_PROJECT_NAME_SOURCE=git-remote` to derive a stable `org/repo` slug from the origin URL (survives directory renames and unifies a repo's worktrees). Clean rebase, typecheck clean, 29 tests pass.

thedotmack:

> step 1 explicitly adopts opt-in git-remote slug; open P1s (.git ordering, file:// remotes, worktree compositing) must be fixed against that design

thedotmack:
Second pass — routing holds at **#3608 (plan-20, Project Identity & Injection Scope)**, and this is the closest existing implementation of step 1: an opt-in stable slug instead of the folder basename, which is the exact identity drift the master is built around. It does not duplicate #2737 — that one maps glob patterns to shared environment names, a different resolver strategy for a different problem. Still open are the P1s from review (`.git` stripping order, `file://` remotes, worktree compositing), and those should be fixed against the plan-20 design rather than in isolation.

PR #2828, opened 2026-06-08
feat(worker): add CLAUDE_MEM_WORKER_AUTOSTART to opt out of hook worker lazy-spawn

surfingdoggo:
## Problem

Hooks lazy-spawn the worker daemon via `ensureWorkerAliveOnce()` whenever the worker port is dead. That's correct for worker-runtime users — but **server-beta-only or externally-managed deployments have no way to stop hook activity from continually resurrecting the worker daemon.**

## Fix

Add `CLAUDE_MEM_WORKER_AUTOSTART` (default `true`). When `false`, `ensureWorkerAliveOnce()` returns `false` without spawning.

- **Default `true` preserves existing behavior exactly** — no-op for current users.
- Opting out (`false`) is for operators who drive the worker out-of-band and knowingly give up worker-only features (data viewer, corpus/skills, semantic injection — already unsupported on server-beta per `docs/server-beta-parity-map.md`).

## Changes

- Setting + opt-out short-circuit at the single `ensureWorkerAliveOnce()` chokepoint.
- Unit tests for the opt-out / default / explicit-`true` cases, plus a `resetAliveCache()` test helper mirroring the existing `clearPortCache()`.
- Documented in the configuration reference.

## Note on CI

`main` currently has unrelated pre-existing failures (PID-file / process-manager and logger-standards tests, plus a duplicate `ModeManager` import typecheck error). This change is isolated to the worker lazy-spawn path; the new test suite passes.

### DIFF
--- a/docs/public/configuration.mdx
+++ b/docs/public/configuration.mdx
@@ -20,6 +20,7 @@ Settings are managed in `~/.claude-mem/settings.json`. The file is auto-created 
 | `CLAUDE_MEM_CONTEXT_OBSERVATIONS` | `50`                        | Number of observations to inject      |
 | `CLAUDE_MEM_WORKER_PORT`      | `37700 + (uid % 100)`           | Worker service port (per-user default; override for fixed port) |
 | `CLAUDE_MEM_WORKER_HOST`      | `127.0.0.1`                     | Worker service host address           |
+| `CLAUDE_MEM_WORKER_AUTOSTART` | `true`                          | When `false`, hooks won't lazy-spawn the worker daemon — an opt-out for server-beta-only or externally-managed deployments. Default `true` preserves existing behavior (you lose worker-only features like the viewer, corpus/skills, and semantic injection when off) |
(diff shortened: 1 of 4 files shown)


surfingdoggo:
Rebased onto latest `main` (v13.5.x). One change needed against the rebase: your refactor made `ProcessManager` import `getSupervisor` from `supervisor/index.js`, so the test's partial module mock (which only stubbed `validateWorkerPidFile`) broke namespace resolution. I added a `getSupervisor` stub to the mock. Feature still absent upstream. Typecheck clean, 3 tests pass.

thedotmack:

> step 2 (`CLAUDE_MEM_DISABLE`/opt-out short-circuits before any work); P1: opt-out must not feed the unreachable counter

thedotmack:
Second pass — routing holds at **#3605 (plan-17, Hook Wrapper Contract)**, step 2, since the lever is the hook-side `ensureWorkerAliveOnce()` chokepoint rather than the worker's own port authority. Nothing open duplicates it and no merged work supersedes the flag. Two things before it moves: the P1 stands that an intentional opt-out must not feed the worker-unreachable failure counter, and per your own note the `ProcessManager`/`getSupervisor` refactor on main breaks the test's partial module mock, so it needs another rebase pass.

PR #2835, opened 2026-06-08
feat: add 'claude' as alias for 'claude-code' platform

mas5464:
## Summary

- Adds `case 'claude':` in `getPlatformAdapter` so hook commands like `hook claude observation` route to the same adapter as `hook claude-code`
- Adds `case 'claude':` in `makeIDETask` so `npx claude-mem install --ide claude` works identically to `--ide claude-code`
- Extends `claude-code` IDE detection to also check for `claude-code` binary in PATH (in addition to `claude`)

## Motivation

The Claude Code CLI binary is named `claude`, so many users naturally try `--ide claude`. Currently that falls through to the `raw` adapter. This PR makes `claude` a first-class alias so the install and hook pipeline work without knowing the internal `claude-code` ID.

## Test plan

- [ ] `npx claude-mem install --ide claude` completes successfully
- [ ] Hook command `hook claude observation` resolves to `claudeCodeAdapter` 
- [ ] Existing `--ide claude-code` behavior unchanged

### DIFF
--- a/src/cli/adapters/index.ts
+++ b/src/cli/adapters/index.ts
@@ -8,6 +8,7 @@ import { windsurfAdapter } from './windsurf.js';
 
 export function getPlatformAdapter(platform: string): PlatformAdapter {
   switch (platform) {
+    case 'claude':
     case 'claude-code': return claudeCodeAdapter;
     case 'codex': return codexAdapter;
     case 'cursor': return cursorAdapter;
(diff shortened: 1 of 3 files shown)


thedotmack:

> `--ide claude` alias does not work end-to-end (pre-validation rejects it)

thedotmack:
Second pass — routing holds at **#3611 (plan-23, Host Integration Contracts)**: platform-id aliasing across `getPlatformAdapter`, `makeIDETask` and IDE detection is the adapter-contract surface that master owns. Not a duplicate of anything open and not superseded on main. The blocker from the earlier review is unchanged: `--ide claude` still does not work end to end because install-time pre-validation rejects the value before `makeIDETask` ever sees it, so the three `case 'claude':` additions are necessary but not sufficient.

PR #534, opened 2026-01-04
refactor: decompose monolithic services into modular architecture

thedotmack:
## Summary

Major architectural refactoring that decomposes several monolithic services into modular, maintainable components:

- **Worker Service**: Extracted infrastructure (GracefulShutdown, HealthMonitor, ProcessManager), server layer (ErrorHandler, Middleware, Server), and integrations (CursorHooksInstaller)
- **Context Generator**: Split into ContextBuilder, ContextConfigLoader, ObservationCompiler, TokenCalculator, formatters (Color/Markdown), and section renderers (Header/Footer/Summary/Timeline)
- **Search System**: Extracted SearchOrchestrator, ResultFormatter, TimelineBuilder, and strategy pattern (Chroma/SQLite/Hybrid search strategies) with dedicated filters (Date/Project/Type)
- **Agent System**: Extracted shared logic into ResponseProcessor, ObservationBroadcaster, FallbackErrorHandler, and SessionCleanupHelper
- **SQLite Layer**: Decomposed SessionStore into domain modules (observations, prompts, sessions, summaries, timeline) with proper type exports

### Additional Changes
- Fixed duplicate observation storage bug (observations stored multiple times when messages were batched)
- Added duplicate observation cleanup script for production database remediation
- Fixed FOREIGN KEY constraint and missing `failed_at_epoch` column issues
- Improved error handling across all refactored modules
- Various hook input parsing and error logging improvements

## Test plan
- [ ] Verify worker service starts correctly with `npm run build-and-sync`
- [ ] Confirm context generation produces correct output format
- [ ] Test search functionality across all strategies (SQLite, Chroma, Hybrid)
- [ ] Validate observation storage no longer creates duplicates
- [ ] Run cleanup script in dry-run mode to verify duplicate detection logic

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2845, opened 2026-06-09
chore(sqlite): remove dead legacy migration system (migrations.ts)

darion-yaphet:
## What
Removes `src/services/sqlite/migrations.ts` (549 lines) — the legacy version-object migration system (`migration001`…`migration010` + the `migrations[]` array).

## Why
Orphaned by the #534 `MigrationRunner` refactor. Verified **zero importers**:
- No code imports `{ migrations }` or any `migration00X` symbol.
- `registerMigration()` (the only consumer of the `migrations[]` array) has zero external callers → `DatabaseManager`'s version-object migration loop always iterated an empty array.
- The file's `export { MigrationRunner }` re-export was redundant — real consumers import from `migrations/runner.js` / `Database.js` directly.

## Scope (deliberately conservative)
Only the confirmed-dead file. The parallel decomposed modules (`sessions/`, `observations/`, `summaries/`, `timeline/`, `import/`, `transactions.ts`, `MigrationRunner`, `ClaudeMemDatabase`) all have live consumers and are left untouched — they belong to a separate "finish the #534 decomposition" effort.

## Verification
- `npm run build` — green (all targets compile; Rule A shell-template canonical check passes).
- `bun test` — 2111 pass / 11 fail. The 11 failures are pre-existing full-suite test-pollution (HealthMonitor / ProcessManager / redis-config / welcome-hint), identical on `main`, and all pass in isolation. **Zero regression from this change.**

## Diff
```
src/services/sqlite/migrations.ts   -549
src/services/sqlite/index.ts        -2  (barrel re-export)
```


thedotmack:
Merged via #2865 (source intact, bundles regenerated once). Thanks!

Issue #2695, opened 2026-05-28
Codex CLI integration fails on Windows: spawnSync ENOENT for codex.cmd


RocStarr21:
## Bug
claude-mem v13.3.0 install fails to set up Codex CLI integration on Windows with `spawnSync codex ENOENT`, even though `codex` is installed globally and accessible from the shell.

## Environment
- OS: Windows
- claude-mem: v13.3.0
- Node/npm: npm global install of `@openai/codex` v0.134.0
- Shell: PowerShell

## Repro
1. `npm install -g @openai/codex`
2. Verify install: `codex --version` → returns `codex-cli 0.134.0`
3. Verify PATH: `where.exe codex` → returns `C:\Users\<user>\AppData\Roaming\npm\codex` and `codex.cmd`
4. Run `npx claude-mem install` and select Codex CLI
5. Install fails with: `Installation failed: spawnSync codex ENOENT`

## Root Cause
On Windows, npm-installed global CLIs are wrapped as `.cmd` files (e.g. `codex.cmd`). Node's `child_process.spawnSync('codex', ...)` does not resolve PATHEXT extensions on Windows unless `shell: true` is passed, or the full filename (`codex.cmd`) is used.

## Suggested Fix
In the Codex CLI integration installer, either:
- Pass `{ shell: true }` to the spawnSync call, or
- On `process.platform === 'win32'`, explicitly resolve to `codex.cmd`

All other IDE integrations (Claude Code, Gemini CLI, OpenCode, OpenClaw, Copilot CLI, Antigravity, Warp) install successfully on the same machine.

thedotmack:
Shipped in **v13.4.0** via #2701 (merged to main). Closing as resolved.

Release: https://github.com/thedotmack/claude-mem/releases/tag/v13.4.0

PR #2847, opened 2026-06-09
fix(chroma): spawn uvx directly on Windows instead of via cmd.exe

siddy6969:
# fix(chroma): spawn uvx directly on Windows instead of via `cmd.exe`

## Summary

On Windows, semantic (Chroma) search silently degrades to keyword-only. The
`chroma-mcp` subprocess never connects — the deep probe sits in a perpetual
`Connection closed` → 10s-backoff loop — so every query falls back to SQLite
keyword search with no error surfaced to the user.

## Root cause

`ChromaMcpManager.connectInternal()` spawns `chroma-mcp` on Windows via:

```
cmd.exe /c uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 ...
```

`cmd.exe` interprets the `>` / `<` in the dep-override specs (`onnxruntime>=1.20`,
`protobuf<7`) as **shell redirection**. The existing #2696 fix wraps those args in
double quotes (`quoteForCmdExe`), but `StdioClientTransport` spawns through
Node's `child_process`, whose Windows cmd.exe **arg-quoting re-mangles the
already-quoted specs**. `cmd.exe` then parses `>` as a redirect to an invalid
target and exits in **~10 ms** with `The directory name is invalid.`

The MCP transport reports `Connection closed` (MCP error -32000); the manager
records a connection failure and enters its `RECONNECT_BACKOFF_MS` (10s) loop.
Result: `connected:false` forever, keyword-only fallback.

### Reproduction (Node, mirroring StdioClientTransport's spawn)

```js
spawn('cmd.exe',
  ['/c','uvx','--python','3.13','--with','"onnxruntime>=1.20"','--with','"protobuf<7"',
   'chroma-mcp==0.2.6','--client-type','persistent','--data-dir', dataDir],
  { stdio: ['pipe','pipe','pipe'] });
// -> [9ms] EXIT code=1   STDERR: "The directory name is invalid."
```

The same args spawned **directly against `uvx.exe`** (no `cmd.exe`) start the
server normally and stay alive.

## Fix

Spawn `uvx` **directly** (no `cmd.exe /c` shell) on every platform. With no
shell, `>` / `<` are passed through literally and never interpreted.

Because Node's shell-less spawn does not PATHEXT-resolve a bare `uvx` on Windows,
a new `ChromaMcpManager.resolveUvxCommand()` resolves the **absolute `uvx.exe`
path** from the same uv bin dirs `ensureUvOnPath()` already uses (and honours
`CLAUDE_MEM_CHROMA_UVX_PATH` when it points straight at a binary), falling back
to bare `uvx.exe`. On non-Windows it returns bare `uvx` (unchanged behaviour).

The now-obsolete `quoteForCmdExe` / `CMD_EXE_METACHARACTERS` helper is removed,
and its Windows spawn-contract test is rewritten to assert the new invariant:
**the command resolves to `uvx.exe`, never `cmd.exe`.**

## Verification

- **Windows 11** (uv 0.11.18, Python 3.13, chroma-mcp 0.2.6): with this change
  the deep probe goes from the perpetual `Connection closed` backoff to
  `{"status":"healthy","connected":true,"probe":{"ok":true,"stage":"done"}}`
  (connect → list → query round-trip), and the `mem-search` MCP tool returns
  semantic results instead of the keyword fallback.
- `bun test tests/services/integrations/spawn-contract-windows.test.ts` → 4 pass.
- `tsc --noEmit` → clean.

## Files changed

- `src/services/sync/ChromaMcpManager.ts` — add `resolveUvxCommand()`; spawn uvx
  directly with raw args; remove `quoteForCmdExe` / `CMD_EXE_METACHARACTERS`.
- `tests/services/integrations/spawn-contract-windows.test.ts` — rewrite the
  #2696 block to assert the direct-uvx-spawn contract (the #2695 codex test is
  unchanged).

### DIFF
--- a/src/services/sync/ChromaMcpManager.ts
+++ b/src/services/sync/ChromaMcpManager.ts
@@ -43,23 +43,13 @@ const CHROMA_MCP_DEP_OVERRIDES: ReadonlyArray<string> = [
   'protobuf<7',
 ];
 
-// Issue #2696: on Windows the chroma-mcp subprocess is spawned via
-// `cmd.exe /c uvx ...`. cmd.exe interprets `<`, `>`, `|`, `&`, `^`, and `(` `)`
-// as shell metacharacters BEFORE uvx ever sees them, so a dep-override spec
-// like `protobuf<7` is parsed as an input redirection (`protobuf < 7`) and the
-// command line breaks. Node's spawn arg-quoting only quotes args containing
-// spaces/quotes, not these cmd.exe operators, so we must wrap any arg that
-// contains one in double quotes ourselves. Args without metacharacters are
-// returned unchanged so the normal command line is byte-identical to before.
-const CMD_EXE_METACHARACTERS = /[<>|&^()]/;
-export function quoteForCmdExe(arg: string): string {
-  if (!CMD_EXE_METACHARACTERS.test(arg)) {
-    return arg;
-  }
-  // Escape any embedded double quotes, then wrap. Inside double quotes cmd.exe
-  // does not perform redirection/grouping on these metacharacters.
-  return `"${arg.replace(/"/g, '\\"')}"`;
-}
+// Issue #2696 (revised): chroma-mcp is now spawned by invoking uvx DIRECTLY on
+// every platform — see ChromaMcpManager.resolveUvxCommand(). The previous
+// `cmd.exe /c uvx ...` path, and the cmd.exe metacharacter-quoting helper that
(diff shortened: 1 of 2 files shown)


tomeyerman:
Thank you for getting this merged, @thedotmack ! Any timeline for publishing a release?

PR #2848, opened 2026-06-09
fix(worker): honor settings.json API request timeout (closes #2822)

### LINKED ISSUE #2822: query_corpus times out at 30s; CLAUDE_MEM_API_TIMEOUT_MS in settings.json has no effect (MCP server reads env only)
## Before submitting
- [x] I searched existing issues and confirmed this is not a duplicate (closest: #2447, #1922, #1712 — all closed; #2447's `CLAUDE_MEM_API_TIMEOUT_MS` workaround does not reach this path, see below)

## Bug Description
`query_corpus` consistently fails with `Error calling Worker API: Request timed out after 30000ms`. `prime_corpus` / `build_corpus` / `list_corpora` succeed; only `query_corpus` (which runs the SDK Q&A over the corpus) crosses 30s.

The #2447 workaround — setting `CLAUDE_MEM_API_TIMEOUT_MS` in `~/.claude-mem/settings.json` — has **no effect**, because the MCP server never reads it:

- `scripts/mcp-server.cjs`: `Zx = nu("CLAUDE_MEM_API_TIMEOUT_MS", Rr(le.API_REQUEST /*=30000*/), {min:500,max:3e5})`.
- `nu(t,e,r)` reads **only `process.env[t]`**; no `settings.json` fallback — unlike `su()`/`Fx()` (`CLAUDE_MEM_WORKER_PORT`/`HOST`) which use `et.loadFromFile(settings.json)`.
- `.mcp.json` registers `mcp-search` as `sh -c '… exec node scripts/mcp-server.cjs'` with **no `env` block**, so no `CLAUDE_MEM_*` is injected.
- Verified on the live process: `tr '\0' '\n' < /proc/<mcp-pid>/environ | grep -c CLAUDE_MEM_` → `0`.

Net: the corpus-query timeout is pinned to the 30000ms default with no supported way to raise it via settings.json.

## Steps to Reproduce
1. `build_corpus` (~30K-token corpus).
2. `prime_corpus` → succeeds.
3. `query_corpus` with a non-trivial question.
4. ~30s → `Request timed out after 30000ms`.
5. Set `CLAUDE_MEM_API_TIMEOUT_MS=120000` in `~/.claude-mem/settings.json`, fully restart Claude Code, repeat → still times out at 30000ms.

## Expected Behavior
`query_corpus` honors `CLAUDE_MEM_API_TIMEOUT_MS` from `settings.json` (as worker port/host do), and/or `.mcp.json` injects the relevant `CLAUDE_MEM_*` into the MCP server `env`, and/or long-running corpus endpoints use a higher default than `API_REQUEST` (30000ms). At minimum, document the supported way to raise the knowledge-agent timeout.

## Environment
- Claude-mem version: 13.4.0
- Claude Code version: 2.1.168
- OS: Arch Linux (kernel 7.0.11-arch1-1)
- Platform: Linux x86_64
- `CLAUDE_MEM_RUNTIME=worker`, `CLAUDE_MEM_PROVIDER=claude` (subscription), `CLAUDE_MEM_MODEL=claude-haiku-4-5`

## Additional Context
Likely interacts with #2381 (Worker Env Isolation). Suggest reading `CLAUDE_MEM_API_TIMEOUT_MS` via the same settings-file accessor used for `CLAUDE_MEM_WORKER_PORT`/`HOST` so the documented workaround takes effect for the MCP corpus path.

Note: the `npm run bug-report` generator could not be used to file this — it fails with `ERR_MODULE_NOT_FOUND: Cannot find package '@anthropic-ai/claude-agent-sdk'` (dependency not installed in the marketplace dir).

rodboev:
## Summary

Worker API calls could still time out at the default request budget even after a longer timeout was written to `settings.json`. The request timeout resolver now reads the persistent `CLAUDE_MEM_API_TIMEOUT_MS` setting while preserving the existing environment-variable override and validation bounds.

## Why

`readTimeoutEnv()` only reads `process.env` at `src/shared/worker-utils.ts:14-29`, and `workerHttpRequest()` used that env-only module-level timeout whenever a caller omitted `options.timeoutMs`. Host and port already load the same user's `settings.json`, so API timeout handling was the inconsistent path.

## Scope

This is limited to the default API request timeout used by worker HTTP calls, including MCP calls that do not pass an explicit timeout. Health checks and readiness waits remain on their existing timeout knobs.

## Risk

Longer configured timeouts can keep a request pending longer before failing, which is the intended behavior for slow corpus queries. Explicit per-call timeouts still win, invalid values still fall back inside the existing bounds, and `CLAUDE_MEM_API_TIMEOUT_MS` in the environment remains the highest-precedence override.

## Verification

- [x] `npm install`
- [x] `npm run build`
- [x] `bun test tests/shared/worker-utils-timeout.test.ts tests/shared/worker-utils-version-recycle.test.ts`
- [x] `npm run lint:hook-io`
- [ ] `npm run strip-comments:check` still fails locally with repo-content output `Changed: 301`
- [ ] Current red CI is the shared `redis-config` plus `search-routes-welcome-hint` baseline failure reproduced from `main`, not a timeout-branch regression

Closes #2822


thedotmack:

PR #2851, opened 2026-06-09
fix(hooks): strip unsupported suppressOutput from Codex hook output (closes #2844)

### LINKED ISSUE #2844: Codex hook output rejected: unsupported suppressOutput field
When running claude-mem under the Codex CLI native hooks integration, every tool call produces a hook failure in the Codex UI:

  • PostToolUse hook (failed)
    error: PostToolUse hook returned unsupported suppressOutput
  
  The memory capture itself still works (observations are enqueued and persisted), but Codex marks the hook as failed on every tool invocation, producing constant red error noise.

  Root cause
  
  The hook handlers return a JSON payload containing suppressOutput: true:

  { continue: true, suppressOutput: true, exitCode: ... }

  suppressOutput is a Claude Code-specific hook-output field. The Codex CLI validates hook output against its own schema and rejects any unsupported field, so it fails the hook.

  This appears intentional for Claude Code — the CHANGELOG notes "Verified all 7 handlers return suppressOutput: true (#598, #784)" — but the shared output formatter (formatOutput / the suppressOutput-emitting path in
  worker-service.cjs) is reused for the Codex platform branch without stripping the field. Codex's hook output schema does not accept suppressOutput.

  Grep of the installed worker-service.cjs (13.4.1) shows 7 return sites that unconditionally include suppressOutput, used across both the Claude Code and Codex code paths. The Codex branches (hook codex observation, hook codex 
  summarize, hook codex session-init, hook codex file-context) all flow through these.

  Environment
  
  - claude-mem version: 13.4.1 (latest on npm)
  - Host: Codex CLI (native plugin/hooks integration via claude-mem-local marketplace, [features] hooks = true)
  - OS: macOS (darwin 25.5.0), shell: zsh
  - Node: v22.x, bun: 1.3.10
  - Plugin path: ~/.codex/plugins/cache/claude-mem-local/claude-mem/13.4.1/

  Reproduction

  1. Install claude-mem into Codex (enabled = true, hooks = true in ~/.codex/config.toml).
  2. Run any session and invoke any tool (e.g. a Bash command).
  3. After the tool runs, the PostToolUse hook fires and Codex reports PostToolUse hook returned unsupported suppressOutput.

  (Same applies to the other Codex hook events that return suppressOutput.)

  Expected behavior
  
  When the target platform is Codex, hook handlers should emit only fields supported by Codex's hook-output schema and omit suppressOutput (and any other Claude-Code-only fields). The Claude Code path can keep emitting
  suppressOutput as before.

  Suggested fix
  
  In the output formatter, gate Claude-Code-only fields behind a platform check, e.g. strip suppressOutput (and any other unsupported keys) from the serialized output when platform === 'codex', rather than including them
  unconditionally in the shared formatOutput.

  Impact
  
  - Functional: low — observations are still captured/persisted.
  - UX: high — every tool call shows a failed-hook error in Codex, which is constant and alarming.

rodboev:
## Summary

Codex native hooks were marking claude-mem hook invocations as failed because the Codex adapter serialized the shared internal `suppressOutput` field. This changes the Codex output formatter to omit that Claude Code-only field while leaving the shared handler/default hook result contract unchanged.

## Why

The Codex formatter built every output from `buildBaseOutput()`, and that helper copied `result.suppressOutput` into the serialized object. Codex rejects that field even though the shared internal hook contract uses it.

## Scope

This PR does not remove `suppressOutput` from the shared handler contract, the standard hook response, or Claude Code behavior. It only strips the field at the Codex adapter boundary.

## Risk

The behavior change is limited to Codex hook output serialization. Handler execution, observation enqueueing, Stop hook behavior, and Claude Code output shaping remain unchanged.

## Verification

- [x] `npm run build`
- [x] `bun test tests/hook-lifecycle.test.ts --test-name-pattern "suppressOutput|Stop outputs"`
- [x] `npm run lint:hook-io`
- [x] Full `tests/hook-lifecycle.test.ts` has an unrelated failure that also reproduces on `main`
- [ ] Current red CI is the shared `redis-config` plus `search-routes-welcome-hint` baseline failure reproduced from `main`, not a Codex-hook regression

Closes #2844


thedotmack:

PR #2852, opened 2026-06-09
fix(sqlite): preserve legacy NULL observation hashes during migration 29 (closes #2812)

### LINKED ISSUE #2812: Migration 29 deletes ~98% of legacy observations on upgrade (NULL content_hash rows collapsed by dedupe)
## Summary

Upgrading from an older version (10.5.5) directly to 13.4.0 caused **schema migration 29** to delete ~98% of stored observations (26,445 → 552 rows). The migration dedupes on `(memory_session_id, content_hash)`, but rows created by older versions have `content_hash = NULL`, so the dedupe collapses each session's entire history into its first observation.

## Environment

- Upgrade path: 10.5.5 → 13.4.0 (Linux, bun worker, `claude plugin update`)
- DB: ~234MB, 26,445 observations accumulated since March

## What happened

On first 13.4.0 worker start, migration 29 ran:

```sql
DELETE FROM observations
 WHERE id NOT IN (
   SELECT MIN(id) FROM observations
    GROUP BY memory_session_id, content_hash
 )
```

followed by creation of `ux_observations_session_hash` (unique on `memory_session_id, content_hash`).

For databases written by pre-content_hash versions, `content_hash` is NULL on every legacy row. In SQLite, `GROUP BY memory_session_id, content_hash` with NULL hashes groups all of a session's rows together, so `MIN(id)` keeps exactly **one observation per session** and deletes the rest.

Result on our DB: 26,445 → 552 observations (552 = number of distinct sessions), uniformly across all projects. `session_summaries`, `user_prompts`, and `sdk_sessions` were untouched.

Note: the logged "v12.4.3 pollution cleanup" was not the cause — its counters all reported 0 and it has both a marker file and a `CLAUDE_MEM_SKIP_CLEANUP_V12_4_3` env gate. Migration 29 has no gate and no skip mechanism.

## Expected

Migration 29 should not treat NULL hashes as identical content. Either:

1. Restrict the dedupe: `... GROUP BY memory_session_id, content_hash HAVING content_hash IS NOT NULL` (and only delete rows whose `content_hash IS NOT NULL`), or
2. Backfill `content_hash` for legacy rows (hash of the row content) **before** the dedupe runs, or
3. At minimum, take an automatic pre-migration backup (the v12.4.3 cleanup already does this — migration 29 doesn't).

## Recovery (worked for us, may help others)

The pre-upgrade `.backup` allowed restoring into the already-migrated DB without re-triggering the migration:

```sql
ATTACH 'backup.db' AS bak;
INSERT OR IGNORE INTO main.observations (<shared columns>, content_hash)
SELECT <shared columns>, COALESCE(content_hash, 'legacy-'||id)
FROM bak.observations
WHERE id NOT IN (SELECT id FROM main.observations);
INSERT INTO observations_fts(observations_fts) VALUES('rebuild');
```

The synthetic per-row hash keeps the new unique index satisfied.

Anyone upgrading from pre-12.x with data they care about should run `UPDATE observations SET content_hash='legacy-'||id WHERE content_hash IS NULL;` (or equivalent) before first 13.x worker start — or at least take a backup.

## Impact

Silent, near-total loss of historical observations for any long-time user upgrading across this migration. No warning, no automatic backup at this step, and the worker continues normally afterwards so the loss is easy to miss.

rodboev:
## Summary

Migration 29 previously grouped `NULL`-valued `content_hash` values during legacy upgrade, which could collapse all `NULL`s in a memory session to a single row and silently drop history. This change repairs `NULL` hashes to deterministic `legacy-<id>` values inside the migration-29 dedupe step, then runs the existing duplicate-removal and unique-index creation.

## Why

In legacy databases where `content_hash` existed but still contained `NULL` values, migration 29 grouped by `(memory_session_id, content_hash)`, so all `NULL` hashes in the same session could collapse together before the unique index was created.

## Scope

- Preserves rows from legacy databases that already have `content_hash` columns with `NULL` values.
- Preserves existing non-`NULL` dedupe behavior for `(memory_session_id, content_hash)`.
- Does not attempt recovery for rows already deleted in earlier successful migration-29 runs.

## Risk

Legacy databases with non-unique `NULL` hashes now keep all rows, which is the intended compatibility behavior for pre-29 state. Modern observations with non-`NULL` content hashes are unchanged.

## Verification

- [x] `npm install`
- [x] `npm run build`
- [x] `bun test tests/services/sqlite/migration-runner.test.ts tests/sqlite/data-integrity.test.ts`
- [ ] Current red CI is the shared `redis-config` plus `search-routes-welcome-hint` baseline failure reproduced from `main`, not a migration-29 regression

Closes #2812


thedotmack:

PR #2855, opened 2026-06-09
fix(hooks): restore OpenCode capture on the current plugin hook contract (closes #2832)

### LINKED ISSUE #2832: OpenCode plugin loads but captures nothing on opencode 1.16.2 (dead chat.message hook + un-awaited POSTs)
## Summary
On opencode **1.16.2**, the claude-mem OpenCode plugin loads but **captures no observations or sessions**. No `/api/sessions/*` request ever reaches the worker. This is separate from the install/registration issue fixed in #2670 (that makes the plugin *load*); this is about *capture*.

## Environment
- opencode 1.16.2
- claude-mem 13.4.0 (worker/SQLite runtime), worker on its configured port
- Plugin registered in `opencode.json` (`"plugin": ["./plugins/claude-mem.js"]`), confirmed loading (`[claude-mem] OpenCode plugin loading` prints)
- Note: `src/integrations/opencode-plugin/index.ts` is unchanged since `90f70499` (2026-05-28), so this affects current `main` as well, not a stale version.

## Repro
1. Install + register the plugin so it loads.
2. Run any session with a tool call (e.g. `opencode run "run: echo hi"`) or an interactive session (quit/relaunch so the registration is picked up).
3. Check the worker DB (`sdk_sessions`) and worker log: **zero** opencode sessions/observations; no `/api/sessions/init` or `/api/sessions/observations` POST arrives.

## Root cause (`src/integrations/opencode-plugin/index.ts` on `main`)
1. **`chat.message` is not a current opencode plugin hook.** Per https://opencode.ai/docs/plugins, message capture is via `message.updated` / `message.part.updated`. The plugin binds `chat.message`, which never fires, so message-only sessions capture nothing.
2. **Capture POSTs are fire-and-forget** (`workerPostFireAndForget`, ~line 104, not awaited). The async hooks dispatch `fetch()` without awaiting it, so the request is abandoned before it sends; even the valid `tool.execute.after` hook's POST never lands.

## Proposed fix
- Replace `chat.message` with `message.updated` / `message.part.updated` (read assistant text from message parts).
- `await` the worker POSTs inside the hooks so opencode's hook-await keeps the process alive until the POST completes (makes `opencode run` and interactive both deterministic).
- Update `REGISTERED_OPENCODE_HOOKS` + the contract test to the real hook set.

Happy to open a PR with this if it's useful.

rodboev:
## Summary

The OpenCode plugin could load successfully while still capturing no sessions or observations on 1.16.2. This PR switches the assistant-message capture path to the current OpenCode hook names and makes the worker POSTs awaitable so `opencode run` cannot exit before the init and observation requests are sent.

## Why

`REGISTERED_OPENCODE_HOOKS` in `src/integrations/opencode-plugin/index.ts:37-43` and the returned plugin hook map in `src/integrations/opencode-plugin/index.ts:190-270` still bound `chat.message`, which is not the current message hook on OpenCode 1.16.2. Even on valid hook invocations, the capture routes flowed through `workerPostFireAndForget()` at `src/integrations/opencode-plugin/index.ts:104-118`, so the async hook body could resolve before the POST reached the worker. Using the real message hooks and awaiting the worker requests restores deterministic capture for both tool and assistant-message sessions.

## Scope

This PR only fixes OpenCode capture contract drift and request lifetime. It does not change the worker API, bus-event coverage, or the search-tool response parser.

## Risk

The only behavioral change is that OpenCode capture hooks now await their worker requests instead of discarding them after scheduling. That may slightly lengthen hook completion time, but it is the intended tradeoff because the current behavior silently loses capture on short-lived sessions.

## Verification / Test plan

- [x] `bun test tests/integrations/opencode-plugin-contract.test.ts` - 16 passed; verifies the current OpenCode hook names and awaited worker POST lifetime.
- [x] `bun test tests/integration/hook-execution-e2e.test.ts` - 10 passed.
- [x] `npm run build` - passed locally and regenerated `plugin/scripts/worker-service.cjs`.
- [ ] `npm run strip-comments:check` - failed against the existing repo-wide comment-stripping baseline in check mode (`Changed: 327`), unrelated to this OpenCode branch.

Closes #2832


rodboev:
Split the hooks so `message.part.updated` only forces session init while `message.updated` remains the sole assistant-message capture path, added non-OK workerPost warnings, and added contract coverage for `message.part.updated` await behavior.

thedotmack:
Review finding that blocks this one: **the new message hooks never fire**. `message.updated` / `message.part.updated` are registered as top-level plugin hooks, but in opencode v1.16.2's plugin contract (`packages/plugin/src/index.ts` at the v1.16.2 tag) the `Hooks` interface has no such keys — those are *bus events*, delivered only via the `event` hook (the runtime does `void hook[\"event\"]?.({event})` and never dispatches hooks by event name). So assistant-message capture stays dead, and the PR also removes `chat.message`, which *does* exist in the v1.16.2 contract. The updated contract test passes only because the fake hook names were added to its own `REAL_OPENCODE_HOOK_NAMES` allowlist, and `pendingAssistantMessageIds` ends up write-only.\n\nThe awaited-POST change is genuinely correct and fixes the tool-capture half of #2832 — worth keeping. The message-capture half should handle `message.updated`/`message.part.updated` *inside* the `event` hook instead. Happy to merge a revision along those lines.

rodboev:
Fixed in the latest push (`e978751d`). `message.updated` and `message.part.updated` are now handled exclusively inside the `event` hook's switch statement (lines 306-325 of `index.ts`), where OpenCode actually delivers them as bus events with `event.type` as the discriminant. The top-level hook registration only exports `tool.execute.after`, `event`, and `experimental.session.compacting`, which are the three real OpenCode hook names.

The awaited-POST fix from the earlier push is preserved. `REAL_OPENCODE_EVENT_TYPES` is now a separate allowlist of bus event type strings, and the contract test asserts the plugin only switches on names from that list so a future typo against the runtime API fails CI rather than silently dropping capture.

thedotmack:
Closing as superseded by #3208, the current OpenCode lifecycle integration. The useful awaited-delivery work is represented there against the current hook contract.

PR #2856, opened 2026-06-09
fix(context): fold submodule cwd sessions into the parent repo memory (closes #2842)

### LINKED ISSUE #2842: Submodule cwd resolves to empty leaf project — "no memory yet" despite parent repo history
## Summary

When a session's working directory is inside a **git submodule**, claude-mem resolves the project to the submodule's leaf directory name (a brand-new, empty project) instead of the parent repository. The SessionStart/clear/compact context hook then finds zero observations and injects the **"This project has no memory yet"** onboarding banner — even though the parent repo has thousands of observations and capture is otherwise working correctly.

Worktrees are handled; submodules are not.

## Environment

- claude-mem `12.6.0` (plugin, bun worker), Claude Code
- macOS (darwin 25.5.0)
- Monorepo layout: superproject `react/` with git submodules (`peerless/`, `main-api/`, etc.)

## Root cause

`wL()` (worktree detector) reads `<cwd>/.git` and only recognizes **worktrees**:

```js
let i = n.match(/^gitdir:\s*(.+)$/);                       // .git file -> gitdir path
if (!i) return Wf;
let o = i[1].match(/^(.+)[/\\]\.git[/\\]worktrees[/\\]([^/\\]+)$/);  // worktrees ONLY
if (!o) return Wf;                                          // submodule -> bails out here
```

A submodule's `.git` file points at `…/.git/modules/<name>`, not `…/.git/worktrees/<name>`, so the regex never matches and `wL()` returns "not a worktree."

`cr()` then treats the submodule as a standalone top-level project named after its directory:

```js
function cr(t){
  let e = Aie(t);                                  // basename-derived name, e.g. "peerless"
  let n = wL(t);
  if (n.isWorktree && n.parentProjectName)
    return { primary: `${n.parentProjectName}/${e}`, parent: n.parentProjectName, isWorktree:true, ... };
  return { primary: e, parent: null, isWorktree:false, allProjects:[e] };  // submodule lands here -> "peerless"
}
```

The SessionStart context hook resolves the project via `cr(cwd).primary`. Inside a submodule that's `"peerless"` (0 observations) instead of `"react"`, so the onboarding/no-memory branch fires.

## Reproduction

1. A git repo `parent/` with an active claude-mem history (many observations under project `parent`).
2. Add a submodule: `parent/sub/` (so `parent/sub/.git` is a file containing `gitdir: …/parent/.git/modules/sub`).
3. Start (or `/compact`) a Claude Code session while the working directory is inside `parent/sub/`.
4. Observed: `# context` injection shows **"This project has no memory yet"** and zero history.
5. Start/compact with cwd at `parent/` → full history injects correctly.

Confirmed by running the hook directly both ways:

```
printf '{"hook_event_name":"SessionStart","source":"startup","cwd":"<DIR>"}' \
  | node scripts/bun-runner.js scripts/worker-service.cjs hook claude-code context
```

- `cwd = parent`      → `# [parent] recent context` (50 obs, full index)
- `cwd = parent/sub`  → `no memory yet`

Note the working directory is *sticky and follows shell `cd`*, so this also bites users who launch from the parent root but whose agent `cd`s into a submodule before an (auto-)compact fires — there's no way to guarantee cwd location at compact time.

## Expected behavior

A session whose cwd is inside a submodule should resolve to (or at least have access to) the **superproject's** memory, the same way worktrees are folded under their parent.

## Proposed fix

Extend submodule detection alongside the existing worktree path. Two viable approaches:

1. **Parse the `.git` gitdir for the `modules` form** (mirrors the worktree branch):
   ```js
   let mod = i[1].match(/^(.+)[/\\]\.git[/\\]modules[/\\](.+)$/);
   if (mod) {
     const superRoot = mod[1];                       // …/parent
     return { isSubmodule: true, submoduleName: path.basename(t),
              parentRepoPath: superRoot, parentProjectName: path.basename(superRoot) };
   }
   ```
2. **Or use git directly:** `git rev-parse --show-superproject-working-tree` — non-empty output means cwd is in a submodule; its basename is the parent project.

Then in `cr()`, decide the namespacing policy for submodules. Worktrees currently namespace as `parent/<worktree>`. For submodules, the more useful default is usually to **resolve to the superproject project itself** (so the parent's accumulated memory is available), or at minimum expose the parent in `allProjects` so context injection can fall back to it. A config flag (e.g. `CLAUDE_MEM_SUBMODULE_MODE = collapse | namespace`) would cover both preferences.

Happy to test a patch against the monorepo layout above.

rodboev:
## Summary

Sessions launched inside a git submodule were resolving to the submodule leaf directory as a brand-new project, so context injection hit the empty-state path even when the parent repo already had substantial memory. This change teaches the `.git` indirection parser to recognize `.git/modules/*` layouts and updates `getProjectContext()` to include the parent repo project for submodule cwd sessions.

## Why

`detectWorktree()` in `src/utils/worktree.ts:20-55` only recognized `.git/worktrees/*`, so a submodule `.git` file pointing at `.git/modules/*` always fell through to `NOT_A_WORKTREE`. `getProjectContext()` in `src/utils/project-name.ts:83-96` then returned `allProjects: [cwdProjectName]`, and `generateContext()` in `src/services/context/ContextBuilder.ts:107-125` queried only that empty leaf project. Recognizing the submodule parent in the same parsing path fixes the lookup without adding route-specific fallback logic.

## Scope

This PR is limited to project-context resolution for submodule cwd sessions. It does not add a configurable submodule namespacing mode, and it does not change current worktree naming, context rendering, or search-query semantics outside the project list produced by `getProjectContext()`.

## Risk

The behavior change is confined to cwds whose `.git` file points at `.git/modules/*`. Ordinary repos and current worktree handling should remain unchanged, but the project-context tests need to keep both branches covered because the parent-project merge logic now has two valid `.git` indirection shapes instead of one.

## Verification / Test plan

- [x] `bun test tests/utils/project-name.test.ts --test-name-pattern "submodule|worktree isolation"` - 7 passed; covers the submodule parent semantics and existing worktree-isolation behavior.
- [x] `npm run build` - passed locally and regenerated bundled context/worker artifacts.
- [ ] `bun test tests/utils/project-name.test.ts` - not rerun as a full file; a previous full-file run had pre-existing Windows `~` expectation failures.
- [ ] `npm run strip-comments:check` - failed against the existing repo-wide comment-stripping baseline in check mode (`Changed: 327`), unrelated to this submodule branch.

Closes #2842

### DIFF
--- a/src/utils/project-name.ts
+++ b/src/utils/project-name.ts
@@ -93,5 +93,14 @@ export function getProjectContext(cwd: string | null | undefined): ProjectContex
     };
   }
 
+  if (worktreeInfo.isSubmodule && worktreeInfo.parentProjectName) {
+    return {
+      primary: worktreeInfo.parentProjectName,
+      parent: worktreeInfo.parentProjectName,
+      isWorktree: false,
+      allProjects: [worktreeInfo.parentProjectName]
+    };
+  }
+
   return { primary: cwdProjectName, parent: null, isWorktree: false, allProjects: [cwdProjectName] };
 }
--- a/src/utils/project-name.ts
+++ b/src/utils/project-name.ts
@@ -99,7 +99,11 @@ export function getProjectContext(cwd: string | null | undefined): ProjectContex
   const expandedCwd = expandTilde(cwd);
   const contextRoot = findNearestGitContextRoot(expandedCwd);
   const repoRoot = findGitRepoRoot(expandedCwd);
-  const worktreeProbePath = contextRoot ?? repoRoot ?? expandedCwd;
+  const validContextRoot = contextRoot
+    && (!repoRoot || contextRoot === repoRoot || contextRoot.startsWith(`${repoRoot}${path.sep}`))
+    ? contextRoot
+    : null;
+  const worktreeProbePath = validContextRoot ?? repoRoot ?? expandedCwd;
   const worktreeInfo = detectWorktree(worktreeProbePath);
   const cwdProjectName = repoRoot
     ? path.basename(repoRoot)
--- a/src/utils/project-name.ts
+++ b/src/utils/project-name.ts
@@ -32,6 +32,23 @@ function findGitRepoRoot(dir: string): string | null {
   }
 }
 
+function findNearestGitContextRoot(dir: string): string | null {
+  let current = dir;
+
+  while (true) {
+    const worktreeInfo = detectWorktree(current);
+    if (worktreeInfo.isWorktree || worktreeInfo.isSubmodule) {
+      return current;
+    }
+
+    const parent = path.dirname(current);
+    if (parent === current) {
+      return null;
+    }
+    current = parent;
+  }
+}
+
 export function getProjectName(cwd: string | null | undefined): string {
   if (!cwd || cwd.trim() === '') {
     logger.warn('PROJECT_NAME', 'Empty cwd provided, using fallback', { cwd });
@@ -74,14 +91,21 @@ export interface ProjectContext {
 }
 
 export function getProjectContext(cwd: string | null | undefined): ProjectContext {
-  const cwdProjectName = getProjectName(cwd);
-
   if (!cwd) {
+    const cwdProjectName = getProjectName(cwd);
     return { primary: cwdProjectName, parent: null, isWorktree: false, allProjects: [cwdProjectName] };
   }
 
   const expandedCwd = expandTilde(cwd);
-  const worktreeInfo = detectWorktree(expandedCwd);
+  const contextRoot = findNearestGitContextRoot(expandedCwd);
+  const repoRoot = findGitRepoRoot(expandedCwd);
+  const worktreeProbePath = contextRoot ?? repoRoot ?? expandedCwd;
+  const worktreeInfo = detectWorktree(worktreeProbePath);
+  const cwdProjectName = repoRoot
+    ? path.basename(repoRoot)
+    : contextRoot
+      ? path.basename(contextRoot)
+      : getProjectName(cwd);
 
   if (worktreeInfo.isWorktree && worktreeInfo.parentProjectName) {
     const composite = `${worktreeInfo.parentProjectName}/${cwdProjectName}`;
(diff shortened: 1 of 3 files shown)


rodboev:
Reworked `getProjectContext()` to probe the nearest git-context root before falling back, so nested submodule and worktree subdirectories inherit the parent repo context without the redundant early `git rev-parse`. I also kept `parent: null` for submodule sessions and added nested-submodule coverage.

rodboev:
Addressed the ancestor-worktree false positive Greptile flagged.

`getProjectContext()` now only trusts the nearest `.git` file root when it matches the repo root, lives inside it, or `git rev-parse` cannot resolve a repo root at all, so nested submodules still inherit the parent repo context while unrelated home-directory worktrees no longer take over ordinary repos below them.

Focused checks after the patch:
- `bun test tests/utils/project-name.test.ts` — new ancestor-worktree regression passes; the only remaining red is the unchanged Windows `~` expectation baseline already present on this branch
- `bun test tests/hooks/file-context.test.ts tests/worker/http/routes/search-routes-welcome-hint.test.ts`
- `npm run build`
- `npm run lint:hook-io`
- `npm run lint:spawn-env`

thedotmack:
Review finding that blocks this one: **relative `gitdir:` paths break the parse — and create a cross-project memory leak**. Real submodule `.git` files usually contain *relative* paths (`gitdir: ../.git/modules/sub`), and the new regex in `src/utils/worktree.ts` doesn't resolve them against cwd, yielding `parentRepoPath: \"..\"` and `parentProjectName: \"..\"`. Consequence: the issue stays unfixed for git's default layout, and worse — every submodule session in every unrelated repo on the machine collapses into one shared project literally named `..`, so observations from repo A's submodule get injected into repo B's submodule sessions. The PR's tests only cover an absolute `gitdir:` path, which is why this slipped through.\n\nThe concept is sound (the gitdir pointer does guarantee the parent is the superproject, and the repoRoot-bounded ancestor walk in `project-name.ts` is good). Fix: `path.resolve(dirname(gitFilePath), gitdirPath)` before matching, plus a relative-path test case. Happy to merge with that.

rodboev:
Fixed in the latest push (`f879976b`). The relative gitdir path now resolves via `path.resolve(path.dirname(gitPath), gitdirPath)` before either the worktree or submodule regex runs, so both absolute and relative `gitdir:` layouts (git's default for submodules) produce the correct parent repo path. Added a relative-path test case that exercises the `../` form and confirms the parent project name resolves correctly rather than falling through as a literal `..`.

The ancestor-worktree false positive Greptile flagged is also handled: `getProjectContext()` now only trusts the nearest `.git` file root when it matches or lives inside the repo root, so an unrelated home-directory worktree no longer takes over ordinary repos below it.

thedotmack:

PR #2857, opened 2026-06-09
fix(observer): accept prose-empty acknowledgements without poison respawn (closes #2817)

### LINKED ISSUE #2817: v13.4.0 (Linux): observer still poison-loops on prose-empty responses despite plan-11 recovery — root cause (#2749 expected behavior) not addressed
### Summary

On v13.4.0 (post plan-11 `0b4611e89`), the observer SDK still enters a continuous poison/respawn loop when the model legitimately replies with prose like "No observations to record at this time." for queued tool calls. plan-11 added kill+respawn with preserved pending messages, which prevents data loss — but the root cause asked for in #2749 was the opposite: classify prose-empty as a valid empty result (ack the batch, advance the queue), not count it toward the poison threshold. Because that part was not shipped, every investigation-heavy or low-signal session burns ~1 SDK call every 10–20 s in a respawn cycle that can never converge.

### Environment

- claude-mem: **13.4.0** (npm `latest`, cache built 2026-05-29 20:45 UTC — includes `0b4611e89`)
- Claude Code: 2.1.159
- OS: Linux 6.12.85 (Alpine container, Home Assistant OS add-on)
- `CLAUDE_MEM_RUNTIME=worker`, `CLAUDE_MEM_MODEL=claude-haiku-4-5-20251001`
- SDK authMethod: `Claude Code OAuth token (read from system keychain at spawn)`

### Reproduction

Any session whose tool calls are investigative (`ps`, `grep`, `cat`, `ls`) instead of producing concrete code outputs. The observer correctly identifies "nothing decisional to record" and emits prose; parser flags as `outputClass=prose`/`idle`; after 3 consecutive invalids → poison + respawn; the fresh SDK session reproduces the same prose for the next batch.

### Logs (verbatim, today's session)

```
[09:54:04] [SDK   ] ← Response received (119 chars) (No observations to record — no tool executions or technical findings from the primary session have been provided yet.)
[09:54:04] [PARSER] SDK returned non-XML prose response — ignoring queued batch {outputClass=prose, consecutiveInvalidOutputs=2}
[09:54:11] [PARSER] SDK returned non-XML idle response — ignoring queued batch {outputClass=idle, consecutiveInvalidOutputs=3}
[09:54:11] [ERROR ][SESSION] SDK session poisoned — killing and respawning, pending messages preserved {threshold=3}
[09:54:11] [WARN  ][SDK_SPAWN] child emitted error event {errorName=AbortError, errorCode=ABORT_ERR} The operation was aborted.
[09:54:19] [PARSER] SDK returned non-XML idle response — ignoring queued batch {outputClass=idle, consecutiveInvalidOutputs=1}
[09:54:20] [SDK   ] ← Response received (49 chars) (no observations - investigation not yet started)
[09:54:20] [PARSER] SDK returned non-XML prose response — ignoring queued batch {consecutiveInvalidOutputs=2}
[09:54:37] [PARSER] SDK returned non-XML idle response {consecutiveInvalidOutputs=3}
[09:54:37] [ERROR ][SESSION] SDK session poisoned — killing and respawning, pending messages preserved
... (repeats every 10–20s for the entire session lifetime)
```

Daily count from my logs (Linux, same install):

```
Jun 1: 734 poison events  ← investigation-heavy day
Jun 2:   0                ← code-implementation day
Jun 3: 1318               ← investigation-heavy
Jun 4–6: 0                ← code-implementation days
Jun 7: hundreds and counting (current session)
```

### Expected (carried over from #2749)

`outputClass=prose` matching a well-formed "no observations" acknowledgement should:

1. Be treated as a valid empty result for the queued batch (ack, advance queue).
2. **Not** increment `consecutiveInvalidOutputs`.
3. **Not** count toward the poison threshold.

Equivalently, the system prompt could pin a contract requiring `<observations></observations>` even when empty, and the parser would then accept the empty tag — but the current state (prose returned + parser rejects + recovery loop) leaves both ends out of sync.

### Impact

- Observations table stalls during low-signal session phases (today: no new rows since 2:34 AM despite an active 2-hour session).
- Continuous Haiku API calls on the user's subscription — every loop iteration costs one observer call.
- Log volume: ~260 KB of WARN/ERROR per session in my case.

### Notes

- plan-11's kill+respawn protects data integrity (good), but doesn't terminate the loop or stop billable calls (the original ask in #2749).
- Threshold (`YH=3` in bundled `worker-service.cjs`) still hardcoded; no env override exposed.
- Per the closing note on #2784 ("please open a fresh issue for any new observer-output fidelity problem") — filing this as the follow-up.

rodboev:
## Summary

Observer sessions could still respawn after valid no-op responses because `processAgentResponse()` counted every non-XML prose or idle output as invalid before distinguishing empty acknowledgements. This revision treats bounded empty-ack prose and idle responses as valid no-op completions: it confirms the claimed batch, resets stale invalid-output debt, and leaves real malformed or poisoned output on the existing respawn path.

## Why

`ResponseProcessor.ts` classifies non-XML output before deciding whether to confirm the batch or respawn the SDK session. The bug was that issue-reported replies such as `No observations to record at this time.`, `no observations - investigation not yet started`, and `No observations to record — no tool executions or technical findings… have been provided yet.` landed in the same invalid-output counter as malformed prose. After three repeats, `INVALID_OUTPUT_RESPAWN_THRESHOLD` killed and respawned the session even though the observer was saying there was no work to store.

The matcher now recognizes only prose/idle no-op forms with `no observations...` prefixes, a 200-character cap, and rejection of content-bearing continuations such as `No observations, but I found...` or `No observations to record, however I identified...`. Poisoned closure strings and unrecognized prose still increment the counter and respawn exactly as before.

## Scope

This PR is now scoped to observer empty-ack handling only. It does not change telemetry consent caching, search-route welcome-hint caching, XML parsing, poison-string classification, or the invalid-output respawn threshold.

The generated `plugin/scripts/worker-service.cjs` bundle is included because `npm run build` regenerates it from the changed worker source.

## Risk

The main risk is accepting prose that actually contains useful findings. The matcher is bounded by output class, length, known empty-ack prefixes, and explicit rejection of ack-prefixed content signals. Tests cover the issue-reported strings, idle output, the 200-character cap, ordinary prose, and ack-prefixed substantive prose.

## Verification / Test plan

- [x] `bun test tests/worker/agents/response-processor.test.ts tests/worker/poison-respawn.test.ts` — 29 passed, 4 existing skips
- [x] `npm run build` — passed
- [x] `npm run lint:hook-io` — passed
- [x] `npm run lint:spawn-env` — passed
- [ ] `npm run strip-comments:check` — fails on the existing repo-wide baseline, `Changed: 327` in check mode

## Upstream

Closes #2817.
Reported by @Patch76.

### DIFF
--- a/src/services/worker/agents/ResponseProcessor.ts
+++ b/src/services/worker/agents/ResponseProcessor.ts
@@ -24,6 +24,51 @@ import { telemetryBuffer } from '../../telemetry/buffer.js';
  */
 export const INVALID_OUTPUT_RESPAWN_THRESHOLD = 3;
 
+const MAX_EMPTY_ACK_LENGTH = 200;
+const EMPTY_ACK_PREFIXES = [
+  'no observations to record',
+  'no new observations to record',
+  'no observations found',
+  'no new observations found',
+  'no observations',
+  'no new observations',
+] as const;
+const EMPTY_ACK_CONTENT_SIGNAL = /\b(?:but|however|although|except|identified|discovered|learned|recorded|captured|stored|noted|issue|bug|fix|error|failure)\b/;
+
+function hasEmptyAckPrefix(normalized: string, prefix: string): boolean {
+  return normalized === prefix ||
+    normalized.startsWith(`${prefix} `) ||
+    normalized.startsWith(`${prefix}.`) ||
+    normalized.startsWith(`${prefix}!`) ||
+    normalized.startsWith(`${prefix}-`) ||
+    normalized.startsWith(`${prefix} -`) ||
+    normalized.startsWith(`${prefix} —`) ||
+    normalized.startsWith(`${prefix} –`);
+}
+
+function isRecognizedEmptyObserverAck(text: string, outputClass: ReturnType<typeof classifyObserverOutput>): boolean {
+  if (outputClass === 'idle') {
+    return true;
+  }
+
+  if (outputClass !== 'prose') {
+    return false;
+  }
+
+  const normalized = text.trim().toLowerCase();
+  if (normalized.length === 0 || normalized.length > MAX_EMPTY_ACK_LENGTH) {
+    return false;
+  }
+
+  const matchedPrefix = EMPTY_ACK_PREFIXES.find(prefix => hasEmptyAckPrefix(normalized, prefix));
+  if (!matchedPrefix) {
+    return false;
+  }
+
+  const remainder = normalized.slice(matchedPrefix.length);
+  return !EMPTY_ACK_CONTENT_SIGNAL.test(remainder);
+}
+
 export async function processAgentResponse(
   text: string,
   session: ActiveSession,
@@ -57,6 +102,24 @@ export async function processAgentResponse(
     // (plan-11, #2485). Attach a preview for diagnostics.
     const outputClass = classifyObserverOutput(text);
     const preview = previewOutput(text);
+    const recognizedEmptyAck = isRecognizedEmptyObserverAck(text, outputClass);
+
+    if (recognizedEmptyAck) {
+      // Treat recognized empty acknowledgements as healthy output so a long
+      // run of valid no-op responses does not preserve stale respawn debt.
+      session.consecutiveInvalidOutputs = 0;
+
+      logger.warn('PARSER', `${agentName} returned non-XML empty-ack ${outputClass} response, confirming claimed batch without respawn`, {
+        sessionId: session.sessionDbId,
+        outputClass,
+        preview,
+        consecutiveInvalidOutputs: session.consecutiveInvalidOutputs,
+      });
+
+      await sessionManager.confirmClaimedMessages(session.sessionDbId);
+      session.earliestPendingTimestamp = null;
+      return;
+    }
 
     session.consecutiveInvalidOutputs = (session.consecutiveInvalidOutputs ?? 0) + 1;
 
--- a/src/services/worker/agents/ResponseProcessor.ts
+++ b/src/services/worker/agents/ResponseProcessor.ts
@@ -34,16 +34,52 @@ const EMPTY_ACK_PREFIXES = [
   'no new observations',
 ] as const;
 const EMPTY_ACK_CONTENT_SIGNAL = /\b(?:but|however|although|except|identified|discovered|learned|recorded|captured|stored|noted|issue|bug|fix|error|failure)\b/;
+const EMPTY_ACK_NEUTRAL_REMAINDER_WORDS = new Set([
+  'at',
+  'this',
+  'time',
+  'for',
+  'batch',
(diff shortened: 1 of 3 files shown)


rodboev:
Tightened the empty-ack matcher to full-response acknowledgements so longer prose that merely mentions "no observations" no longer bypasses the invalid-output counter, while keeping the known "No observations to record for this batch." reply recognized.

thedotmack:
Review finding that blocks this one: **the matcher is too narrow to fix the issue it closes**. The anchored patterns in `ResponseProcessor.ts` (e.g. `/^\s*no (?:new )?observations?(?: to record(?: for this batch)?| found)?[.!]?\s*$/i`) fail to match every verbatim example in #2817: \"No observations to record at this time.\", \"no observations - investigation not yet started\", and \"No observations to record — no tool executions or technical findings… have been provided yet.\" all have trailing text the regex rejects. The tests pass only on the PR's own invented string \"No observations to record for this batch.\" Also `idle` (empty) responses — which dominate the issue's logs — still count toward respawn.\n\nThe gating approach itself is right (prose-only, so poisoned closure strings still respawn). Suggest a recall-oriented matcher instead: prose class + case-insensitive prefix `no observations` + bounded length (say <200 chars), and consider exempting `idle` from the counter too. Happy to merge a revision.

rodboev:
Replaced the anchored regex patterns with the prefix-based matcher you suggested: prose class + case-insensitive `startsWith('no observations')` + the existing 200-char length gate. The `EMPTY_ACK_PATTERNS` array is gone entirely.

All three verbatim examples from #2817 now match: "No observations to record at this time.", "no observations - investigation not yet started", and "No observations to record — no tool executions or technical findings… have been provided yet." `idle` responses were already exempted from the counter in the prior push (line 31-33 of the old code, preserved here); that path returns `true` and clears accumulated respawn debt without incrementing.

Tests updated to cover the exact issue strings, the 200-char boundary, and rejection of non-prefix prose. 23 response processor tests pass, build and lint clean.

rodboev:
Cleaned this up to rebase on current `main` and scoped it back to observer empty acknowledgements only.

Also reworked to remove the unrelated telemetry and search-route changes, keeping the issue-reported prose and idle cases covered, and adding regression coverage for the prior overmatch risk. Ack-prefixed content such as `No observations, but I found...` now stays on the invalid-output path instead of being confirmed as an empty no-op.

Focused tests, build, and the hook/spawn lint gates pass. `strip-comments:check` still fails on the repo-wide baseline with `Changed: 327` in check mode, so I left that checklist item unchecked in the PR body.

thedotmack:
Closing as superseded by current main’s general rule that incompatible observer output is ignored without poison or respawn accounting.

PR #2859, opened 2026-06-09
fix(installer): require explicit consent before disabling Claude Code auto-memory (closes #2836)

### LINKED ISSUE #2836: BUG: Silent disabling of Claude Code auto_memory on install causes complete memory loss
## Summary

Installing claude-mem v13.4.0 silently disables Claude Code's built-in auto_memory system by writing `CLAUDE_CODE_DISABLE_AUTO_MEMORY=1` to `~/.claude/settings.json`. This causes users who rely on the native `memory/MEMORY.md` system to lose **all accumulated memory** on their next session — with zero warning or consent.

## Steps to Reproduce

1. Use Claude Code with the built-in auto_memory system (memory/ folder with MEMORY.md index, 75+ memory files accumulated over months)
2. Install claude-mem plugin: `npx claude-mem install`
3. Restart Claude Code
4. Claude has zero memory — doesn't know who you are, your preferences, your projects, or your rules

## Root Cause

`src/npx-cli/commands/install.ts` → `disableClaudeAutoMemory()` writes:

```json
"env": { "CLAUDE_CODE_DISABLE_AUTO_MEMORY": "1" }
```

to `~/.claude/settings.json` without **any user prompt or confirmation**.

The install summary shows "Auto-memory: disabled" as a line item, but this is buried in output that scrolls by during installation — not a confirmation dialog.

## Impact — Critical

- **75 memory files** (user profile, 25+ behavioral rules, 20+ active projects, tool preferences, shared memory protocols) became invisible overnight
- User spent significant time debugging "why did all my memory disappear"
- The memory files are NOT deleted (still on disk), but Claude Code no longer reads them at startup — making them functionally dead
- claude-mem's own memory store is empty on first session, so the user ends up with **zero memory from either system**
- For power users who have built up months of context, this is a catastrophic data loss from their perspective

## Expected Behavior

1. **Ask before disabling**: The installer should explicitly warn: *"claude-mem will disable Claude Code's built-in auto_memory. You have X memory files that will no longer be loaded. Continue? [y/N]"*
2. **Offer migration**: Before disabling, offer to import existing memory/ files into claude-mem's system
3. **At minimum**: Make the disable behavior opt-in, not automatic. Add a flag like `--disable-auto-memory` for users who explicitly want it

## Current Workaround

Manually remove from `~/.claude/settings.json`:
```json
"env": { "CLAUDE_CODE_DISABLE_AUTO_MEMORY": "1" }
```

## Environment

- Claude Code version: latest (Opus 4.6)
- claude-mem version: 13.4.0
- OS: Windows 11
- Number of existing memory files: 75
- Time to diagnose: ~30 minutes of a power user's session

## Suggestion

The comment in install.ts says *"the built-in MEMORY.md system creates shadow state"* — but for users who actively manage their MEMORY.md, it's not shadow state, it's their **primary knowledge base**. Please respect existing user data.

---
Reported by a power user (Tony) via Claude Code session 2026-06-08.

rodboev:
## Summary

The installer was silently writing `CLAUDE_CODE_DISABLE_AUTO_MEMORY=1` into Claude Code settings for every Claude Code install, which could make an existing native memory corpus disappear on the next session. This PR moves that write behind an explicit user decision, keeps non-interactive installs untouched unless they pass an explicit disable flag, and reports the choice clearly in the install summary.

## Why

`disableClaudeAutoMemory()` in `src/npx-cli/commands/install.ts:159-169` is correct as a settings mutation primitive, but `runInstallCommandInner()` called it unconditionally for every Claude Code install at `src/npx-cli/commands/install.ts:1489-1508`. The bug was therefore not in file I/O or idempotency; it was that the installer had no consent gate before mutating `~/.claude/settings.json`.

## Scope

This PR only changes the decision flow around disabling native auto-memory. It does not implement import or migration of existing memory files, and it does not alter plugin registration, worker startup, or non-Claude-Code IDE installs.

## Risk

The behavioral change is deliberate: fresh installs will now preserve Claude Code's native auto-memory unless the user opts in to disabling it. The only compatibility risk is automation that implicitly relied on the old silent disable, which is why the non-interactive path should gain an explicit flag rather than losing the capability outright.

## Verification / Test plan
- [x] `bun test tests/install-disable-auto-memory.test.ts` — 18 passed; covers the explicit consent flow, the non-interactive default, and the new `left enabled` summary state.
- [x] `bun test tests/install-non-tty.test.ts` — 39 passed; verifies `--disable-auto-memory` parsing and help text.
- [x] `npm run build` — passed; no tracked generated artifacts changed for this fix.
- [x] `npm run lint:hook-io && npm run lint:spawn-env` — passed.
- [x] `npm run strip-comments:check` — fails on the existing repo-wide baseline (`Changed: 301` in check mode), unrelated to this branch.

Closes #2836


rodboev:
Updated the stale inline docs around `autoMemoryStatus` so they match the consent-gated flow and the four actual summary states.

PR #2860, opened 2026-06-09
fix(sqlite): bound stored prompt payloads to stop prompt-table bloat (closes #2793)

### LINKED ISSUE #2793: Database grows unbounded: prompt_text stored verbatim (×2), no retention, no VACUUM
## Summary

The SQLite database (`~/.claude-mem/claude-mem.db`) grows **unbounded** because full prompt submissions are stored verbatim with no size cap, no retention/pruning, and no `VACUUM`. On a long-running install it reached **136 GB** (plus ~122 GB of observer-session transcripts), almost all of it recoverable junk.

Installed version: **12.1.6** (also reviewed release notes through 13.4.0 — none address this).

## Root cause (from inspecting the schema + `worker-service.cjs`)

1. **`user_prompts.prompt_text TEXT NOT NULL`** stores the *entire* UserPromptSubmit payload verbatim. With injected context (CLAUDE.md, rule files, the full MCP tool list, memory injections, etc.), each row averaged **~990 KB**, with the largest at **11.6 MB**. 82k rows = **78 GB**.
2. **The same text is duplicated** in `sdk_sessions.user_prompt` — another **39 GB** across 24k rows.
3. **No retention/pruning** of `user_prompts`, `sdk_sessions`, or `observations`. The only `DELETE` statements in the worker are on `pending_messages` (on confirm/abort). Old rows accumulate forever.
4. **No `VACUUM`** anywhere — even when rows are deleted, the file never shrinks (freelist grows but disk is never reclaimed).
5. **`pending_messages` queue jam** — rows with `status='failed'` are never cleaned up. This install had **175,251** failed rows.

### Real footprint observed
| Location | Size |
|---|---|
| `claude-mem.db` (`user_prompts.prompt_text` 78 GB + `sdk_sessions.user_prompt` 39 GB + FTS) | 136 GB |
| `~/.claude/projects/*observer-sessions` (28k transcript files) | 122 GB |
| `chroma/` | 8.4 GB |

The actually-useful data (`observations`) was only **10,602 rows ≈ 0.02 GB**.

### Manual recovery that worked (DB 136 GB → 48 MB, observations preserved)
```sql
-- (stop worker first)
DROP TRIGGER user_prompts_ai; DROP TRIGGER user_prompts_ad; DROP TRIGGER user_prompts_au;
DELETE FROM user_prompts;
INSERT INTO user_prompts_fts(user_prompts_fts) VALUES('rebuild');
-- recreate the 3 triggers
UPDATE sdk_sessions SET user_prompt = NULL;
DELETE FROM pending_messages WHERE status='failed';
VACUUM;
```

## Suggested fixes
1. **Cap / truncate** stored `prompt_text` (e.g. strip injected context, or store only the user's actual message; or a configurable max length).
2. **Don't duplicate** the prompt in both `user_prompts` and `sdk_sessions`.
3. **Add a retention policy** (configurable max age / max rows) with periodic pruning of `user_prompts`, `sdk_sessions`, old `observations`, and `failed` `pending_messages`.
4. **Run `VACUUM`** (or `PRAGMA auto_vacuum=INCREMENTAL`) after pruning so disk is actually reclaimed.
5. **Cap observer-session transcript retention** under `~/.claude/projects/*observer-sessions`.

Note: v13's Postgres backend changes *where* the data lives but, without retention + the prompt-storage cap, the same unbounded growth would apply to Postgres.

## Environment
- claude-mem 12.1.6
- macOS (Apple Silicon)
- Claude Code

rodboev:
## Summary

SQLite prompt storage was amplifying database growth by persisting the full prompt payload twice, once on `sdk_sessions.user_prompt` and once on `user_prompts.prompt_text`, with FTS mirroring the second copy again. This PR adds a shared prompt-normalization cap before both write paths so new prompt rows keep useful user-authored content without preserving the full injected context blob verbatim.

## Why

The current schema stores prompt text in both `sdk_sessions.user_prompt` (`src/services/sqlite/schema.sql:28-44`) and `user_prompts.prompt_text` (`src/services/sqlite/schema.sql:157-165`). `createSDKSession()` writes the session copy at `src/services/sqlite/sessions/create.ts:63-70`, and `saveUserPrompt()` writes the prompt-history copy at `src/services/sqlite/prompts/store.ts:5-21`. Because `createUserPromptsTable()` in `src/services/sqlite/SessionStore.ts:407-473` also mirrors prompt history into FTS, every oversized prompt payload is multiplied across multiple SQLite structures. Bounding the payload before either insert stops the dominant growth source without requiring a schema rewrite.

## Scope

This PR is intentionally limited to forward-write normalization of stored prompt text. It does not add pruning, transcript retention, or `VACUUM` scheduling, and it does not remove existing columns or rewrite historical rows.

## Risk

The tradeoff is that stored prompt text becomes a summarized or truncated representation rather than a verbatim copy of the full injected payload. That is acceptable for this subsystem because prompt history is used for UI/search recall, while the full durable memory value comes from observations and summaries, not from replaying giant prompt wrappers.

## Verification / Test plan
- [x] `bun test tests/sqlite/prompts.test.ts` — 9 passed; verifies tag-stripped oversized prompt payloads are bounded before prompt-history storage.
- [x] `bun test tests/sqlite/sessions.test.ts` — 16 passed; verifies the session row stores the same normalized prompt text as prompt history.
- [x] `bun test tests/session_store.test.ts` — 8 passed; the `SessionStore` convenience path still behaves correctly after sharing the normalizer.
- [x] `npm run build` — passed; regenerated `plugin/scripts/context-generator.cjs` from the shared SQLite/session-store source change.
- [x] `npm run lint:hook-io && npm run lint:spawn-env` — passed.
- [x] `npm run strip-comments:check` — fails on the existing repo-wide baseline (`Changed: 301` in check mode), unrelated to this branch.

Closes #2793


thedotmack:

PR #2877, opened 2026-06-10
refactor: reorganize HTTP server directories to remove server/servers/services-server ambiguity

darion-yaphet:
## Why

Three directories were named almost identically with no way to tell them apart:

| Directory | Actual role |
|---|---|
| `src/services/server/` | Express HTTP runtime shared by the worker and the server-beta stack |
| `src/server/` | server-beta cloud stack (auth, Postgres routes, queues, jobs) |
| `src/servers/` | a single file — the stdio MCP server entrypoint |

Worse, the dependency direction was tangled both ways:
- 7 files in the newer `src/server/` tree imported `RouteHandler` (and `ServerBetaService` imported the `Server` class itself) from the older `src/services/server/Server.ts`
- `src/services/server/Server.ts` imported `src/server/queue/queue-health-types.ts` back from the new tree

Neither tree could be refactored or retired without touching the other.

## What

Pure move + import-path updates. **No behavior change.**

- `src/services/server/` → `src/shared/http/` — the shared HTTP runtime now lives in a neutral location both consumers can depend on
- `src/server/queue/queue-health-types.ts` → `src/shared/http/` — its only consumer is `Server.ts`; moving it breaks the old→new back-edge entirely
- `src/servers/mcp-server.ts` → `src/entrypoints/mcp-server.ts` — the directory name now says what it is: a bundled entrypoint
- `tests/servers/` → `tests/entrypoints/`; `scripts/build-hooks.js` entry path updated
- `src/server/` (cloud stack) is untouched — with the other two gone, its name is no longer ambiguous

After this PR the dependency direction is one-way: `src/server/` and `src/services/worker-service.ts` both depend on `src/shared/http/`, and nothing in `src/shared/http/` reaches back into `src/server/`.

## Test plan

- [x] `tsc --noEmit` clean (the `posthog-node` resolution error preexists on a clean tree before `npm install`)
- [x] `npm run build` succeeds; bundle size guardrails pass (mcp-server.cjs / worker-service.cjs budgets)
- [x] Full suite: 2197 pass / 2 fail — the same 2 telemetry test-isolation flakes fail identically on `main` (pass when run in isolation)


darion-yaphet:
Should I submit to another branch?

thedotmack:
Closing as outdated. A repository-wide directory reorganization now conflicts with substantial subsequent architecture work and should not be revived as a 42-file stale refactor.

PR #2878, opened 2026-06-10
refactor: unify SQLite persistence under src/storage/

darion-yaphet:
## Why

The codebase had two parallel SQLite layers in sibling trees with no shared home:

| Tree | Contents | Schema mechanism |
|---|---|---|
| `src/services/sqlite/` | worker store: observations, sessions, summaries, prompts, timeline | versioned migration ladder (`migrations/runner.ts`) |
| `src/storage/sqlite/` | server-beta repositories: projects, teams, api_keys, audit_log, … | idempotent `ensureServerStorageSchema()` |

What makes the split actively misleading: **both table sets live in the same database file** (`~/.claude-mem/claude-mem.db`). The worker's migration runner already calls `ensureServerStorageSchema()` (`runner.ts`, migration v33) to create the server-owned tables — the two trees co-own one database, but nothing in the layout says so. Meanwhile the Postgres repositories already lived under `src/storage/postgres/`, so persistence was split across `services/` and `storage/` for no structural reason.

## What

Pure move + import-path updates. **No behavior change, no schema change.**

```
src/storage/sqlite/          worker-owned tables + migration ladder (was src/services/sqlite/)
src/storage/sqlite/server/   SERVER_OWNED_TABLES + ensure-schema     (was src/storage/sqlite/*.ts)
src/storage/postgres/        server-beta Postgres repositories       (untouched)
```

- The worker store becomes the package root, matching its role as the database's primary owner; server-beta repositories become an explicit `server/` subpackage.
- `src/storage/sqlite/index.ts` now documents which owner controls which table set and where each schema source lives, so the single-file/two-owners contract is written down instead of implied.
- `tests/services/sqlite/` → `tests/storage/sqlite/` to keep the mirror.
- The migration runner's cross-import collapses from `../../../storage/sqlite/schema.js` to `../server/schema.js`.

Deliberately **out of scope**: entity-level unification (`server_sessions` vs `sdk_sessions`, `memory_items` vs `observations`) — that is a product/data-model decision, not a layout fix.

Note: #2877 touches one shared line (`SessionSearch.ts`'s ErrorHandler import). Whichever lands second gets a trivial one-line rebase; happy to do it.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `npm run build` passes, all bundle-size guardrails green
- [x] Full suite: 2203 pass / 0 fail


darion-yaphet:
Closing: superseded by the `edc5cf7d` audit on main, which already removed the duplicate worker SQLite classes (~10.4k lines) that this PR set out to consolidate. After rebasing onto current main, the only remaining change is a pure file relocation (`src/services/sqlite/*` → `src/storage/sqlite/*`, server-beta files → `src/storage/sqlite/server/`) with import-path churn across 58 files and no behavior change. Not worth the churn/conflict surface. Abandoning the branch.

Issue #2818, opened 2026-06-07
[Windows] Chroma vector sync never connects — cmd.exe /c uvx mangles --with ">=" / "<" args (MCP -32000 Connection closed)


liu91168:
## Summary

On Windows, the Chroma MCP integration never connects. Every observation logs:

```
[CHROMA_MCP] Connecting to chroma-mcp via MCP stdio {command=C:\WINDOWS\system32\cmd.exe,
  args=/c uvx --python 3.11 --with "onnxruntime>=1.20" --with "protobuf<7" chroma-mcp==0.2.6
       --client-type persistent --data-dir C:/Users/<me>/.claude-mem/chroma}
[CHROMA_MCP] Connection failed, killing subprocess tree ... {error=MCP error -32000: Connection closed}
[CHROMA] ... chroma sync failed, continuing without vector search
```

The child process dies in **~20 ms** (so it is *not* the 30 s MCP handshake timeout). Vector search silently and permanently degrades to "continuing without vector search"; only SQLite FTS remains.

## Environment

- Windows 11, claude-mem **v13.4.0** (plugin / worker runtime)
- `uv`/`uvx` 0.11.15 in `~/.local/bin`, Python 3.11 (user-installed)
- Provider: `openrouter` — LLM calls work perfectly; **only Chroma is affected**

## Root cause

The Windows branch launches Chroma through `cmd.exe /c uvx ...`
(`worker-service`: `s = process.platform === "win32" ? process.env.ComSpec || "cmd.exe" : "uvx"`).

The hard-coded `--with` constraints contain shell metacharacters `>` and `<`:

```
--with onnxruntime>=1.20    --with protobuf<7
```

The code quotes any arg matching `[<>|&^()]` (the `Fse()` helper) to protect it from cmd redirection. But the arg then passes through **two escaping layers** — Node's `child_process.spawn` Windows command-line builder, then `cmd.exe` re-parsing — and the quoting ends up mangled, so `uvx`/`cmd` receives broken arguments and exits instantly.

## Reproduction / evidence

Isolated `child_process.spawn` probes against the *exact* args the worker uses:

| # | Spawn | Result |
|---|-------|--------|
| 1 | `cmd /c uvx --version` | ✅ exit 0 — PATH / uvx fine |
| 2 | `cmd /c uvx ... chroma-mcp==0.2.6 --help` (**no `--with`**) | ✅ exit 0 — chroma-mcp itself fine |
[...]

PR #2880, opened 2026-06-10
fix(chroma): spawn chroma-mcp via `--from` so uv < 0.5.31 works

### LINKED ISSUE #2879: Chroma semantic search never connects with uv < 0.5.31 — bare `uvx <pkg>==<ver>` positional rejected → MCP -32000 Connection closed (repro: macOS)
## Before submitting
- [x] I searched existing issues and confirmed this is not a duplicate. The existing chroma-mcp spawn issues (#2696, #2790, #2818, #2861, …) are all **Windows / cmd.exe** shell-escaping or PATH problems. This is a distinct, **cross-platform, uv-version** axis and is not fixed by those.

## Summary
Where the worker resolves a **uv older than 0.5.31**, Chroma semantic search never connects. Reproduced on **macOS** (details below); the failure is gated by the **uv version**, not the OS — see Root cause. The worker spawns `chroma-mcp` via `uvx … chroma-mcp==0.2.6 …`, the subprocess dies in ~13 ms, and the only surfaced signal is `MCP error -32000: Connection closed`. The connection enters reconnect backoff and search silently degrades to keyword-only. Capture/DB are unaffected.

## Environment
- claude-mem 13.4.0 / 13.5.5 (reproduced on both)
- macOS (Darwin arm64); not Windows
- uv resolved by the worker: `0.5.11` and `0.5.29` (mise-managed, global-pinned). Both **< 0.5.31**.
- Chroma mode: local (persistent), python 3.13

## Symptom
`GET /api/chroma/status?deep=1`:
```json
{"status":"unhealthy","connected":false,
 "details":"chroma-mcp deep probe failed at stage 'list'",
 "probe":{"ok":false,"stage":"list","error":"chroma-mcp connection in backoff (6s remaining)"}}
```
Worker log (`~/.claude-mem/logs/`):
```
[CHROMA_MCP] Connecting to chroma-mcp via MCP stdio {args=--python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 --client-type persistent --data-dir …}
[CHROMA_MCP] Connection failed, killing subprocess tree to prevent zombie {error=MCP error -32000: Connection closed}
```
Each attempt fails in ~13–80 ms (far too fast for chroma-mcp to boot Python).

## Root cause
`ChromaMcpManager.buildCommandArgs()` invokes the package as a **bare positional**:
```
uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 --client-type persistent …
```
The `uvx <pkg>==<ver>` positional form (PEP 508 specifier in a tool request) was only added in **uv 0.5.31** (["Allow PEP 508 requirements in tool requests" #11337](https://github.com/astral-sh/uv/pull/11337), ["Bring parity to `uvx` and `uv tool install` requests" #11345](https://github.com/astral-sh/uv/pull/11345)). On older uv it is rejected:
```
error: Not a valid package or extra name: "chroma-mcp==0.2.6".
       Names must start and end with a letter or digit …
```
uvx exits immediately → the MCP stdio handshake gets `Connection closed`.

A secondary problem makes this hard to diagnose: the transport is created with `stderr: 'pipe'` but nothing ever reads it, so the actual uv error (`Not a valid package…`) is swallowed and only the generic `-32000 Connection closed` reaches the logs.

## Reproduction (deterministic)
Same args, two uv versions:
```bash
# uv 0.5.29  → FAILS
uvx --python 3.13 --with 'onnxruntime>=1.20' --with 'protobuf<7' \
    chroma-mcp==0.2.6 --client-type persistent --data-dir ~/.claude-mem/chroma
# → error: Not a valid package or extra name: "chroma-mcp==0.2.6".

# uv 0.8.23  → boots, returns serverInfo {"name":"chroma","version":"1.6.0"}
```
Boundary matches the changelog exactly: **uv ≤ 0.5.30 fails, uv ≥ 0.5.31 works.**

## Fix
Use the canonical invocation that works on **every** uv version (and is the form uv's own docs recommend for version specifiers):
```
uvx … --from chroma-mcp==0.2.6 chroma-mcp --client-type persistent …
```
Verified booting chroma-mcp on uv 0.5.29 **and** 0.8.23 with identical surrounding args (macOS). This is orthogonal to the Windows fixes (#2696 direct-spawn, #2790 PATH): it touches only the package-invocation form in `buildCommandArgs()`, contains no cmd.exe metacharacters, and doesn't regress the new-uv path. The gate is the uv version (per astral-sh/uv#11337), which is platform-agnostic, so a Windows host on uv ≤ 0.5.30 should benefit too — though I have **not** tested on Windows.

PR attached.

## Suggested secondary improvement (optional)
Log `transport.stderr` on connect failure so the underlying uv/chroma error surfaces in `~/.claude-mem/logs/` instead of being masked by `-32000 Connection closed`.

husniadil:
## What
Change the chroma-mcp invocation in `ChromaMcpManager.buildCommandArgs()` from the bare positional form to the `--from` form:
```diff
- `chroma-mcp==${CHROMA_MCP_PINNED_VERSION}`,
+ '--from', `chroma-mcp==${CHROMA_MCP_PINNED_VERSION}`, 'chroma-mcp',
```
Applied to both the remote (`http`) and local (`persistent`) arg builders via a shared `packageInvocation` constant, with an explanatory comment.

## Why
The bare `uvx <pkg>==<ver>` positional (PEP 508 specifier in a tool request) was only added in **uv 0.5.31** ([astral-sh/uv#11337](https://github.com/astral-sh/uv/pull/11337), [#11345](https://github.com/astral-sh/uv/pull/11345)). On any host whose worker resolves uv ≤ 0.5.30, uvx rejects the spec:
```
error: Not a valid package or extra name: "chroma-mcp==0.2.6".
```
The subprocess dies in ~13 ms and the MCP client only sees `MCP error -32000: Connection closed` → reconnect backoff → semantic search silently falls back to keyword. (`stderr` is piped but never read, so the real uv error is masked.)

`--from <pkg>==<ver> <cmd>` is the long-standing version-specifier form that uv's docs recommend, and is verified here on uv 0.5.29 and 0.8.23 — so it works across uv versions where the bare positional does not.

## Distinct from the Windows issues
All prior chroma-mcp spawn fixes (#2696 direct-spawn, #2790 PATH, #2818/#2861/… cmd.exe quoting) address Windows shell/PATH mechanics and leave the **invocation form** unchanged. This is a separate, cross-platform **uv-version** axis:

| host uv | before | after |
|---|---|---|
| ≥ 0.5.31 | works | works (no change) |
| ≤ 0.5.30 (e.g. mise/pinned 0.5.x) | **MCP -32000, search offline** | **works** |

No regression to the new-uv or Windows paths: the change touches only the args array, adds no cmd.exe metacharacters, and the existing `resolveUvxCommand()` direct-spawn is untouched.

## Testing
Booted `chroma-mcp==0.2.6` over MCP stdio with the new args on **uv 0.5.29 and uv 0.8.23** — both return `serverInfo {"name":"chroma","version":"1.6.0"}`. End-to-end: built 13.5.5, reinstalled locally, `GET /api/chroma/status?deep=1` → `{"status":"healthy", … "stage":"done"}` (semantic round-trip ~88 ms).

Checks run locally on this change:
- `npm run typecheck` — passes
- `npm run build` — passes
- `bun test` over the chroma/spawn suites (`tests/services/sync/`, `spawn-contract-windows`, `chroma-search-strategy`) — 37 pass / 0 fail
- No dependency change, so `plugin/bun.lock` is unchanged (frozen-lockfile CI check is a no-op).

> Verified on **macOS only**. Linux and Windows are not tested; the gate is the uv version (astral-sh/uv#11337), which is platform-agnostic (same Rust binary), so those hosts on uv ≤ 0.5.30 are expected to benefit — flagging the untested platforms for transparency.

Fixes #2879

### DIFF
--- a/src/services/sync/ChromaMcpManager.ts
+++ b/src/services/sync/ChromaMcpManager.ts
@@ -204,6 +204,15 @@ export class ChromaMcpManager {
 
     const depOverrideFlags = CHROMA_MCP_DEP_OVERRIDES.flatMap(spec => ['--with', spec]);
 
+    // Invoke the package via `--from chroma-mcp==<ver> chroma-mcp` rather than the
+    // bare `chroma-mcp==<ver>` positional. The bare-positional sugar (`uvx PKG==VER`)
+    // only parses on uv >= 0.5.31 (PEP 508 specifiers in tool requests landed in
+    // astral-sh/uv#11337); on older uv (e.g. mise-pinned 0.5.x) it's rejected with
+    // `error: Not a valid package or extra name`, the subprocess dies in ~13ms, and
+    // it surfaces only as `MCP error -32000: Connection closed`. The `--from` form
+    // is the canonical invocation and works on every uv version.
+    const packageInvocation = ['--from', `chroma-mcp==${CHROMA_MCP_PINNED_VERSION}`, 'chroma-mcp'];
+
     if (chromaMode === 'remote') {
       const chromaHost = settings.CLAUDE_MEM_CHROMA_HOST || '127.0.0.1';
       const chromaPort = settings.CLAUDE_MEM_CHROMA_PORT || '8000';
@@ -215,7 +224,7 @@ export class ChromaMcpManager {
       const args = [
         '--python', pythonVersion,
         ...depOverrideFlags,
-        `chroma-mcp==${CHROMA_MCP_PINNED_VERSION}`,
+        ...packageInvocation,
         '--client-type', 'http',
         '--host', chromaHost,
         '--port', chromaPort
@@ -241,7 +250,7 @@ export class ChromaMcpManager {
     return [
       '--python', pythonVersion,
       ...depOverrideFlags,
-      `chroma-mcp==${CHROMA_MCP_PINNED_VERSION}`,
+      ...packageInvocation,
       '--client-type', 'persistent',
       '--data-dir', DEFAULT_CHROMA_DATA_DIR.replace(/\\/g, '/')
     ];


thedotmack:
Closing as already implemented on current main: the Chroma uvx invocation uses --from and the explicit chroma-mcp entrypoint.

PR #2883, opened 2026-06-10
fix(sqlite): make cwd-remap reconciliation idempotent to fix orphaned worktree observations (closes #2864)

### LINKED ISSUE #2864: Worktree observations orphaned by one-shot `.cwd-remap-applied-v1` migration; unreachable after the worktree is removed (v13.3.0)
## Summary

Observations recorded while working inside a `git worktree` are written under a composite `parent/worktree` project key (expected, per the current resolver). The problem is **recovery**: the one-time remap that is supposed to reconcile project keys (`nie()` → `oJe()`, guarded by the `~/.claude-mem/.cwd-remap-applied-v1` marker) only ever runs **once per install**. Any worktree created *after* that marker is written is never reconciled, and `merged_into_project` is left `NULL`. Combined with the read side resolving the parent only from a *live* worktree directory, this means:

- Worktree observations never reach the parent project, **and**
- Once the worktree directory is removed, they become **unreachable by any read path** — effectively orphaned/lost.

This is distinct from the worktree→parent **write** feature tracked in #2711 / roadmap #2785. That feature is about *where new writes go*; this report is about (a) the recovery migration being a permanent one-shot, and (b) data becoming unreachable after worktree removal. Even if the write-side opt-in ships, existing installs that have already passed the one-shot migration won't benefit, and already-orphaned data stays orphaned.

## Environment

- claude-mem **13.3.0** (plugin, bun/cjs worker)
- Claude Code, macOS (darwin 25.5.0)
- Standard (non-bare) repo `myrepo`, with a worktree `myrepo-feature-x`

## What I observed

The repo's history is split across two project keys, and the parent key cannot see the worktree's observations:

```sql
sqlite3 ~/.claude-mem/claude-mem.db \
 "SELECT project, COUNT(*) FROM observations GROUP BY project;"
-- myrepo               | 4
-- myrepo/myrepo-feature-x | 11
```

All worktree rows have `merged_into_project = NULL`:

```sql
SELECT DISTINCT merged_into_project
FROM observations WHERE project = 'myrepo/myrepo-feature-x';
-- (null)
```

The worker read API for the parent returns only the parent rows — the worktree rows are not surfaced:

```
curl -s "http://localhost:$PORT/api/context/inject?project=myrepo&full=true"
-- Stats: 4 obs ... (the worktree observations are absent)
```

The worktree directory has since been removed (normal cleanup after the branch merged), so there is no longer a live worktree from which the read-side `parent,worktree` composite can be reconstructed.

## Root cause

**1. Capture composes a `parent/worktree` key.** The worktree-aware project resolver returns `parent_basename/worktree_basename` for a worktree checkout (in 13.3.0's bundle this is the `kind: "worktree"` branch that returns `` `${basename(commonDirParent)}/${basename(topLevel)}` ``). That yields `myrepo/myrepo-feature-x`. Expected given current design.

**2. Reconciliation is a permanent one-shot.** The remap routine is gated by a marker file:

```js
// nie(dataDir):
if (existsSync(".cwd-remap-applied-v1")) {           // marker present → return immediately
  log("cwd-remap marker exists, skipping");
  return;
}
// ... else run oJe(db, dataDir, marker) and write the marker ...
```

In my case the marker is dated **2026-05-21**, while the worktree observations were created **2026-06-08** — 18 days later:

```
$ stat -f "%Sm %N" ~/.claude-mem/.cwd-remap-applied-v1
May 21 11:18:04 2026 .../.cwd-remap-applied-v1
$ cat ~/.claude-mem/.cwd-remap-applied-v1
2026-05-21T08:18:04.980Z
```

Because the marker already exists, the remap never runs again, so the worktree rows are never reconciled and `merged_into_project` is never populated. The migration cannot, by design, catch any worktree created after its single run.

**3. Read-side recovery depends on a live worktree dir.** The parent↔worktree association is recomputed from the current working directory at read time (the `?projects=parent,worktree` composite). After the worktree directory is deleted, there is nothing from which to recompute the parent, so the orphaned rows can't be reattached through normal reads either.

## Impact

- Worktree-based workflows silently fragment project history; the parent project under-reports its own work.
- After routine worktree cleanup (delete the worktree post-merge), those observations are **unreachable** — not just mis-keyed. For users who do most work in worktrees this is effective data loss.
- A future write-side fix (#2711) does not retroactively help: existing installs have already burned the one-shot migration, and already-orphaned rows have no `merged_into_project` pointer and no live worktree to recompute from.

## Suggested directions

- Make reconciliation **idempotent / repeatable** instead of a one-shot marker — e.g. run a lightweight "fold `parent/*` worktree keys into `parent`" pass on worker start (or behind a versioned marker that bumps when the logic changes), so installs that already passed v1 still benefit.
- On **worktree removal**, promote that worktree's observations to the parent (the "promote after merge" idea from the now-closed #2018 / #1462) before the association is lost.
- At minimum, when a read resolves a parent, also surface rows under `parent/*` (prefix match) and/or honor `merged_into_project`, so removal of the worktree dir doesn't strand data.
- A documented, supported manual reconciliation command (`UPDATE observations SET project=:parent WHERE project LIKE :parent || '/%'`) would at least give users a non-DBA escape hatch.

## Related

- #2711 — option to write worktree observations to the parent project (closed → consolidated into #2785). Complementary; this issue is about reconciliation/recovery, not write destination.
- #2785 — plan-12 roadmap (tracks the write-side feature).
- #2018 / #1462 — promote worktree observations to parent after PR merge (closed). Directly relevant to the "unreachable after removal" half.
- #2663 / #1939 — project key derived from `basename(cwd)`, fragmenting worktrees/monorepos (closed).
- #2842 — sibling cwd-resolution defect for **submodules** (open). Same family (cwd → wrong project key), different code path.

Happy to provide a sanitized DB dump for the affected project or test a patch.

rodboev:
## Summary

The `.cwd-remap-applied-v1` marker guards `runOneTimeCwdRemap()` so it runs at most once per
install. Any worktree created after that first run accumulates observations under a
`parent/leaf` compound project key with `merged_into_project = NULL`. Once the worktree
directory is deleted, `WorktreeAdoption` cannot resolve it back to a parent repo, and those
observations become permanently unreachable via the `(o.project = ? OR o.merged_into_project = ?)`
read-side query. This PR removes the one-shot guard, making the reconciliation idempotent so
it runs on every startup and repairs any previously-missed compound-key rows.

## Why

`ProcessManager.ts:301–303` early-returns if `existsSync(markerPath)` is true. The marker is
written at line 401 on first success. Every subsequent startup exits before `executeCwdRemap()`
is called, so observations from worktrees created after the first run never have their
`merged_into_project` backfilled. The reconciliation body in `executeCwdRemap()` (lines 346–395)
is already idempotent: the session-filter loop at line 372–376 skips rows where
`old_project === c.project`, so re-runs produce zero DB writes for already-correct rows. The
only non-idempotent side effect (DB backup at lines 342–344) is moved inside the
`targets.length > 0` branch.

## Scope

This PR does not:
- Change `WorktreeAdoption.ts`; the live-worktree merge path is correct and complementary.
- Alter the `merged_into_project` column schema or the `ObservationCompiler` read query.
- Change any other one-shot migration (`runOneTimeChromaMigration`); those are destructive
  wipes that must remain one-shot.
- Delete existing `.cwd-remap-applied-v1` marker files from user data directories.

## Verification

- [ ] `bun test tests/infrastructure/process-manager.test.ts` - ran locally; 53 passed and 5 skipped, with 1 pre-existing Windows baseline failure in `resolveWorkerRuntimePath`.
- [x] `npm run build` - passed locally.
- [ ] `npm run strip-comments:check` - failed against the existing repo-wide comment-stripping baseline in check mode (`Changed: 315`), unrelated to this cwd-remap branch.

The dead marker constant has been removed. Re-running the reconciliation does not accumulate backups when there is no work, because the backup is created only inside the `targets.length > 0` branch. The per-startup git classification cost is bounded by caching each distinct cwd in `byCwd`.
Closes #2864

### DIFF
--- a/src/services/infrastructure/ProcessManager.ts
+++ b/src/services/infrastructure/ProcessManager.ts
@@ -182,6 +182,87 @@ export function getPlatformTimeout(baseMs: number): number {
   return process.platform === 'win32' ? Math.round(baseMs * WINDOWS_MULTIPLIER) : baseMs;
 }
 
+export async function getChildProcesses(parentPid: number): Promise<number[]> {
+  if (process.platform !== 'win32') {
+    return [];
+  }
+
+  if (!Number.isInteger(parentPid) || parentPid <= 0) {
+    logger.warn('SYSTEM', 'Invalid parent PID for child process enumeration', { parentPid });
+    return [];
+  }
+
+  try {
+    const cmd = `powershell -NoProfile -NonInteractive -Command "Get-CimInstance Win32_Process -Filter 'ParentProcessId=${parentPid}' | Select-Object -ExpandProperty ProcessId"`;
+    const { stdout } = await execAsync(cmd, { timeout: HOOK_TIMEOUTS.POWERSHELL_COMMAND, windowsHide: true });
+    return stdout
+      .split('\n')
+      .map(line => line.trim())
+      .filter(line => line.length > 0 && /^\d+$/.test(line))
+      .map(line => parseInt(line, 10))
+      .filter(pid => pid > 0);
+  } catch (error: unknown) {
+    if (error instanceof Error) {
+      logger.error('SYSTEM', 'Failed to enumerate child processes', { parentPid }, error);
+    } else {
+      logger.error('SYSTEM', 'Failed to enumerate child processes', { parentPid }, new Error(String(error)));
+    }
+    return [];
+  }
+}
+
+export function parseElapsedTime(etime: string): number {
+  if (!etime || etime.trim() === '') return -1;
+
+  const cleaned = etime.trim();
+  let totalMinutes = 0;
+
+  const dayMatch = cleaned.match(/^(\d+)-(\d+):(\d+):(\d+)$/);
+  if (dayMatch) {
+    totalMinutes = parseInt(dayMatch[1], 10) * 24 * 60 +
+                   parseInt(dayMatch[2], 10) * 60 +
+                   parseInt(dayMatch[3], 10);
+    return totalMinutes;
+  }
+
+  const hourMatch = cleaned.match(/^(\d+):(\d+):(\d+)$/);
+  if (hourMatch) {
+    totalMinutes = parseInt(hourMatch[1], 10) * 60 + parseInt(hourMatch[2], 10);
+    return totalMinutes;
+  }
+
+  const minMatch = cleaned.match(/^(\d+):(\d+)$/);
+  if (minMatch) {
+    return parseInt(minMatch[1], 10);
+  }
+
+  return -1;
+}
+
+const CHROMA_MIGRATION_MARKER_FILENAME = '.chroma-cleaned-v10.3';
+
+export function runOneTimeChromaMigration(dataDirectory?: string): void {
+  const effectiveDataDir = dataDirectory ?? DATA_DIR;
+  const markerPath = path.join(effectiveDataDir, CHROMA_MIGRATION_MARKER_FILENAME);
+  const chromaDir = path.join(effectiveDataDir, 'chroma');
+
+  if (existsSync(markerPath)) {
+    logger.debug('SYSTEM', 'Chroma migration marker exists, skipping wipe');
+    return;
+  }
+
+  logger.warn('SYSTEM', 'Running one-time chroma data wipe (upgrade from pre-v10.3)', { chromaDir });
+
+  if (existsSync(chromaDir)) {
+    rmSync(chromaDir, { recursive: true, force: true });
+    logger.info('SYSTEM', 'Chroma data directory removed', { chromaDir });
+  }
+
+  mkdirSync(effectiveDataDir, { recursive: true });
+  writeFileSync(markerPath, new Date().toISOString());
+  logger.info('SYSTEM', 'Chroma migration marker written', { markerPath });
+}
+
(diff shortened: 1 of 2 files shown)


rodboev:
Looks like the red CI here is unrelated to the cwd-remap change. Looks like a transient port collision in the test.

The failed `typecheck · build · test · bundle-size` job got through the suite with one failure:

```text
Worker API Endpoints Integration > Server State Management > should track MCP ready state dynamically

error: Failed to start server. Is port 49194 in use?

2246 pass
18 skip
1 fail
```

The PR-specific local checks in the body are still the ones covering the cwd-remap behavior.

rodboev:
Closing this one myself after rebasing it onto community-edge and checking what still survives there.

The repeatable cwd-remap behavior from this PR is already present on community-edge, so the core marker-guard fix here is no longer unique. After the rebase, what remains is mostly unrelated ProcessManager drift plus the worker endpoint ephemeral-port test cleanup.

The underlying deleted-worktree recovery gap from #2864 still looks real, but this PR no longer addresses that remaining gap cleanly. If I come back to it, I’ll open a fresh narrower PR against community-edge.

PR #2885, opened 2026-06-10
fix(hooks): cap summarize Stop-hook retry loop with per-session disabled sentinel (closes #2846)

### LINKED ISSUE #2846: Hook infinite retry loop on summarize event (Permission denied)
## Bug Description

When the claude-mem worker becomes unreachable during a session, the `summarize` hook enters an infinite retry loop that never stops. Each retry generates a new "Stop hook feedback" message that triggers another Claude response, which triggers another hook attempt, creating an endless cycle.

## Environment
- OS: Windows 11 Home
- Claude Code: v2.1.169
- claude-mem: v13.4.0
- Shell: Bash (via Git for Windows)

## Steps to Reproduce
1. Start a Claude Code session with claude-mem installed
2. The worker becomes unreachable (in this case due to `printf: write error: Permission denied`)
3. Any tool call triggers the hook, which fails
4. On session end/summarize, the hook retries indefinitely

## Observed Behavior

The hook script fails with:
```
claude-mem worker unreachable for N consecutive hooks.
```

Where N increments from 1 to 33+ without stopping. The `summarize` event hook keeps retrying on every Claude response, creating an infinite feedback loop that makes the session unusable.

## Expected Behavior

After N consecutive failures (e.g., 3-5), the hook should **stop retrying** for the remainder of the session, or at minimum not retry on the `summarize` event which creates a feedback loop.

## Impact
- Wastes API tokens (each retry triggers a Claude response)
- Makes the session difficult to close gracefully
- User must force-quit with Ctrl+C

## Suggestion
Add a max retry cap or exponential backoff. If the worker is unreachable after N attempts, disable hooks for the rest of the session and log a single warning instead of retrying indefinitely.

rodboev:
## Summary

When the worker becomes unreachable, `recordWorkerUnreachable` calls `emitBlockingError`
(exit 2) once the failure threshold is hit. Claude Code treats exit 2 as blocking feedback,
generates a new model turn, the session ends again, and the Stop hook fires — still failing —
in a tight infinite loop. This PR adds a `hooks-disabled-until.json` sentinel file written the
first time the threshold is crossed. `executeWithWorkerFallback` returns early on every
subsequent invocation during the same session, breaking the loop. The sentinel expires after
30 minutes so a restarted worker automatically re-enables hooks.

## Why

`hookCommand` (src/cli/hook-command.ts:79) calls into `summarizeHandler.execute` →
`executeWithWorkerFallback` (src/shared/worker-utils.ts:425) → `ensureWorkerAliveOnce` returns
`false` → `recordWorkerUnreachable` is called. Once `consecutiveFailures >= threshold`,
`recordWorkerUnreachable` calls `emitBlockingError` which exits 2. Claude Code surfaces that as
model-visible feedback, generating a response that ends the session and re-triggers the Stop
hook — which hits the same path again because the worker is still unreachable. There is no
guard in `executeWithWorkerFallback` that checks whether hooks have already been disabled.

The fix adds `areHooksDisabledForSession()` checked at the top of `executeWithWorkerFallback`,
and `writeHooksDisabledSentinel()` called inside `recordWorkerUnreachable` when the threshold
is first crossed. `resetWorkerFailureCounter` clears the sentinel when the worker comes back.

## Scope

- Does not change the fail-loud threshold or the `CLAUDE_MEM_HOOK_FAIL_LOUD_THRESHOLD` setting.
- The blocking error message is still surfaced exactly once — at the threshold crossing.
- Does not affect startup retry logic in `ensureWorkerStarted` (src/services/worker-spawner.ts).

## Verification
- [x] `bun test tests/shared/worker-utils-version-recycle.test.ts` — 2 passing
- [x] `bun test tests/shared/worker-utils-timeout.test.ts` — 3 passing
- [x] `npm run build` — bundle within guardrail

Closes #2846

### DIFF
--- a/src/shared/worker-utils.ts
+++ b/src/shared/worker-utils.ts
@@ -1,5 +1,6 @@
 import path from "path";
-import { readFileSync, existsSync, writeFileSync, renameSync, mkdirSync } from "fs";
+import { readFileSync, existsSync, writeFileSync, renameSync, mkdirSync, unlinkSync } from "fs";
+import { execSync } from "child_process";
 import { spawnHidden } from "./spawn.js";
 import { logger } from "../utils/logger.js";
 import { HOOK_TIMEOUTS, getTimeout } from "./hook-constants.js";
@@ -503,6 +504,10 @@ function getHookFailuresPath(): string {
   return path.join(getStateDir(), 'hook-failures.json');
 }
 
+function getHooksDisabledPath(): string {
+  return path.join(getStateDir(), 'hooks-disabled-until.json');
+}
+
 function readHookFailureState(): HookFailureState {
   try {
     const raw = readFileSync(getHookFailuresPath(), 'utf-8');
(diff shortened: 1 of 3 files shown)


thedotmack:
Closing as superseded by #3168 and #2892, which address the Stop-hook loop and worker-outage contract at the hook boundary.

PR #2886, opened 2026-06-10
fix(context): drop sub-day clock from injected headers to stop busting prefix cache (closes #2872)

### LINKED ISSUE #2872: Volatile minute-granularity timestamps in injected context defeat Anthropic prompt caching in long sessions
## Summary

claude-mem's context injections embed a **live, minute-granularity clock**, which makes the injected blocks non-deterministic. Anthropic's prompt cache is prefix-based, so any re-render of the same conversation point (retry, resume, fork) produces different bytes → cached suffix is invalidated and rewritten at 1.25× input price.

## Where

- `scripts/worker-service.cjs` (v13.5.0): PreToolUse(Read) file-context template begins with `Current: ${o} ${a} ${c}` → e.g. `Current: 2026-06-10 12:02pm GMT+8`. Two Reads in the same session a few minutes apart get different injected headers.
- SessionStart recent-context block is also stamped to the minute (`# [project] recent context, 2026-06-10 10:41am GMT+8`).

## Evidence (correlational, from our transcripts)

Auditing 3 days of Claude Code transcripts on one project (deduped by `requestId`): 276 mid-session cache collapses, ~32.6M tokens rewritten. In the largest loop session, **34 of 40 large collapses (cache_read falls to the ~17k static prefix, cache_creation ≈ full context) immediately follow a prompt/injection event**. claude-mem is not the only injector in play, but its blocks are the only ones we found carrying a live minute clock.

## Ask

Make injected blocks deterministic where possible:
1. Drop the live clock from the per-Read file-context template (or reduce granularity to day).
2. Same for the SessionStart recent-context header.
3. Optionally pin the rolling observation list per session instead of re-deriving it per injection.

The timestamp adds little for the model but guarantees cache-hostile bytes in every injection. Happy to provide the audit script / raw numbers.

## Environment
claude-mem 13.5.0 · Claude Code (macOS, Darwin 25.5.0)

rodboev:
## Summary

Every PreToolUse(Read) file-context block and every SessionStart recent-context header embeds a live wall-clock timestamp at minute granularity. The injected text changes every 60 seconds, so Anthropic's prefix cache treats each turn as a new prompt prefix and can never reuse cached KV state. This PR reduces all injected clock references to date-only (YYYY-MM-DD), which changes at most once per 24 hours and preserves cache coherence across the entire session.

## Why

`formatFileTimeline()` in `src/cli/handlers/file-context.ts:109–119` calls `new Date()` and formats `hour + minute + timezone` into the string `Current: <date> <HH:MMa> <TZ>` that opens every file-context injection. The identical `formatHeaderDateTime()` helper is independently duplicated in `AgentFormatter.ts:12–22` and `HumanFormatter.ts:12–22`; both produce `# [project] recent context, <date> <HH:MMa> <TZ>` for the SessionStart header via `renderAgentHeader` / `renderHumanHeader`. All three call sites resolve to model-visible text that changes every minute.

## Scope

This PR targets only the three injected-clock call sites. It does not:
- Change the `<requested_at>` fields in `buildInitPrompt` / `buildContinuationPrompt`; these are already date-only ISO strings.
- Change per-observation timestamps within the timeline body, which are derived from stored epoch values (stable across turns).
- Alter session routing, DB schema, or hook invocation strings.

## Verification

- [x] `bun test tests/context/formatters/agent-formatter.test.ts` - 40 passed.
- [x] `npm run build` - passed after rebasing onto current `origin/main`; regenerated bundled context/worker artifacts.
- [ ] `bun test tests/sdk/prompts.test.ts` and `bun test tests/context-injection.test.ts` - not rerun.
- [ ] `npm run strip-comments:check` - failed against the existing repo-wide comment-stripping baseline in check mode (`Changed: 327`), unrelated to this timestamp branch.

Closes #2872

### DIFF
--- a/src/cli/handlers/file-context.ts
+++ b/src/cli/handlers/file-context.ts
@@ -106,17 +106,7 @@ function formatFileTimeline(
     return aEpoch - bEpoch;
   });
 
-  const now = new Date();
-  const currentDate = now.toLocaleDateString('en-CA'); 
-  const currentTime = now.toLocaleTimeString('en-US', {
-    hour: 'numeric',
-    minute: '2-digit',
-    hour12: true
-  }).toLowerCase().replace(' ', '');
-  const currentTimezone = now.toLocaleTimeString('en-US', { timeZoneName: 'short' }).split(' ').pop();
-
   const lines: string[] = [
-    `Current: ${currentDate} ${currentTime} ${currentTimezone}`,
     `This file has prior observations — supplementary context follows. The Read result below is the full requested section.`,
     `- **Need details on a past observation?** get_observations([IDs]) — ~300 tokens each.`,
     `- **Need a structural map first?** smart_outline("${safePath}") — line numbers only, cheaper than re-reading.`,
--- a/src/services/context/formatters/AgentFormatter.ts
+++ b/src/services/context/formatters/AgentFormatter.ts
@@ -9,21 +9,10 @@ import type {
 import { ModeManager } from '../../domain/ModeManager.js';
 import { formatObservationTokenDisplay } from '../TokenCalculator.js';
 
-function formatHeaderDateTime(): string {
-  const now = new Date();
-  const date = now.toLocaleDateString('en-CA'); 
-  const time = now.toLocaleTimeString('en-US', {
-    hour: 'numeric',
-    minute: '2-digit',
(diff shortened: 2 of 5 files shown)


thedotmack:

PR #2887, opened 2026-06-10
fix(build): bundle zod into worker-service.cjs to survive node_modules churn (closes #2831)

### LINKED ISSUE #2831: worker-service.cjs has external zod requires — breaks when node_modules missing
## Problem

When a new version of claude-mem lands in the plugin cache (`~/.claude/plugins/cache/thedotmack/claude-mem/<version>/`), if `bun install` / `npm install` is not run (or fails silently), the plugin crashes immediately:

```
Cannot find module 'zod/v3' from '.../scripts/worker-service.cjs'
```

This has happened on multiple version upgrades. The failure is silent — Claude Code marks the version as active (`.in_use`) and spawns processes against it, which all fail.

## Root cause

`worker-service.cjs` has external `require()` calls for zod:

```js
require("zod")
require("zod/v3")
require("zod/v4")
require("zod/v4-mini")
```

These resolve against `node_modules`, which may not exist in the cache directory.

Meanwhile, `mcp-server.cjs` **inlines** zod — the zod source code is bundled directly into the file. So `mcp-server.cjs` works fine without `node_modules`, but `worker-service.cjs` does not.

This inconsistency means the plugin is half-bundled: some files are self-contained, others depend on an install step that has no guaranteed lifecycle in the Claude Code plugin cache model.

## Suggested fix

Bundle zod into `worker-service.cjs` the same way it's already bundled into `mcp-server.cjs`. This would make the plugin fully self-contained with zero external `require()` calls, eliminating the dependency on a consumer-side install step.

This is likely a bundler configuration issue — `worker-service.cjs` probably marks zod as external while `mcp-server.cjs` does not.

## Workaround

Manually run `bun install --frozen-lockfile` in the affected version directory. But this has to be repeated on every version upgrade, which is the problem.

## Environment

- claude-mem version: 13.4.1 (also seen on prior versions)
- OS: macOS
- Runtime: bun 1.3.14
- zod: ^4.4.3 (in package.json)

rodboev:
## Summary

When plugin upgrades leave `plugin/node_modules/zod` briefly missing or mismatched, any `worker-service.cjs` build that still treats `zod` as external can crash hook clients with a missing-module error. This PR removes `zod` from the worker-service esbuild `external` list and adds a post-build guard that fails the build if a `require("zod...")` ever reappears in the worker output.

## Why

`scripts/build-hooks.js` still allowed the worker bundle to emit an external `require("zod")` path. The MCP build already protects against this class of regression with an output check, but the worker build did not. Removing `zod` from the worker external list forces it to inline into `worker-service.cjs`, and the new `workerZodRequireRegex` guard keeps that invariant locked in on future edits.

## Scope

This PR changes only the worker-service build entry in `scripts/build-hooks.js`. It does not change the server-beta, context-generator, or transcript-watcher bundling strategy, and it keeps the existing plugin/package dependency declarations intact for the other bundled consumers that still resolve `zod` at runtime.

## Verification

- [x] `npm run build` - passed locally; `worker-service.cjs` built at 2754.35 KB and the new guard confirmed no external `require("zod...")` remains.
- [x] `bun test tests/infrastructure/plugin-distribution.test.ts --test-name-pattern "Build Script Verification|Required Files"` - 8 passed.
- [x] `npm run strip-comments:check` - matched the current repo baseline in check mode (`Changed: 330`).

Closes #2831

### DIFF
--- a/scripts/build-hooks.js
+++ b/scripts/build-hooks.js
@@ -54,14 +54,6 @@ function stripHardcodedDirname(filePath) {
   }
 }
 
-/**
- * Rule A canonical-template manifest: maps each host-managed config file's
- * command string to the buildShellCommand() options that generate it. The
- * build asserts the hand-maintained files still match the generator output so
- * the defensive shell prelude can't drift between the three files (issues
- * #1215, #1533). See src/build/hook-shell-template.ts and CLAUDE.md →
- * "Spawn-Contract Resolution".
- */
 function shellTemplateManifest(buildShellCommand) {
   const ccTrailing = (...tail) => [
     'node', '"$_P/scripts/bun-runner.js"', '"$_P/scripts/worker-service.cjs"', ...tail,
@@ -111,8 +103,6 @@ function shellTemplateManifest(buildShellCommand) {
     'plugin/.mcp.json': {
       kind: 'mcp',
       command: buildShellCommand({
-        // The mcp Node launcher derives its spawn target from requireFile, so
-        // no trailingCommand is needed (it is ignored for this host).
         host: 'mcp', requireFile: 'mcp-server.cjs',
         notFoundMessage: 'claude-mem: mcp server not found',
         mcpExtraCandidates: ['$PWD/plugin', '$PWD'],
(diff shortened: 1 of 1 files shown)


thedotmack:

PR #2889, opened 2026-06-10
fix(launcher): exit 0 on signal-kill for hook event subcommands (addresses #2823)

### LINKED ISSUE #2823: Every hook reports "Failed with non-blocking status code: No stderr output" on Claude Code 2.1.168 (10.4.0); 13.4.0 unusable (missing deps) — both released paths broken
## Summary
On the latest Claude Code, **claude-mem 10.4.0 fires a hook error on essentially every event** — `Stop`, `PostToolUse`, `UserPromptSubmit` all show:

> ⎿ hook error: Failed with non-blocking status code: No stderr output

e.g. "Ran 4 stop hooks" with 2 errors every turn. The underlying work still completes (observations store, session-start context injects), so it's **cosmetic — but it's on every turn and erodes trust in the plugin.** Crucially, **users can't escape it by upgrading**, because 13.4.0 is broken-packaged (details below).

## Environment
- claude-mem: **10.4.0** (pinned; see "13.4.0 broken" below for why)
- Claude Code: **2.1.168**
- Node **v20.19.2**, Bun **1.3.9**, macOS 26.5.1 (arm64)

## Root cause — 10.4.0 hooks
Two distinct issues, both surfacing as the same opaque "non-blocking status code / no stderr":

1. **`worker-service.cjs start` is not idempotent.** It runs on `SessionStart`, `UserPromptSubmit`, `PostToolUse`, AND `Stop`. After `SessionStart` launches the daemon, every later `start` fails:
   `[ERROR] [SYSTEM] ✗ Worker failed to start  Failed to start server. Is port 37777 in use?`
   This goes to `~/.claude-mem/logs/…`, **not stderr**, so Claude Code shows "non-blocking status code, no stderr output" with no detail.

2. **The work hooks don't exit cleanly.** `summarize`, `session-complete`, `observation`, `session-init` make their HTTP call successfully (daemon logs `← 200 /api/sessions/...`) but the spawned `bun` client process then exits non-zero / is killed at the hook timeout (observed `code=143` SIGTERM / `137` SIGKILL, no output). So **every** hook — not just `start` — is flagged by Claude Code, even though the work succeeded.

Net: removing the redundant `start` from a hook lowers the error count by one, but the remaining work-hooks still error — i.e. there's no per-hook workaround.

## Why upgrading doesn't help — 13.4.0 is broken-packaged
`13.4.0` (current npm `latest`) ships an incomplete `node_modules`: declared deps are **missing** (`zod`, `shell-quote`, ~25 native `tree-sitter-*` grammars), and the worker also `require()`s **undeclared** modules (`better-auth`, `@better-auth/api-key`). There's no working install step, so the worker crashes on `SessionStart` with `Cannot find module 'zod/...'`. So 10.4.0 users are stuck: stay and get hook-error spam, or upgrade and the memory worker won't start at all.

## Repro
1. Install claude-mem 10.4.0 on Claude Code 2.1.x.
2. Run any session; observe a hook-error line on (almost) every turn.
3. `tail -f ~/.claude-mem/logs/claude-mem-*.log` → repeated "Is port 37777 in use?" plus successful `← 200 /api/sessions/...` (work completing) interleaved.

## Suggested fixes
- Make `start` idempotent: detect an already-running daemon / port-in-use and **exit 0** (or drop `start` from non-`SessionStart` events entirely — `SessionStart` already launches it).
- Ensure hook client processes **`process.exit(0)`** after their request completes, so an open daemon socket / event-loop handle doesn't leave them to be SIGKILLed at the hook timeout.
- For 13.x: declare AND bundle all runtime deps (`zod`, `shell-quote`, the `tree-sitter-*` grammars, `better-auth`, `@better-auth/api-key`), or restore a real install step, so the worker starts on a clean install.

rodboev:
## Summary

Hook client processes were exiting with code 143 (SIGTERM) or 137 (SIGKILL) even when the daemon successfully processed the HTTP request, causing Claude Code to flag hook event invocations as failed. `bun-runner.js` already clamped signal-exit codes to 0 for the `start` subcommand, but not for `hook` event subcommands. This PR extends the signal-exit guard only to the hook event path.

## Why

`plugin/scripts/bun-runner.js` previously limited the signal-kill guard to the `start` lifecycle command. When a hook event subprocess is killed by the hook timeout after making a successful HTTP call, some platforms report the exit as a numeric code (143 or 137) rather than a Node.js `signal` string. The new guard applies that clamp when `args[1] === 'hook'`, so lifecycle commands outside the hook event path keep their normal exit semantics.

## Scope

This PR changes one line in `plugin/scripts/bun-runner.js` and keeps the signal clamp scoped to hook event commands. Start-command idempotency is in a companion PR. Neither PR alone fully the upstream issue together they do.

## Verification

- [x] `bun test tests/cli/hook-io.test.ts tests/cli/hook-io-discipline.test.ts` - 14 passed.
- [x] `npm run build` - passed locally, including the pre-ES2020 parser-compat guard.
- [ ] `npm run strip-comments:check` - failed against the existing repo-wide comment-stripping baseline in check mode (`Changed: 315`), unrelated to this hook-exit branch.

## Upstream

Closes #2823.
Reported by @ogiberstein.

### DIFF
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -231,7 +231,8 @@ child.on('error', (err) => {
 });
 
 child.on('close', (code, signal) => {
-  if (signal || code > 128) {
+  const isHookEventCommand = args[1] === 'hook';
+  if ((signal || code > 128) && isHookEventCommand) {
     process.exit(0);
   }
   process.exit(code || 0);
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -231,7 +231,7 @@ child.on('error', (err) => {
 });
 
 child.on('close', (code, signal) => {
-  if ((signal || code > 128) && args.includes('start')) {
+  if (signal || code > 128) {
     process.exit(0);
   }
   process.exit(code || 0);


thedotmack:

PR #2890, opened 2026-06-10
fix(windows): suppress console window flash in findBun() spawnSync call (closes #2869)

### LINKED ISSUE #2869: [Windows] Hook execution flashes black console window on every tool use
## Environment
- OS: Windows 11
- Claude Code: VS Code Extension
- claude-mem: v13.5.0

## Problem
Every hook execution causes a black console window to briefly flash on screen.
The window appears and disappears instantly, but is very disruptive during
active use since hooks fire constantly throughout a session.

## Root Cause (Confirmed via Testing)
All hooks use `"shell": "bash"`. On Windows, spawning a bash process from a
GUI application (VS Code) causes the OS to momentarily display a console window.
This is standard Windows behavior — it requires `CREATE_NO_WINDOW` to suppress.

## Reproduction Steps
1. Install claude-mem on Windows 11 with VS Code extension
2. Open any project and ask Claude any question that involves file reads
3. Observe repeated black console window flashes throughout processing

## Impact by Hook (Confirmed)

| Hook | Matcher | Frequency | Impact |
|------|---------|-----------|--------|
| PostToolUse | `*` | Every single tool call | **Primary cause** — most disruptive |
| PreToolUse | `Read` | Every file read | Secondary cause |
| UserPromptSubmit | — | Once per prompt | Minor |
| SessionStart | — | Once per session | Minor |

**Disabling PostToolUse(`*`) alone caused a dramatic reduction in flashes.**
Disabling PreToolUse(Read) alone had minimal visible effect.

## Suggested Fix
On Windows, suppress console window creation when spawning bash processes.

Option A — Use `CREATE_NO_WINDOW` process flag in the Node.js hook runner:
```js
// In bun-runner.js / worker-service.cjs
spawn(cmd, { windowsHide: true })  // Node.js child_process option
```

Option B — Detect platform and use PowerShell with hidden window:

```js
{
  "shell": "powershell",
  "command": "Start-Process bash -ArgumentList '-c','<cmd>' -WindowStyle Hidden -Wait"
}
```

Option C — Use the args exec form (no shell spawn = no window):
```js
{
  "command": "node",
  "args": ["path/to/bun-runner.js", "path/to/worker-service.cjs", "hook", "claude-code", "observation"]
}
```

rodboev:
## Summary

Every hook invocation on Windows causes a black console window to flash because `findBun()` in `plugin/scripts/bun-runner.js` calls `spawnSync('where bun', { shell: true })` without `windowsHide: true`. Adding `windowsHide: true` suppresses the window. The main bun-runner spawn already has it; only `findBun()` was missing it.

## Why

`findBun()` uses `spawnSync('where bun', { shell: true })` on Windows, spawning `cmd.exe` without `windowsHide: true`. The equivalent TypeScript lookup `resolveBunRuntime()` in `src/shared/worker-utils.ts` already passes `windowsHide: true`.

## Scope

- Does not address the `shell: "bash"` flash from Claude Code's own bash.exe spawn.
- Does not change the main bun spawn at bun-runner.js line 146.

## Verification
- [x] `npm run build` — bundle within guardrail, parser-compat check passes

Closes #2869

### DIFF
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -67,6 +67,10 @@ function isPluginDisabledInClaudeSettings() {
     const settingsPath = join(configDir, 'settings.json');
     if (!existsSync(settingsPath)) return false;
     const settings = JSON.parse(readFileSync(settingsPath, 'utf-8'));
+    // No optional chaining (?.) here: this launcher must parse on the oldest
+    // Node that any host might invoke it with. Some Claude Code installs run
+    // hooks under a bundled pre-ES2020 Node whose ESM loader throws
+    // "SyntaxError: Unexpected token '.'" on `?.` (issue #2791).
     return Boolean(
       settings &&
       settings.enabledPlugins &&
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -25,7 +25,8 @@ function findBun() {
     ? spawnSync('where bun', {
         encoding: 'utf-8',
         stdio: ['pipe', 'pipe', 'pipe'],
-        shell: true
+        shell: true,
+        windowsHide: true
       })
     : spawnSync('which', ['bun'], {
         encoding: 'utf-8',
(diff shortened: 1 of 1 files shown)


thedotmack:

PR #2892, opened 2026-06-10
fix: never block Claude Code harness on worker outages

DNA:
## Problem

When the claude-mem background worker is unreachable, hooks exit with code 2 — Claude Code's "block this action" contract. After just **3 consecutive failures** (`FAIL_LOUD_DEFAULT_THRESHOLD`), the harness starts rejecting Stop and UserPromptSubmit events, then overrides after 9 blocks with:

> "A hook blocked the turn from ending 9 consecutive times — overriding and ending turn."

This erases the user's in-progress prompt. The only escape is removing the plugin entirely.

The failure counter never resets unless a hook succeeds, so a stale counter from a past outage blocks the very first prompt of a new session even after the worker recovers.

## Root cause

`recordWorkerUnreachable()` in `src/shared/worker-utils.ts` calls `emitBlockingError()` (→ `process.exit(2)`) at threshold. This is the wrong contract for an optional background-memory service. Claude Code's own hook docs recommend non-blocking JSON output (`{ continue: true, systemMessage: "..." }`) for advisory failures.

### `src/shared/worker-utils.ts`
- **`recordWorkerUnreachable`**: replace `emitBlockingError` (exit 2) with `emitDiagnostic` (stderr-only, no exit). Worker outages never block the harness.
- **`executeWithWorkerFallback`**: treat degraded 5xx/429 responses as unreachable instead of silently resetting the counter — they should surface the offline banner too.
- **`consumeWorkerOutageHint(sessionId?, bypassThrottle?)`**: new exported helper returning a one-line user-facing banner when `consecutiveFailures > 0`, throttled to once per session / 30 min.

### `src/cli/handlers/session-init.ts` + `src/cli/handlers/summarize.ts`
Wire `consumeWorkerOutageHint` + `withUserHint` into the fallback branches. Session-init bypasses throttle (always shows at session start if worker is down); summarize is throttled (won't spam on every Stop).

## Tests

- 7 new tests in `tests/cli/worker-outage.test.ts`: throttle, bypass, expiry, new-session, and no-exit-2 guarantee.
- Updated contract test in `tests/cli/hook-stream-discipline.test.ts` to assert `emitDiagnostic` routing (not `emitBlockingError`).

Test suite: was 6 failing → now 1 failing (pre-existing Chroma test, unrelated).

## Before / After

**Before** — prompt erased:
```
⏺ Stop hook error: claude-mem worker unreachable for 16 consecutive hooks.
⏺ A hook blocked the turn from ending 9 consecutive times — overriding and ending turn.
```

**After** — turn completes, one-time advisory shown:
```
claude-mem: background worker offline — memory features (search, observations, context) unavailable this session.
```

### DIFF
--- a/src/cli/handlers/session-init.ts
+++ b/src/cli/handlers/session-init.ts
@@ -3,7 +3,8 @@
 // console.* / process.exit. logger.* calls are DIAGNOSTIC; thrown errors are
 // caught by hookCommand and routed through emitBlockingError.
 import type { EventHandler, NormalizedHookInput, HookResult } from '../types.js';
-import { executeWithWorkerFallback, isWorkerFallback } from '../../shared/worker-utils.js';
+import { executeWithWorkerFallback, isWorkerFallback, consumeWorkerOutageHint } from '../../shared/worker-utils.js';
+import { withUserHint } from '../../shared/hook-io.js';
 import { getProjectContext } from '../../utils/project-name.js';
 import { logger } from '../../utils/logger.js';
 import { HOOK_EXIT_CODES } from '../../shared/hook-constants.js';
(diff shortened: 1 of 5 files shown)


thedotmack:

> step 2 (availability failures cannot block: systemMessage instead of exit 2)

thedotmack:
Second pass — routing holds at **#3605 (plan-17, Hook Wrapper Contract)**, step 2. Important update since this was opened: #4033 (`bf25c8972`) landed on main and fixes the specific failure mode in your report — `recordWorkerUnreachable()` now blocks only on the threshold transition via a `thresholdTripped` latch, so a sustained outage no longer exits 2 on every subsequent hook and can no longer rack up the nine blocks that erase the prompt. What is still unresolved is your broader thesis, that an optional background service should never exit 2 at all and should use `{ continue: true, systemMessage }` instead. Leaving this open for that design call, but it needs a rebase — the file has moved substantially.

PR #2895, opened 2026-06-11
fix(worker): detect real port occupation cross-platform and reclaim stale workers

emilzo:
## Problem

On Windows, `isPortInUse()` probed `/api/health` instead of testing the socket. A stale/zombie worker that still holds the port but no longer answers health checks therefore read as **"port free"**. Every hook then spawned a duplicate daemon, which collided on `bind` with `EADDRINUSE`, logged `Worker failed to start. Is port <port> in use?`, and exited — repeating on every subsequent hook in a tight loop until the zombie was killed manually.

Observed in the wild on port `37777`: the worker re-spawned and failed every ~30s for minutes, with no automatic recovery.

## Root cause

`isPortInUse` conflated two different questions:

- **Is the port occupied?** — a TCP-level fact
- **Is a healthy worker answering?** — an application-level fact

The POSIX branch answered the first correctly (`bind` + `EADDRINUSE`); the Windows branch answered the second (`fetch /api/health`). So a port held by an unhealthy/zombie process was invisible to every "don't spawn a duplicate" guard on Windows.

## Fix

1. **`HealthMonitor.isPortInUse`** — detect real TCP occupation on every platform via a bind attempt (the previous POSIX behaviour). Health stays a separate concern handled by `waitForHealth`. *This alone stops the loop*: the duplicate daemon now hits the `isPortInUse` guard and exits cleanly (`Port already in use, refusing to start duplicate`) instead of colliding on bind.
2. **`ProcessManager.reclaimStaleWorkerPort`** — automatic recovery: when the port is held but nothing healthy answers, find the owning PID, **confirm it belongs to claude-mem** (recorded PID file *or* a `worker-service`/`claude-mem` command-line marker — it will never kill an unrelated process), terminate it, wait for the port to free, and clear the PID file.
3. **`worker-spawner`** — use the reclaim in the "port in use but unhealthy" branch, then fall through to a fresh spawn; if reclaim fails, give up cleanly instead of re-spawning on every hook.

## Notes

- Cross-platform: PID discovery uses `Get-NetTCPConnection` on Windows and `lsof` on POSIX; termination uses `taskkill /F` / `SIGKILL`.
- The reclaim is deliberately conservative — it refuses to act unless ownership is positively confirmed, so it can never kill an unrelated process bound to the port.
- Build artifacts (`plugin/scripts/*.cjs`) are intentionally omitted; CI regenerates them via `npm run build`.

## Testing

- `tsc --noEmit`: clean for the changed files (pre-existing errors elsewhere are unrelated).
- `tests/infrastructure/health-monitor.test.ts` (covers `isPortInUse`): passes.
- Rebuilt the worker bundle from this source and verified on Windows that it starts identically to the original (no regression) and that the duplicate-spawn loop no longer occurs.

### DIFF
--- a/src/services/infrastructure/HealthMonitor.ts
+++ b/src/services/infrastructure/HealthMonitor.ts
@@ -21,28 +21,17 @@ async function httpRequestToWorker(
 }
 
 export async function isPortInUse(port: number): Promise<boolean> {
-  if (process.platform === 'win32') {
-    try {
-      const response = await fetch(`http://127.0.0.1:${port}/api/health`);
-      return response.ok;
-    } catch (error) {
-      if (error instanceof Error) {
-        logger.debug('SYSTEM', 'Windows health check failed (port not in use)', {}, error);
-      } else {
-        logger.debug('SYSTEM', 'Windows health check failed (port not in use)', { error: String(error) });
-      }
-      return false;
-    }
-  }
-
+  // Detect real TCP occupation on every platform by attempting to bind. This is
+  // deliberately NOT conflated with "a healthy worker is answering": a stale or
+  // zombie worker can hold the socket without responding to /api/health (health
+  // is a separate question, answered by waitForHealth). The previous Windows
+  // branch probed /api/health here, so a zombie holding the port read as "free",
+  // every caller spawned a duplicate that died on bind with EADDRINUSE, and the
+  // worker re-spawn looped forever until the zombie was killed by hand.
   return new Promise((resolve) => {
     const server = net.createServer();
     server.once('error', (err: NodeJS.ErrnoException) => {
-      if (err.code === 'EADDRINUSE') {
-        resolve(true);
-      } else {
-        resolve(false);
-      }
+      resolve(err.code === 'EADDRINUSE');
     });
     server.once('listening', () => {
       server.close(() => resolve(false));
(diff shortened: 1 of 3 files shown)


tomeyerman:
Part 1 is right; the reclaimStaleWorkerPort recovery has a Windows edge case it can't recover from — an ownerless listen socket

Confirming the core diagnosis from Windows 11 (bun runtime, port 37777): the old isPortInUse health-probe really does misread a held-but-unhealthy port as free, symptom matches exactly — duplicate daemon colliding on bind with EADDRINUSE, re-spawning every ~30s. Switching isPortInUse to a real TCP bind probe (part 1) is correct and on its own stops the tight loop.

The gap is in part 2 (reclaimStaleWorkerPort). Its model is find owning PID → confirm it's claude-mem → terminate → wait for the port to free — which presumes a live zombie you can kill. There's a reproducible Windows case where that doesn't hold: after an abrupt death the listen socket stays LISTENING while its owning PID is already gone (netstat shows the PID; tasklist/Win32_Process show "No tasks"). Walking that through reclaim:

1. PID discovery returns an OwningProcess — pointing at a corpse.
2. Ownership confirmation can't run the cmdline-marker path (no process to inspect).
3. taskkill /F <dead pid> is a no-op and frees nothing.
4. Wait-for-free never completes — Windows won't release a dead-owner LISTENING socket short of a reboot.

So reclaim correctly declines, the spawner gives up cleanly (better than the loop), but the user is wedged with no recovery and a misleading duplicate-guard message that reads as "healthy worker exists."

Suggestion: when the bind probe fails and Get-NetTCPConnection's OwningProcess doesn't resolve to a live process, emit a distinct actionable error ("port held by orphaned socket, PID N no longer exists — reboot or set CLAUDE_MEM_WORKER_PORT") instead of falling through. Keeps the conservative no-kill behavior; gives a true diagnosis. Offer to test the branch/canary against the repro.

thedotmack:
Closing in favor of the newer, more conservative occupied-port handling in #3219 and Windows recovery work in #3009. Automatically reclaiming an uncertain listener has undesirable process-safety effects.

PR #2901, opened 2026-06-11
Avoid no-op SDK respawns

NRCP-HarrisonBooth:
Idle/prose observer output is a no-op, not a poisoned SDK session. This keeps claimed batches confirmed without respawning unless the classifier returns poisoned. Tests: bun test tests/worker/poison-respawn.test.ts tests/worker/agents/response-processor.test.ts tests/sdk/output-classifier.test.ts; npm run typecheck; git diff --check. Full bun test was attempted but exceeded 5 minutes locally without a failure summary.


thedotmack:
Closing as superseded by current main’s no-op response handling, which ignores incompatible output without respawning the observer.

Issue #2900, opened 2026-06-11
[Windows] git subprocess calls in worker-service.cjs lack windowsHide -> console windows flash constantly


Orbita-Media:
## Problem

On Windows, while claude-mem is running, **empty console windows flash on the desktop constantly** – every few seconds when idle, and near-continuously when multiple Claude Code sessions are active. Each flash is a short-lived (~16 ms) `conhost.exe` window that pops up at varying screen positions and interrupts typing/workflow.

## Root cause

The worker (`scripts/worker-service.cjs`) makes frequent `git` calls: a periodic per-repo `git status` on every memory operation, plus `rev-parse`, `ls-files`, and `cat-file`. Several of these `child_process` spawns are **missing `windowsHide: true`**, so Windows allocates a visible console window for each invocation.

The flag is applied **inconsistently** across call sites. Some git spawns have it, others do not:

**Has `windowsHide` (good)**
- The main git runner: `spawnSync("git", args, { ..., windowsHide: true, shell: false })`
- `execSync("git rev-parse --show-toplevel", { ..., windowsHide: true })`

**Missing `windowsHide` (the bug)**
- `spawnSync("git", ["-C", repo, ...args], { encoding: "utf8", timeout: 5000 })` (the per-repo helper used for `git status`, called frequently)
- a second `spawnSync("git", ["-C", repo, ...args], { encoding: "utf8", timeout: <var> })`
- `execFileSync("git", ["rev-parse", "--show-toplevel"], { cwd, encoding: "utf-8", stdio: [...] })`
- `execSync("git ls-files", { cwd, encoding: "utf-8", maxBuffer: ... })`
- `execFileSync("git", ["cat-file", "-e", "${hash}^{commit}"], { cwd, stdio: "ignore", timeout: 5000 })`

Because the per-repo `git status` helper runs often (frequency scales with active sessions times observation processing), the flashing is effectively constant on a busy machine.

## Environment

- Windows 11
- claude-mem 13.5.6 (also observed on 13.4.x and 13.5.5)
- Claude Code (native binary), bun-based worker

## Reproduction

1. Windows, claude-mem running with one or more active Claude Code sessions
2. Watch the desktop: empty console windows flash every few seconds

## Suggested fix

[...]

PR #2902, opened 2026-06-11
docs: add Windows Terminal flash troubleshooting note

caioribeiroclw-pixel:
## Summary

Adds a Windows troubleshooting note for the case reported in #2900 where short-lived console flashes continue even after `windowsHide: true` patches are applied.

The key nuance from the live issue thread is that Windows Terminal hosting can surface `CASCADIA_HOSTING_WINDOW_CLASS` windows; `windowsHide` suppresses classic Console Host windows but does not hide Windows Terminal-hosted windows. The doc points users to switch the default terminal application to Windows Console Host, then keep the `windowsHide` fixes in place.

## Validation

- `git diff --check`
- Tried `npm run typecheck`, but this fresh shallow clone does not have dependencies installed (`tsc: not found`). The change is docs-only MDX prose.

## Context

This is intended as a low-risk support doc companion to #2900 rather than a replacement for the spawn-site code fixes.

### DIFF
--- a/docs/public/troubleshooting.mdx
+++ b/docs/public/troubleshooting.mdx
@@ -112,6 +112,28 @@ The skill includes comprehensive diagnostics, automated repair sequences, and de
    npm run worker:restart
    ```
 
+
+### Windows: console windows still flash even after `windowsHide` fixes
+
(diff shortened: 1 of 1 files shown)


caioribeiroclw-pixel:
Addressed the Greptile note in 14fa497 by changing the rendered registry path from `HKCU\\Console\\%%Startup` to the human-readable `HKCU\\Console\\%Startup`. Re-ran `git diff --check`; still docs-only.

thedotmack:

Issue #2785, opened 2026-06-05
[plan-12] Provider & Extensibility Roadmap — net-new capabilities, not defects


thedotmack:
# [plan-12] Provider & Extensibility Roadmap — net-new capabilities, not defects

## Defect

This is a roadmap master, not a defect cluster: it aggregates net-new capabilities across providers, ingestion/filtering, observability, auth, and UX so they ship as coherent slices rather than one-off PRs.

### Providers & auth
- #2522 — Vertex AI support for the Gemini provider (GCP service-account / ADC + Vertex endpoint)
- #2704 — auth-helper command for refreshable gateway tokens (apiKeyHelper equivalent)

### Ingestion & filtering
- #2690 — backfill / ingest existing Claude Code session JSONL files
- #2498 — incremental scan skill for changes made outside a Claude session
- #2463 — `tool_response`-level filter (extension / size / content heuristic) to prevent binary/Playwright blowups
- #2423 — per-directory disable support
- #2711 — option to write worktree observations to the parent project
- #2736 — skip / throttle subagent observations (Dynamic Workflow `workflow-subagent`) — re-raise of #2303

### Observability & logging
- #2566 — MCP grammar-introspection routes + worker provider retry telemetry + audit log
- #2513 — clarify logger audit policy and codify CI behavior
- #2702 — logging cleanup: type-safe Component union, dead-log removal, noise reduction

### UX / semantics / extensibility
- #2645 — i18n support for startup UI labels
- #2467 — PreToolUse:Read injection treated as a turn boundary in connected tool sequences (semantics call)
- #2418 — enable OpenHarness integration / allow PR from fork

## Fix sequence

Design doc: `plans/12-provider-and-extensibility-roadmap.md`. Ship per sub-area as independent slices; each slice carries its own tests. No single PR closes this master — it closes when its sub-areas land.

## Out of scope

All defect clusters (plan-01..11, plan-13).

NRKZOZW:
Re the **Ingestion & filtering → #2736 (skip / throttle subagent observations)** line:

A complete, self-contained implementation of this slice is already up in #2741 (tested + built) — offering it as the base for the plan-12 slice rather than a from-scratch reimplementation.

**What it does** (all filtered *before* any worker round-trip / provider call):
- `CLAUDE_MEM_SKIP_SUBAGENT_OBSERVATIONS` — global toggle, gated on `agentId` (provably safe for main-session).
[...]

PR #2905, opened 2026-06-11
fix(observer): harden non-Claude observation flow

ojungo69:
## Summary

This hardens the worker response path against XML-shaped provider drift and high-volume observation loops from non-Claude providers.

- Enforce the expected output kind from `lastGeneratorSource`: `init`/`ingest` must produce observations, while `summarize` must produce summaries.
- Treat wrong-root XML as an invalid output path instead of persisting it as the wrong record type.
- Recognize explicit no-op observation XML such as `<type>skip</type>` only when the worker response path opts in, so server-beta summary parsing still rejects it by default.
- Batch already-buffered same-agent observation messages for OpenRouter/Gemini so stateless providers do not make one expensive model call per tool event.
- Tighten observation prompts to prefer explicit no-op XML for routine status/log/queue/database/process checks instead of storing low-value observations.
- Regenerate the plugin worker/server bundles.

## Why

Non-Claude providers can return parseable XML with the wrong root or create low-value observations for every routine tool event. Before this change:

- a summary prompt returning `<observation>...</observation>` could store an observation instead of a summary;
- an observation prompt returning `<summary>...</summary>` could store a summary and confirm/drop the observation batch;
- `<observation><type>skip</type>...</observation>` could fall back to the first configured observation type and become a stored observation;
- OpenRouter/Gemini processed each queued observation as a separate full-history multi-turn request, which made queues drain slowly and encouraged over-observation during bursts of read-only diagnostic tools.

That makes the queue appear to move while increasing low-value observations and degrading summary quality. This complements #2857, which handles prose/idle empty acknowledgements; this PR covers XML-shaped contract violations and non-Claude observation throughput.

Related to #2817 and the provider resilience work in #2785.

## Test plan

- [x] `bun test tests/sdk/prompts.test.ts tests/services/worker/session-message-buffer.test.ts tests/shared/settings-defaults-manager.test.ts`
- [x] `bun test tests/sdk/parser.test.ts tests/sdk/output-classifier.test.ts tests/worker/agents/response-processor.test.ts tests/gemini_provider.test.ts tests/worker/provider-classifiers.test.ts tests/worker/provider-errors.test.ts`
- [x] `bun test tests/server/server.test.ts`
- [x] `bun test tests/sdk/parser.test.ts`
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `cubic review -j` — returned `issues: []`

Additional check:

- `bun test` full-suite run reached `2248 pass`, `18 skip`, `8 fail`. Re-running the relevant failures showed `tests/server/server.test.ts` and `tests/sdk/parser.test.ts` pass in isolation; the remaining `openclaw/src/index.test.ts` failures are timeout failures that reproduce when that file is run alone and are outside this diff.


ojungo69:
Follow-up pushed in `267d5eb5` to address the live `Untitled` / over-observation regression seen with non-Claude providers.

What changed:
- Added a shared observation content helper so context/search/viewer/OpenClaw/CLAUDE.md generation derive display titles from `title -> narrative -> subtitle -> text -> facts -> concepts` instead of falling back to `Untitled`.
- Added durable-observation filtering for context injection and `/api/observations`, filtering legacy no-op rows like `No durable observation to record.` before rendering.
- Added pre-dispatch and worker-ingest filtering for low-signal observations so routine read-only commands, meta tools, and all-low-signal `multi_tool_use.parallel` calls do not enqueue without requiring users to expand `CLAUDE_MEM_SKIP_TOOLS`.
- Preserved failed read-only commands, validation/build/test commands, mutating commands, mutating local worker API calls, and per-child failures inside `multi_tool_use.parallel`.
- Updated no-op parsing so worker and server-beta generation accept explicit no-op observation XML as successful zero-observation output, including `type=skip` with a no-op title.
- Fixed observation content hashing so subtitle/facts/concepts/files-only observations no longer dedupe-collide, while preserving the legacy hash for title+narrative-only observations.
- Updated `sync-marketplace.cjs` to sync the Codex cache path as well as the Claude marketplace/cache paths; this was required because Codex hooks were still running stale bundled code.

Local verification:
- `npm run typecheck`
- `bun test tests/sqlite/data-integrity.test.ts tests/shared/observation-content.test.ts tests/shared/observation-filter.test.ts tests/sdk/parser.test.ts tests/context/formatters/agent-formatter.test.ts tests/server/generation/process-generated-response.test.ts` (Postgres integration test skipped without `CLAUDE_MEM_TEST_POSTGRES_URL`)
- `npm run build`
- `git diff --check`
- `cubic review -j` -> `{"issues":[]}`
- `npm run build-and-sync` -> worker restart verified at version `13.5.6`
- Verified installed Claude/Codex bundles contain no `Untitled` fallback matches.
- Verified `/api/context/inject?projects=claude-mem` has no `Untitled` or no-op observation text matches.
- Verified read-only `/api/observations` and context curl probes are logged as hooks but no longer enqueue after the latest local sync.

Also re-scanned the staged diff for obvious secret patterns (`sk-*`, `Bearer ...`, private key headers); no matches.

ojungo69:
Follow-up pushed in `b8c8c311` after checking whether the low-signal filter could affect Claude/Codex observation behavior.

Additional fix:
- Refined string-only tool-response failure detection so successful read-only searches/log inspections containing words like `error` or `[ERROR]` are still skipped as low-signal.
- Preserved failed read-only command observations for string-only outputs such as `Process exited with code 22`, `curl: (22) ... error: 404`, `curl: (7) Failed to connect`, `jq: parse error`, `regex parse error`, and `Failed to open file`.
- Added regression coverage for both sides of that boundary.

Claude-provider impact check:
- Provider selection code was not changed.
- ClaudeProvider still routes SDK text through `processAgentResponse`; the no-op observation parsing remains additive and only applies to observer-response handling.
- Claude-specific resume/error/provider-classifier tests pass.

Verification:
- `bun test tests/shared/observation-filter.test.ts` -> 13 pass, 0 fail
- `npm run typecheck`
- `npm run build`
- `bun test tests/sqlite/data-integrity.test.ts tests/shared/observation-content.test.ts tests/shared/observation-filter.test.ts tests/sdk/parser.test.ts tests/context/formatters/agent-formatter.test.ts tests/server/generation/process-generated-response.test.ts tests/claude-provider-resume.test.ts tests/claude-provider-error-classifier.test.ts tests/worker/provider-classifiers.test.ts tests/worker/agents/response-processor.test.ts` -> 162 pass, 5 skip, 0 fail
- `git diff --check`
- `cubic review -j` -> `{"issues":[]}`
- `npm run build-and-sync` -> local Claude/Codex caches synced and worker restart verified at version `13.5.6`

ojungo69:
Final follow-up pushed in `d39f423f`.

This tightens the string-response failure boundary after live verification showed that reading logs containing example text like `Process exited with code 22` could still be misclassified as a failed read-only command. The filter now treats those phrases as failures only when they appear as line-start tool/shell failure output, while still preserving real failures such as `bash: rg: command not found`, `Command failed with exit code 1`, `curl: (22) ...`, `jq: parse error`, and `Failed to open file`.

Additional verification after the final commit:
- `bun test tests/shared/observation-filter.test.ts` -> 13 pass, 0 fail
- `npm run typecheck`
- `npm run build`
- `cubic review -j` -> `{"issues":[]}`
- `bun test tests/sqlite/data-integrity.test.ts tests/shared/observation-content.test.ts tests/shared/observation-filter.test.ts tests/sdk/parser.test.ts tests/context/formatters/agent-formatter.test.ts tests/server/generation/process-generated-response.test.ts tests/claude-provider-resume.test.ts tests/claude-provider-error-classifier.test.ts tests/worker/provider-classifiers.test.ts tests/worker/agents/response-processor.test.ts` -> 162 pass, 5 skip, 0 fail
- `npm run build-and-sync` -> local Claude/Codex caches synced and worker restart verified at version `13.5.6`
- Live log check: the read-only log-inspection command that previously enqueued now logs only `PostToolUse` and no `QUEUE ENQUEUED`.

ojungo69:
Follow-up pushed in `95d495c5` for the recurring `Untitled` / over-observation report.

Additional fixes:
- `/api/observations` now derives paginated titles from durable content instead of returning `null` titles that downstream UIs can render as `Untitled`.
- Routine verification no-op narratives such as `All routine verification commands ... No debugging findings or root cause analysis to record.` are filtered at the shared content layer and SQL pagination layer.
- Low-signal command filtering now handles `|| true` tails and ignores `>` inside quoted SQL/strings, so read-only probes like `sqlite3 ... "WHERE created_at_epoch > ..."` and `curl ... | rg ... || true` do not enqueue.

Verification:
- `git diff --check`
- `npm run build`
- `npm run typecheck`
- `bun test --isolate tests/shared/observation-content.test.ts tests/shared/observation-filter.test.ts tests/session_store.test.ts tests/context/formatters/agent-formatter.test.ts tests/context/observation-compiler.test.ts tests/worker/search/result-formatter.test.ts tests/worker/SearchManager.timeline-anchor.test.ts tests/worker/agents/response-processor.test.ts tests/server/generation/process-generated-response.test.ts tests/sdk/parser.test.ts tests/sqlite/data-integrity.test.ts` -> 170 pass, 5 skip, 0 fail
- `cubic review -j` -> `{ "issues": [] }`
- secret-pattern scan over `src`, `tests`, `plugin/scripts`, `dist`, plugin manifests, and `package.json` -> no matches
- `npm run build-and-sync` -> worker restart verified at version `13.5.6`
- Live checks: exact `title == null` / `title == "Untitled"` / routine no-op matches returned `[]`; context injection had no exact `Untitled`/no-op rows; post-sync read-only API probes show hook receipt without `QUEUE ENQUEUED`.

Note: a non-isolated broad Bun run can expose an existing cross-file `ModeManager` mock interference in `tests/sdk/parser.test.ts`; the same suite passes with `--isolate`, and `parser.test.ts` also passes on its own.

ojungo69:
Follow-up pushed in `d336e13c` for the platform_source confusion found during local verification.

What changed:
- Made `GET /api/sessions/status` read-only for unknown `contentSessionId` values. It now looks up an existing session id instead of calling `createSDKSession(..., "", "")`, so a status poll can no longer pre-create a default `platform_source=claude` row.
- Added a regression test proving status-before-Codex-init does not create a row and the later Codex session remains `platform_source=codex`.
- Addressed cubic review feedback in `ObservationBatchSizeResolver` so numeric JSON settings like `CLAUDE_MEM_OBSERVATION_BATCH_SIZE: 7` continue to work, while preserving logger coverage.
- Rebuilt synced plugin bundles.

Verification:
- RED confirmed first: new status regression failed with `Expected: 0, Received: 1` before the fix.
- `bun test tests/worker/http/routes/session-routes-status.test.ts tests/worker/observation-batch-size.test.ts tests/logger-usage-standards.test.ts tests/sqlite/sessions.test.ts tests/session_store.test.ts` -> 35 pass, 0 fail
- `npm run typecheck`
- `npm run build`
- `git diff --check`
- `cubic review -j` on source/test diff -> `{ "issues": [] }`
- `npm run build-and-sync` -> marketplace/cache sync and worker restart verified at version `13.5.6`
- Live worker probe: `/api/sessions/status?contentSessionId=codex-status-readonly-...` returned `not_found`; DB check for that content_session_id returned `0` rows.

Note: a broad `bun test --isolate` run still hits pre-existing timeout failures in OpenClaw/CORS tests unrelated to this diff; the targeted regression and affected suites pass.

ojungo69:
Follow-up pushed in `775781c0` for the recurring Codex summaries showing up as Claude.

Root cause:
- The worker opens one shared Bun SQLite `Database` in `DatabaseManager` and passes it into `SessionStore`.
- `SessionStore` only enabled `PRAGMA foreign_keys = ON` when it opened the DB itself. Existing migrated runtime DBs using the shared connection therefore had foreign-key enforcement off.
- When non-Claude providers rotated `memory_session_id`, `sdk_sessions.memory_session_id` was updated but existing `observations` / `session_summaries` did not receive the `ON UPDATE CASCADE`. Those rows became orphaned, and read paths using a fallback `COALESCE(..., 'claude')` could render them as Claude.

Fix:
- Enable `PRAGMA foreign_keys = ON` for shared `Database` connections too.
- Added a regression test that reproduces the migrated/shared-connection path and proves `session_summaries.memory_session_id` cascades after a provider id update.
- Rebuilt synced plugin bundles.

Verification:
- RED confirmed before fix: the new shared-connection cascade test kept the old `memory_session_id`.
- `bun test tests/fk-constraint-fix.test.ts tests/services/sqlite/search-platform-source-scoping.test.ts tests/session_id_usage_validation.test.ts tests/worker/http/routes/session-routes-status.test.ts` -> 17 pass, 0 fail
- `npm run typecheck`
- `npm run build`
- `npm run build-and-sync` -> local worker restart verified at version `13.5.6`
- `git diff --check`
- `cubic review -j` -> `{ "issues": [] }`
- Local DB repair was limited to provider-generated orphan ids that still embedded a matching `content_session_id`; after repair, provider-format orphan candidates are `0`, and current Codex summaries join as `platform_source=codex`.

ojungo69:
Closing this PR as superseded by later changes on `main` and the canonical tracking issue #2960.

The core poison-loop behavior this PR was originally trying to mitigate — benign empty/prose no-op observer outputs accumulating invalid-output debt and triggering repeated respawns — appears to have been addressed on `main` by treating non-XML idle/prose outputs as claimed no-op batches instead of respawn debt.

However, this PR also contained additional hardening that does not appear to be fully present on `main` yet:

- rejecting wrong-root XML based on `lastGeneratorSource`
- treating explicit no-op observation XML such as `<type>skip</type>` as a valid no-op only in observation paths
- batching same-agent observation messages for OpenAI-compatible/stateless providers

Rather than trying to rebase this large stale branch, I plan to split the remaining useful pieces into smaller, current-`main` PRs.

PR #2908, opened 2026-06-12
feat: add Kimi Code CLI support

T3mmyKay:
- Add Kimi platform adapter and register it for kimi/kimi-code hooks
- Add TOML-based hook installer for ~/.kimi-code/config.toml
- Wire worker-service dispatch for kimi-code subcommands and hook events
- Add kimi-code to npx CLI IDE detection/install path
- Add Kimi plugin manifest (.kimi-plugin) and onboarding skill
- Include .kimi-plugin in published package files
- Add adapter and installer tests
- Fix schema-repair tests to use sqlite3 CLI instead of Python's sqlite3 (Python 3.9.6 + SQLite 3.51.0 disallows modifying sqlite_master)


silvioprog:
@T3mmyKay really great work here!
Just wanted to let you know that this PR has some merge conflicts.

usbrandon:
Thanks @T3mmyKay for getting Kimi Code support started — I tested this against a live Kimi Code CLI install and found a few payload mismatches that would prevent it from working end-to-end (Kimi sends `tool_output` rather than `tool_response`, its `Stop` payload has no `transcript_path`, and it appends raw hook stdout to the model context so the JSON envelope gets injected verbatim).

This branch also no longer merges cleanly onto `main` (59 conflicts, mostly built artifacts), so I ported your source changes onto current main and applied the fixes on top: #3547. You're credited via co-author trailer on the port commit.

Feel free to pull any of those corrections back into this PR if you'd rather land it here — happy either way.

thedotmack:
Closing in the PR consolidation pass — superseded: ported with fixes to #3547 by usbrandon (credited); 59 conflicts here. The analysis here fed the plan-23 design (#3611); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

Issue #2379, opened 2026-05-08
[plan-04] Installer Failure Transparency — cross-IDE error taxonomy + 12×4 test matrix


thedotmack:
## Goal

Stop the universal installer (`npx claude-mem install`) from silently swallowing real failures and falsely reporting "installed successfully" on all 12 IDEs. Convert every error-suppression site to a single `installerError(severity, ctx)` decision point driven by an explicit taxonomy. Make `tree-sitter` ERESOLVE conflicts, missing `uv`, broken Python/ONNX deps, and per-IDE setup failures fail loudly with platform-specific remediation.

## Net effect

- "Installation Complete" only prints when every ABORT-level dependency was satisfied.
- `runNpmInstallInMarketplace` runs strict first; `--legacy-peer-deps` only on confirmed `ERESOLVE` token, with the fallback announced loudly.
- Postinstall regression guard: any new transitive dep with `scripts.postinstall` outside an explicit allowlist fails the build (prevents recurrence of v12.6.1 `tree-sitter-swift` hang).
- Cross-IDE test matrix: 12 IDEs × 4 scenarios (happy / ERESOLVE / missing uv / missing bun) = 48 cells, each asserting exit code, summary text, and remediation presence.
- ONNX runtime / chroma-mcp dependency pins refreshed (issue #2371).
- Per-IDE failures classified WARN_CONTINUE vs ABORT instead of bucketed into a single yellow tip.

## Plan

Full phased plan in branch `sequoia-territory`: `plans/04-installer-transparency.md`.

## Consolidation

This issue consolidates the cluster of installer-time failures, missing-dependency surfacing, post-upgrade-state issues, and per-IDE setup bugs. Full child-issue list will be posted as a comment.

thedotmack:
## Consolidated child issues

Each entry: one sentence on what it is, one sentence on why it's the same root cause.

- **#2371 — ONNX runtime fails to load embedding model on macOS arm64 / Python 3.13 (chroma-mcp 0.2.6 dep).** The chroma-mcp transitive dep ships an ONNX runtime too old to parse the bundled PyTorch-2.0 model IR; the plan's dependency-pin phase + ABORT-with-remediation surfacing covers this.
- **#2331 — Installing through claude market requires `bun install` upon every claude-mem upgrade.** Post-upgrade install-state marker isn't being refreshed across versions; the plan's install-marker cleanup phase fixes the recurring bun-install requirement.
[...]

PR #2918, opened 2026-06-13
fix(setup): restore tree-sitter CLI on clean Windows installs (closes #2910)

### LINKED ISSUE #2910: v13.5.6: smart_search/smart_outline/smart_unfold return 0 symbols on clean Windows install - tree-sitter-cli binary never downloaded (--ignore-scripts), #1247's extensionless existsSync check still unfixed
## Before submitting

Closest prior issues, all closed, none covering this mechanism: #1247, #2617 (consolidated into #2379), #2520, #2410. See "Relationship to prior issues" below.

### Bug Description

On a clean Windows install of claude-mem 13.5.6, all three smart-explore tools silently fail:

- `smart_search` → `Scanned 444 files, found 0 symbols` (on a Next.js/TypeScript repo)
- `smart_outline` / `smart_unfold` → `Could not parse <file>. File may use an unsupported language or be empty.` (on an ordinary 7-line `.ts` file)

No error is surfaced anywhere — not in the tool result, not in `~/.claude-mem/logs/`.

**Root cause: the tree-sitter CLI binary does not exist on the user's machine, and no supported install path can ever produce it.**

The chain:

1. `scripts/mcp-server.cjs` parses by shelling out to the tree-sitter CLI — `execFileSync(bin, ["query", "-p", grammarDir, queryFile, ...files])` — resolving `bin` to `node_modules/tree-sitter-cli/tree-sitter`.
2. The `tree-sitter-cli` npm package ships **no binary**. It contains only `cli.js`, `dsl.d.ts`, `install.js`, and `package.json`. The binary is downloaded from GitHub releases by `install.js`, which is its `install` (postinstall) script.
3. The plugin's bundled `node_modules` (shipped since the missing-node_modules fixes around #2520/#2456) was packed without running that script, and the universal installer's marketplace step runs `npm install` with `--ignore-scripts` (visible in the published `dist/npx-cli/index.js`: `["install", "--omit=dev", "--ignore-scripts"]`). So the download never happens on any user machine, on any platform.
4. Binary resolution falls through to a bare `tree-sitter` PATH lookup, which fails on machines without a globally installed tree-sitter CLI. `execFileSync` throws, and the `catch { return new Map() }` in `runBatchQuery` swallows it — producing the "0 symbols" silent failure already called out in #1247.
5. Independently: the resolver checks `existsSync(join(dir, "tree-sitter"))` — extensionless — so even when `tree-sitter.exe` is present on Windows, the resolved path is discarded. This is #1247's "Root Cause 2", still present in the 13.5.6 bundle.

This appears to be a **regression introduced by the installer hardening**: in the v10.x era the binary *was* present on user machines (#1247's reporter explicitly lists `tree-sitter-cli/tree-sitter.exe` as bundled), meaning install scripts ran back then. The `--ignore-scripts` policy (plan-04 / #2379, added after the `tree-sitter-swift` postinstall hang) removed the only mechanism that fetched the binary, and the smart tools' silent-catch design hid the breakage.

### Steps to Reproduce

1. Windows 11, Node 22.x, no global `tree-sitter` on PATH
2. Install/update claude-mem to 13.5.6 via the Claude Code plugin marketplace (`claude plugin update claude-mem@thedotmack`)
3. Observe `node_modules/tree-sitter-cli/` in the new version dir contains no `tree-sitter.exe` (only `cli.js`, `install.js`, `dsl.d.ts`, `package.json`)
4. Call `smart_search(query="anything", path="<any TS repo>")` → 0 symbols; `smart_outline` on any `.ts` file → "Could not parse"

### Expected Behavior

Smart tools extract symbols; or, failing that, the tool result says *why* (binary missing / spawn failed) instead of reporting 0 symbols as if the scan succeeded.

### Environment

- **Claude-mem version**: 13.5.6 (also reproduced on 13.4.0)
- **Claude Code version**: 2.1.175
- **OS**: Windows 11 Home 10.0.26200
- **Platform**: win32-x64, Node v22.17.1

### Logs

Nothing relevant is logged: the spawn failure is caught and logged at `debug` level only (`tree-sitter query failed for N file(s)`), below the default `INFO` threshold, and the MCP tool returns a normal-looking empty result.

### Verified workaround (two steps, no compiler needed)

```bash
cd ~/.claude/plugins/cache/thedotmack/claude-mem/13.5.6/node_modules/tree-sitter-cli
node install.js   # downloads official tree-sitter.exe v0.26.9 from GitHub releases
cp tree-sitter.exe tree-sitter   # satisfy the extensionless existsSync check
```

After this, against the same repo: `smart_search` → `Scanned 81 files, found 369 symbols`, correct ranked matches with signatures; `smart_outline` produces correct outlines. (Verified by driving `scripts/mcp-server.cjs` directly over stdio.)

The workaround is wiped by every plugin update, since each update creates a fresh version directory.

### Suggested fixes

1. **Ship or fetch the binary**: either bundle per-platform tree-sitter CLI binaries, or add `tree-sitter-cli` to the existing postinstall allowlist so its `install.js` runs (the allowlist mechanism from plan-04 already exists for exactly this kind of vetted exception).
2. **Fix the Windows resolver**: check `tree-sitter.exe` (or use `process.platform === "win32"`) in `getTreeSitterBin()` — #1247's Root Cause 2.
3. **Fail loud**: when the CLI can't be spawned, say so in the tool result instead of returning 0 symbols — same transparency principle as plan-04, applied to the MCP server.

### Relationship to prior issues

- #1247 (closed): same symptom, same silent catch, same extensionless-check bug — closed without a code fix ("reopen if still relevant"); the fix that helped users was a community workaround in the comments. In that era the binary was present, so only the resolver/compiler issues bit.
- #2617 (closed): "fix Windows tree-sitter CLI resolution" — consolidated into #2379; the extensionless check is still in the 13.5.6 bundle.
- #2520 (closed): same silent no-op on Linux from missing `node_modules` — fixed by bundling `node_modules`, but bundling cannot include this binary because it's fetched by an install script that never runs.
- #2410 (closed): suggested WASM (`web-tree-sitter`) — would eliminate this entire class of native-binary/platform issues.

rodboev:
## Summary
Clean Windows installs were suppressing the trusted `tree-sitter-cli` bootstrap, then the smart-file-read parser ignored the resulting `tree-sitter.exe` even when it existed locally. This patch restores Bun's trusted install-script path in the installer, resolves the package-local Windows binary before PATH fallback, and adds focused regression coverage for both behaviors.

## Why
`installPluginDependencies()` in `src/npx-cli/install/setup-runtime.ts:403-469` previously invoked Bun through a string command with `--ignore-scripts`, which prevented Bun from running the trusted `tree-sitter-cli` install hook at all. The fix switches that call to array-based `spawn()` args, keeps `--frozen-lockfile`, drops `--ignore-scripts`, and preserves the existing timeout plus stdout/stderr error reporting so the installer still fails loudly when Bun hangs or exits non-zero.

`getTreeSitterBinForTests()` in `src/services/smart-file-read/parser.ts:483-528` now checks `tree-sitter.exe` before the extensionless binary when the resolver is running on Windows. `runBatchQuery()` at `src/services/smart-file-read/parser.ts:551-561` continues to share the same cached path and PATH fallback behavior, but it no longer discards a valid package-local Windows binary.

The regression tests anchor both execution paths. `tests/setup-runtime.test.ts:155-160` asserts that the install argv preserves `--frozen-lockfile` and no longer includes `--ignore-scripts`, and `tests/services/smart-file-read/tree-sitter-bin.test.ts:9-40` verifies Windows `.exe` preference plus bare-PATH fallback.

## Scope
This PR does not change the existing empty-result behavior when tree-sitter still cannot be executed for some other reason. It also does not rewrite smart-file-read to use WASM or bundle per-platform binaries; it keeps the native CLI path and only repairs the supported installer plus Windows resolver flow.

## Risk
The installer now permits Bun to execute trusted dependency scripts again, which is the intended package policy and is narrower than enabling arbitrary npm scripts. The spawn path remains array-based with `shell: false`, the timeout guard is preserved, and the parser's binary lookup only adds the Windows `.exe` candidate ahead of the existing fallback chain.

## Verification / Test plan
- [x] `bun test tests/setup-runtime.test.ts tests/services/smart-file-read/tree-sitter-bin.test.ts` - 17 passing
- [x] `npm run build`
- [ ] `npm run lint:hook-io && npm run lint:spawn-env && npm run strip-comments:check` - hook IO and spawn-env passed, `strip-comments:check` reports a pre-existing repo-wide failure (`Changed: 330`) in check mode

## Upstream

Closes #2910.
Reported by @jyepe.

### DIFF
--- a/src/npx-cli/install/setup-runtime.ts
+++ b/src/npx-cli/install/setup-runtime.ts
@@ -1,5 +1,5 @@
 import { existsSync, readFileSync, writeFileSync } from 'fs';
-import { execSync, spawn, spawnSync } from 'child_process';
+import * as childProcess from 'child_process';
 import { createRequire } from 'module';
 import { join } from 'path';
 import { homedir } from 'os';
@@ -72,7 +72,7 @@ function markerPath(targetDir: string): string {
 
 function getBunPath(): string | null {
   try {
-    const result = spawnSync('bun', ['--version'], {
+    const result = childProcess.spawnSync('bun', ['--version'], {
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
       shell: IS_WINDOWS,
@@ -94,7 +94,7 @@ function getBunVersion(): string | null {
   if (!bunPath) return null;
 
   try {
-    const result = spawnSync(bunPath, ['--version'], {
+    const result = childProcess.spawnSync(bunPath, ['--version'], {
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
       shell: IS_WINDOWS,
@@ -107,7 +107,7 @@ function getBunVersion(): string | null {
 
 function getUvPath(): string | null {
   try {
-    const result = spawnSync('uv', ['--version'], {
+    const result = childProcess.spawnSync('uv', ['--version'], {
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
       shell: IS_WINDOWS,
@@ -129,7 +129,7 @@ function getUvVersion(): string | null {
   if (!uvPath) return null;
 
   try {
-    const result = spawnSync(uvPath, ['--version'], {
+    const result = childProcess.spawnSync(uvPath, ['--version'], {
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
       shell: IS_WINDOWS,
@@ -157,13 +157,13 @@ function describeExecError(error: unknown): string {
 function installBun(): void {
   try {
     if (IS_WINDOWS) {
-      execSync('powershell -c "irm bun.sh/install.ps1 | iex"', {
+      childProcess.execSync('powershell -c "irm bun.sh/install.ps1 | iex"', {
         stdio: 'pipe',
         timeout: INSTALL_TIMEOUT_MS,
         shell: process.env.ComSpec ?? 'cmd.exe',
       });
     } else {
-      execSync('curl -fsSL https://bun.sh/install | bash', {
+      childProcess.execSync('curl -fsSL https://bun.sh/install | bash', {
         stdio: 'pipe',
         timeout: INSTALL_TIMEOUT_MS,
         shell: '/bin/bash',
@@ -189,13 +189,13 @@ function installBun(): void {
 function installUv(): void {
   try {
     if (IS_WINDOWS) {
-      execSync('powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"', {
+      childProcess.execSync('powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"', {
         stdio: 'pipe',
         timeout: INSTALL_TIMEOUT_MS,
         shell: process.env.ComSpec ?? 'cmd.exe',
       });
     } else {
-      execSync('curl -LsSf https://astral.sh/uv/install.sh | sh', {
+      childProcess.execSync('curl -LsSf https://astral.sh/uv/install.sh | sh', {
         stdio: 'pipe',
         timeout: INSTALL_TIMEOUT_MS,
         shell: '/bin/bash',
@@ -414,7 +414,7 @@ export async function installPluginDependencies(targetDir: string, bunPath: stri
     // Async spawn (not spawnSync): a blocked event loop freezes the installer's
     // clack spinner for the duration of the install, which reads as a stall.
     await new Promise<void>((resolve, reject) => {
-      const child = spawn(bunPath, installArgs, {
+      const child = childProcess.spawn(bunPath, installArgs, {
         cwd: targetDir,
         shell: false,
         stdio: ['pipe', 'pipe', 'pipe'],
--- a/src/npx-cli/install/setup-runtime.ts
+++ b/src/npx-cli/install/setup-runtime.ts
@@ -422,7 +422,9 @@ export async function installPluginDependencies(targetDir: string, bunPath: stri
 
       const stdoutChunks: Buffer[] = [];
       const stderrChunks: Buffer[] = [];
+      let timedOut = false;
       const timer = setTimeout(() => {
+        timedOut = true;
         child.kill();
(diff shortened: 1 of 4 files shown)


rymalia:
@rodboev — flagging overlap so we coordinate rather than collide. #3095 touches the same two functions you do (`installPluginDependencies` in `setup-runtime.ts`, `getTreeSitterBin` in `parser.ts`) for the same root cause (#2910).

Different call on the install invocation: yours drops `--ignore-scripts` globally and leans on Bun's trusted-dependency allowlist; #3095 keeps `--ignore-scripts` and runs only `tree-sitter-cli`'s install script explicitly. My reasoning for the narrower path is that `--ignore-scripts` was added by #2379 specifically to stop the `tree-sitter-swift` postinstall hang — dropping it globally reopens that surface, whereas running one vetted script doesn't. #3095 also adds a `tree-sitter --version` probe so a present-but-non-executable binary (e.g. the glibc prebuilt on musl) is caught instead of assumed-good.

Your Windows `tree-sitter.exe` resolver fix is the piece I didn't have and clearly need — I've folded that behavior into #3095 and credited it here. So the two aren't really competing; #3095 is close to a superset. Whichever the maintainers prefer to land, I'm glad to rebase the other onto it (or pull your resolver test over verbatim). What's your preference?

rodboev:
I think this is a question for @thedotmack. Both PRs are open, both address #2910, they differ on the `--ignore-scripts` approach. The maintainer should pick whichever they prefer. I'm happy to rebase if there are conflicts after one lands.

rymalia:
Update: I've since ported your injectable resolver tests into #3095 (`tree-sitter-bin.test.ts` — `tree-sitter.exe` preference on Windows, bare-PATH fallback, and non-Windows resolution), using the same `setTreeSitterBinDepsForTests` seam you introduced. So the Windows path now has real coverage, not just the behavior. Credit to you for the seam. Still happy to converge whichever way the maintainers prefer.

rodboev:
Closing this from my side. The linked issue now points at the fuller fix path in #3095, and that work was accepted onto the maintainer's `community-edge` branch. This older branch is no longer the active vehicle for the bug.

PR #2920, opened 2026-06-13
fix(chroma): prewarm uvx installs before the MCP connect deadline (closes #2897)

### LINKED ISSUE #2897: chroma-mcp 30s connect timeout kills uvx mid-install, causing infinite retry loop that leaked ~111GB of uv temp files (macOS)
## Summary

The worker connects to chroma-mcp by spawning `uvx` with a **hardcoded 30s MCP connect timeout** (`mF=3e4` in `scripts/worker-service.cjs`, still present in 13.5.6). On a cold uv cache, `uvx --python 3.13 --with 'onnxruntime>=1.20' --with 'protobuf<7' chroma-mcp==0.2.6` must download a managed CPython 3.13 (~70MB) plus chromadb/onnxruntime/kubernetes wheels (hundreds of MB), which easily exceeds 30s. The worker then kills the whole subprocess tree, uv is interrupted mid-extraction, and **the partially-built environment is never cached — so the next attempt starts cold again**, producing an infinite kill/retry loop under reconnect backoff.

Each killed attempt leaks partial extractions to disk. Over ~1 month this accumulated on my machine:

- **69 GB** / 6,389 `.tmpXXXXXX` dirs in `~/.cache/uv/` (interrupted wheel extractions)
- **42 GB** in `~/.local/share/uv/python/.temp/` (~600 partial CPython 3.13 installs)

**Total: ~111 GB** of leaked temp files.

## Environment

- claude-mem 13.4.1 → 13.5.6 (both affected; timeout & uvx args identical)
- macOS (darwin arm64), uv installed
- Slow-ish PyPI connectivity (mainland China) — makes cold installs reliably exceed 30s

## Evidence from `~/.claude-mem/logs/`

```
[2026-06-02 21:27:13.618] [WARN ] [CHROMA_MCP] Connection failed, killing subprocess tree to prevent zombie {error=MCP connection to chroma-mcp timed out after 30000ms}
[2026-06-02 21:28:15.220] [WARN ] [CHROMA_MCP] Connection failed, killing subprocess tree to prevent zombie {error=MCP connection to chroma-mcp timed out after 30000ms}
[2026-06-02 21:29:08.143] [WARN ] [CHROMA_MCP] Connection failed, killing subprocess tree to prevent zombie {error=MCP connection to chroma-mcp timed out after 30000ms}
[2026-06-02 21:31:59.175] [WARN ] [CHROMA_MCP] chroma-mcp subprocess closed unexpectedly, applying reconnect backoff
```

Even with a warm cache the margin is thin — a successful connect on 13.5.6 took 23s of the 30s budget:

```
[2026-06-11 11:50:30.336] [INFO ] [CHROMA_MCP] Connecting to chroma-mcp via MCP stdio {command=uvx, args=--python 3.13 --with onnxruntime>=1.20 --with protobuf<7 chroma-mcp==0.2.6 --client-type persistent --data-dir ...}
[2026-06-11 11:50:53.663] [INFO ] [CHROMA_MCP] Connected to chroma-mcp successfully
```

Because `onnxruntime>=1.20` / `protobuf<7` are range constraints, any new release on PyPI re-triggers a cold download and the loop starts again.

## Suggested fixes

1. **Make the connect timeout configurable** (e.g. `CLAUDE_MEM_CHROMA_CONNECT_TIMEOUT_MS`) and/or raise the default — 30s is not enough for a cold `uvx` install of chromadb+onnxruntime+managed Python.
2. **Pre-warm outside the MCP timeout**: run an explicit install step (`uv python install 3.13` + `uvx ... --help` or `uv tool install`) with a generous/no timeout before the MCP stdio connect, so the connect timeout only ever covers process startup, not package installation.
3. **Pin exact dependency versions** (e.g. `onnxruntime==x.y.z`) so the environment stays cache-hit across PyPI releases.
4. Optionally detect repeated connect-timeout kills and surface a warning, since the failure mode silently eats disk at GB/day rates.

Happy to provide more logs or test a patch.

rodboev:
## Summary

Cold `chroma-mcp` starts currently spend the same fixed 30 second budget on `uvx` package installation and on the MCP stdio handshake, so slow first installs get torn down as if the server had hung. This change moves the package-prepare work out of the handshake timeout, adds configurable timeouts for both phases, and keeps the existing reconnect cleanup logic intact once the real MCP connect has started.

## Why

`src/services/sync/ChromaMcpManager.ts:19` hard-codes `MCP_CONNECTION_TIMEOUT_MS = 30_000`, and `connectInternal()` arms that timeout before `this.client.connect(this.transport)` at `src/services/sync/ChromaMcpManager.ts:142-163`. The command being raced is not a cheap process start: `buildCommandArgs()` at `src/services/sync/ChromaMcpManager.ts:200-247` always asks `uvx` to resolve managed Python 3.13 plus `onnxruntime>=1.20` and `protobuf<7` around `chroma-mcp==0.2.6`. On a cold cache, that install work can legitimately exceed 30 seconds before the MCP handshake even begins.

When the timeout fires, the failure path at `src/services/sync/ChromaMcpManager.ts:153-160` logs a connect error and immediately calls `disposeCurrentSubprocess()`. That is correct for a real hung transport, but it is the wrong classification for a slow environment build because the partial install never reaches a reusable cache state. This PR splits those phases apart: a best-effort prewarm prepares the `uvx` environment first, then the existing MCP timeout is applied only to the real stdio handshake.

This approach follows the upstream design direction proposed in the upstream issue. It also stacks cleanly with open PR `#2880`, which changes the `uvx` invocation form for older `uv` versions but does not address the timeout race.

## Scope

This PR does not change orphan cleanup, reconnect backoff, or the existing tree-kill semantics once a transport connect has genuinely failed. It also does not tighten transitive dependency pins or add temp-file cleanup; those are adjacent concerns, not part of the timeout/prewarm fix. Open PR `#2880` remains the owner of the invocation-form compatibility surface.

## Risk

Startup now has an explicit prewarm phase, which means the first cold Chroma connect may launch one additional `uvx` process before the real MCP transport starts. The risk is bounded because the prewarm uses the same resolved executable and array-based args as the real spawn, and warm reconnects should skip it after the first successful prepare. The main behavioral change is intentional: slow installs now fail with a prewarm/config-timeout error instead of being misreported as a hung MCP handshake.

## Verification

- [x] `bun test tests/services/sync/chroma-mcp-manager-ssl.test.ts tests/services/sync/chroma-mcp-manager-singleton.test.ts tests/services/sync/chroma-mcp-manager-timeout.test.ts` — 14 passing
- [x] `npm run build`
- [x] `npm run lint:hook-io && npm run lint:spawn-env && npm run strip-comments:check` — `lint:hook-io` and `lint:spawn-env` passed; `strip-comments:check` still reports the repo-wide existing comment-removal baseline (`Changed: 330`)

## Upstream

Closes #2897.
Reported by @LXChild.


rodboev:
Closing this from my side. Current main already prewarms chroma-mcp before connect, carries a configurable prewarm timeout, and launches uvx directly in `src/services/sync/ChromaMcpManager.ts`. This branch no longer adds a distinct fix.

PR #2923, opened 2026-06-13
fix: make installer launchable, add uninstall + viewer memory deletion

Soundman-Coder:
## Summary

Fixes two issues: the installer failing to launch, and the missing ability to delete memories.

### Installer
- **`build.mjs`** now sets the execute bit on `dist/index.js`, so the shebang'd bin works when invoked via `npx`/directly (esbuild does not set it).
- Added an **Uninstall** option to the welcome menu (shown when claude-mem is already installed).
- New **uninstall step** (`src/steps/uninstall.ts`): stops the worker, removes the plugin from Claude Code's `settings.json` / `installed_plugins.json` / `known_marketplaces.json`, deletes the marketplace + cache directories, and optionally deletes `~/.claude-mem` memories (opt-in, defaults to keeping them).

### Viewer memory deletion
- **Worker**: new `DELETE /api/observation/:id`, `/api/summary/:id`, `/api/prompt/:id` endpoints (`DataRoutes.ts`) backed by `deleteObservation` / `deleteSessionSummary` / `deleteUserPrompt` in `SessionStore.ts`. FTS indexes stay in sync via the existing `AFTER DELETE` triggers. Each delete emits an `item_deleted` SSE event.
- **Viewer**: a `DeleteButton` (trash icon + confirmation) on each card, an `App`-level delete handler (calls the endpoint, drops the item from local state, refreshes stats), `useSSE` handling of `item_deleted` to sync other open tabs, and button styling in `viewer-template.html`.

## Verification
- ✅ Full plugin build passes (`npm run build` — worker, MCP, context generator, viewer); changes confirmed present in the bundles.
- ✅ Installer typechecks and builds; `dist/index.js` is now executable.
- ⚠️ No end-to-end runtime test (Bun was not installed in the dev environment, so the worker could not be started to exercise the DELETE endpoints live). The DB delete path was validated against the FTS `AFTER DELETE` triggers, which reference only existing columns.


Soundman-Coder:
Closing in favor of two focused PRs split by concern: #2924 (installer launcher) and #2925 (viewer memory deletion).

PR #2924, opened 2026-06-13
fix(installer): make launchable + add uninstall option

Soundman-Coder:
## Summary
Installer-only changes (split out from #2923).

- **`build.mjs`** now sets the execute bit on `dist/index.js`, so the shebang'd bin works via `npx`/direct invocation (esbuild does not set it). This was the likely cause of the installer "not launching".
- Added an **Uninstall** option to the welcome menu (shown when claude-mem is already installed).
- New **uninstall step** (`src/steps/uninstall.ts`): stops the worker, removes the plugin from Claude Code's `settings.json` / `installed_plugins.json` / `known_marketplaces.json`, deletes the marketplace + cache directories, and optionally deletes `~/.claude-mem` memories (opt-in, defaults to keeping them).

## Verification
- Installer typechecks and builds; `dist/index.js` is now `-rwxr-xr-x`.

### DIFF
--- a/installer/build.mjs
+++ b/installer/build.mjs
@@ -1,4 +1,7 @@
 import { build } from 'esbuild';
+import { chmodSync } from 'fs';
+
+const outfile = 'dist/index.js';
 
 await build({
   entryPoints: ['src/index.ts'],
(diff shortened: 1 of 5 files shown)


thedotmack:
Closing as superseded by the current installer and dedicated uninstall command. This older branch no longer matches the distribution architecture.

PR #2925, opened 2026-06-13
feat(viewer): delete memories from the viewer

Soundman-Coder:
## Summary
Viewer + worker changes (split out from #2923): the ability to delete a memory from the viewer.

### Worker
- New `DELETE /api/observation/:id`, `/api/summary/:id`, `/api/prompt/:id` endpoints (`DataRoutes.ts`) backed by `deleteObservation` / `deleteSessionSummary` / `deleteUserPrompt` in `SessionStore.ts`. FTS indexes stay in sync via the existing `AFTER DELETE` triggers. Each delete emits an `item_deleted` SSE event.

### Viewer
- A `DeleteButton` (trash icon + confirmation) on each card.
- App-level delete handler: calls the endpoint, drops the item from local state, refreshes stats.
- `useSSE` handles `item_deleted` to sync other open tabs.
- Button styling in `viewer-template.html`.

Includes rebuilt plugin artifacts (`worker-service.cjs`, `mcp-server.cjs`, `context-generator.cjs`, `viewer-bundle.js`, `viewer.html`).

## Verification (live, end-to-end)
- `DELETE` existing observation → **200** `{success, deleted:true}`; stats `1 → 0`; subsequent GET → **404**.
- `DELETE` non-existent id → **404**; invalid id → **400**.
- The `AFTER DELETE` FTS trigger runs without error.


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#2785 ([plan-12] Provider & Extensibility Roadmap)** as **needs rework**: the diagnosis is right, but the item noted below blocks it. If you'd like to keep it moving, address that against the design in the plan doc; otherwise the fix will be absorbed into the plan-12 bundle.

> viewer delete feature; P1: SQLite delete leaves Chroma vectors + prompt numbering

thedotmack:
Second pass — routing holds at **#2785 (plan-12, Provider & Extensibility Roadmap)**: viewer delete is a UX capability, not a defect in any of the architectural clusters. No open PR duplicates it and main still has no delete endpoints in `DataRoutes.ts`, so it is not superseded. The blocking P1 stands though — the SQLite delete leaves the Chroma vectors orphaned, so a deleted memory keeps coming back through semantic search; that needs a companion delete against the vector store (and the prompt-numbering gap) before this can land.

PR #2927, opened 2026-06-13
fix(sdk): break observer respawn loop on haiku skip-prose

rovencap:
Haiku-4.5 frequently returns conversational refusals like "(no observations - insufficient data in this observation window)" or "I observe an empty primary session state... Skipping observation generation". These were classified as `prose`, incremented `consecutiveInvalidOutputs`, and at threshold triggered respawn — which re-queued the same low-signal batch into the fresh session, producing the same prose, looping forever (observed in field logs as repeated "SDK session poisoned — killing and respawning" with `outputClass=idle/prose`).

Two-part fix:

1. New `'skip'` output class in `classifyObserverOutput()` for known polite-refusal markers the prompt itself warns against ("no observations", "insufficient data", "no substantive", "no tool execution", "skipping", etc.). Skip → confirm batch, no counter bump, no respawn. Order matters: the XML gate runs first so a real `<observation>` mentioning "no observations" in its narrative still parses cleanly.

2. For non-poison threshold-respawns (accumulated idle / unrecognized prose), confirm the offending batch *before* respawn so the fresh SDK session moves on to new work. `'poisoned'` (context wedge) still preserves the buffer — that case is genuinely recoverable with a clean context window.

Tests cover: skip-prose recognition, skip never increments counter or respawns over 3x threshold, prose-threshold confirms-before-respawn with strict call ordering, poisoned still preserves pending.


thedotmack:
Closing as superseded by current main’s general non-XML output rule. Provider-specific skip-prose recovery is no longer the right abstraction.

PR #2928, opened 2026-06-14
fix(install): mergeSettings preserves top-level keys on env-nested files

antfleet-ops:
## Bug

`mergeSettings` (src/npx-cli/commands/install.ts:743) silently destroys non-env top-level keys when the user has the Claude-Code-style nested layout in `~/.claude/settings.json`.

If the existing file looks like:
```json
{
  "env": { "CLAUDE_MEM_RUNTIME": "worker" },
  "hooks": [...],
  "permissions": {...},
  "apiKeyHelper": "..."
}
```

Then any call to `mergeSettings({ FOO: "1" })` rewrites the file as:
```json
{ "CLAUDE_MEM_RUNTIME": "worker", "FOO": "1" }
```

`hooks`, `permissions`, `apiKeyHelper`, etc. are gone — and any subsequent claude-mem operation that calls `mergeSettings` (which happens at install time and via `disableClaudeAutoMemory` from `writeJsonFileAtomic`) wipes user configuration that `readRawStoredAuthMethod` itself reads from `(raw.env ?? raw)`.

## Root cause

```ts
// before
if (parsed.env && typeof parsed.env === 'object') {
  current = { ...parsed.env };           // ← cloned subset, then…
}
// …
writeFileSync(path, JSON.stringify(current, null, 2), 'utf-8');  // ← overwrote whole file
```

The function read the env subtree into a separate object and wrote that as the entire file. Top-level peers had no path to survive the round-trip.

## Fix

Track whether the file used the env-nested shape and mutate the relevant subtree **in place** on the parsed document, preserving the rest. The write-back persists the full document, not just one subtree.

## How this was found

AntFleet two-model review on `bench-claude-mem` — Claude Opus 4.7 and GPT-5 unanimously flagged this as data-loss / high severity:
- https://github.com/AntFleet/bench-claude-mem/pull/1#issuecomment-4700643021

## Test plan

- [ ] Hand-test: `mkdir -p /tmp/cmem-test/.claude && echo '{"env":{"FOO":"1"},"hooks":[],"permissions":{}}' > /tmp/cmem-test/.claude/settings.json && HOME=/tmp/cmem-test node -e "..."` — confirm `hooks` and `permissions` survive the merge
- [ ] Existing unit tests pass (no new tests added — the bug is positional logic, captured by the diff itself; happy to add a regression test on request)

### DIFF
--- a/src/npx-cli/commands/install.ts
+++ b/src/npx-cli/commands/install.ts
@@ -10,7 +10,7 @@ import { homedir } from 'os';
 import { dirname, join } from 'path';
 import { SettingsDefaultsManager, type SettingsDefaults } from '../../shared/SettingsDefaultsManager.js';
 import { USER_SETTINGS_PATH } from '../../shared/paths.js';
-import { writeJsonFileAtomic as writeSettingsJsonAtomic } from '../../shared/atomic-json.js';
+import { parseJsonWithBom, writeJsonFileAtomic as writeSettingsJsonAtomic } from '../../shared/atomic-json.js';
 import { loadClaudeMemEnv, saveClaudeMemEnv } from '../../shared/EnvManager.js';
 import { ensureWorkerStarted, type WorkerStartResult } from '../../services/worker-spawner.js';
 import { formatHostForUrl } from '../../shared/worker-utils.js';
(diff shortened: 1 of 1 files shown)


antfleet-ops:
Hey @thedotmack — ping on this trio (#2928, #2929, #2930). All three are data-integrity fixes flagged by AntFleet's two-model bench review. #2928 + #2929 silently drop top-level `settings.json` keys on env-nested files (data loss); #2930 guards NaN config values that silently truncate output. Happy to rework any of them if the shape doesn't fit, or close out if they're not landing.

antfleet-ops:
Rebased onto latest `main` to clear the merge conflict.

Heads up while resolving: the `mergeSettings` refactor now on `main` (`readFlatSettings` + `writeSettingsJsonAtomic`) still carries the original data-loss bug this PR targets. `readFlatSettings(path)` unwraps an env-nested `settings.json` to just its `.env` subtree, and `mergeSettings` then writes that flattened view back as the *entire* file — so on env-nested files it silently drops every non-`env` top-level key (`hooks`, `permissions`, `apiKeyHelper`, `model`, `statusLine`, …) and collapses the `env` wrapper.

The rebased fix reads the **full** document (BOM-aware, via `parseJsonWithBom`), mutates only the `env` subtree when the file is env-nested (root otherwise), and writes the whole document back through the atomic writer — so it keeps your atomic-write + BOM-parse improvements and preserves top-level keys. Net diff is +25/−6, one file.

PR #2929, opened 2026-06-14
fix(server-beta-bootstrap): preserve top-level keys + close chmod TOCTOU

antfleet-ops:
## Two adjacent bugs in `persistServerBetaSettings`

Both surfaced by AntFleet's two-model review of claude-mem at AntFleet/bench-claude-mem#3 (unanimous between claude-opus-4-7 and gpt-5).

### 1. Data-loss on env-nested settings (high)

When `~/.claude-mem/settings.json` uses the Claude-Code-style nested shape:
```json
{
  "env": { ... },
  "hooks": [...],
  "permissions": {...}
}
```
the function was extracting `existing.env` into a `flat` object and writing `flat` back as the **entire file** — silently dropping every non-env top-level key. Same pattern fixed in `mergeSettings` (PR #2928).

```ts
// before
const flat = (existing.env && typeof existing.env === 'object'
  ? existing.env
  : existing);
flat.CLAUDE_MEM_SERVER_BETA_API_KEY = values.apiKey;
// ...
writeFileSync(settingsPath, JSON.stringify(flat, null, 2));  // ← clobbered the document
```

Fix: track whether the source used env-nested shape, mutate the relevant subtree in place, write the full document back.

### 2. `writeFile`/`chmod` TOCTOU on API key (medium / security)

The previous sequence was:
```ts
writeFileSync(path, ...)     // ← file created with default mode (e.g. 0644)
chmodSync(path, 0o600)       // ← restricts permissions
```
Between those two syscalls the API key plaintext was readable by any other local user. The window is short but it exists every install/rotation.

Fix: pass `{ mode: 0o600 }` directly to `writeFileSync` so the file is **created** with restrictive permissions atomically. The `chmod` stays as a belt-and-braces fallback for the case where the file already existed with looser permissions before this write.

## Test plan

- [ ] Hand-test: pre-populate `~/.claude-mem/settings.json` with `{env:{...},hooks:[],permissions:{}}`, call `persistServerBetaSettings`, confirm `hooks` and `permissions` survive
- [ ] Verify `ls -la` shows `0600` on the file immediately after `writeFileSync` returns (no race)

## How this was found

https://github.com/AntFleet/bench-claude-mem/pull/3#issuecomment-4700599686

### DIFF
--- a/src/services/hooks/server-bootstrap.ts
+++ b/src/services/hooks/server-bootstrap.ts
@@ -145,8 +145,13 @@ export function persistServerSettings(
       existing = {};
     }
   }
-  // Settings file format: prefer the flat shape (modern). The migration in
-  // SettingsDefaultsManager.loadFromFile already collapses nested → flat.
+  // Settings file format: support both the flat shape (modern) and the
+  // env-nested shape (Claude-Code-style: { env: {...}, hooks: [...], ... }).
+  // `flat` is a *reference* into `existing` — the env subtree when nested, or
+  // the root document otherwise — so mutating `flat` mutates `existing` in
+  // place. We then write the full `existing` document below (NOT `flat`), so
+  // non-env top-level keys (hooks, permissions, apiKeyHelper, ...) survive.
+  // Writing `flat` back as the whole file silently dropped them (data loss).
   const flat = (existing.env && typeof existing.env === 'object'
     ? existing.env
     : existing) as Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


antfleet-ops:
Rebased onto latest `main` (v13.11.0) to clear the merge conflict.

Heads up while resolving: the bootstrap refactor now on `main` renamed `persistServerBetaSettings` → `persistServerSettings` (in `server-bootstrap.ts`) and adopted `writeJsonFileAtomic`, but it **reintroduced the original data-loss bug** this PR targets. On an env-nested `settings.json`, it derives `flat = existing.env` and then calls `writeJsonFileAtomic(settingsPath, flat)` — writing only the `env` subtree back as the whole file, silently dropping every sibling top-level key (`hooks`, `permissions`, `apiKeyHelper`, …).

The rebased fix keeps `flat` as a reference into `existing` (so the key writes still land in the right subtree) and writes the full `existing` document. Net change is one line (`flat` → `existing`) plus comments. The atomic temp-file+rename writer already covers write-atomicity; I left the trailing `chmodSync(…, 0o600)` in place and documented the residual first-creation permission window (new files are created under the process umask until the chmod narrows them). Happy to add a regression test if useful.

Issue #2891, opened 2026-06-10
claude-mem worker unreachable — hooks fail silently on every tool call after session start


mykle0:
**Plugin version:** 13.5.5
**OS:** macOS (darwin)
**Frequency:** Persistent — occurs in 3+ separate sessions, survives app restarts

## Symptom

Every tool call in a session produces this error in the hook output:

    claude-mem worker unreachable for N consecutive hooks.

The counter increments on every hook (`file-context`, `observation`, `summarize`). Once it starts, it does not recover within the session.

## Hooks affected
- `file-context`
- `observation`
- `summarize`

## What was tried
- Restarted Claude Code — same behavior on next session
- Occurred across 3+ separate sessions on different days

## Impact
- All memory/observation logging silently drops for the entire session
- Hook noise pollutes every tool call response (stop hook feedback messages surfaced in UI)
- MCP server (`plugin_claude-mem_mcp-search`) appears to still load but hook-based logging is non-functional

## Expected behavior
Worker recovers or restarts automatically; hooks fail gracefully without surfacing raw error output to the user.

## Worker path
`~/.claude/plugins/cache/thedotmack/claude-mem/13.5.5/scripts/bun-runner.js` + `worker-service.cjs`

Goro2030:
Confirming this still reproduces on **13.6.1** (npm `latest`), on **WSL2/Ubuntu** — so it's not macOS-specific. I traced the mechanism and have a tested workaround + a concrete fix proposal.

## Environment
- claude-mem **13.6.1**, WSL2 Ubuntu (`systemd=true`), bun 1.3.11, node v22.19.0

## Reproduction of the broken recovery path
Once the worker is unreachable, every recovery through `worker-cli.js` fails:

```
$ node scripts/worker-cli.js start
Failed to start: Process died during startup
$ node scripts/worker-cli.js restart
Failed to restart: Process died during startup
```

…yet the worker runs perfectly fine when launched directly. So the worker isn't the problem — the **launch/health-gating logic is**.

## Root cause (`scripts/worker-cli.js`)
`startWithBun()` (Linux branch) spawns the worker detached, pipes its stdout/stderr into a stream **owned by the short-lived `worker-cli` process**, then gates readiness on the *spawned* PID:

```js
const child = spawn(bun, [workerScript], {
  detached: true,
  stdio: ["ignore", "pipe", "pipe"],   // piped into a stream owned by worker-cli
[...]

PR #2937, opened 2026-06-15
fix: remove stale plugin/scripts/worker-cli.js (regressed #2747/#2891 "Process died during startup")

### LINKED ISSUE #2747: worker-cli.js start 总是失败 'Process died during startup' — waitForHealth 检查了错误的 PID
## Summary

`worker-cli.js start` reports `Failed to start: Process died during startup` on macOS (v13.4.0), but the underlying worker is actually healthy and listening on the readiness port. The bug is in `waitForHealth` — it checks whether the *spawned parent* PID is alive, but `start` mode makes that parent `bun` process immediately `fork-exit` and daemonize the real worker as a grandchild.

## Reproduction

```bash
# Clean state
pkill -f "worker-service.cjs"; sleep 1

# Run the official CLI command
node ~/.claude/plugins/marketplaces/thedotmack/plugin/scripts/worker-cli.js start
# → Failed to start: Process died during startup
```

The CLI never even reaches the readiness fetch (or does it once, then bails on the first liveness check).

## Expected vs Actual

- **Expected:** CLI should print `Worker started (PID: <real-worker-pid>)` and return exit 0.
- **Actual:** Returns exit 1 with `Failed to start: Process died during startup` after ~50ms.

## Root cause

In `scripts/worker-cli.js`, `waitForHealth(t, e)` uses the PID returned from `spawn("bun", [..., "worker-service.cjs", "start"])` as the liveness probe. But the `start` subcommand in `worker-service.cjs` does the standard double-fork pattern: the immediate child (returned by spawn) is the launcher, which exits once the daemonized grandchild is up.

By the time `waitForHealth` runs `isProcessAlive(t)`, `t` is already dead. The function short-circuits to the `"Process died during startup"` branch *before* ever hitting the readiness fetch:

```js
// scripts/worker-cli.js — waitForHealth (minified, restored)
static async waitForHealth(t, e) {
  for (let r = 0; r < 15; r++) {
    if (this.isProcessAlive(t)) {
      if ((await fetch(`http://127.0.0.1:${e}/api/readiness`,
                       { signal: AbortSignal.timeout(1000) })).ok) {
        return { success: true, pid: t };
      }
    } else {
      return { success: false, error: "Process died during startup" };
      //   ^^^ returns here immediately because launcher already exited
    }
    await new Promise(c => setTimeout(c, 200));
  }
  return { success: false, error: `Readiness check timed out…` };
}
```

## Workaround (confirmed working)

Skip `worker-cli.js` and start the worker directly in daemon mode:

```bash
nohup ~/.bun/bin/bun \
  ~/.claude/plugins/marketplaces/thedotmack/plugin/scripts/worker-service.cjs \
  --daemon </dev/null >/dev/null 2>&1 &
disown
```

→ `curl http://127.0.0.1:37777/api/readiness` returns `{"status":"ready","mcpReady":true}` within ~1s.

## Suggested fix

Two changes in `waitForHealth`:

1. **Don't fail fast on dead parent PID.** The `else` branch on a dead PID should be removed or only fall through to a slower check; treat dead parent as "still starting" and keep polling readiness.
2. **Readiness is the source of truth.** Drop the `isProcessAlive(t)` gate entirely — just keep polling `GET /api/readiness` until it returns 200 or the timeout elapses. The PID written to `worker.pid` should be resolved from the readiness response or the daemonized grandchild, not the launcher.

Optional: in `start`, persist the daemonized grandchild's PID to `worker.pid` (likely already done by the `start` subcommand, but worth verifying the CLI reads from there rather than the spawn handle).

## Environment

- claude-mem v13.4.0 (`84636894`)
- macOS (darwin)
- Bun 1.3.14
- Node v25.9.0

## Related

- Files identical between `marketplaces/thedotmack/plugin/scripts/` and `cache/thedotmack/claude-mem/13.4.0/scripts/` (MD5 verified).
- Hook flow uses `bun-runner.js` which spawns the worker differently and is unaffected — this issue is specific to the user-facing `worker-cli.js start` path.
- Discovered while debugging hook errors `RangeError: Maximum call stack size exceeded` (separate issue, likely #2188 / #623 in `bugfixes-2026-01-10.md`).

Goro2030:
## What

Removes `plugin/scripts/worker-cli.js`, a stale committed bundle that ships the
"Process died during startup" bug from #2747 (and the surface symptom in #2891).

## Why

`worker-cli.js` (and its source `src/cli/worker-cli.ts`) were **already deleted
in v8.2.1** — per the changelog:

> Code Consolidation (-580 lines) … `src/cli/worker-cli.ts` (81 lines) — CLI
> handling now in worker-service … All hooks now use `worker-service.cjs` CLI
> directly instead of the deleted `worker-cli.js`.

The committed bundle `plugin/scripts/worker-cli.js` has since **regressed back
into the tree**, carrying the old pre-consolidation code. It is:

- **not built** — `npm run build` (`sync-plugin-manifests` + `build-hooks` +
  `gen-plugin-lockfile`) never emits it; `build-hooks.js` only bundles
  `worker-service.cjs`,
- **not referenced** by any runtime path — only by
  `tests/plugin-scripts-line-endings.test.ts`,
- but **still shipped** (it matches the `plugin/scripts/*.js` files allowlist).

### The bug it ships

Its `startWithBun()` (Linux branch) spawns the worker detached, pipes the
child's stdout/stderr into a `WriteStream` **owned by the short-lived CLI
process**, then gates readiness on the **launcher PID**:

```js
const child = spawn(bun, [workerScript], { detached: true, stdio: ["ignore","pipe","pipe"], ... });
const out = createWriteStream(logFile, { flags: "a" });
child.stdout?.pipe(out); child.stderr?.pipe(out); child.unref();
return this.waitForHealth(child.pid, port);   // gates on the launcher PID
```

```js
waitForHealth(pid, port, timeout = 1e4) {
  while (Date.now() - start < timeout) {
    if (!this.isProcessAlive(pid))
      return { success: false, error: "Process died during startup" };
    // ...readiness fetch...
  }
}
```

The launcher exits as soon as the worker re-execs to `--daemon`, so
`waitForHealth` sees `isProcessAlive(launcherPid) === false` and returns
`"Process died during startup"` **before any readiness fetch runs** — and the
CLI process exiting tears down the piped stdio, killing the detached daemon too.
A user running the obvious manual recovery
`node plugin/scripts/worker-cli.js start|restart` gets the error **plus a dead
worker**. (Reproduced on WSL2/Ubuntu, claude-mem 13.6.1.)

### Live paths are unaffected

The actual recovery paths are already correct and untouched by this PR:
- hooks → `ensureWorkerRunning()` — lazy-spawn with `stdio: 'ignore'` + `detached`,
- `spawnDaemon()` — `setsid` + `stdio: 'ignore'` + `detached` + `unref()`.

So this file is the only remaining broken surface: dead, unbuilt, shipped.

## Changes

- Delete `plugin/scripts/worker-cli.js`.
- Drop it from `tests/plugin-scripts-line-endings.test.ts` (now **8 pass / 0 fail**).
- Remove the `worker-cli.js` line from the architecture file-tree doc
  (`docs/public/architecture/overview.mdx`).

## Testing

```
$ bun test tests/plugin-scripts-line-endings.test.ts
 8 pass / 0 fail
```

## Note

If a standalone CLI entry point is still desired, it should be re-derived from
the current `worker-service` path rather than restored from this pre-8.2.1
artifact. Happy to follow up with that if preferred.

Closes #2747 (and addresses the manual-recovery surface of #2891).


thedotmack:
Closing as already implemented on current main: the stale worker-cli.js artifact is no longer shipped.

Issue #2485, opened 2026-05-15
v13.2.0: observer SDK responses dropped by parser as non-XML; observations table stays at 0


Yacine-dev:
## Bug Description

v13.2.0 fresh install: every observer SDK response is dropped by the parser as `non-XML/empty`, so the `observations` table stays at 0 rows even though all upstream stages are healthy. The observer Claude SDK is replying conversationally instead of emitting `<observation>` XML, and the parser silently discards every batch.

**Pipeline status after a clean install + valid OAuth (`claude` CLI re-login, fresh `Claude Code-credentials` keychain `mdat`):**

| Stage | Status |
|---|---|
| Hooks fire (`HOOK -> PostToolUse: Read/Bash/...`) | OK |
| Queue enqueues with full tool metadata | OK |
| SDK auth (`Injected fresh CLAUDE_CODE_OAUTH_TOKEN at spawn-time`) | OK |
| Generator (Claude SDK returns 49-318 char responses) | OK |
| **XML parser** | FAIL - drops every batch as `non-XML/empty` |
| `observations` table | FAIL - stays at 0 |
| `session_summaries` table | FAIL - stays at 0 |
| `npx claude-mem search` | FAIL - no hits (nothing indexed) |

Example dropped response (conversational English instead of `<observation>` XML):

> "I'm observing the primary session, but the provided context only shows a user request to 'please ret...'"

Looks like a prompt-honoring regression in v13.2.0 -- the observer model isn't following the system prompt's instruction to emit structured XML.

## Steps to Reproduce

```
1. npx claude-mem install
2. claude                     # complete OAuth flow so Claude Code-credentials keychain entry is fresh
3. npx claude-mem restart
4. Use Claude Code normally for ~30s (Read, Bash, ToolSearch will trigger hooks)
5. sqlite3 ~/.claude-mem/claude-mem.db "SELECT COUNT(*) FROM observations;"        # => 0
6. grep -c "non-XML/empty" ~/.claude-mem/logs/claude-mem-$(date +%Y-%m-%d).log     # => climbs by >=1 per batch
```

## Expected Behavior

Observer emits `<observation>` XML that the parser accepts; `observations` row count climbs as hooks fire during normal Claude Code usage; `npx claude-mem search <term>` returns hits after a few minutes of activity.

## Environment

- **Claude-mem version**: 13.2.0
- **Claude Code version**: 2.1.142 (also reproduced with 1.0.120 earlier in the same install)
[...]

PR #2943, opened 2026-06-15
fix(worker): stop respawning sessions on benign "nothing to record" output

Dmtrch:
## Problem

A healthy observer session that legitimately has **no memory-worthy content** answers with empty/idle output or plain prose such as `No observations to record.`

In `ResponseProcessor.processAgentResponse`, *every* non-XML output — `idle` **and** `prose` alike — incremented `consecutiveInvalidOutputs`. After 3 such benign responses the session was classified as `poisoned`, killed and respawned. On a session that keeps having nothing to record this becomes a kill/respawn loop that saturates the worker and surfaces to hooks as repeated **"worker unreachable"**.

Observed in the wild:
```
[PARSER] SDK returned non-XML idle  ... consecutiveInvalidOutputs=1
[PARSER] SDK returned non-XML prose ... consecutiveInvalidOutputs=2   preview="No observations to record."
[SESSION] SDK session poisoned — killing and respawning ... threshold=3
[SDK_SPAWN] child emitted error event ... AbortError
```

This contradicts the classifier's own documented intent (plan-11 #2485): *"avoid respawn churn on benign idle output, and trigger recovery on a poisoned session."*

## Fix

- **`output-classifier.ts`** — recognize plain-prose "nothing to record" skips (`no observations to record`, `nothing to summarize`, `no summary needed`, …) as the existing benign `idle` class instead of `prose`. The SDK sometimes answers in prose instead of emitting `<skip_summary/>`.
- **`ResponseProcessor.ts`** — `idle` no longer counts toward the respawn threshold. Only `prose` (a possible wedge) accumulates; `poisoned` still respawns immediately. This realigns the code with the classifier's documented intent.

Genuine conversational prose and real `poisoned` closure strings are unaffected — the wedge protection is preserved.

## Tests

- `tests/sdk/output-classifier.test.ts` — benign-skip prose classifies as `idle`.
- `tests/worker/poison-respawn.test.ts` — repeated idle / "nothing to record" outputs never trigger a respawn, regardless of count.

Both suites pass (`bun test` — 18/18 on the two affected files).

> Note: only source + tests are included; rebuilt `plugin/scripts/*.cjs` bundles are intentionally omitted (esbuild reorders the bundle, producing hundreds of lines of unrelated churn).

### DIFF
--- a/src/sdk/output-classifier.ts
+++ b/src/sdk/output-classifier.ts
@@ -31,6 +31,21 @@ const POISONED_MARKERS: string[] = [
   'this session has ended',
 ];
 
+// Benign "nothing to record" phrases the SDK emits as plain prose instead of
+// <skip_summary/>. They mean the batch had no memory-worthy content — a healthy
+// outcome — so they must be treated as idle, never as a wedge that accumulates
+// toward a respawn. Kept lowercase; matched case-insensitively against the raw
+// output.
+const BENIGN_SKIP_MARKERS: string[] = [
+  'no observations to record',
+  'nothing to record',
+  'no observations to make',
+  'nothing to summarize',
+  'no summary needed',
+  'nothing worth recording',
+  'no memory-worthy',
+];
+
 /**
  * Returns a short, single-line preview of raw output for diagnostics/logging so
  * a dropped batch is visible instead of silent.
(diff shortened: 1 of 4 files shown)


thedotmack:
Closing as superseded by current main: benign and incompatible observer prose is ignored and never causes SDK respawn debt.

Issue #2708, opened 2026-05-29
where bun 在Windows 用户名为文中文时上的 GBK/UTF-8 编码问题


11zzx:
where bun 在Windows 用户名为文中文时上的 GBK/UTF-8 编码问题

thedotmack:
Consolidating into #2779 (plan-02). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

PR #2944, opened 2026-06-15
fix(windows): hide cmd window in `where bun` spawnSync (findBun)

budriamanda:
## Problem

On Windows, `findBun()` in both `plugin/scripts/bun-runner.js` and `plugin/scripts/version-check.js` resolves Bun with:

```js
spawnSync('where bun', { encoding: 'utf-8', stdio: ['pipe','pipe','pipe'], shell: true })
```

With `shell: true` and no `windowsHide`, Node launches `cmd.exe /c where bun`, which **flashes a visible console window**. Because `bun-runner.js` runs on every hook event (SessionStart, UserPromptSubmit, PostToolUse on Write/Edit, Stop), users see constant window blinking during normal Claude Code use.

This is the same class of issue tracked around #2708 / the Windows spawn surfaces — this PR fixes the two `where bun` lookups specifically.

## Fix

Add `windowsHide: true` to both `where bun` `spawnSync` calls.

- The main `spawn` in `bun-runner.js` and `codexSpawn` in `CodexCliInstaller.ts` already pass `windowsHide: true` — these two lookups were the remaining surfaces that escaped it.
- POSIX `which` path is untouched (no shell, no window).

## Test

`tests/bun-runner.test.ts` asserts `source.toContain("spawnSync('where bun'")` — still satisfied (the call signature is unchanged apart from the added option).

### DIFF
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -25,7 +25,8 @@ function findBun() {
     ? spawnSync('where bun', {
         encoding: 'utf-8',
         stdio: ['pipe', 'pipe', 'pipe'],
-        shell: true
+        shell: true,
+        windowsHide: true
       })
     : spawnSync('which', ['bun'], {
         encoding: 'utf-8',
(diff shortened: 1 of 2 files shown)


thedotmack:

PR #2948, opened 2026-06-16
fix(codex): remove unsupported hooks description

### LINKED ISSUE #2947: Codex 0.140.0 rejects codex-hooks.json root description field
## Bug

`plugin/hooks/codex-hooks.json` in the current codebase has a root-level `description` field:

```json
{
  "description": "claude-mem Codex CLI hook integration",
  "hooks": { ... }
}
```

Codex CLI 0.140.0 rejects this hooks config with an unknown-field error:

```text
unknown field description
```

This prevents Codex from loading the claude-mem Codex hook configuration until the root `description` field is removed.

## Confirmed current state

- Local CLI: `codex-cli 0.140.0`
- Current source file: `plugin/hooks/codex-hooks.json` has root keys `description` and `hooks`
- `origin/main:plugin/hooks/codex-hooks.json` also still contains the root-level `description`
- The installed/cache copy works after manually removing the root-level `description`, leaving only root `hooks`

## Root cause

The Codex hooks file includes metadata that Codex 0.140.0 does not accept at the hook-config root. The build verification in `scripts/build-hooks.js` currently validates hook event names under `codexHooks.hooks`, but it does not validate the allowed root keys for Codex hook configs, so this incompatible root field can ship.

## Suggested fix

Remove the root-level `description` from `plugin/hooks/codex-hooks.json` and keep the file shaped as:

```json
{
  "hooks": { ... }
}
```

Optionally add a build-time assertion that Codex hook configs only contain root keys accepted by Codex 0.140.0, so future metadata fields do not regress this.

a652:
## Summary
- remove the unsupported root-level `description` field from `plugin/hooks/codex-hooks.json`
- add build-time validation so future unsupported Codex hook root fields fail before release

Fixes #2947

## Tests
- `bun test tests/infrastructure/plugin-distribution.test.ts`
- `npm run build`

### DIFF
--- a/plugin/hooks/codex-hooks.json
+++ b/plugin/hooks/codex-hooks.json
@@ -1,5 +1,4 @@
 {
-  "description": "claude-mem Codex CLI hook integration",
   "hooks": {
     "SessionStart": [
       {
--- a/scripts/build-hooks.js
+++ b/scripts/build-hooks.js
@@ -629,6 +629,12 @@ async function buildHooks() {
       }
     }
     const codexHooks = JSON.parse(fs.readFileSync('plugin/hooks/codex-hooks.json', 'utf-8'));
+    const validCodexHookRootKeys = new Set(['hooks']);
+    for (const rootKey of Object.keys(codexHooks)) {
+      if (!validCodexHookRootKeys.has(rootKey)) {
+        throw new Error(`plugin/hooks/codex-hooks.json contains unknown Codex hook root field: ${rootKey}`);
+      }
+    }
     for (const eventName of Object.keys(codexHooks.hooks ?? {})) {
       if (!validCodexHookEvents.has(eventName)) {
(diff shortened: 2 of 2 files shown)


Glucksberg:
Validated this shape against Codex CLI 0.142.0 with a local claude-mem install. The current manifest fails to load on the top-level `description`; removing it and keeping the build-time root-key guard lets new Codex sessions ingest observations again.

Also ran `npm run build` successfully after applying the same two-file patch locally. The separate local requirement was enabling Codex hooks in `~/.codex/config.toml`; that is installer/doctor follow-up rather than part of this manifest fix.

thedotmack:
Both changes here — dropping the top-level \`description\` field and the \`validCodexHookRootKeys\` build guard in \`scripts/build-hooks.js\` — are already present on \`main\`. Closing as redundant. The remaining half of this bug class (SessionStart no-op fallback leaking an invalid \`{"continue":true}\` to Codex) is fixed in #3127.

PR #2951, opened 2026-06-16
test(worker): guard in-memory buffer reclaim

anavarro-bb:
## Summary

- Add regression coverage for `SessionMessageBuffer.resetClaimed()` re-yielding an in-flight message without requiring a durable `pending_messages.status='processing'` row.
- This guards the worker queue failure class observed in Liberty [LIB-284](/LIB/issues/LIB-284): orphaned durable processing rows wedging the drain.

## Context

The live Liberty box is running claude-mem plugin `v10.6.2`, where the installed generated worker bundle still has the durable pending queue behavior. Upstream `main` (`v13.6.1`) has already moved the worker to the in-memory `SessionMessageBuffer` path, which removes the cross-process orphaned `processing` row failure mode. This PR keeps that behavior reviewable and locked down with a focused test instead of proposing a stale patch against the old generated bundle.

## Verification

- `/Users/alex/.bun/bin/bun test tests/services/worker/session-message-buffer.test.ts`
- `git diff --check`

## Not done

- Did not apply or restart the live Liberty claude-mem worker.
- Did not mutate the live sqlite queue.
- Did not merge this PR.

### DIFF
--- a/tests/services/worker/session-message-buffer.test.ts
+++ b/tests/services/worker/session-message-buffer.test.ts
@@ -72,6 +72,36 @@ describe('SessionMessageBuffer (in-RAM observation buffer)', () => {
     expect(second.map(m => m.tool_name)).toEqual(['Write']);
   });
 
+  test('resetClaimed re-yields an in-flight message without a durable processing row', async () => {
+    const buffer = new SessionMessageBuffer();
+    const controller = new AbortController();
+    const iterator = buffer.drain({
+      sessionDbId: 1,
+      signal: controller.signal,
+      idleTimeoutMs: 10_000,
+    });
+
+    const messageId = buffer.enqueue(1, obs('Read', 'tool-1'));
+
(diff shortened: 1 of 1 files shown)


thedotmack:

> test-only; main already covers the drainAll variant

thedotmack:
Second pass — this routes to **#3609 (plan-21, SQLite Schema Evolution & Queue State Integrity)**, the queue-state half of that master. Closing as superseded: main already carries the exact assertion this adds. `61fe70a20` ("refactor: remove retry/observation queue, replace with SessionMessageBuffer") landed `tests/services/worker/session-message-buffer.test.ts` with `test('confirm removes a message; resetClaimed makes claimed messages re-drainable')`, which is the same re-yield guard, and the durable `pending_messages.status='processing'` failure mode this was written against no longer exists on that path. Thanks for chasing it down — the coverage you wanted is in place.

PR #2953, opened 2026-06-16
Fix claude-mem codex-hooks.json for current Codex: drop `description` (0.140 parser) + filter SessionStart no-op

shaneli2016:
## Problem
`plugin/hooks/codex-hooks.json` breaks on current Codex CLI in two ways:

1. **Codex 0.140 fails to even parse the file** — its stricter `deny_unknown_fields` hooks parser rejects the top-level `description` key:
   ```
   failed to parse plugin hooks config .../codex-hooks.json: unknown field `description`, expected `hooks`
   ```
2. **On 0.139 it parses but SessionStart fails** — the context hook pipes the worker's `{"continue":true}` (a Claude Code no-op) to Codex, which rejects it: `SessionStart hook (failed): hook returned invalid session start JSON output`.

## Fixes (all in `plugin/hooks/codex-hooks.json`)
- **Remove the top-level `description` field** so the file parses on Codex 0.140+ (only `hooks` remains; `matcher`/`timeout`/`statusMessage` are still accepted).
- **worker-start hook**: append `>/dev/null` so its stdout isn't read as hook output.
- **context hook**: pipe through a filter that only forwards a real SessionStart payload, dropping the `{"continue":true}` no-op:
  ```
  ... hook codex context | node -e 'let s="";process.stdin.on("data",d=>s+=d);process.stdin.on("end",()=>{s=s.trim();if(!s)return;try{const j=JSON.parse(s);if(j&&j.hookSpecificOutput&&j.hookSpecificOutput.hookEventName==="SessionStart")process.stdout.write(JSON.stringify(j));}catch(e){process.stdout.write(s);}})'
  ```

## Preferred root fix (maintainer's call)
Cleaner would be in `worker-service.cjs`: when invoked as `hook codex context`, never emit `{"continue":true}` — return a valid Codex `hookSpecificOutput` payload or nothing.

## Testing
```
printf 'Reply with exactly: OK\n' | codex exec --skip-git-repo-check
```
Verified on **Codex CLI 0.140.0** + claude-mem 13.6.1: 0 parse errors, all `SessionStart` hooks `Completed`, no `invalid session start JSON output`.


shaneli2016:
Thanks for the review — both points on the context hook are valid and now addressed (latest commit):

1. **`catch` re-emitting non-JSON** → the `catch` branch now writes to **stderr** instead of stdout, so malformed/plain-text worker output can never reach Codex's hook reader (no risk of re-triggering "invalid session start JSON output").
2. **Pipeline exit code masked** → the worker call is now captured with `_CTX=$(… hook codex context) || exit $?` before filtering, so a crashed/failed worker surfaces its non-zero exit to Codex instead of being swallowed by the filter's `exit 0`.

Tested the four paths locally: no-op `{"continue":true}` → empty/exit 0; real `SessionStart` payload → forwarded; non-JSON → stderr only; worker non-zero exit → propagated.

(The simplest root fix is still worker-side — emit a valid Codex payload or nothing for `hook codex context` — but this keeps the shim self-contained.)

Glucksberg:
I validated the parser side of this on Codex CLI 0.142.0: the top-level `description` removal is enough to get the hooks manifest accepted again, and that same minimal fix is already isolated in #2948.

This PR still covers the extra SessionStart output-filtering behavior, so I would treat it as a follow-up after #2948 lands or rebase/drop the duplicate `description` commit to keep the scopes clean.

thedotmack:
Superseded by #3127, which fixes the root cause in \`worker-service.cjs\` (the SessionStart no-op fallback in \`hook-command.ts\` now emits a valid \`hookSpecificOutput\` instead of leaking a bare \`{"continue":true}\`) rather than patching the shipped \`codex-hooks.json\` shell script. The top-level \`description\` field this PR also addressed was already dropped separately. Closing in favor of #3127.

PR #2957, opened 2026-06-16
fix(sdk): bound Claude provider observer context to prevent overflow (#2956)

### LINKED ISSUE #2956: Claude provider (SDKAgent) has no context bound — observer overflows with "Prompt is too long" and saves zero memory on long sessions
## Summary

When using the **Claude provider** (`CLAUDE_MEM_PROVIDER=claude`, `SDKAgent`), the observer agent accumulates every observation into a single, ever-growing SDK conversation with **no sliding window, no summarization, and no message/token cap**. On long primary sessions the conversation eventually exceeds the model's context window, the SDK returns "Prompt is too long", and the session is aborted — silently producing **zero** memory for that whole session.

The OpenRouter provider already solves exactly this problem; the Claude provider just never got the same treatment.

## Version / environment

- claude-mem `v10.6.3`
- `CLAUDE_MEM_PROVIDER=claude`, `CLAUDE_MEM_MODEL=claude-sonnet-4-5` (~200K context)
- Triggered reliably by long sessions with large tool outputs (big `Read`s, long `Bash`/test logs, web fetches).

## Root cause

Each observation is rendered with the **full** pretty-printed `tool_input` and `tool_output` embedded in the prompt:

`src/sdk/prompts.ts` — `buildObservationPrompt` (~L114-118)
```ts
return `<observed_from_primary_session>
  <what_happened>${obs.tool_name}</what_happened>
  <occurred_at>${new Date(obs.created_at_epoch).toISOString()}</occurred_at>${obs.cwd ? `\n  <working_directory>${obs.cwd}</working_directory>` : ''}
</observed_from_primary_session>`;
```

These are fed into the SDK conversation one-by-one by the message generator, with no bound on count or cumulative size:

`src/services/worker/SDKAgent.ts` — message generator (~L371-408)
```ts
for await (const message of this.sessionManager.getMessageIterator(session.sessionDbId)) {
  if (message.type === 'observation') {
    const obsPrompt = buildObservationPrompt({ /* ... */ });
    yield { type: 'user', message: { role: 'user', content: obsPrompt }, /* ... */ };
  }
  // no truncation, no rolling summary, no max-observations/max-tokens guard
}
```

A `grep` for `truncate` / `MAX_CONTEXT` / `MAX_ESTIMATED` / `sliding` in `SDKAgent.ts` finds nothing relevant — the only `truncate` references there are for shortening a single response in a **log line**, not the conversation history.

When the overflow hits, it is detected but the run just aborts with no memory saved:

`src/services/worker/SDKAgent.ts` (~L200-206, L253-255)
```ts
if (textContent.includes('prompt is too long') || textContent.includes('context window')) {
  logger.error('SDK', 'Context overflow detected - terminating session');
  return;
}
// ...
if (typeof textContent === 'string' && textContent.includes('Prompt is too long')) {
  throw new Error('Claude session context overflow: prompt is too long');
}
```

## The fix already exists for the other provider

`OpenRouterAgent` bounds its history with a sliding window keyed on configurable limits:

`src/services/worker/OpenRouterAgent.ts` — `truncateHistory` (~L296-329)
```ts
const MAX_CONTEXT_MESSAGES = parseInt(settings.CLAUDE_MEM_OPENROUTER_MAX_CONTEXT_MESSAGES) || DEFAULT_MAX_CONTEXT_MESSAGES; // 20
const MAX_ESTIMATED_TOKENS = parseInt(settings.CLAUDE_MEM_OPENROUTER_MAX_TOKENS) || DEFAULT_MAX_ESTIMATED_TOKENS;

if (history.length <= MAX_CONTEXT_MESSAGES) {
  const totalTokens = history.reduce((sum, m) => sum + this.estimateTokens(m.content), 0);
  if (totalTokens <= MAX_ESTIMATED_TOKENS) return history;
}
// keep most-recent messages within both limits, drop the rest
```

`SDKAgent` has no equivalent. That asymmetry is the bug.

## Suggested fix

Port the same bounding into the Claude provider. A few complementary options:

1. **Sliding window in `SDKAgent`** — mirror `truncateHistory`: cap the conversation by a configurable `CLAUDE_MEM_CLAUDE_MAX_CONTEXT_MESSAGES` / `CLAUDE_MEM_CLAUDE_MAX_TOKENS` (or a provider-agnostic `CLAUDE_MEM_OBSERVER_MAX_*`), keeping most-recent observations.
2. **Rolling summary** — periodically (every N observations) replace older raw observations with a compact summary, so memory of early events survives compaction instead of being dropped.
3. **Trim per-observation payloads** — `buildObservationPrompt` embeds entire `tool_output` JSON; a per-field size cap (e.g. truncate any single `parameters`/`outcome` block over K bytes) would cut the dominant bloat source for `Read`/`Bash`/web tools without losing the event itself.
4. **Graceful degradation** — on detected overflow, emit a minimal "observed but truncated" record (or trigger compaction and continue) rather than aborting with zero output, so a long session still yields *some* memory.

## Workaround for users on the Claude provider today

Add the high-volume, low-signal read tools to `CLAUDE_MEM_SKIP_TOOLS` in `~/.claude-mem/settings.json`, e.g. append `Read,Grep,Glob,WebFetch,WebSearch`. These are dropped at queue time in `SessionRoutes.ts` (~L507-510) before entering the observer conversation, which significantly slows the accumulation. It reduces the failure rate but does not eliminate it — a long session with heavy `Bash`/test output can still overflow, which is why the bounding logic above is the real fix.

dndungu:
Fixes #2956.

## Problem

When using the Claude provider (`CLAUDE_MEM_PROVIDER=claude`, `ClaudeProvider` / Agent SDK), the observer agent accumulates every observation into a single long-lived SDK conversation that we resume via `memorySessionId`. There is **no cumulative cap** on how large that context grows. On long primary sessions it eventually exceeds the model window, the SDK returns "Prompt is too long", and the run aborts — saving **zero** memory for the whole session. The OpenRouter and Gemini providers already bound their context; the Claude provider never got an equivalent. That asymmetry is the bug.

## Why OpenRouter's fix doesn't port directly

`OpenRouterProvider.truncateHistory` slices `session.conversationHistory` before each stateless HTTP request because OpenRouter resends the full history every turn. In `ClaudeProvider`, that array is only ever **pushed to**, never resent — the SDK owns the real context internally. Slicing it would be a no-op. The only lever for the SDK path is to start a **fresh SDK session** once the context grows too large.

## What's already handled (not touched)

- **Per-observation payload trimming** already exists: `src/sdk/prompts.ts` `truncateObservationField` caps each `<parameters>`/`<outcome>` block at 16k chars (issue option 3, added for #2468).
- A **reactive** overflow handler already exists in `ClaudeProvider.startSession` (resets + aborts after the model echoes "Prompt is too long"). It fires only after the wall is hit.

This PR adds the missing **proactive cumulative bound**.

## The fix

After each turn, read the SDK-reported full context size (`input_tokens + cache_creation_input_tokens + cache_read_input_tokens`) and, if it exceeds the new `CLAUDE_MEM_CLAUDE_MAX_TOKENS` cap (default `150000`), call `resetSessionForFreshStart` (clears `memorySessionId`, sets `forceInit`) and abort — mirroring the existing in-loop quota guard. The next observation ingest starts a fresh, small SDK session. The guard fires **only after** the current turn's memory has been emitted, so it loses nothing, and it stops the observer **before** the hard overflow wall.

### Changes
- `SettingsDefaultsManager`: add `CLAUDE_MEM_CLAUDE_MAX_TOKENS` (default `150000`), on par with `CLAUDE_MEM_GEMINI_MAX_TOKENS` / `CLAUDE_MEM_OPENROUTER_MAX_TOKENS`.
- `SettingsRoutes`: allowlist + validate (1000..1000000, mirroring `GEMINI_MAX_TOKENS`).
- `ClaudeProvider`: proactive context-bound guard in the `startSession` message loop.
- `worker-types` / `SessionRoutes` / `scrub` / `telemetry`: add a distinct `context_bound` abort reason (proactive reset vs reactive `overflow`).

## Tests
- `tests/shared/settings-defaults-manager.test.ts`: asserts the new default.
- `tests/claude-provider-context-bound.test.ts`: threshold predicate, full-context accounting, and the `resetSessionForFreshStart` effect reused by the guard.

`bun test` (provider + settings + telemetry + session-route suites) green; `tsc --noEmit` clean; `npm run build` succeeds.

### DIFF
--- a/src/npx-cli/commands/telemetry.ts
+++ b/src/npx-cli/commands/telemetry.ts
@@ -71,7 +71,7 @@ const COLLECTED_FIELDS = [
   'invalid_output_class   xml / idle / prose / poisoned (never the output)',
   'consecutive_invalid_outputs   unusable outputs in a row before recovery',
   'respawn_triggered      whether the compression agent was restarted',
-  'abort_reason     idle / shutdown / overflow / restart_guard / quota / poisoned / none',
+  'abort_reason     idle / shutdown / overflow / context_bound / restart_guard / quota / poisoned / none',
   'previous_shutdown      crash / clean / unknown (detected at worker start)',
   'previous_uptime_seconds / uptime_seconds',
   '                 worker uptime in whole seconds (previous run / at stop)',
--- a/src/services/telemetry/scrub.ts
+++ b/src/services/telemetry/scrub.ts
@@ -89,7 +89,8 @@ export const ALLOWED_PROPERTY_KEYS: Set<string> = new Set([
   // session_compressed trust signals — booleans, counters, and our own
   // closed enums (invalid_output_class: xml | idle | prose | poisoned, where
   // 'xml' means XML-shaped output that still failed to parse; abort_reason:
-  // idle | shutdown | overflow | restart_guard | quota | poisoned | none).
+  // idle | shutdown | overflow | context_bound | restart_guard | quota |
+  // poisoned | none).
   // Never model output, never raw abort strings.
   'fabrication_detected',
   'fabricated_count',
--- a/src/services/worker-types.ts
+++ b/src/services/worker-types.ts
@@ -39,7 +39,7 @@ export interface ActiveSession {
   lastSummaryStored?: boolean;
   pendingAgentId?: string | null;
   pendingAgentType?: string | null;
-  abortReason?: 'idle' | 'shutdown' | 'overflow' | 'context_bound' | 'restart-guard' | 'quota' | string | null;
+  abortReason?: 'idle' | 'shutdown' | 'overflow' | 'context-bound' | 'restart-guard' | 'quota' | string | null;
   respawnTimer?: ReturnType<typeof setTimeout>;
   /** When the latest compression prompt was dispatched to the model — telemetry compression_ms. */
   lastPromptSentAt?: number | null;
--- a/src/services/worker-types.ts
+++ b/src/services/worker-types.ts
@@ -39,7 +39,7 @@ export interface ActiveSession {
   lastSummaryStored?: boolean;
   pendingAgentId?: string | null;
   pendingAgentType?: string | null;
-  abortReason?: 'idle' | 'shutdown' | 'overflow' | 'restart-guard' | 'quota' | string | null;
+  abortReason?: 'idle' | 'shutdown' | 'overflow' | 'context_bound' | 'restart-guard' | 'quota' | string | null;
   respawnTimer?: ReturnType<typeof setTimeout>;
   /** When the latest compression prompt was dispatched to the model — telemetry compression_ms. */
   lastPromptSentAt?: number | null;
--- a/src/services/worker/ClaudeProvider.ts
+++ b/src/services/worker/ClaudeProvider.ts
@@ -202,6 +202,15 @@ export class ClaudeProvider {
 
     const settings = SettingsDefaultsManager.loadFromFile(USER_SETTINGS_PATH);
     const maxConcurrent = parseInt(settings.CLAUDE_MEM_MAX_CONCURRENT_AGENTS, 10) || 2;
+    // #2956 — proactive cumulative context bound. Unlike the OpenRouter/Gemini
+    // providers (which resend a sliced history each stateless turn), the Agent
+    // SDK owns the conversation internally and we resume it via memorySessionId,
+    // so session.conversationHistory cannot be truncated to bound context. The
+    // only lever is to start a fresh SDK session once the SDK-reported context
+    // grows too large. We read the real per-turn context size from usage and, if
+    // it exceeds this cap, reset to a fresh start *before* hitting the hard
+    // "Prompt is too long" wall (which otherwise saves zero memory).
+    const maxObserverTokens = parseInt(settings.CLAUDE_MEM_CLAUDE_MAX_TOKENS, 10) || 150000;
     await waitForSlot(maxConcurrent, session.abortController.signal);
 
     const isolatedEnv = sanitizeEnv(await buildIsolatedEnvWithFreshOAuth());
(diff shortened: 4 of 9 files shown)


dndungu:
Thanks for the review. Both points addressed in ce1b06b:

**1. Tests now exercise the real guard logic.** Extracted the guard's three decision units into exported functions on `ClaudeProvider` — `resolveObserverMaxTokens`, `computeFullContextTokens`, and `observerContextExceeded` — and the guard itself now calls them. The tests import these instead of reimplementing them, so a drift like `>` → `>=` or dropping `cache_creation_input_tokens` from the sum would now fail a test rather than pass silently. Added a cap-resolution suite too.

**2. Abort-reason naming aligned.** Renamed the internal value `'context_bound'` → `'context-bound'` to match the existing hyphenated convention (`'restart-guard'`). `normalizeAbortReason` maps it to the underscore telemetry enum `context_bound`, exactly as it does `restart-guard` → `restart_guard`, keeping the internal-hyphen / external-underscore boundary consistent.

`tsc --noEmit` clean; context-bound + settings + telemetry + session-route suites green.

dndungu:
Could I please get a review here?

thedotmack:

> step 3 (token budget as first-class object; proactive cap for the SDK path)

thedotmack:
Second pass — routing holds at **#3606 (plan-18, Observer Response Pipeline)**. Bounding the Agent SDK observer context by starting a fresh SDK session once the cumulative window is exceeded is step 3 of that master, and the reasoning about why `truncateHistory` cannot port to `ClaudeProvider` is correct. Explicitly not a duplicate of #3151, which bounds the resent history on the OpenAI-compatible path; these are the two halves of the same cap and both are wanted. Nothing on main supersedes it.

PR #2980, opened 2026-06-17
fix worker restart port cleanup on Windows

cccsaber:


thedotmack:
Closing as superseded by the newer Windows worker/port recovery work in #3009 and #3219. This older branch overlaps heavily with the current supervisor path.

Issue #2292, opened 2026-05-04
Hooks silently block every Claude Code event with `No stderr output` once `recordWorkerUnreachable` threshold trips


ruoye-ruolan:
## Symptom

Every Claude Code lifecycle event (UserPromptSubmit, PostToolUse, Stop, …) is blocked by the claude-mem hook with this opaque message:

```
blocking error from command "[…] hook claude-code <event>": No stderr output
```

The user has no idea why their session is broken.

## Root cause

Two compounding bugs:

### Bug A — invisible fail-loud diagnostic (this issue's PR fixes this)

`hookCommand()` in `src/cli/hook-command.ts` replaces `process.stderr.write` with a no-op for the duration of the hook to keep third-party stderr noise out of the hook output. `recordWorkerUnreachable()` in `src/shared/worker-utils.ts:347-352` writes its `claude-mem worker unreachable for N consecutive hooks` diagnostic via `process.stderr.write(...)` and then immediately calls `process.exit(BLOCKING_ERROR)`. `process.exit` runs **before** the `finally` block in `hookCommand()` restores the real stderr writer, so the diagnostic is always swallowed and Claude Code shows the empty-stderr blocking error.

### Bug B — zombie worker, no self-recovery (separate, recommended follow-up)

When the worker daemon's PID is alive but its HTTP port is unresponsive (zombie / hung event loop), `validateWorkerPidFile` reports "alive", `worker-spawner.ts:89` refuses to lazy-spawn a duplicate, and every hook fails forever. Failures accumulate silently in `~/.claude-mem/state/hook-failures.json` until the fail-loud threshold trips.

Reproduced locally on macOS, plugin v12.6.0:

```json
{"consecutiveFailures":642,"lastFailureAt":1777894510572}
```

…meaning **642 consecutive hooks** had been blocking the user's Claude Code sessions before they realized something was wrong.

Logs from this state:

```
Worker PID file points to a live process, skipping duplicate spawn
Live PID detected but worker did not become healthy before timeout — likely still starting
Worker port did not open after lazy-spawn within 3 attempts
```

## Reproduction

1. Let the worker daemon hang (or simulate via `kill -STOP <worker_pid>`).
2. Open a new Claude Code session in a project where claude-mem is installed.
3. After `CLAUDE_MEM_HOOK_FAIL_LOUD_THRESHOLD` consecutive hook failures (default 3), every hook returns blocking exit with `No stderr output`.

## Expected behavior

Either:
[...]

PR #2981, opened 2026-06-17
refactor(settings): route SettingsDefaults diagnostics through emitDiagnostic

darion-yaphet:
## Summary

`SettingsDefaultsManager` is a base-layer library module (runs under the worker daemon and hooks) that emitted its `[SETTINGS]` diagnostics via raw `console.warn`. Those lines can be swallowed by the Phase 2 hook stderr buffer (#2292).

Replace the 5 `console.warn` calls with `emitDiagnostic` (from `hook-io`):

- **Survives the Phase 2 hook stderr buffer (#2292)** — the same channel `logger` itself uses, instead of being swallowed by raw console.
- **Preserves the deliberate stderr-not-stdout behavior** that protects the machine-readable JSON contract on stdout (CLI commands like `start` promise JSON on stdout to the hook framework).

## Why `emitDiagnostic`, not `logger`

`logger` transitively depends on this module (`logger → shared/paths → SettingsDefaultsManager`). Importing `logger` here would **invert the layering and risk a circular dependency at bootstrap**. `hook-io` is a leaf module (type-only import), so there is no cycle.

## Scope note

This was scoped deliberately. Of 546 `console.*` calls in `src/`, the vast majority are correct and intentionally left as-is:

- `worker-service.ts` / `ServerBetaService.ts` — CLI entrypoints; `console.log(JSON.stringify(...))` is the program's stdout output contract. Routing to the file logger would silence the CLI.
- `npx-cli/**`, `*Installer.ts`, `bin/**` — user-facing CLI / install-time output.
- `ui/viewer/**` — browser React code; the Node file-based logger cannot run there.
- `utils/logger.ts`, `shared/hook-io.ts` — the logger's own stderr fallback.

## Verification

- `tsc --noEmit` clean
- `npm run build` green
- `bun test` — 39 pass / 0 fail (incl. `settings-defaults-manager.test.ts` and `logger-usage-standards.test.ts`)
- `[SETTINGS]` diagnostics confirmed still printing to stderr in test output

### DIFF
--- a/src/shared/SettingsDefaultsManager.ts
+++ b/src/shared/SettingsDefaultsManager.ts
@@ -3,6 +3,7 @@ import { readFileSync, writeFileSync, existsSync, mkdirSync } from 'fs';
 import { join, dirname } from 'path';
 import { homedir } from 'os';
 import { HOOK_TIMEOUTS, getTimeout } from './hook-constants.js';
+import { emitDiagnostic } from './hook-io.js';
 
 export interface SettingsDefaults {
   CLAUDE_MEM_MODEL: string;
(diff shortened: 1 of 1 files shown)


thedotmack:

PR #2990, opened 2026-06-17
fix: map content field correctly in server-beta observation payload builder

### LINKED ISSUE #2987: `observation_add` / `memory_add` fail with 400 in server-beta mode
## Before submitting

---

### Bug Description

`observation_add` and `memory_add` MCP tools always fail with a 400 ValidationError in server-beta mode. The `content` field is provided by the caller but arrives as `undefined` at the server.

```
Server beta POST /v1/memories returned 400:
{"error":"ValidationError","issues":[{"expected":"string","code":"invalid_type","path":["content"],"message":"Invalid input: expected string, received undefined"}]}
```

**Root cause:** In `mcp-server.cjs`, the `buildAddObservationPayload` method in the `ServerBetaClient` class maps the content value to the legacy field name `narrative` instead of `content`:

```javascript
buildAddObservationPayload(e) {
  let r = e.content,           // ← reads content correctly
      n = e.kind ?? "manual",
      ...
  return {
    projectId: e.projectId,
    kind: n,
    type: n,
    narrative: r,              // ← BUG: should be `content: r`
    ...
  }
}
```

The server-beta `POST /v1/memories` endpoint expects `content`, not `narrative`. The field `narrative` is the legacy worker-mode name and was not updated when the server-beta payload builder was written.

Read operations (`observation_search`, `get_observations`, `timeline`) are unaffected — they use a separate `buildSearchPayload` method that doesn't touch the content/narrative field.

### Steps to Reproduce
1. Configure claude-mem in server-beta mode (PostgreSQL + Valkey/BullMQ)
2. Call the `observation_add` MCP tool with a `content` parameter
3. Receive 400 ValidationError — `content` is `undefined`
4. Same result with `memory_add` (and its `narrative` legacy alias)

### Expected Behavior
The observation should be created and returned with the provided content.

### Environment
- **Claude-mem version**: 13.6.1 (latest on npm)
- **Claude Code version**: 2.1.179
- **OS**: Linux 6.12.90+deb13.1-amd64 x86_64
- **Platform**: Debian 13, server-beta mode with PostgreSQL + Valkey

### Logs
The 400 error is surfaced client-side in the MCP tool result, not in the worker log. Worker log shows no corresponding error for these calls — the request is rejected by the server-beta HTTP layer before reaching the queue.

```
Server beta POST /v1/memories returned 400:
{"error":"ValidationError","issues":[{"expected":"string","code":"invalid_type","path":["content"],"message":"Invalid input: expected string, received undefined"}]}
```

### Additional Context
- **Fix confirmed locally:** changing `narrative: r` to `content: r` in `buildAddObservationPayload` resolves the issue — write-read round-trip verified after patching and restarting the session.
- Only the write path is affected. All read/search tools work correctly.
- The `memory_add` compatibility alias constructs `content` from `t?.content ?? t?.narrative ?? ""` then feeds it into the same broken `buildAddObservationPayload`, so it fails identically.

arclabch:
## Summary

Fixes #2987.

`observation_add` and `memory_add` were failing with HTTP 400 `ValidationError: expected string, received undefined at path ["content"]` in server-beta mode.

In `buildAddObservationPayload` (`src/services/hooks/server-beta-client.ts`), the observation content was mapped to the legacy `narrative` field name. Server-beta's `CreateMemoryItemSchema` requires `content`, so the field arrived `undefined` and the request was rejected.

## Change

- Renamed the payload field `narrative: content` → `content` in `buildAddObservationPayload`.
- Updated the explanatory comment, which previously (and now inaccurately) described `narrative` as the FTS-indexed write field.

One file, minimal change. Read operations were unaffected and remain unchanged.

## Verification

- Fix matches the reporter's locally-verified change (`narrative` → `content`).
- Code review: confirmed the rename is correct, minimal, and that no other `narrative` references in `src/services/hooks/` belong to the add-observation write path (remaining occurrences are read-side / SQLite-schema / legacy-import concerns, independent of this fix).
- Security review: PASS — comment + key rename only; value is typed `string`, flows into a JSON HTTP body (no injection surface), no secrets introduced.

## Test plan

- [ ] In server-beta mode, call `observation_add` and `memory_add` and confirm they return 200 instead of 400.
- [ ] Confirm the persisted record's searchable text is populated.

### DIFF
--- a/src/services/hooks/server-beta-client.ts
+++ b/src/services/hooks/server-beta-client.ts
@@ -278,13 +278,12 @@ export class ServerBetaClient {
   buildAddObservationPayload(
     input: ServerBetaAddObservationRequest,
   ): Record<string, unknown> {
-    // Write-path contract (#2684): /v1/memories persists a `memory_items` row
-    // whose searchable text lives in `narrative` (the FTS trigger copies it
-    // into memory_items_fts). The MCP `observation_add` surface speaks in terms
-    // of `content`; map it onto `narrative` so the row is never empty and the
-    // FTS index always has something to match. `type` is REQUIRED by
-    // CreateMemoryItemSchema; default it from `kind` so a manual insert that
-    // only supplied content still persists instead of 400-ing.
+    // Write-path contract (#2684, #2987): /v1/memories validates against
+    // CreateMemoryItemSchema, whose searchable text field is `content`
+    // (received undefined -> 400 ValidationError). The MCP `observation_add`
+    // surface also speaks in terms of `content`, so pass it straight through.
+    // `type` is REQUIRED by the schema; default it from `kind` so a manual
+    // insert that only supplied content still persists instead of 400-ing.
     const content = input.content;
     const kind = input.kind ?? 'manual';
     const metadataTitle = typeof input.metadata?.title === 'string' ? input.metadata.title : undefined;
@@ -292,7 +291,7 @@ export class ServerBetaClient {
       projectId: input.projectId,
       kind,
       type: kind,
-      narrative: content,
+      content,
       ...(metadataTitle ? { title: metadataTitle } : {}),
       ...(input.serverSessionId !== undefined ? { serverSessionId: input.serverSessionId } : {}),
       ...(input.metadata !== undefined ? { metadata: input.metadata } : {}),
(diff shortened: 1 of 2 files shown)


arclabch:
Closing — project uninstalled, no longer pursuing these changes.

PR #3001, opened 2026-06-18
Add configurable OpenRouter reasoning effort support to disable unnecessary reasoning tokens

### LINKED ISSUE #2995: Feature request: support disabling reasoning for OpenRouter / OpenAI-compatible providers
## Before submitting

- [x] I searched existing issues (queries: `reasoning`, `reasoning_effort`, `empty response openrouter`) and confirmed this is not a duplicate.

---

When the OpenRouter provider is pointed at a reasoning model (e.g. GLM-4.5-Air, DeepSeek-R1), reasoning is on by default, which hurts an observation-extraction workload that doesn't need it:

1. **Slow processing** — single-session generation takes several minutes because every request spends time in the thinking phase.
2. **Empty responses** — intermittent `[ERROR] [SDK] Empty response from OpenRouter` / `Empty OpenRouter init response`. On a longer observation task the reasoning phase can consume the entire `max_tokens: 4096` budget, leaving zero tokens for the final answer in `choices[0].message.content`, so the worker treats the response as empty.

A probe against GLM-4.5-Air shows how lopsided this is — for a trivial reply it burns ~89 of 90 completion tokens on reasoning:

```
POST /chat/completions
  messages:[{role:user, content:"Think briefly, then reply with exactly: PONG"}]
  max_tokens: 2048
→ message { content:"PONG", reasoning_content:"We are being asked to think..." }
→ usage.completion_tokens = 90   (content ≈ 1 token, reasoning ≈ 89 tokens)
```

claude-mem only reads `choices[0].message.content`, so all reasoning tokens are discarded overhead.

**Root cause** — the OpenRouter request body is hard-coded with no reasoning control, and there is no settings key to influence it. `src/services/worker/OpenRouterProvider.ts` (671de5e3):

```ts
body: JSON.stringify({
  model,
  messages,
  temperature: 0.3,   // Lower temperature for structured extraction
  max_tokens: 4096,
}),
```

No `reasoning` / `reasoning_effort` / `max_completion_tokens` field is sent, so reasoning behavior is decided entirely by the model's default (on, for reasoning models).

Add a setting that maps to OpenRouter's unified `reasoning` parameter:

| Setting | Values | Default | Effect |
| --- | --- | --- | --- |
| `CLAUDE_MEM_OPENROUTER_REASONING_EFFORT` | `none` / `minimal` / `low` / `medium` / `high` | unset (model default) | When set, emits `reasoning: { effort }` in the request body. `none` disables reasoning. |

This mirrors how `CLAUDE_MEM_OPENROUTER_MAX_CONTEXT_MESSAGES` already makes a previously hard-coded behavior tunable (#2416 is a good precedent for this kind of "make the hard-coded configurable" change).

Ideally the field would be emitted only when the model advertises support for it (via OpenRouter's `supported_parameters`), to avoid HTTP 400 on models that reject `reasoning`.

- **Switch to a non-reasoning model** — works, but users have to discover on their own that claude-mem can't disable reasoning, and some provider plans (e.g. Zhipu GLM Coding Plan) ship primarily reasoning models.
- **Raise `max_tokens`** — doesn't address the wasted reasoning tokens or the latency, and the field is hard-coded today anyway.

**Additional context**

- Prior art: [claude-code-router PR #1222](https://github.com/musistudio/claude-code-router/pull/1222) ("sanitize reasoning parameter for openrouter") already does per-model reasoning handling — useful reference.
- OpenRouter's unified reasoning parameter: https://openrouter.ai/docs/guides/best-practices/reasoning-tokens (`reasoning: { effort: "none" }` disables reasoning).
- Related symptom already tracked in #2291 ("Skipped/empty observer responses can loop forever") — that fix addressed the loop, not this root cause.
- Confirmed in v13.6.2 (bundled `worker-service.cjs`) and current `main` `OpenRouterProvider.ts`.
- Environment: `CLAUDE_MEM_PROVIDER=openrouter` via `CLAUDE_MEM_OPENROUTER_BASE_URL` → Zhipu GLM Coding Plan, model `glm-4.5-air` (`reasoning: true`, context 131072, cost $0).

Iyanuoluwa007:
Related Issue

Closes #2995
https://github.com/thedotmack/claude-mem/issues/2995

Problem

Currently, when claude-mem uses the OpenRouter provider with reasoning models (for example GLM-4.5-Air or DeepSeek-R1), reasoning is enabled by default because the OpenRouter request does not expose any reasoning configuration.

The current implementation in:

src/services/worker/OpenRouterProvider.ts

sends requests with:

{
  model,
  messages,
  temperature: 0.3,
  max_tokens: 4096
}

but does not provide any way to control OpenRouter reasoning behavior.

This causes:

Increased latency because requests spend time generating reasoning tokens that are not required for observation extraction.
Empty responses when reasoning tokens consume the available completion budget before the final response content is generated.
Unnecessary token usage because claude-mem only consumes message.content and does not use the reasoning output.
Solution

Added configurable OpenRouter reasoning effort support through:

CLAUDE_MEM_OPENROUTER_REASONING_EFFORT

Supported values:

none
minimal
low
medium
high

Example:

CLAUDE_MEM_OPENROUTER_REASONING_EFFORT=none

When configured, the OpenRouter request includes:

{
  "reasoning": {
    "effort": "none"
  }
}

This allows users to disable reasoning for workloads where only the final structured output is needed.

Implementation

Changes include:

Added OpenRouter reasoning effort configuration support.
Added request payload handling for the reasoning parameter.
Preserved existing behavior when the setting is not configured.
Added tests covering default behavior and configured reasoning values.
Testing

All related tests passed successfully.

Ran:

npx tsc --noEmit

Result:

✅ Passed

Ran OpenRouter test:

bun test ./tests/shared/openrouter-base-url.test.ts

Result:

10 pass
0 fail
12 expect() calls
Ran 10 tests across 1 file

The implementation also keeps existing OpenRouter behavior unchanged when no reasoning configuration is provided.

Summary

This change makes OpenRouter usage more efficient for reasoning models by allowing users to disable or control reasoning tokens, reducing unnecessary latency and preventing empty responses caused by reasoning consuming the completion budget.

### DIFF
--- a/src/services/worker/OpenRouterProvider.ts
+++ b/src/services/worker/OpenRouterProvider.ts
@@ -513,6 +513,7 @@ export class OpenRouterProvider {
     const messages = this.conversationToOpenAIMessages(truncatedHistory);
     const totalChars = truncatedHistory.reduce((sum, m) => sum + m.content.length, 0);
     const estimatedTokens = this.estimateTokens(truncatedHistory.map(m => m.content).join(''));
+    const reasoningEffort = getReasoningEffort();
 
     logger.debug('SDK', `Querying OpenRouter multi-turn (${model})`, {
       turns: truncatedHistory.length,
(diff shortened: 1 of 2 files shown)


thedotmack:

> step 4 (`reasoning:{effort}` only when advertised); P1: default 'none' injects reasoning into every request

thedotmack:
Second pass — routing holds at **#3606 (plan-18, Observer Response Pipeline)**, step 4 (provider config surface). Flagging that #3040 covers the same motivating case — disabling reasoning so the completion budget is not eaten before content is emitted — via a generic `CLAUDE_MEM_OPENROUTER_EXTRA_BODY`, with no open P1 against it. This PR still carries the P1 that a default of `none` injects a `reasoning` object into every request including models that never advertise it. Leaving open since a typed knob is a defensible alternative, but it needs that default fixed and a decision against #3040 before it can land.

PR #3002, opened 2026-06-19
feat(sqlite): default fresh databases to incremental auto_vacuum (refs #2793)

laihenyi:
## Summary

Every claude-mem database is created with SQLite's default **`auto_vacuum=NONE`** — none of the three primary connection paths set it:

- `ClaudeMemDatabase` (`src/services/sqlite/Database.ts`)
- the sqlite `DatabaseManager` (`src/services/sqlite/Database.ts`)
- the worker `DatabaseManager` (`src/services/worker/DatabaseManager.ts`)

As #2793 documents, deleted rows then leave their pages on the freelist and the file never shrinks.

This PR adds `enableIncrementalAutoVacuumIfFresh()` and calls it at all three open sites, switching **brand-new** databases to `INCREMENTAL` before WAL/schema creation. Freed pages then become reclaimable via the cheap `PRAGMA incremental_vacuum` instead of only a full, exclusive-lock `VACUUM`.

## Why gated on a fresh database

auto_vacuum can only switch away from `NONE` on an empty database (otherwise a full `VACUUM` is required), so the helper gates on an empty `sqlite_master`. **Existing `NONE` databases are deliberately left untouched** — flipping the pragma without a full VACUUM makes `PRAGMA auto_vacuum` *report* `2` while `incremental_vacuum` is a silent no-op, which would strand their existing free pages and break the reclaim path. Leaving them at `NONE` keeps a full `VACUUM` as the correct reclaim for legacy databases.

## Relationship to in-flight work

- **#2904** (reclaim free pages after cleanup): its `maintenance.ts` branches `autoVacuumMode === 2 ? incremental_vacuum : VACUUM`. Without this PR nothing ever sets mode `2`, so on real installs that branch can only ever take the full-VACUUM path. With this PR, fresh installs get the cheap incremental path.
- **#2849** (`busy_timeout`): the full-VACUUM path needs an exclusive lock. On a 100 GB+ database that window collides with hook-spawned workers (`database is locked`), which is exactly what `busy_timeout` mitigates.

This PR touches only the connection-open sites, so it is complementary to both (it may need a trivial rebase depending on merge order, since #2849 also edits these pragma blocks).

## Empirical motivation

On a long-running v13.6.1 install: live DB `131 GB`, `page_count=32,076,486`, **`freelist_count=28,911,332` (90.1% free pages)** for ~12.5k observations. A manual `VACUUM` reclaimed it to **14.5 GB** (`integrity_check ok`, identical row counts). The freed pages came from a bulk cleanup (`CleanupV12_4_3` deletes) with no following VACUUM and `auto_vacuum=NONE`.

## Testing

- New `tests/services/sqlite/auto-vacuum.test.ts`:
  - enables `INCREMENTAL` on a fresh DB and verifies it persists through WAL + table creation
  - leaves an existing `NONE` DB untouched (guards the no-op-incremental_vacuum trap)
  - verifies `incremental_vacuum` actually returns free pages to the OS once enabled
- `npm run typecheck` and the `tests/services/sqlite/` suite pass.

### DIFF
--- a/src/services/sqlite/SessionSearch.ts
+++ b/src/services/sqlite/SessionSearch.ts
@@ -14,6 +14,7 @@ import {
   ObservationRow,
   UserPromptRow
 } from './types.js';
+import { enableIncrementalAutoVacuumIfFresh } from './autoVacuum.js';
 
 export class SessionSearch {
   private db: Database;
@@ -26,6 +27,9 @@ export class SessionSearch {
     } else {
       ensureDir(DATA_DIR);
       this.db = new Database(dbPathOrDb);
+      // Must precede journal_mode = WAL: auto_vacuum can only be switched on an
+      // empty database, before the first WAL-mode write locks the mode in.
+      enableIncrementalAutoVacuumIfFresh(this.db);
       this.db.run('PRAGMA journal_mode = WAL');
     }
 
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -17,6 +17,7 @@ import { parseFileList } from './observations/files.js';
 import { DEFAULT_PLATFORM_SOURCE, normalizePlatformSource, sortPlatformSources } from '../../shared/platform-source.js';
 import { findRecentDuplicateUserPrompt as findRecentDuplicateUserPromptRecord } from './prompts/get.js';
 import { normalizeStoredPromptText } from './prompt-storage.js';
+import { enableIncrementalAutoVacuumIfFresh } from './autoVacuum.js';
 
 function resolveCreateSessionArgs(
   customTitle?: string,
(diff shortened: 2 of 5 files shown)


laihenyi:
Thanks @greptile-apps — both findings were spot-on and are addressed in `8ec69ccd`:

1. **Freshness gate (`Database.ts`)** — now requires `page_count <= 1` in addition to an empty `sqlite_master`, so a previously-committed DB whose tables were all dropped is no longer mistaken for fresh (which would have reproduced the exact report-`2`-but-no-op trap). Added a regression test covering that dropped-all-tables case.
2. **Test checkpoint** — switched `wal_checkpoint(PASSIVE)` → `TRUNCATE` so every WAL frame is flushed before the `freelist_count` assertion, removing the non-deterministic failure mode.

`npm run typecheck` and the `auto-vacuum` suite (4 tests) pass.

laihenyi:
Rebased onto current `main` and reauthored (force-pushed `55e52418`). Heads-up since the diff moved:

`main` removed `src/services/sqlite/Database.ts` in the Ponytail audit (#3021), which is where this logic previously lived — hence the modify/delete conflict. The fix is unchanged in behavior; it now lives in a dedicated **`src/services/sqlite/autoVacuum.ts`** module and is wired into the three open paths in the new structure:

- `SessionStore` and `SessionSearch` constructors — before `journal_mode = WAL` (path branch only).
- `DatabaseManager.initialize()` — right after `new Database(DB_PATH)`, before `new SessionStore(this.db)`. This is the one that matters for the worker: it opens the shared connection and hands the instance to SessionStore/SessionSearch, which then take the `instanceof Database` branch and skip their own pragma setup — so a fresh worker DB would otherwise never get opted in.

The freshness gate (`tableCount === 0 && page_count <= 1`) and the `TRUNCATE`-checkpoint test from the previous review round are carried over verbatim. `npm run typecheck` is clean; the auto-vacuum suite (4 tests) and the broader sqlite + worker suites (246 tests) pass locally.

laihenyi:
Closing — `main` now sets `auto_vacuum = INCREMENTAL` on fresh databases via the centralized `applySqliteConnectionPragmas` in `src/services/sqlite/connection.ts`, which supersedes this PR. Thanks for tackling the #2793 bloat root cause on your side! 🙏

PR #3005, opened 2026-06-19
Add support for projectName override via .claude-mem.json

timoxd7:
## Info
Project names are derived from the git repo root (or cwd basename), which splits independent copies/clones of the same project into separate memories just because their folders are named differently.

## How?
Add an explicit override: a `.claude-mem.json` with `{ "projectName": "..." }` (read by walking up from cwd to the home dir) pins the project name regardless of folder name. It takes precedence over the git-root/basename derivation and collapses worktree composites to the single configured name, so sibling copies can share one memory. Accepts `project_name` as a snake_case alias; empty or non-string values are ignored.


thedotmack:

PR #3014, opened 2026-06-20
fix(opencode): full OpenCode v1.17.8 plugin compatibility

DRCOMPUTER60290:
- Add required `id: "claude-mem"` field to plugin default export (OpenCode v1.17.8 rejects plugins without an explicit id)
- Fix API signature mismatch: `args` moved from output to input in `tool.execute.after`; `sessionID` moved from output.message to input in `chat.message`
- Fix port resolution: read ~/.claude-mem/settings.json directly so the OpenCode process (which has no inherited env vars) uses the same port as the worker
- Pass first user message as session prompt so the viewer shows the actual text instead of "[media prompt]"
- Add `registerOpenCodeMCPServer()` to auto-register the MCP server in opencode.json during install
- Rebuild all plugin bundles


thedotmack:
Closing as superseded by #3208, which targets the current OpenCode lifecycle and adds context injection, deterministic capture, retries, and substantially stronger contract coverage.

PR #2829, opened 2026-06-08
feat(memory): zero-spend ingest of Claude Code auto-memory into observations

surfingdoggo:
## What

Adds `claude-mem memory ingest`, which imports Claude Code's own "auto-memory"
markdown (`~/.claude/projects/<encoded-cwd>/memory/*.md`) directly into the
memory database as observations.

## Why

Auto-memory is *already distilled* prose, the same kind of artifact the
observation generator produces. So this stores each topic file directly through
the existing observation seam (content-hash dedup + Chroma sync) and **does not
run the Haiku generation pipeline**. Re-generating over already-distilled prose
would be lossy and pay for negative value. The `MEMORY.md` link index is skipped.

This is the mechanical, **zero-model-spend** sibling to the transcript backfill
(#2690): transcripts need generation, distilled memory does not.

## How

- `src/services/memory/ingest.ts`: enumerate + parse (hand-rolled frontmatter
  parser, no YAML dep) + content-hash store orchestrator. Dry-run is DB-free and
  worker-free.
- `src/services/memory/cli.ts` + npx wiring: `memory ingest [--source <dir> |
  --all] [--dry-run] [--require-cwd]`. Dry-run runs client-side; the real store
  drives the worker over HTTP.
- `src/services/worker/http/routes/MemoryIngestRoutes.ts`: `POST /api/memory/ingest`.
- Tests: frontmatter parsing, scan, mapping, ingest orchestrator (13 cases).
- Docs: `usage/memory-ingest.mdx` + nav entry.

All changes are additive; no existing behavior changes.

## Testing

- `bun test tests/memory/memory-ingest.test.ts` -> 13 pass
- `tsc --noEmit` clean

We have been running this in our own claude-mem deployment (worker/SQLite) since
Thursday 4 June. We used it to import an existing project's accumulated memory
files, which made the transition painless: we did not have to start memory from
scratch. It was quick and easy.

Note: `main` currently has pre-existing unrelated red CI.

PR #3027, opened 2026-06-21
Disable Claude Code native auto-memory by default + memory transfer

thedotmack:
## Why

Claude Code's native auto-memory **conflicts with claude-mem** — it creates shadow state outside the user's control and competes with claude-mem for context-window tokens. The installer's default was backwards: it *kept native memory enabled* and even labeled that "(Recommended)". Worse, non-interactive/CI installs **silently left it on**.

This flips the default and pairs it with a transfer so users don't lose notes they've already accumulated.

### 1. Disable native auto-memory by default (the fix)
- `resolveClaudeAutoMemoryChoice`: interactive default is now **Disable** ("(Recommended)"), and the **non-interactive/CI path now disables too** (was the silent bug).
- New `--keep-auto-memory` opt-out (works interactively and in CI); `--disable-auto-memory` kept for symmetry.
- Updated prompt/help/summary copy. The actual toggle (`CLAUDE_CODE_DISABLE_AUTO_MEMORY=1`) is unchanged.

### 2. `npx claude-mem migrate-memory` (the transfer)
Imports existing native auto-memory (`~/.claude/projects/<encoded-cwd>/memory/*.md`) into claude-mem, **per project**.

**Mechanism — no new compression API.** Each note is POSTed to the existing observation-ingest endpoint (`POST /api/sessions/observations`) — the *same endpoint the PostToolUse hook uses* — so claude-mem's normal generator compresses it into a real observation. Works in default worker/SQLite mode (no server-beta, no Postgres, no extra API key). This path was traced and **adversarially verified** against source before building.

- **Per-project routing:** attribution comes only from the payload `cwd`, so the command recovers each project's real path — transcript `cwd` (exact) → filesystem probe (resolves the `claude-mem` dash ambiguity) → naive fallback — and passes it.
- **Dates:** notes are dated from the file's `mtime` (frontmatter has no date fields) via a new **optional** `timestampEpoch` threaded through the ingest path into the `overrideTimestampEpoch` plumbing `storeObservations` already had. **Strictly opt-in** — the live hook path is unchanged (still stamps `now()`).
- **Non-destructive:** reads only; never modifies or deletes native files. Skips the redundant `MEMORY.md` index files. Deterministic per-project session id so re-runs dedupe instead of duplicating.

## Verification
- `tsc --noEmit` clean
- 377 tests pass (376 existing + 1 new buffer back-dating test), 0 fail
- `npm run build` passes (bundles regenerated)
- Dry-run on real machine data correctly discovered 2 projects / 3 notes and decoded the dash-containing `claude-mem` path:
  ```
  claude-mem (/Users/alexnewman/Scripts/claude-mem)
    • feedback_stop_asking_just_do.md
    • feedback_dont_verify_auth.md
  fireside-chat (/Users/alexnewman/Scripts/fireside-chat)
    • project_fireside_clips.md
  ```

## Remaining step
The live migration run (`build-and-sync` to deploy the worker change, then `migrate-memory`) writes into the local claude-mem DB and restarts the worker, so it's left as an explicit post-merge action rather than done in-PR.


thedotmack:
## Verified end-to-end against a live worker (commit `11d54d85b`)

I ran the actual migration against a running worker and fixed every gap that static checks missed. Confirmed in the SQLite DB:

- **Notes compress into correct observations** — e.g. `feedback_stop_asking_just_do.md` → "User Preference: Stop Asking, Just Do It"; `feedback_dont_verify_auth.md` → "User Preference: Do Not Verify Auth".
- **Attributed to the right project** (`claude-mem`, `fireside-chat`) — not the cwd the command runs in.
- **Dated by the source file's mtime** (2026-03-13, 2026-03-15), not import time.
- **Archive only when confirmed stored** — when one note couldn't be confirmed in time, the project's native files were correctly left in place (no data loss).

### Bugs found only by running it (all fixed)
1. **Session init required** — posting to a fresh session returned idle responses and stored nothing. Now calls `/api/sessions/init` first, like the live hook.
2. **Wrong worker port** — used `SettingsDefaultsManager.get()` (compiled default) instead of the configured port; fixed to read the settings file.
3. **Dates didn't stick** — the enqueue-time timestamp didn't survive the agent's idle turns. Replaced with an authoritative post-store backdate (`POST /api/sessions/backdate`).
4. **False-positive archive** — a drained-but-empty queue looked like success. `GET /api/sessions/status` now returns `storedObservations`; archiving waits for confirmed storage.
5. **Stale install tests** — updated the two tests that encoded the old opt-in default.

Full suite: **2154 pass, 0 fail**. Build clean. Test machine restored to a clean state afterward (the live run was done during an active session, which is why one note timed out under queue backlog — a standalone run isn't affected).

thedotmack:
Routing this into **#2785 (plan-12, Provider & Extensibility Roadmap)**: the installer default flip plus `npx claude-mem migrate-memory` is an ingestion/product capability, which is what the roadmap master collects. One overlap to flag — #2829 also imports `~/.claude/projects/<encoded-cwd>/memory/*.md` into observations, but mechanically and with zero model spend, where this PR pushes each note through the observation-ingest endpoint and compresses it. Same source data, two different bets on whether distilled prose should be re-generated; they should be reconciled on the master rather than both landing. Leaving both open.

PR #3028, opened 2026-06-22
fix: ignore unparseable observer output instead of poisoned respawn

### LINKED ISSUE #3022: Observer respawns ~1/3 of Haiku sessions on a keyword false-positive ("poisoned") — it should just ignore unparseable output
## Principle being violated

The observer parser should **ignore any output it can't parse** — that's the whole job. It must not inspect the *content* of unparseable output and act on it.

Today it does: when the SDK returns non-XML, a substring classifier labels some of it **`poisoned`** and **kills + respawns the SDK session**. "Poisoned" is a case-insensitive `includes()` against 11 hardcoded English phrases (`session exhausted`, `context window`, `prompt is too long`, `this session has ended`, …) — `src/sdk/output-classifier.ts:20-32`. This is a brittle keyword match standing in for "ignore," and it over-fires badly.

## Evidence it's broken (live telemetry, 30d)

`poisoned` aborts as a share of all compression turns, by model:

| Model | poisoned % of turns | context |
|---|---|---|
| haiku (4.5) | **41.1%** | 200K |
| claude-haiku-4-5-20251001 | **33.9%** (9.0M turns) | 200K |
| claude-sonnet-4-5 | 17.7% | 200K–1M |
| claude-sonnet-4-6 | 9.6% | 200K–1M |
| claude-opus-4-7 | 5.9% | 200K |
| gemini-2.5-flash-lite | **0.1%** | 1M |
| openai/gpt-oss-120b:free | **0.1%** | — |
| gemini-3-flash-preview / xiaomi mimo-flash:free | **~0%** | small |

Total ≈ **13.2M poisoned aborts / 30d ≈ 22% of all turns.**

**This is the inverse of real context exhaustion.** If "poisoned" meant running out of room, the *small-context* models would top the list. Instead they're ~0% and the *large-context Claude* models are 6–41% — the classifier is matching Claude's closure-phrase wording (e.g. "I cannot continue this session"), not session health. Net effect: claude-mem kills + respawns ~1/3 of Haiku sessions on a false positive, throwing away work and re-spending tokens.

(The word "poisoned" is also wrong — it implies tainted/corrupted input, not a wedged session. The concept should go.)

## The fix — "parse XML or drop it," nothing content-based

The correct behavior **already exists** at `ResponseProcessor.ts:115-120` ("Plain-text skip responses are intentionally ignored" → drops the batch, returns). The violation is the respawn trigger at `ResponseProcessor.ts:75-77`:

```js
const mustRespawn =
  outputClass === 'poisoned' ||                                  // ← remove
  session.consecutiveInvalidOutputs >= INVALID_OUTPUT_RESPAWN_THRESHOLD; // ← see decision
```

### Recommended: pure ignore (remove all content-driven respawn)
1. `src/sdk/output-classifier.ts` — delete `POISONED_MARKERS` and the `poisoned` class; keep `previewOutput()` for logging only (never behavior).
2. `src/services/worker/agents/ResponseProcessor.ts` — remove the `mustRespawn`/respawn block (`:71-113`); unparseable always falls through to the existing drop path. Keep the warn-log so drops stay visible.
3. `src/services/worker/SessionManager.ts` — remove `respawnPoisonedSession` (`:252`) / `abortReason='poisoned'` (`:273`) if dead.
4. `src/services/worker/http/routes/SessionRoutes.ts:37,44`, `src/services/telemetry/scrub.ts:90-99`, `src/npx-cli/commands/telemetry.ts:71,74` — drop the `poisoned` enum value from mappings/docs.
5. Tests — remove poison-classification + immediate-respawn tests; add one asserting unparseable output (even containing "context window") is **dropped, not respawned**.

**Loop-safety:** the drop path calls `confirmClaimedMessages` — the batch is not re-queued, so a stuck session can't loop "until quota exhausted." The 3-strikes respawn guards a loop the drop path already prevents.

### Fallback (if a safety net is wanted)
Remove only the `outputClass === 'poisoned' ||` clause + `POISONED_MARKERS`; keep the content-agnostic `consecutiveInvalidOutputs >= 3` structural respawn. Kills the false positives, retains a structural circuit-breaker. (Owner leans pure-ignore.)

## Verification
- `bun test` (output-classifier + ResponseProcessor + telemetry) green; `tsc --noEmit` clean.
- `grep -rin "poison" src` returns nothing behavioral.
- Manual: feed non-XML containing "context window" → asserted ignored/dropped, session NOT respawned.
- Post-ship telemetry: `abort_reason='poisoned'` / `invalid_output_class='poisoned'` volume → 0 on updated installs; Haiku throughput rises.

Iyanuoluwa007:
Fixes #3022

Issue:
https://github.com/thedotmack/claude-mem/issues/3022

## Summary

Remove content-based "poisoned" output detection from the observer pipeline and make unparseable SDK output safely ignored instead of triggering session respawns.

## Problem

The observer parser should ignore output it cannot parse. Previously, non-XML SDK responses were inspected for hardcoded phrases like:

- "session exhausted"
- "context window"
- "prompt is too long"
- "this session has ended"

These keyword checks caused false positives because normal SDK responses could contain similar wording. This resulted in unnecessary session kills, respawns, wasted tokens, and lost session progress.

## Changes

- Removed `poisoned` output classification and keyword matching
- Changed unparseable observer output handling to drop/ignore instead of respawn
- Removed content-based session recovery behavior
- Kept structural recovery using consecutive invalid output threshold
- Renamed abort reason handling from `poisoned` to `invalid_output`
- Updated telemetry mappings and documentation
- Restored commit hash verification documentation comment
- Updated tests to ensure closure-like messages are treated as normal prose

## Testing

Passed:

```bash
bun test tests/sdk/output-classifier.test.ts
bun run typecheck


thedotmack:
Thanks for this 🙏 v13.9.1 ships an equivalent fix — the observer now drops unparseable/benign output instead of doing a poisoned respawn (commit ad4bd6f7a, https://github.com/thedotmack/claude-mem/releases/tag/v13.9.1). Closing as superseded by the released fix; please reopen if it still reproduces on 13.9.1+.

PR #3033, opened 2026-06-22
fix(windows): strip UTF-8 BOM in all settings.json readers (#3013)

### LINKED ISSUE #3013: Windows: settings.json written with UTF-8 BOM causes JSON parse failure on every hook
## Bug Description

On Windows with PowerShell 5.1, `settings.json` is written with a **UTF-8 BOM** (`EF BB BF`). Node.js's JSON parser treats the BOM byte as an unrecognized token, causing the worker to fail loading its config on every startup.

## Error Message

```
[LOGGER] Failed to load log level from settings: JSON Parse error: Unrecognized token ''
claude-mem worker unreachable for N consecutive hooks.
```

The empty string in `Unrecognized token ''` is the invisible BOM character (`﻿`).

## Root Cause

PowerShell 5.1 (Windows PowerShell, not PowerShell Core) defaults to UTF-8 with BOM for certain write operations. When `settings.json` is written by PowerShell-based tooling, the BOM is prepended.

Confirmed via:
```powershell
$bytes = [System.IO.File]::ReadAllBytes("$env:USERPROFILE\.claude-mem\settings.json")
# Result: first 3 bytes = 0xEF 0xBB 0xBF  (UTF-8 BOM)
# File content is structurally valid JSON — BOM prefix breaks the parser
```

## Impact

- Worker fails to start on **every** Claude Code session restart
- Each hook call waits for timeout before failing → massive latency penalty on all tool calls (Glob, Read, Grep, Agent, etc.)
- After 24 consecutive failures, `CAPTURE_BROKEN` flag is set and worker is **permanently disabled** until manual reset
- Only workaround: manually change port + reset `hook-failures.json` + delete `CAPTURE_BROKEN`
- Long-running Agent tasks accumulate the full timeout overhead on every single tool call

## Suggested Fixes

**Option 1 — strip BOM on read (minimal change):**
```javascript
const raw = fs.readFileSync(settingsPath, 'utf8').replace(/^﻿/, '');
const settings = JSON.parse(raw);
```

**Option 2 — enforce utf8 without BOM on write (preferred):**
```javascript
fs.writeFileSync(settingsPath, JSON.stringify(settings, null, 2), { encoding: 'utf8' });
// Node.js fs.writeFileSync with 'utf8' never writes a BOM
```

**Option 3 — self-healing recovery (bonus):**
Instead of setting `CAPTURE_BROKEN` permanently after N failures, auto-reset `consecutiveFailures` so the worker retries on next session start.

## Environment

- Shell: PowerShell 5.1 (Windows PowerShell)
- Node.js: invoked via `bun-runner.js` → `worker-service.cjs`
- Claude Code: latest

## Related

On Windows, TCP sockets enter TIME_WAIT state (default 240s) after the worker process exits abnormally. This compounds the BOM issue: even if the JSON bug were fixed, the port remains unavailable for ~4 minutes after a crash. Workaround: set `TcpTimedWaitDelay=30` in the Windows registry to reduce TIME_WAIT to 30s.

Neclode:
Fixes #3013.

**Problem.** On Windows, external tooling (PowerShell 5.1, some editors/formatters) rewrites `~/.claude-mem/settings.json` with a UTF-8 BOM (`EF BB BF`). Node/Bun decode it to `U+FEFF` at offset 0 on `utf-8` read, and `JSON.parse` rejects it — so the worker fails to load config on **every** hook, eventually tripping `CAPTURE_BROKEN` and disabling capture entirely.

**Root cause.** BOM stripping was added to exactly **one** of seven settings readers (`SettingsDefaultsManager`). The other six still parsed raw and threw — including `logger.getLevel()`, which is the source of the `Failed to load log level from settings` error in the report.

**Why read-side, not write-side.** Our own writes use `writeFileSync(..., 'utf-8')`, which never emits a BOM — so a write-side fix is a no-op and wouldn't repair the already-BOM'd files sitting on users' disks. The fix has to be read-side resilience.

**Changes**
- Add shared `stripBom()` to `utils/json-utils` and route `readJsonSafe` through it
- Strip BOM in the six remaining `settings.json` readers (`logger`, `paths`, `plugin-state`, npx `server`, `server-beta-bootstrap`, `SettingsRoutes`)
- Collapse `SettingsDefaultsManager`'s inline BOM regex onto the shared helper
- Drop a dead `logger` import so `json-utils` stays an import-cycle-free leaf
- Test: `readJsonSafe` parses a BOM-prefixed `settings.json`

**Verification:** `bun test` 39/39 pass · `tsc --noEmit` clean.

### DIFF
--- a/src/npx-cli/commands/server.ts
+++ b/src/npx-cli/commands/server.ts
@@ -11,6 +11,7 @@ import {
   runStatusCommand,
   runStopCommand,
 } from './runtime.js';
+import { stripBom } from '../../utils/json-utils.js';
 
 const UNSUPPORTED_SERVER_COMMANDS = new Set([
   'logs',
@@ -148,7 +149,7 @@ async function runServerBetaKeysRotateCommand(): Promise<void> {
   let previousApiKeyId: string | null = null;
   if (existsSync(settingsPath)) {
     try {
-      const raw = JSON.parse(readFileSync(settingsPath, 'utf-8')) as Record<string, unknown>;
+      const raw = JSON.parse(stripBom(readFileSync(settingsPath, 'utf-8'))) as Record<string, unknown>;
       const flat = (raw.env && typeof raw.env === 'object' ? raw.env : raw) as Record<string, unknown>;
       const previousKey = flat.CLAUDE_MEM_SERVER_BETA_API_KEY;
       if (typeof previousKey === 'string' && previousKey.length > 0) {
--- a/src/services/hooks/server-beta-bootstrap.ts
+++ b/src/services/hooks/server-beta-bootstrap.ts
@@ -23,6 +23,7 @@
 import { createHash, randomBytes } from 'crypto';
 import { chmodSync, existsSync, readFileSync, writeFileSync, mkdirSync } from 'fs';
 import { dirname } from 'path';
+import { stripBom } from '../../utils/json-utils.js';
 import { createPostgresPool, type PostgresPool } from '../../storage/postgres/pool.js';
 import { parsePostgresConfig } from '../../storage/postgres/config.js';
 import { PostgresAuthRepository } from '../../storage/postgres/auth.js';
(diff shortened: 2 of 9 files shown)


thedotmack:
Closing as already implemented on current main: settings readers now use the shared BOM-safe atomic JSON parser.

PR #3039, opened 2026-06-24
fix: prevent a broken/partial dependency install from bricking the worker

rajivsinclair:
## Summary

Hardens worker startup so a **partially-failed dependency install can no longer brick the plugin**. This closes a failure mode that took a real install fully down: every tool call blocked behind a worker that could not start.

Two complementary layers:
1. **Install-time** — required pure-JS deps (`zod`, `shell-quote`) can no longer be taken out by an unrelated native-build failure.
2. **Runtime** — the worker supervisor refuses to kill a healthy worker (or burn the cold-boot budget) to start one whose dependencies can't resolve, and reports *why*.

---

## Background: the failure this fixes

Observed on a real machine after the plugin auto-updated:

1. The plugin's dependency install **aborted on a failed native `tree-sitter` build** (node-gyp; e.g. the grammars don't compile against newer Node whose v8 headers break the build). The install died partway, so the required pure-JS dep **`zod` was never installed**.
2. The bundled worker imports `zod` subpaths (`zod/v3`, `zod/v4`, `zod/v4-mini`, pulled in transitively by the MCP / agent SDKs), so it **crashed on startup with `Cannot find module 'zod/v3'`** and never bound its port.
3. The version-mismatch recycler then **killed the healthy previous-version worker** to "upgrade" into the broken one — turning a partial install into a total outage.
4. After 3 consecutive unreachable hooks, the fail-loud guard began **blocking every tool call** (`worker unreachable for N consecutive hooks`).

The root problems: one optional native build could abort the whole install, and the supervisor would recycle/spawn into a worker that cannot run.

---

### 1. Install resilience — `scripts/build-hooks.js`
Split the generated `plugin/package.json` so the **required** pure-JS deps (`zod`, `shell-quote`) stay in `dependencies`, while the **native `tree-sitter` grammars move to `optionalDependencies`**. npm/bun treat a failed optional-dependency build as non-fatal, so a grammar that can't compile on a given platform/Node no longer aborts the install before the hard deps land — it simply disables that language's parsing. (The worker already degrades gracefully when a grammar is absent.)

### 2. Runtime preflight — `src/shared/worker-utils.ts`
New `checkInstalledWorkerDependencies()`: a spawn-free check that the worker's hard runtime deps actually **resolve** before we act on them. It uses real `require.resolve` (honoring each package's `exports` map) over the full critical set — every `zod` subpath the bundle needs, plus `shell-quote` — anchored at the worker script with candidate `node_modules` paths walked up the tree (so it's correct whether deps live at `<marketplace>/node_modules`, `<plugin>/node_modules`, or a hoisted dev root). Wired into `ensureWorkerRunning()`:
- **Recycle path:** never kill a *healthy* worker to swap in a target that can't boot. If the install is broken, keep serving the current (version-mismatched) worker and log remediation; the next hook retries once deps are fixed.
- **Lazy-spawn path:** fail fast with the real cause instead of a silent ~15 s cold-boot wait followed by a generic "unreachable".

### 3. Single source of truth — `src/shared/zod-required-subpaths.ts` (new) + `src/npx-cli/install/setup-runtime.ts`
Extracted the critical `zod` subpath set into `ZOD_REQUIRED_SUBPATHS`, now imported by **both** the install-time post-check (`verifyCriticalModules`) and the runtime preflight, so the two guards can never drift.

---

## Why this approach
- **Defense in depth.** `optionalDependencies` is best-effort install hygiene; it relies on package-manager behavior. The runtime preflight is the *hard* guarantee that a partial/stale install can never silently brick the worker — and when something is wrong, the operator gets an actionable message instead of a wedged session.
- **Resolve, don't stat.** A present `zod/` directory does not imply its subpath exports resolve. Mirroring the installer's `require.resolve` approach catches exactly the partial-install case that caused the incident.
- **No new divergence.** Sharing `ZOD_REQUIRED_SUBPATHS` keeps the install-time and runtime checks identical by construction.

## Verification
- `bun install --frozen-lockfile --ignore-scripts` (the runtime install command in `setup-runtime.ts`) succeeds with the new dependency buckets.
- All of `zod/v3`, `zod/v4`, `zod/v4-mini`, and `shell-quote` resolve against a real `zod@4` install via the new preflight.
- Confirmed the worker boots normally with the native grammars absent (graceful degradation).
- Reproduced the original outage and recovered it by completing the `zod` install.

## Risk & scope
- **Happy path is unchanged:** with deps fully installed, the preflight passes and recycle/spawn proceed exactly as before.
- Behavior only changes in the broken-install case — and strictly for the better (keep serving / fail fast with diagnostics instead of bricking).
- No schema, API, or data changes.

## Reviewer's guide
1. **`src/shared/worker-utils.ts`** — `checkInstalledWorkerDependencies()` and its two call sites in `ensureWorkerRunning()` (the core logic).
2. **`scripts/build-hooks.js`** — the `dependencies` vs `optionalDependencies` split.
3. **`src/shared/zod-required-subpaths.ts`** + **`src/npx-cli/install/setup-runtime.ts`** — the shared constant.
4. The `plugin/scripts/*.cjs` changes are **generated esbuild bundles** — no need to read them by hand.

The three commits are the initial fix followed by two rounds addressing automated review (Codex, Greptile): robust multi-layout dependency resolution, the full `zod` subpath set, and the `shell-quote` check.


rajivsinclair:
Thanks for the reviews — addressed in `3c7f1d24`:

**Codex P1 — resolve zod from the worker script dir / Greptile P2 — fixed node_modules path false negatives**
`checkInstalledWorkerDependencies` no longer hard-codes `<plugin>/node_modules/zod`. It now walks up from the worker script directory (`findInstalledPackageJson`), mirroring Node/Bun resolution, so it finds deps at `<marketplace>/node_modules` (marketplace root), `<plugin>/node_modules`, **or** a hoisted dev repo root. That removes the false "broken deps" negatives on cold-start/recycle and in dev installs.

**Greptile P2 — shell-quote also required**
Preflight now also checks `shell-quote`, so a worker that would crash on `require('shell-quote')` is caught up front instead of burning the ~15s cold-boot wait.

**Codex P1 — regenerate the plugin lockfile**
Regenerated `plugin/bun.lock` against the new buckets and verified the actual runtime command `bun install --frozen-lockfile --ignore-scripts` (setup-runtime.ts:417) **succeeds** (exit 0, `zod` installed). Bun 1.2's lockfileVersion-1 text format lists workspace deps in a single block, so the file is byte-identical, but it's freshly resolved against the `optionalDependencies` split and frozen-install passes — no stale-lock divergence.

Note on the deeper guarantee: `optionalDependencies` is best-effort install hygiene; the hard guarantee that a partial install can't brick the worker is the runtime preflight + graceful recycle/spawn guards in this PR (verified the worker boots fine with the native grammars absent).

rajivsinclair:
Good catch — fixed in `17a6f645`.

The preflight now verifies the **full critical set** (`zod/v3`, `zod/v4`, `zod/v4-mini`) plus `shell-quote`, using real `require.resolve` (honoring each package's `exports` map) instead of parsing zod's exports for just `./v3`. So a partial install that has `./v3` but is missing `./v4`/`./v4-mini` is now correctly rejected.

To prevent the two guards from drifting, I extracted the set into `src/shared/zod-required-subpaths.ts` (`ZOD_REQUIRED_SUBPATHS`), now imported by **both** the install-time check (`setup-runtime.ts` `verifyCriticalModules`) and the runtime preflight (`worker-utils.ts`). Verified all three subpaths + `shell-quote` resolve against a real zod@4 install.

thedotmack:
Thanks! v13.9.1 ships dependency-health preflight + graceful degradation so a broken/partial dependency install no longer bricks the worker (commits c520c8f2c and f7aa16d60, https://github.com/thedotmack/claude-mem/releases/tag/v13.9.1). Closing as superseded by the released fix.

PR #3040, opened 2026-06-24
feat(openrouter): add CLAUDE_MEM_OPENROUTER_EXTRA_BODY for provider-specific request parameters

pillumina:
## Motivation

Users connecting through OpenAI-compatible endpoints (DeepSeek, MiniMax, local models, etc.) sometimes need to pass provider-specific parameters in the chat completion request body — for example, `{"thinking":{"type":"disabled"}}` to disable reasoning on models that would otherwise consume the output token budget with chain-of-thought.

Currently there is no way to inject custom body fields without patching the source.

## Changes

Adds a new optional setting `CLAUDE_MEM_OPENROUTER_EXTRA_BODY` that accepts a JSON object string. The parsed object is spread **last** into the request body, so it can add or override any field:

```json
{
  "CLAUDE_MEM_OPENROUTER_EXTRA_BODY": "{\"thinking\":{\"type\":\"disabled\"}}"
}
```

**Implementation** (2 files, +26/-1):

- `SettingsDefaultsManager.ts`: new field in the interface + default (`""`)
- `OpenRouterProvider.ts`:
  - `extraBody?: Record<string, unknown>` added to `OpenRouterConfig`
  - `getOpenRouterConfig()` parses the JSON, validates it's an object, logs a warning if invalid
  - `extraBody` spread into the request body after `usage.include` so it's always last

**Error handling**:
- Missing setting: no extra fields spread (unchanged behavior)
- Invalid JSON: logged warning, no extra fields spread
- Valid JSON that is not an object (e.g. array, string, number): logged warning, ignored

## Backward compatibility

Fully compatible — empty/missing setting means no extra fields are spread. All existing requests are unchanged.

## Usage examples

Disable thinking on reasoning models:
```json
{ "CLAUDE_MEM_OPENROUTER_EXTRA_BODY": "{\"thinking\":{\"type\":\"disabled\"}}" }
```

Lower reasoning effort (for providers that support `reasoning_effort`):
```json
{ "CLAUDE_MEM_OPENROUTER_EXTRA_BODY": "{\"reasoning_effort\":\"low\"}" }
```

### DIFF
--- a/src/services/worker/OpenRouterProvider.ts
+++ b/src/services/worker/OpenRouterProvider.ts
@@ -198,7 +198,7 @@ export class OpenRouterProvider extends OpenAICompatibleProvider<OpenRouterConfi
   }
 
   protected async query(history: ConversationMessage[], config: OpenRouterConfig): Promise<ProviderQueryResult> {
-    return this.queryOpenRouterMultiTurn(history, config.apiKey, config.model, config.apiUrl, config.siteUrl, config.appName);
+    return this.queryOpenRouterMultiTurn(history, config.apiKey, config.model, config.apiUrl, config.siteUrl, config.appName, config.extraBody);
   }
 
   private async queryOpenRouterMultiTurn(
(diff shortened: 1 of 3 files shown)


pillumina:
@thoraxe ready for review

thedotmack:

> step 4 (OpenAI-compatible config surface: extra request-body fields)

thedotmack:
Second pass — routing holds at **#3606 (plan-18, Observer Response Pipeline)**: a generic escape hatch on the OpenAI-compatible request body is the step-4 config surface that lets the observer stop burning its output budget on reasoning tokens. Worth noting the overlap with #3001, which adds a typed `CLAUDE_MEM_OPENROUTER_REASONING_EFFORT` knob for the same disable-reasoning use case; this PR's `extra_body` subsumes that case generically and has no open P1, so it is the better base. Not closing #3001 outright since a typed knob is a different ergonomic bet, but the two should not both land as-is.

PR #3041, opened 2026-06-24
feat(worker): honor ?project= in GET /api/stats

laihenyi:
## Summary

`GET /api/stats` (`handleGetStats` in `DataRoutes.ts`) always ran unfiltered `COUNT(*)` queries and **ignored the `project` query parameter**. Any project-scoped request returned the identical global counts — even `?project=<nonexistent>` returned the global total.

This breaks status-line / dashboard consumers that fetch both a global and a per-project stat to render something like `136 obs | 14480 total`: both fields showed the same global number.

## Repro (before)

```
GET /api/stats                       -> observations: 14480
GET /api/stats?project=claude-mem    -> observations: 14480   # should be project-scoped
GET /api/stats?project=zzz-none      -> observations: 14480   # should be 0
```

## Fix

Read `req.query.project` and scope the observation/session/summary counts to it when present. The appended `WHERE project = ?` clause is constant and the value is bound as a parameter — no dynamic SQL. With no param, behavior is unchanged. This mirrors how `parsePaginationParams` already reads `project` for the paginated routes.

## After

```
GET /api/stats                       -> observations: 14480
GET /api/stats?project=claude-mem    -> observations: 136
GET /api/stats?project=zzz-none      -> observations: 0
```

## Testing

- New `tests/worker/http/routes/data-routes-stats-project.test.ts`: asserts global counts with no param, project-scoped counts with `?project=`, and that the `COUNT(*)` queries are parameterized by project.
- `npm run typecheck` passes; existing `tests/worker/http/routes/` suite (34 tests) stays green.
- Verified against a live worker (counts above are real).

### DIFF
--- a/src/services/sqlite/observations/recent.ts
+++ b/src/services/sqlite/observations/recent.ts
@@ -33,14 +33,27 @@ export function getAllRecentObservations(
   return stmt.all(limit) as AllRecentObservationRow[];
 }
 
-export function getFirstObservationCreatedAt(db: Database): string | null {
-  const stmt = db.prepare(`
-    SELECT created_at
-    FROM observations
-    ORDER BY created_at_epoch ASC
-    LIMIT 1
-  `);
+export function getFirstObservationCreatedAt(db: Database, project?: string): string | null {
+  // When a project is supplied, scope to that project's earliest row — including
+  // worktree observations adopted into it via merged_into_project — so the value
+  // matches the project-scoped counts instead of leaking the global earliest row.
+  const stmt = project
+    ? db.prepare(`
+        SELECT created_at
+        FROM observations
+        WHERE (project = ? OR merged_into_project = ?)
+        ORDER BY created_at_epoch ASC
+        LIMIT 1
+      `)
+    : db.prepare(`
+        SELECT created_at
+        FROM observations
+        ORDER BY created_at_epoch ASC
+        LIMIT 1
+      `);
 
-  const row = stmt.get() as { created_at: string } | undefined;
+  const row = (project ? stmt.get(project, project) : stmt.get()) as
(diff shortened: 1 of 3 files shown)


laihenyi:
Thanks @greptile-apps — both findings were correct and are fixed in `bfe7d7b0`. I aligned the endpoint with `PaginationHelper`'s reader scoping rather than inventing new semantics:

1. **Adopted worktree undercount** — observations and summaries now match the readers' `(project = ? OR merged_into_project = ?)`, so a project that adopted a worktree's rows counts them. `sdk_sessions` has no `merged_into_project` column, so it keeps plain `project = ?` (matching the prompts reader).
2. **Unfiltered `firstObservationAt`** — now scoped to the project (and its adopted worktrees) via `getFirstObservationCreatedAt(db, project)`, returning `null` when the project has none instead of leaking the global earliest row.

One related change for full reader-consistency: the **global** (no-`project`) counts now exclude the internal `observer-sessions` bookkeeping project (`WHERE project != ?`), the same way every reader's no-project branch does — so the status line's "total" matches what the dashboard list views sum, rather than including internal rows. Happy to drop that part if you'd prefer the raw `COUNT(*)` for the global case.

Tests assert the per-table WHERE clauses and bound params across the global, project-scoped, and `firstObservationAt` paths. `npm run typecheck` is clean and the DataRoutes + sqlite suites (107 tests) pass.

Verified live against a real DB: `?project=claude-mem` → 142 obs with a project-local `firstObservationAt` of `2026-02-18`, while global is `2026-01-17`; `?project=<unknown>` → 0 with `firstObservationAt: null` (no global leak).

thedotmack:

PR #3044, opened 2026-06-24
feat(mcp): add memory_save tool for worker-runtime manual writes

amellman:
## Problem

The `observation_add` / `memory_add` MCP tools are gated to the **server-beta** runtime — they call the `/v1` REST core and throw `requires CLAUDE_MEM_RUNTIME=server-beta` on the default `worker` runtime. That leaves **worker-runtime users with no agent-facing manual-write tool at all**, even though the worker already exposes an ungated `POST /api/memory/save` endpoint (`MemoryRoutes.ts`) that writes straight to SQLite + Chroma.

So an agent told to "save this to memory" on a stock worker install hits a hard error, despite the backend fully supporting the write.

## Change

Adds a `memory_save` MCP tool that proxies `POST /api/memory/save` via the existing `workerHttpRequest` helper — the same plumbing `search` / `timeline` / `get_observations` already use. It:

- works on the **default `worker` runtime** (no server-beta, Postgres, or Docker required);
- introduces **no new imports** and **no Bun-only code** (stays within the MCP bundle's Node constraint and the 600 KB budget — built bundle is ~333 KB);
- mirrors the shape of the existing `memory_*` alias tools.

```
Params: text (required), title?, project?, metadata?
Returns: { success, id, title, project, message }
```

## Test

Built via `scripts/build-hooks.js` and exercised over a real MCP stdio handshake against a running worker:

- `tools/list` includes `memory_save` (required: `["text"]`)
- `tools/call memory_save` → `{ "success": true, "id": <n>, "project": "...", "message": "Memory saved as observation #<n>" }`, verified persisted to SQLite and synced to Chroma.

## Why not just use server-beta?

server-beta requires standing up Docker + Postgres + an API key. For users who deliberately stay on the lightweight worker runtime, this exposes the manual-write capability the worker already has, with zero added infrastructure.

### DIFF
--- a/src/servers/mcp-server.ts
+++ b/src/servers/mcp-server.ts
@@ -643,6 +643,55 @@ NEVER fetch full details without filtering first. 10x token savings.`,
       return handleObservationAdd(merged);
     },
   },
+  // Worker-runtime manual write. Unlike observation_add/memory_add (which are
+  // gated to server-beta and call /v1/memories), this talks to the worker's
+  // ungated POST /api/memory/save — writing straight to SQLite + Chroma. It is
+  // the only manual-write path available on the default `worker` runtime, where
+  // no server-beta server, API key, or project id is configured. Uses the same
+  // workerHttpRequest plumbing as search/timeline, so it shares the worker the
+  // MCP server already auto-starts in worker mode. No Bun-only imports.
+  {
+    name: 'memory_save',
+    description:
+      'Save a manual memory to the local worker (SQLite + Chroma) via POST /api/memory/save. ' +
+      'Worker-runtime path — does NOT require server-beta. Use this on the default worker runtime. ' +
+      'Params: text (required), title, project, metadata.',
+    inputSchema: {
+      type: 'object',
+      properties: {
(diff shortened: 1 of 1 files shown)


thedotmack:
Retargeting this PR to `community-edge`.

The manual-write capability is useful experimentation, but we do not want to expose `memory_save` in the base MCP surface. When the primary agent sees a memory-writing tool, it starts treating persistence as part of its responsibility: deciding what to save, narrating memory operations, and spending attention on bookkeeping instead of the user’s task.

The base product contract is ambient memory: hooks and observers capture and persist work automatically, without asking the primary agent to manage memory. `community-edge` is the right place to explore explicit/opt-in memory tooling without changing that default attention model.

thedotmack:
Closing in the PR consolidation pass — superseded: landed on community-edge as fa3ec99f2 (#3172); maintainer declined for main. The analysis here fed the plan-23 design (#3611); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

PR #3049, opened 2026-06-25
fix: Cursor IDE compat (platform source + empty stdin)

Steaeavean:
Two fixes for Cursor + Claude Code dual-IDE setups on macOS.\n\n1. **Platform source conflict** — update incoming platform_source instead of throwing (observations were dropped with HTTP 500).\n2. **Empty stdin (#2188)** — default no-op for non-lifecycle hooks; restore diagnostics via CLAUDE_MEM_STRICT_STDIN=1.\n\nTested locally on claude-mem 13.8.1.


thedotmack:
Thanks for this. v13.9.1 threads platformSource through the worker/storage/search/session flows (commit 348d9ee40, https://github.com/thedotmack/claude-mem/releases/tag/v13.9.1), which covers the platform-source half here. If the empty-stdin handling for Cursor is still needed on 13.9.1, please reopen or open a focused PR for just that piece — happy to take it.

PR #3055, opened 2026-06-25
feat(worker): add CLAUDE_MEM_WORKER_SCRIPT_PATH to pin the worker bundle

### LINKED ISSUE #3054: Add CLAUDE_MEM_WORKER_SCRIPT_PATH to pin the worker bundle (testability/operability)
**Gap:** `resolveWorkerScriptPath()` (src/shared/worker-utils.ts) resolves the `worker-service.cjs` the daemon spawns/respawns from exactly two hardcoded candidates — the installed marketplace path, then cwd — with no override. Since the installed path is checked first and almost always exists, every respawn resolves to the *installed* bundle, even when the worker was first launched from a local build. There's no documented way to run/test the worker against a local branch build without overwriting the install via `sync-marketplace`.

**Proposal:** add an opt-in env var `CLAUDE_MEM_WORKER_SCRIPT_PATH`, checked first in `resolveWorkerScriptPath()` (guarded by `existsSync`), so all daemon spawn/respawn/handoff/MCP callers inherit a pinned bundle. Unset → byte-identical current behavior. Enables end-to-end testing of the worker (including the respawn path) and helps diagnose bundle-version issues (e.g. dual-install confusion).

Naming follows the project's conventions: `CLAUDE_MEM_WORKER_*` scope (cf. `CLAUDE_MEM_WORKER_PORT`/`_HOST`) + `_PATH` suffix (cf. `CLAUDE_MEM_CHROMA_UVX_PATH`).

Happy to send a PR (tests + docs).

crippledgeek:
Closes #3054.

## Problem
`resolveWorkerScriptPath()` (`src/shared/worker-utils.ts`) resolves the `worker-service.cjs` the daemon spawns/respawns from two hardcoded candidates — the installed marketplace path, then cwd — with no override. The installed path is checked first and almost always exists, so **every respawn resolves to the installed bundle**, even when the worker was launched from a local build. There's no documented way to run/test the worker against a local branch build without overwriting the install via `sync-marketplace`.

## Change
Add an **opt-in** env var `CLAUDE_MEM_WORKER_SCRIPT_PATH`, checked first in `resolveWorkerScriptPath()` (guarded by `existsSync`):

```ts
const candidates = [
  process.env.CLAUDE_MEM_WORKER_SCRIPT_PATH,   // opt-in override, first
  path.join(MARKETPLACE_ROOT, 'plugin', 'scripts', 'worker-service.cjs'),
  path.join(process.cwd(), 'plugin', 'scripts', 'worker-service.cjs'),
];
for (const candidate of candidates) {
  if (candidate && existsSync(candidate)) return candidate;
}
```

- **One site, all daemon spawn paths inherit it** — `worker-utils.ts` lazy-spawn, `worker-service.ts` restart + restart-handoff successor, `mcp-server.ts`. (The `npx` CLI wrapper and hook shell templates intentionally keep resolving the installed bundle as thin entry-clients; they delegate to the daemon, which honors the override.)
- **Opt-in / zero-risk** — unset, empty, or set-but-nonexistent → byte-identical prior behavior. The `candidate &&` guard is also required for the type to compile (`process.env.X` is `string | undefined`).
- **Env-only** — this is a bootstrap resolution (which worker to spawn, before settings load), so it deliberately gets no `SettingsDefaults` entry, mirroring how `MARKETPLACE_ROOT`/`CLAUDE_CONFIG_DIR` are env-derived.
- **Naming** follows the project conventions: `CLAUDE_MEM_WORKER_*` scope (cf. `WORKER_PORT`/`WORKER_HOST`) + `_PATH` suffix (cf. `CLAUDE_MEM_CHROMA_UVX_PATH`).

## Tests
New `tests/shared/worker-script-path.test.ts` (first coverage for this resolver): override-set-and-existing wins; set-but-nonexistent ignored (== unset); empty-string ignored (== unset). Env/temp-file isolated. `npm run typecheck` clean; full suite shows no new failures.

## Docs
`configuration.mdx` (env table) + `development.mdx` (Manual Testing: local-build recipe, with a note that pinning a *different*-version bundle than the installed plugin can trigger the existing version-recycle restart behavior once sessions fire hooks).

### DIFF
--- a/docs/public/configuration.mdx
+++ b/docs/public/configuration.mdx
@@ -20,6 +20,7 @@ Settings are managed in `~/.claude-mem/settings.json`. The file is auto-created 
 | `CLAUDE_MEM_CONTEXT_OBSERVATIONS` | `50`                        | Number of observations to inject      |
 | `CLAUDE_MEM_WORKER_PORT`      | `37700 + (uid % 100)`           | Worker service port (per-user default; override for fixed port) |
 | `CLAUDE_MEM_WORKER_HOST`      | `127.0.0.1`                     | Worker service host address           |
+| `CLAUDE_MEM_WORKER_SCRIPT_PATH` | —                               | Absolute path to a specific `worker-service.cjs` for the daemon to spawn/respawn. Overrides the installed-plugin and cwd resolution. Opt-in; for local-build testing and bundle pinning. |
 | `CLAUDE_MEM_DATA_DIR`         | `~/.claude-mem`                 | Data root — every other path (database, chroma, logs, settings.json, worker.pid) derives from this |
 | `CLAUDE_MEM_SKIP_TOOLS`       | `ListMcpResourcesTool,SlashCommand,Skill,TodoWrite,AskUserQuestion` | Comma-separated tools to exclude from observations |
 
--- a/docs/public/development.mdx
+++ b/docs/public/development.mdx
@@ -112,6 +112,34 @@ npm run worker:logs
 echo '{"session_id":"test-123","cwd":"'$(pwd)'","source":"startup"}' | node plugin/scripts/context-hook.js
 ```
 
+#### Run the worker from a local build (without syncing to the marketplace)
+
+`CLAUDE_MEM_WORKER_SCRIPT_PATH` pins the worker bundle the daemon spawns/respawns,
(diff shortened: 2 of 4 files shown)


thedotmack:

PR #3058, opened 2026-06-25
test(settings): isolate loadFromFile tests from all ambient default env keys

crippledgeek:
## What

Generalize the env isolation in `tests/shared/settings-defaults-manager.test.ts`: snapshot and delete **every** `getAllDefaults()` key from `process.env` in `beforeEach`, restore in `afterEach`.

Closes #3056

## Why

The `loadFromFile` edge-case tests assert `loadFromFile(<empty|corrupt file>)` deep-equals `getAllDefaults()`. But `loadFromFile` applies `process.env` overrides on top of file/defaults, while `getAllDefaults()` returns pure defaults — so any settings-default key present in the environment makes the two diverge.

The suite already stripped `CLAUDE_MEM_DATA_DIR` for this reason, but only that one key. On a contributor machine running a claude-mem install, other keys are exported too (e.g. `CLAUDE_MEM_API_TIMEOUT_MS=120000`), which silently failed 9 of these tests locally while CI — a clean env — stayed green.

```bash
bun test tests/shared/settings-defaults-manager.test.ts            # 9 fail on a dev box
env -u CLAUDE_MEM_API_TIMEOUT_MS bun test tests/shared/settings-defaults-manager.test.ts  # 35 pass
```

## Scope

**Test-only.** No production code changes — `loadFromFile`'s env-override behavior is correct and is already covered by the `environment variable overrides` describe block. The generalized strip is robust to whichever `CLAUDE_MEM_*` vars the host exports and subsumes the existing `CLAUDE_MEM_DATA_DIR` special-case. As a bonus it also de-flakes the `get`/`getInt`/`getBool` tests, which read `process.env` directly.

## Verification

Full suite before (on a dev box exporting `CLAUDE_MEM_*`): **2159 pass / 9 fail**.
After: **2159 pass / 0 fail**. `tsc --noEmit` clean.

Issue #3037, opened 2026-06-23
Generation halts in an infinite poison/respawn loop when the subscription account hits its weekly/usage limit (rate-limit prose misclassified as invalid output)


syx-labs:
## Summary

When the Claude **subscription** account used for generation hits its **weekly/usage limit**, generation enters an **infinite poison → respawn loop** and never recovers, even though raw capture keeps working — so the halt is silent.

The SDK returns the plain-text message `You've hit your weekly limit · resets … ` instead of the expected `<observation>`/`<summary>` XML. The parser classifies this as a generic invalid output (`outputClass=prose`), increments `consecutiveInvalidOutputs`, and at `threshold=3` marks the session **poisoned** → kills + respawns it. The respawn immediately re-sends the pending message → hits the same limit → poisoned again → loop, until the worker is killed.

Net effect: **observation/summary generation stops and does not resume** until the account limit resets. **Restarting the worker does not help** (the respawn just re-hits the limit). Raw event capture (`user_prompts`) keeps advancing because it does not go through the SDK, which is why the failure is invisible from the outside.

## Environment

- claude-mem **13.8.0** (plugin), Claude Code on macOS
- `CLAUDE_MEM_MODEL=sonnet`
- `CLAUDE_MEM_CLAUDE_AUTH_METHOD=subscription` (OAuth token from keychain — **not** `ANTHROPIC_API_KEY`)
- Single worker daemon / HTTP listener

## Mechanism

The generator (`startGeneratorWithProvider`) calls the Claude SDK and the parser only accepts `<observation>`/`<summary>` XML. Any other text is `outputClass=prose|idle` → counted as an invalid output. A **rate/usage-limit response is human prose**, so it is mistaken for a malformed generation and poisons the session instead of being treated as a transient, retryable condition.

## Log evidence (redacted: session id → `NNNN`, project/content ids/pids/tool payloads removed)

```
[2026-06-23 01:37:21.578] [INFO ] [SDK   ] [session-NNNN] ← Response received (71 chars) {promptNumber=4} You've hit your weekly limit · resets Jun 25 at 5pm (America/Sao_Paulo)
[2026-06-23 01:37:21.579] [WARN ] [PARSER] [session-NNNN] SDK returned non-XML prose response — ignoring queued batch {outputClass=prose, preview=You've hit your weekly limit · resets Jun 25 at 5pm, consecutiveInvalidOutputs=3}
[...]

Issue #3007, opened 2026-06-19
SDK session repeatedly poisoned by non-XML idle responses during heavy agentic sessions


tnessffttozero:
# SDK session repeatedly poisoned by non-XML idle responses during heavy agentic sessions

## Problem Statement

During and after heavy agentic sessions, the Claude Agent SDK session repeatedly enters a "poisoned" state due to consecutive non-XML idle responses, causing the worker to kill and respawn the SDK session in a loop. While a related dependency issue (marketplace deps missing after repair) was worked around manually, the core SDK session poisoning/respawn loop persists and continues to disrupt active sessions.

## Environment

- **Claude-mem**: 13.6.2
- **Claude Code**: 2.1.183
- **Node.js**: v25.9.0
- **Bun**: 1.3.14
- **OS**: Linux 7.0.3-arch1-1 (x64)
- **Platform**: linux
- **Dev Mode**: No

**Paths**
- Plugin: `~/.claude/plugins/marketplaces/thedotmack`
- Data Directory: `~/.claude-mem`
- Current Directory: `~/.claude/plugins/marketplaces/thedotmack`

**Worker Status**
- Running: Yes
- PID: 390914
- Port: 37700
- Uptime: 40 minutes
- Active Sessions: 0
- SSE Clients: 5

**Database**
- Path: `~/.claude-mem/claude-mem.db`
- Exists: Yes
- Size: 216.00 KB

**Configuration**
- Settings File: `~/.claude-mem/settings.json`
- Settings Exist: Yes
- `CLAUDE_MEM_MODEL`: `claude-sonnet-4-6`

## Steps to Reproduce

1. Run `claude-mem doctor` — it reports marketplace dependencies missing, even after running the repair command.
2. Manually run `npm install --legacy-peer-deps` inside the plugin directory to resolve the dependency issue.
3. Start a heavy agentic session (high tool-use volume / long-running session).
4. Observe worker logs during and after the session.

## Expected vs Actual Behavior

**Expected:** The SDK session should remain stable during heavy agentic use, processing observations without entering a poisoned/respawn loop. `claude-mem doctor --repair` should also fully resolve missing marketplace dependencies without requiring a manual `npm install --legacy-peer-deps` workaround.

**Actual:** The SDK session is repeatedly poisoned by consecutive non-XML idle responses, triggering a kill-and-respawn loop that recurs during and after heavy agentic sessions. The dependency issue could only be resolved with a manual workaround, not via the built-in repair command.

[...]

PR #3059, opened 2026-06-25
fix(observer): window invalid-output respawn so benign idle can't poison quiet sessions

### LINKED ISSUE #3032: Observation generation stuck in a self-sustaining poison→respawn loop: benign "prose"/"idle" SDK output counts toward the poison threshold, wiping context and dropping all captured work
## Environment

- claude-mem 13.8.0 (observed; appears to be a regression vs 13.6.2, which captured tool work fine)
- Windows 11, bun runtime, worker HTTP on 127.0.0.1:37777
- Generator model: claude-sonnet-4-5

## Summary

Across an entire multi-hour session, the observer/generator captured **only user prompts** and produced essentially **zero observations of actual work** (every generated observation had `files_read: []`). The worker logs show the generator session being **"poisoned" and respawned every ~30–60s** for hours.

Root cause: the parser treats a **benign** generator response (the model saying "nothing to record yet" as plain prose, or returning an empty/idle response) as an **invalid output**. After `Y9 = 3` consecutive such outputs the session is "poisoned": `conversationHistory = []` is wiped and the SDK session respawns (losing context, Issue #817, `preservedPending: 0`). Because the wipe destroys the very context the observer needs to accumulate, the next checkpoint again looks like "just a user prompt → nothing to record → prose", and the loop sustains itself forever. No work is ever captured.

## Code (worker-service.cjs, function `Bc`)

```js
// M9(t): output is "valid" only if it contains <observation>/<summary>/<skip_summary/>
// D9(t): outputClass -> "idle" | "poisoned" (matches Cwe API-failure markers) | "xml" | "prose"
if (!d.valid) {
  let k = D9(t), I = N9(t);
  if (e.consecutiveInvalidOutputs = (e.consecutiveInvalidOutputs ?? 0) + 1,
      g.warn("PARSER", `${a} returned non-XML ${k} response — ignoring queued batch`, {...}),
      k === "poisoned" || e.consecutiveInvalidOutputs >= Y9) {   // <-- BUG
        // poison: respawnPoisonedSession -> conversationHistory=[] wipe
  }
  ...
}
```

The problem is `|| e.consecutiveInvalidOutputs >= Y9`: benign `prose`/`idle` (the model legitimately having nothing to record) accumulates toward the threshold and triggers a context-destroying poison. `Cwe` (the genuine poison markers) are real API failures ("context window", "session exhausted", "prompt is too long", …) — `prose`/`idle` are not failures.

## Evidence (logs)

```
# Every cycle, for hours:
[WARN ] [PARSER] SDK returned non-XML prose response — ignoring queued batch {outputClass=prose, consecutiveInvalidOutputs=3}
[ERROR] [SESSION] SDK session poisoned — killing and respawning {outputClass=prose, consecutiveInvalidOutputs=3, threshold=3}
[WARN ] [SESSION] Respawning poisoned SDK session, preserving pending messages {preservedPending=0}
[WARN ] [SESSION] Discarding stale memory_session_id from previous worker instance (Issue #817)
# prose previews show the model only ever sees the prompt:
"I'm observing the primary session, but I don't see any technical work being performed yet - only the user's request."
# Net result: only prompt-echo observations, all with files_read: []
```

## Suggested fix (one condition)

Only poison on a genuine API-failure marker, never on accumulated benign prose/idle:

```js
-  k === "poisoned" || e.consecutiveInvalidOutputs >= Y9
+  k === "poisoned"
```

(Or, more conservatively: do not increment `consecutiveInvalidOutputs` for `k === "prose" || k === "idle"`; treat them as a no-op skip, equivalent to `<skip_summary/>`.)

## Verification of the fix (applied locally, before/after on the same session 327)

| | Before patch | After patch |
|---|---|---|
| Poison events | every ~30–60s for hours | **none** (count frozen) |
| `consecutiveInvalidOutputs=3/4` | → poison + context wipe | **ignored, session survives** |
| Observations stored | 1 in hours (prompt echo) | **3 in minutes, real content** |
| `files_read` | `[]` | **populated** (actual files read) |

After the patch, the generator survives the early "idle/prose" checkpoints, accumulates the real tool activity, and emits + stores genuine `<observation>` XML (`[DB] STORING | obsCount=1`).

## Related

Distinct from the worker-recycle/zombie issue (#3031); this is the observation-generation pipeline. Also touches Issue #817 (SDK context lost on respawn) — the poison loop makes #817 fire continuously.

crippledgeek:
## What

Replaces the unbounded `consecutiveInvalidOutputs >= 3` respawn counter — which wrongly counted benign `idle`/prose SDK output and poison-looped quiet sessions, wiping context and dropping all captured work — with a **time-windowed burst counter** (modelled on systemd's `StartLimitBurst`/`StartLimitIntervalSec` and Erlang/OTP supervisor `intensity`/`period`).

Closes #3032
Refs #2935 #3007 #3037 #3022 #2955 #2960

## Why

On low-signal sessions the SDK legitimately emits `idle`/empty/prose responses. The old counter treated every one as an "invalid output" and, after 3 in a row, killed and respawned the observer — which on a quiet session just produces more idle output, so the respawn itself re-triggers the counter: a self-sustaining poison→respawn loop. The fix only respawns when invalid outputs arrive as a *burst within a bounded window*, so isolated benign idles can never accumulate to the threshold.

## How

- New module `src/services/worker/agents/respawn-policy.ts`: `evaluateRespawn` (windowed decision with an exhaustiveness guard over output classes), `parseRespawnPolicy`, cached settings-backed `getRespawnPolicy`, and a `FailureWindow` class.
- `ActiveSession.consecutiveInvalidOutputs` → `invalidOutputWindow` (timestamps within the window, not a monotonic count).
- `poisoned` output still respawns **immediately**; `idle` is **exempt by default**.
- Respawn-policy telemetry dimension wired through the scrubber; an orphaned telemetry key retired.

## New settings (all string, in `SettingsDefaultsManager`, defaults reproduce prior behavior minus the bug)

| Key | Default | Range |
|---|---|---|
| `CLAUDE_MEM_INVALID_OUTPUT_EXEMPT_CLASSES` | `idle` | comma-list of output classes |
| `CLAUDE_MEM_INVALID_OUTPUT_RESPAWN_THRESHOLD` | `3` | 1..=100 |
| `CLAUDE_MEM_INVALID_OUTPUT_WINDOW_MS` | `60000` | 1000..=3_600_000 |

## Tests

- `tests/worker/agents/respawn-policy.test.ts` (new, comprehensive: windowing, threshold boundaries, exempt classes, exhaustiveness).
- `tests/shared/settings-respawn-policy.test.ts` (new, settings parse + validation bounds).
- Extended `tests/worker/poison-respawn.test.ts`, `response-processor.test.ts`, `scrub.test.ts`.
- Targeted suite: **71 pass / 0 fail**; `tsc --noEmit` clean.
- **Live soak (prior session):** real Haiku run emitted 5× idle + 3× prose → badCount ≤ 1, **0 poisons** (old code would have respawned).

## Notes

- **Poison-loop cluster:** this PR fully resolves #3032; the other `Refs` (#2935, #3007, #3037, #3022, #2955, #2960) describe the same self-sustaining respawn loop from different triggers. Windowing the invalid-output burst is a contributing mitigation for them, so they're linked as related rather than closed — e.g. #3022's `poisoned`-keyword false-positive is a distinct concern the `poisoned` class still respawns on immediately.
- Independent of #3055/#3057/#3058 (branched cleanly off `main`, no shared commits) — can merge in any order.
- Pre-existing/flaky suite failures (SettingsDefaultsManager env-leak — addressed in #3058 — and GeminiProvider/Worker-API full-suite-concurrency flakes) are unrelated; this diff touches none of those modules.

### DIFF
--- a/docs/public/telemetry.mdx
+++ b/docs/public/telemetry.mdx
@@ -96,7 +96,7 @@ Every event property passes through a strict whitelist scrubber — any key not 
 | `chroma_available` | `true` | Whether the vector-search backend was reachable for a search (false = fell back to full-text search) |
 | `fallback_reason` | `none` | Why a search fell back from vector search: none / chroma_connection / chroma_error / chroma_not_initialized — a closed enum, never an error message |
 | `invalid_output_class` | `idle` | Coarse class of an unusable compression output: xml / idle / prose / poisoned (`xml` = looked like the expected format but failed to parse) — never the output itself |
-| `consecutive_invalid_outputs` | `3` | How many unusable outputs occurred in a row before recovery |
+| `respawn_threshold` | `3` | The unusable-output burst threshold that triggered a session respawn |
 | `respawn_triggered` | `true` | Whether the compression agent was restarted after repeated unusable output |
 | `abort_reason` | `idle` | Why a compression session was aborted: idle / shutdown / overflow / restart_guard / quota / poisoned / none — a closed enum |
 | `previous_shutdown` | `clean` | How the previous worker run ended, detected at startup: crash / clean / unknown |
--- a/src/npx-cli/commands/telemetry.ts
+++ b/src/npx-cli/commands/telemetry.ts
@@ -67,7 +67,7 @@ const COLLECTED_FIELDS = [
   'chroma_available whether vector search was reachable for a search',
   'fallback_reason  none / chroma_connection / chroma_error / chroma_not_initialized',
   'invalid_output_class   xml / idle / prose / poisoned (never the output)',
-  'consecutive_invalid_outputs   unusable outputs in a row before recovery',
+  'respawn_threshold      unusable-output threshold that triggered a respawn',
   'respawn_triggered      whether the compression agent was restarted',
   'abort_reason     idle / shutdown / overflow / restart_guard / quota / poisoned / none',
   'previous_shutdown      crash / clean / unknown (detected at worker start)',
--- a/src/services/telemetry/scrub.ts
+++ b/src/services/telemetry/scrub.ts
@@ -92,7 +92,7 @@ export const ALLOWED_PROPERTY_KEYS: Set<string> = new Set([
   // idle | shutdown | overflow | restart_guard | quota | poisoned | none).
   // Never model output, never raw abort strings.
   'invalid_output_class',
-  'consecutive_invalid_outputs',
+  'respawn_threshold',
   'respawn_triggered',
   'abort_reason',
   // Worker lifecycle health — previous_shutdown (crash | clean | unknown),
--- a/src/services/worker-types.ts
+++ b/src/services/worker-types.ts
@@ -1,5 +1,6 @@
 
 import type { Response } from 'express';
+import type { FailureWindow } from './worker/agents/respawn-policy.js';
 
 export interface ConversationMessage {
   role: 'user' | 'assistant';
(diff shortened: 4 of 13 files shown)


thedotmack:
Thanks! v13.9.1 takes a similar approach — benign idle/prose no longer counts toward the respawn threshold (commit ad4bd6f7a, https://github.com/thedotmack/claude-mem/releases/tag/v13.9.1). Closing as superseded by the released fix.

PR #3060, opened 2026-06-25
fix(supervisor): preserve Bedrock and Vertex skip-auth flags in SDK env (closes #3053)

### LINKED ISSUE #3053: Bug: CLAUDE_CODE_SKIP_BEDROCK_AUTH (and CLAUDE_CODE_SKIP_VERTEX_AUTH) stripped from SDK subprocess env — SDK hangs / 401s for Bedrock-routed gateways with Bearer auth
## Problem

The plugin's env-sanitization allowlist (`ENV_PRESERVE` in `src/supervisor/env-sanitizer.ts`, evident in the bundled output of `worker-service.cjs`, `mcp-server.cjs`, `server-beta-service.cjs`, `transcript-watcher.cjs`) includes `CLAUDE_CODE_USE_BEDROCK` and `CLAUDE_CODE_USE_VERTEX` but **not** `CLAUDE_CODE_SKIP_BEDROCK_AUTH` or `CLAUDE_CODE_SKIP_VERTEX_AUTH`.

When the parent Claude Code session has BOTH `CLAUDE_CODE_USE_BEDROCK=1` and `CLAUDE_CODE_SKIP_BEDROCK_AUTH=1` set (a common config for routing through a LiteLLM / OpenAI-compatible gateway that accepts Bearer auth on the `/bedrock/` path), the claude-mem worker inherits both — but only forwards `CLAUDE_CODE_USE_BEDROCK=1` to its spawned `claude` SDK subprocess. The SDK then:

1. Sees `CLAUDE_CODE_USE_BEDROCK=1` → routes via the Bedrock path
2. Does NOT see `CLAUDE_CODE_SKIP_BEDROCK_AUTH=1` → tries AWS SigV4 signing with no credentials
3. Either hangs indefinitely or hits the gateway unauthenticated and gets `401 Unauthorized`

In the log, the symptom is a ~97-char SDK response: `Failed to authenticate. API Error: 401 Unauthorized: ...` (or a queue that grows without any `[SDK] ← Response received` lines firing).

## Affected versions

- Reproduced on v13.8.1 (current)
- v13.8.0 had the same bug — patched locally and the `*.pre-skipbedrock.bak` files were preserved when v13.8.1 was installed, which is how we identified the regression

## Root cause (in bundled output)

In every spawn site that forwards env to a child, the allowlist Set is:

\`\`\`
\"CLAUDE_CODE_OAUTH_TOKEN\",\"CLAUDE_CODE_GIT_BASH_PATH\",\"CLAUDE_CODE_USE_BEDROCK\",\"CLAUDE_CODE_USE_VERTEX\",\"ANTHROPIC_BEDROCK_BASE_URL\",\"AWS_REGION\",\"AWS_PROFILE\",\"AWS_ACCESS_KEY_ID\",\"AWS_SECRET_ACCESS_KEY\",\"AWS_SESSION_TOKEN\",\"ANTHROPIC_VERTEX_PROJECT_ID\",\"CLOUD_ML_REGION\",\"GOOG...
\`\`\`

`CLAUDE_CODE_SKIP_BEDROCK_AUTH` and `CLAUDE_CODE_SKIP_VERTEX_AUTH` are missing.

## Reproduction

1. Configure the Claude Code session env with `CLAUDE_CODE_USE_BEDROCK=1`, `CLAUDE_CODE_SKIP_BEDROCK_AUTH=1`, `ANTHROPIC_BEDROCK_BASE_URL=<gateway>/bedrock`, `ANTHROPIC_AUTH_TOKEN=<bearer>` (a gateway-bearer-on-bedrock-path setup)
2. Confirm Claude Code itself works (it honors `SKIP_BEDROCK_AUTH=1` and uses the Bearer token)
3. Start a session — observe `~/.claude-mem/logs/claude-mem-<date>.log`
4. Worker enqueues observations but SDK calls either hang indefinitely or return `Failed to authenticate. API Error: 401`
5. Check the SDK subprocess's env: \`tr '\0' '\n' < /proc/<sdk-pid>/environ | grep -E \"USE_BEDROCK|SKIP_BEDROCK\"\` — only \`USE_BEDROCK=1\` is present

## Workaround

Add the missing keys to the allowlist Set in all four bundled scripts under \`~/.claude/plugins/cache/thedotmack/claude-mem/<version>/scripts/\`:

\`\`\`bash
sed -i 's/\"CLAUDE_CODE_USE_BEDROCK\",\"CLAUDE_CODE_USE_VERTEX\"/\"CLAUDE_CODE_USE_BEDROCK\",\"CLAUDE_CODE_SKIP_BEDROCK_AUTH\",\"CLAUDE_CODE_USE_VERTEX\",\"CLAUDE_CODE_SKIP_VERTEX_AUTH\"/g' \\
  worker-service.cjs mcp-server.cjs server-beta-service.cjs transcript-watcher.cjs
\`\`\`

(Patch is lost on every plugin version bump.)

## Proposed fix

In \`src/supervisor/env-sanitizer.ts\` (or wherever \`ENV_PRESERVE\` lives), add the two SKIP_* keys alongside their \`USE_*\` counterparts:

\`\`\`diff
  \"CLAUDE_CODE_USE_BEDROCK\",
+ \"CLAUDE_CODE_SKIP_BEDROCK_AUTH\",
  \"CLAUDE_CODE_USE_VERTEX\",
+ \"CLAUDE_CODE_SKIP_VERTEX_AUTH\",
\`\`\`

These vars are documented Claude Code SDK env vars; any caller using \`USE_BEDROCK=1\` with a non-AWS endpoint needs \`SKIP_BEDROCK_AUTH=1\` to be honored end-to-end.

## Related

- #2999 — same code path (\`env-sanitizer.ts\`), same bug pattern: another category of env var is silently dropped from the allowlist. Whoever fixes that one should probably audit the full list against Claude Code's SDK env-var surface.

rodboev:
## Summary

`sanitizeEnv()` preserves `CLAUDE_CODE_USE_BEDROCK` and `CLAUDE_CODE_USE_VERTEX`, but it strips the paired `CLAUDE_CODE_SKIP_BEDROCK_AUTH` and `CLAUDE_CODE_SKIP_VERTEX_AUTH` flags before the SDK subprocess launches. That leaves bearer-auth gateway routes on the Bedrock or Vertex code path without the explicit skip-auth override, which matches the reported hangs and `401 Unauthorized` failures.

This change adds the two missing skip-auth flags to the supervisor allowlist, regenerates the shipped worker, server-beta, and MCP bundles that embed that preserve set, and covers both the source helper and the packaged runtime surfaces with focused sanitizer regressions. The subprocess launch flow stays unchanged; only the env contract is corrected.

## Why

`ENV_PRESERVE` in `src/supervisor/env-sanitizer.ts` is the authoritative allowlist for `CLAUDE_CODE_*` vars that survive sanitization. The exact-match check preserves `CLAUDE_CODE_USE_BEDROCK` and `CLAUDE_CODE_USE_VERTEX`, then the prefix filter drops every remaining `CLAUDE_CODE_*` var, including the omitted `CLAUDE_CODE_SKIP_*_AUTH` flags.

`ClaudeProvider` builds the SDK subprocess env with `sanitizeEnv(await buildIsolatedEnvWithFreshOAuth())`, so the omission is what reaches the child process. The same preserve set is also baked into the shipped `worker-service.cjs`, `server-beta-service.cjs`, and `mcp-server.cjs` bundles that installed users actually execute. Adding the two missing keys and regenerating those bundles fixes the auth-mode mismatch without widening the rest of the prefix filter.

## Scope

- Keeps scope to the supervisor allowlist, the three shipped bundles that embed it, and the focused sanitizer test surface.
- Does not change `EnvManager.ts` or any subprocess launch site logic.
- Related sanitizer work in [#3018](https://github.com/thedotmack/claude-mem/pull/3018) preserves proxy vars on the same file surface; this PR keeps scope to the two skip-auth flags from [#3053](https://github.com/thedotmack/claude-mem/issues/3053).

## Verification

- [x] `bun test tests/supervisor/env-sanitizer.test.ts` - 14 passing, including shipped-bundle coverage for worker, server-beta, and MCP runtime bundles
- [x] `npm run build` - bundle within guardrail
- [x] `npm run lint:hook-io && npm run lint:spawn-env` - clean
- [ ] `npm run strip-comments:check` - current upstream baseline still reports 329 would-change files, including the pre-existing narrative comment block in `src/supervisor/env-sanitizer.ts`
- [x] Manual: copied the branch sanitizer test file onto an `origin/main` throwaway checkout and confirmed the new skip-auth assertions fail on base but pass on this branch

Closes #3053


OliverViola:
great to have you

thedotmack:
Closing as already implemented on current main: the Bedrock and Vertex skip-auth flags are present in the authoritative environment-sanitizer allowlist with coverage.

Issue #3038, opened 2026-06-24
Near-duplicate observations accumulate: content_hash only dedups byte-identical rows (fuzzy write-time dedup?)


DVN82:
## Summary

`content_hash` only catches **byte-identical** observations, so semantically-identical
observations that are worded slightly differently each get a distinct hash and accumulate
indefinitely. Over time a single recurring piece of work dominates the store as near-duplicate
noise. This is related to but distinct from the concurrency angle in #1260 and the sync-dedupe
in #2240 (both closed) — here the rows are *not* byte-identical, so hash dedup never fires even
in the single-writer case.

## Evidence (real worker-mode DB)

In a `claude-mem.db` with 134 observations, **27** described the same recurring fix; **13** had
byte-identical titles and the rest were paraphrases (e.g. "Hardened on-demand checkpoint against
transcript read race" vs "On-demand checkpoint hardened against transcript read race" vs "Race
condition fixed in checkpoint hook transcript reading"). All 134 had **distinct `content_hash`
values** — so `content_hash` dedup removed none of them. ~20% of the store was near-dup noise.

## Why it matters

- Retrieval quality degrades: one repeated topic crowds out distinct observations in ranked results.
- Storage/embedding cost grows with redundant Chroma docs.
- The signal-to-noise of injected SessionStart context drops.

## Proposed approach

Add an optional **fuzzy/near-duplicate guard at write time** (and/or an enrichment-time sweep),
in addition to the exact `content_hash`:

1. On insert, compare the new observation's normalized title/narrative against recent same-project
   observations using a cheap similarity (token-set Jaccard or `difflib` ratio; or cosine over the
   embedding you already compute for Chroma).
2. Above a threshold, **merge** instead of inserting: keep the newest, increment a `collapsed`
   counter + retain member ids in metadata for audit, recompute salience.
3. Make it opt-in / threshold-configurable to avoid over-merging genuinely distinct work.

Happy to share the standalone cleanup script I wrote (clusters by title similarity, deletes from
SQLite only — FTS is maintained by the `observations_ad`/`_au` triggers, and worker-mode search
hydrates Chroma hits through SQLite so orphaned vectors are inert) if useful as a starting point.

Environment: claude-mem worker runtime, v13.4.x.

crippledgeek:
[...]

PR #3063, opened 2026-06-26
feat(dedup): opt-in near-duplicate observation dedup — Tier-0 auto-merge + Tier-1 candidates (#3038)

crippledgeek:
Refs #3038.

Opt-in, off-by-default near-duplicate dedup for observations. `content_hash` only
catches byte-identical rows, so paraphrased observations of the same recurring work
accumulate as noise. This adds two **deterministic** tiers, validated against a real
7,651-observation DB before any code was written (see the diagnostic on #3038).

> This is the **conservative** interpretation of #3038 (exact auto-merge + fuzzy
> *candidate-flagging*, not fuzzy auto-merge). The data showed fuzzy lexical
> *auto-merge* is unsafe — happy to switch to auto-merge if you'd prefer; I asked on
> the issue. The LLM-adjudication tier that *could* safely resolve the residual cases
> is deliberately deferred to a follow-up PR.

## What it does (all gated on `CLAUDE_MEM_DEDUP_ENABLED`, default `false`)

- **Tier 0 — exact-normalized-title → silent auto-merge.** A new observation whose title
  matches an existing one after lowercase/whitespace-collapse/punctuation-strip is collapsed
  onto the existing row (cross-session) and its new `occurrence_count` is bumped. O(1) via an
  indexed `title_norm_key` (`sha256(project + normalizeTitle)`), mirroring the `content_hash`
  pattern. **NULL key for empty/emoji/punctuation-only titles** so they never collapse into
  each other.
- **Tier 1 — IDF-weighted TF-IDF cosine + an IDF "veto" → review-only candidates.** Records
  near-dup *candidates* in `observation_dedup_candidates` (never auto-merged). The veto
  (a Fellegi–Sunter blocking key) rejects pairs that differ only in a rare discriminating
  token — `rdlp-api` vs `rdlp-plugin`, `ffmpeg-7.1` vs `6.1` — which plain string similarity
  wrongly scores ~0.9.

## Why this shape (validation evidence)

On the real DB at production thresholds: SimHash-over-narrative was useless (false-positive
dominated); the signal lives in the **title**; exact-normalized-title found **35–47 real
redundancies `content_hash` misses with zero false positives**; and fuzzy lexical *auto-merge*
would have destroyed genuinely-distinct work that differs in one token — hence Tier-1 is
review-only. Semantic paraphrases (different words, same meaning) need embeddings and are out
of scope here.

## Surfaces

- Migration v33 (pure SQL, idempotent): `occurrence_count`, `token_df` (IDF model),
  `dedup_meta`, `observation_dedup_candidates`.
- `POST /api/dedup/scan` — opt-in idempotent backfill (so existing DBs participate) + bounded
  inverted-index corpus sweep. Gated on the flag + single-flight + row-cap.
- `GET /api/dedup/candidates?project=` — read-only candidate list.
- 7 `CLAUDE_MEM_DEDUP_*` settings (documented in `configuration.mdx`).

**Disabled ⇒ byte-identical legacy behavior** (regression-tested).

## Tests

Strict TDD, ~20 commits. 175 dedup/sqlite/settings tests green; `tsc --noEmit` clean; full
`bun test` green. Coverage includes the empty-title data-loss guard, cross-project isolation,
cross-session + intra-batch Tier-0, cold-start gate, retry-idempotency (token_df/doc_count stay
flat on a merge or redelivery), and a real-DB-derived golden fixture.

## Reviews

Both ran over the full `BASE..HEAD` diff, fixes applied, then re-reviewed:

- **security-reviewer → PASS.** No injection / auth bypass / data-loss. Initial MEDIUM findings
  (scan concurrency guard; prepared-statement hardening) + LOWs (backfill row cap; trust-model
  doc) all resolved and re-confirmed.
- **code-reviewer → Approve.** Initial findings (occurrence_count retry-idempotency, a
  token_df-flat regression test, scan-when-disabled gating, IDF single-load perf, NaN-config
  clamp, DRY) all resolved with a genuine regression test; re-confirmed merge-ready.

## Trust model

Like all worker routes, `/api/dedup/*` is guarded only by the 127.0.0.1 binding + localhost
CORS (consistent with `DataRoutes`/`SearchRoutes`/`MemoryRoutes`).

## Follow-up (separate PR, only if this is accepted)

An async LLM-adjudication consolidation tier (with a tombstone/audit subsystem) to resolve the
common-token cases lexical methods can't (`Code` vs `Security` review) — kept out of this PR
because a wrong auto-merge in an append-only store is unrecoverable, so it needs its own
reversibility machinery.

### DIFF
--- a/docs/public/configuration.mdx
+++ b/docs/public/configuration.mdx
@@ -66,6 +66,33 @@ Use [LiteLLM Gateway](configuration/litellm-gateway) when you want `CLAUDE_MEM_P
 | `CLAUDE_MEM_PYTHON_VERSION`   | `3.13`                          | Python version for chroma-mcp         |
 | `CLAUDE_CODE_PATH`            | _(auto-detect)_                 | Path to Claude Code CLI (for Windows) |
 
+### Near-Duplicate Deduplication
+
+Opt-in (off by default) dedup of observations whose titles describe the same recurring
+work but are worded slightly differently, so `content_hash` (byte-identical only) misses them.
+Two deterministic tiers:
+
+- **Tier 0 — exact-normalized-title** → safe silent auto-merge: a new observation whose
+  title matches an existing one after lowercasing / whitespace-collapse / punctuation-strip
+  is collapsed onto the existing row and its `occurrence_count` is bumped (cross-session).
+- **Tier 1 — IDF-weighted similarity** → *review-only* near-duplicate **candidates** are
+  recorded in `observation_dedup_candidates` (never auto-merged), so distinct work that
+  differs only in one rare token (`rdlp-api` vs `rdlp-plugin`) is never silently destroyed.
+
+| Setting                              | Default | Description                           |
+|--------------------------------------|---------|---------------------------------------|
+| `CLAUDE_MEM_DEDUP_ENABLED`           | `false` | Master switch. Off = byte-identical legacy behavior |
+| `CLAUDE_MEM_DEDUP_COSINE_THRESHOLD`  | `0.80`  | Min IDF-weighted cosine for a Tier-1 candidate |
+| `CLAUDE_MEM_DEDUP_IDF_VETO_DF`       | `10`    | A title token in ≤ N project rows is "discriminating" and vetoes a merge |
+| `CLAUDE_MEM_DEDUP_MIN_SHARED_TOKENS` | `2`     | Require ≥ N shared tokens before scoring (kills sparse-vector noise) |
+| `CLAUDE_MEM_DEDUP_MIN_PROJECT_DOCS`  | `10`    | Cold-start gate: skip Tier-1 below N docs/project (IDF unreliable) |
+| `CLAUDE_MEM_DEDUP_MAX_SCAN`          | `2000`  | Per-insert Tier-1 candidate-scan cap |
+| `CLAUDE_MEM_DEDUP_MAX_BACKFILL_ROWS` | `50000` | `dedup-scan` skips a project larger than this (memory safety) |
+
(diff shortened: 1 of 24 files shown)


crippledgeek:
Thanks @greptile-apps — good catch, fixed in 4c67a644.

`recordTier1Candidates` / `sweepProjectCandidates` were doing `count++` on every `INSERT OR IGNORE` that classified as a candidate, so a re-run or redelivery reported a phantom count for rows that the `UNIQUE(observation_id, duplicate_of_id)` guard actually ignored. They now accumulate `.run().changes` instead, so the returned count reflects only newly-persisted rows — a second idempotent sweep correctly returns 0 (the persisted data was always idempotent; only the reported number was off).

Tightened the idempotency test to assert the return value (not just the row count) — it fails against the old code and passes now. 140 dedup/sqlite tests green, typecheck clean.

thedotmack:

> step 5 adopts exactly this shape (opt-in normalized-title dedup with occurrence_count; fuzzy tier review-only)

thedotmack:
Second pass — confirming **#3606 (plan-18, Observer Response Pipeline)** as the home for this; observation dedup is the "observation output quality" slice of that master, and step 5 already adopts this exact shape (opt-in normalized-title merge with `occurrence_count`, fuzzy tier as candidates only). No open PR duplicates it and nothing on main supersedes it — `content_hash` is still the only dedup path in `SessionStore`. This remains the reference implementation for that step; the conservative Tier-0-auto / Tier-1-flag split is the right call given the 7.6k-row diagnostic.

Issue #3042, opened 2026-06-24
[Bug] SDK observation generator returns idle responses on every session exit — observations never stored (Ubuntu, Claude Pro subscription auth)


halbodb:
Bug Description
The SDK observation generator consistently returns non-XML idle responses on every session exit, causing the session to be poisoned and no observations to be stored. This happens regardless of model selection (Haiku or Sonnet) and persists across multiple reinstalls.
Steps to Reproduce

Install claude-mem on Ubuntu with Claude Pro subscription auth
Open Claude Code in any project
Do any work (read files, run commands, ask questions)
Exit the session cleanly with exit
Observe the worker console — Stop hook fires, SDK session starts, returns idle 3 times, session poisoned

Expected Behavior
The background SDK observation generator should contact the configured model, receive a valid XML <observation> block in response, and store it to the SQLite database with STORED | obsCount > 0.
Environment

Claude-mem version: 13.8.0
Claude Code version: Latest (OAuth subscription auth via claude.ai)
OS: Ubuntu 24 (Linux)
Platform: x86_64, Python 3.14, Bun 1.3.14, uv 0.11.24
Auth method: Claude Pro subscription (CLAUDE_MEM_CLAUDE_AUTH_METHOD=subscription)
Model tested: claude-haiku-4-5-20251001 (default), claude-sonnet-4-6 (manual override)

Logs
[2026-06-24 16:29:37.138] [INFO ] [WORKER] Broadcasting processing status {isProcessing=true, queueDepth=1, activeSessions=1}
[2026-06-24 16:29:38.128] [WARN ] [PARSER] [session-6] SDK returned non-XML idle response — ignoring queued batch {outputClass=idle, preview=, consecutiveInvalidOutputs=1}
[2026-06-24 16:29:41.347] [WARN ] [PARSER] [session-6] SDK returned non-XML idle response — ignoring queued batch {outputClass=idle, preview=, consecutiveInvalidOutputs=2}
[2026-06-24 16:29:48.162] [WARN ] [PARSER] [session-6] SDK returned non-XML idle response — ignoring queued batch {outputClass=idle, preview=, consecutiveInvalidOutputs=3}
[2026-06-24 16:29:48.163] [ERROR] [SESSION] [session-6] SDK session poisoned — killing and respawning, pending messages preserved {outputClass=idle, consecutiveInvalidOutputs=3, threshold=3}
[...]

PR #3068, opened 2026-06-26
fix(observer): emit <skip_summary/> on skip instead of empty response

funrun11:
## Problem

On quiet / low-signal sessions the SDK observer poison-loops and stores **zero** observations:

```
[PARSER]  SDK returned non-XML idle response — ignoring queued batch {consecutiveInvalidOutputs=1}
[PARSER]  SDK returned non-XML prose response — ignoring queued batch {consecutiveInvalidOutputs=2}
[PARSER]  SDK returned non-XML idle response — ignoring queued batch {consecutiveInvalidOutputs=3}
[ERROR]   SDK session poisoned — killing and respawning, pending messages preserved
```

Reproduced on 13.8.0 and 13.8.1 (Linux + Windows). Worker restart and version bump do **not** fix it.

## Root cause — contract drift

`src/sdk/prompts.ts` instructs the observer:

> Return either one or more `<observation>...</observation>` blocks, or **an empty response** if this tool use should be skipped.

and several mode `skip_guidance` fields repeat *"return an empty response only"* / *"No output necessary if skipping."*

But `src/sdk/output-classifier.ts` classifies an empty/non-XML reply as `idle`/`prose` → **invalid output**, and `ResponseProcessor.ts` poisons the session after 3 consecutive invalids. So a *correct skip* is scored as an invalid output and walks the session straight to the poison threshold.

Meanwhile the parser and the server-side providers **already** standardize on a self-closing `<skip_summary/>` tag:

- `src/sdk/parser.ts` — `&lt;skip_summary(?:\s+reason="([^"]*)")?\s*/&gt;`
- `src/sdk/output-classifier.ts` — treats `<skip_summary/>` as valid `xml`
- `src/server/generation/providers/shared/prompt-builder.ts` — *"return a single self-closing `<skip_summary />`"*

## Fix

Align the SDK observation path and the mode `skip_guidance` with that existing contract: on skip, emit `<skip_summary reason="..."/>` (a recognized, valid output) instead of an empty response. No parser/threshold changes.

- `src/sdk/prompts.ts` — central skip instruction
- `plugin/modes/code.json`, `meme-tokens.json`, `law-study.json`, `law-study--chill.json`, `email-investigation.json` — `skip_guidance`

## Testing

Patched a local install, restarted the worker, ran a quiet file-reading session. Before: poison loop, 0 observations. After: skips emit `<skip_summary reason="..."/>` (stored as valid, `consecutiveInvalidOutputs` stays 0), substantive reads distill into `<observation>` blocks and sync to chroma. No poison events.

## Notes

Refs #3032, #3042. Complementary to the parser-side guards in #3059 / #2943 — this fixes the contract at the source (don't emit the empty reply) rather than tolerating it downstream. The two approaches compose.


thedotmack:
Thanks for this. The skip/empty-response poisoning is handled in v13.9.1 by dropping invalid prose in the output classifier rather than via a <skip_summary/> sentinel (commit ad4bd6f7a, https://github.com/thedotmack/claude-mem/releases/tag/v13.9.1). Closing as superseded — please reopen if the skip contract still bites on 13.9.1+.

PR #3070, opened 2026-06-26
feat(server): remote authenticated MCP recall endpoint (POST /v1/mcp)

ChenglinWei97:
## What
The **secure authenticated MCP link** a user pastes into Claude Code (or Cursor / Codex / any MCP client) to recall their cloud memory from Server Beta:

```
claude mcp add --transport http claude-mem <server-beta-base>/v1/mcp \
  --header "Authorization: Bearer cm_..."
```

Closes the one gap: the MCP server today is **stdio-only** (local worker). Server Beta already has the cloud recall data + API-key auth + REST recall — it just had no remote MCP transport. This adds it.

## Design — reuse, not rebuild
- **Auth:** the existing `readAuth` (`requirePostgresServerAuth`, scope `memories:read`). No new auth — the `cm_` key's team (and project scope) bound every read, same as `/v1/search`.
- **Storage:** the existing `PostgresObservationRepository.search` / `.listByProject` — the same code path as `/v1/search` and `/v1/context`. Identical data, identical guards.
- **Transport:** MCP streamable-HTTP (SDK 1.29), **stateless** — one transport + server per request, so it slots into Express 5 with no session state.

`src/server/mcp/recall-mcp-server.ts` is a pure factory (storage injected as a `RecallBackend`) exposing read-only tools **`search` / `context` / `recent`**. The route in `ServerV1PostgresRoutes` supplies a team-scoped backend. Mutating tools are intentionally omitted — a pasted recall link is read-only.

## Tests
`tests/server/mcp/recall-mcp-server.test.ts` drives a real MCP client over an in-memory transport (no Postgres): tool listing, arg forwarding + limit clamping, context packing, and backend failures surfacing as tool errors. **7/7 pass; full `tsc --noEmit` clean.**

## Before merge — one live check
Per the "verify against the live server" lesson: worth one integration pass hitting `/v1/mcp` on a real Server Beta with a `cm_` key from an MCP client, to confirm the streamable-HTTP + Express 5 wiring end-to-end (the unit test covers the tool logic, not the HTTP transport handshake).

## Open question
Where is Server Beta deployed (the host the link points at), and is `POST /v1/keys` the intended way for a user to mint a `cm_` key? Happy to add a key-issuance route + a Connect snippet in a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3078, opened 2026-06-28
feat(server): paid-readiness — usage metering, rate limiting/quotas, key issuance + connect

ChenglinWei97:
## What
Three Server Beta primitives needed before paid users can hit the cloud, all **opt-in via env** (default off → behavior unchanged) and **fail-open / fire-and-forget** (a metering or limiter hiccup never takes the API down). Pairs with the `/v1/mcp` recall link (#3070).

### Usage metering
- `usage_events` table + `PostgresUsageRepository` (record / total / summarize)
- `meterRequests` middleware — one `request` event per authed call (`CLAUDE_MEM_USAGE_METERING=1`), fire-and-forget. Token/observation metering uses the same repo from the generation worker.
- `GET /v1/usage` — per-kind totals for the team this month

### Rate limiting + quota
- `rate_limit_counters` table + `PostgresRateLimitRepository` — atomic fixed-window UPSERT, correct across instances
- `requireRateLimit` → `429` + `Retry-After` + `X-RateLimit-*` (`CLAUDE_MEM_RATE_LIMIT_PER_MIN`)
- `requireMonthlyQuota` → `402` at the per-team monthly cap (`CLAUDE_MEM_MONTHLY_REQUEST_CAP`)

### Key issuance + connect
- `POST /v1/keys` (**write** scope) — mints a **read-only**, optionally-expiring key + returns the paste-ready `claude mcp add … /v1/mcp` command (raw key shown once). Write-scope gate stops a read key escalating into more keys.
- `GET /v1/connect` (read scope) — same command with a `<YOUR_API_KEY>` placeholder

## Wiring
`readAuth`/`writeAuth` became **middleware arrays** (`[auth, ...guards]`) so the guards slot in after auth with **zero per-route edits** (Express accepts arrays anywhere a handler goes). Guards are only added when their env var is set.

## Tests — all live-verified against real Postgres (11/11), tsc clean
- Repo aggregation; rate-limit 429 + local-dev skip; quota 402/pass; fire-and-forget metering; `GET /v1/usage` end-to-end (proves the array-middleware wiring)
- Key mint → 201 + command shape; **the minted key authenticates** (calls `/v1/usage` → 200); read key can't mint (403); `/v1/connect` placeholder

## Open item (documented in api.md)
Cold-start — minting a team's **first** key still needs a session-gated path. better-auth's `apiKey()` plugin exists but writes to a **separate store** than the Postgres `api_keys` these routes check; wiring the better-auth org → Server Beta team mapping is the remaining piece. The write-scope `POST /v1/keys` covers everything once a team has one key.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -36,6 +36,9 @@ default) means no rate limit, no quota, and no metering, so behavior is unchange
   limit returns `429` with `Retry-After` (and `X-RateLimit-*` headers). Fail-open.
 - `CLAUDE_MEM_MONTHLY_REQUEST_CAP` — max requests per team per calendar month
   (UTC). At the cap, returns `402 quota_exceeded`. Fail-open.
+- `CLAUDE_MEM_MONTHLY_TOKEN_CAP` — max provider tokens per team per month. Gates
+  **writes only** (ingestion drives generation = token spend); reads stay
+  available so a team over budget can still recall. `402` at the cap. Fail-open.
 - `CLAUDE_MEM_USAGE_METERING=1` — record one `request` usage event per
   authenticated call (fire-and-forget). Token/observation metering writes to the
   same `usage_events` table from the generation worker.
--- a/docs/api.md
+++ b/docs/api.md
@@ -20,9 +20,29 @@ Available beta endpoints:
 - `PATCH /v1/memories/:id`
 - `POST /v1/search`
 - `POST /v1/context`
+- `GET /v1/usage`
 - `GET /v1/audit?projectId=<id>`
 
 When `CLAUDE_MEM_AUTH_MODE=api-key`, send `Authorization: Bearer <key>`. Read endpoints require `memories:read`; write endpoints require `memories:write`.
+
+## Rate limiting, quota, and usage metering
+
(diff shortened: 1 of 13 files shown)


ChenglinWei97:
Extended this PR with the cost side of metering (the $/1M-token model):
- **Token + observation metering** (`1a6f134a`) — when a generation job completes, record the provider `tokens` it spent and the `observation` count into `usage_events` (best-effort inside the completion transaction, like the audit row). Threaded `tokensUsed` from the provider result through.
- **Monthly token cap** (`d75477d3`) — `CLAUDE_MEM_MONTHLY_TOKEN_CAP` gates **writes only** (ingestion = token spend); reads stay available so an over-budget team can still recall.
- **X-RateLimit-Reset** (`41f7d6c3`) — your earlier review nit; resolved.

So the metering story is now end-to-end: requests + tokens + observations → quotas (request cap, token cap) → `GET /v1/usage`. All still opt-in via env, fail-open. tsc clean.

thedotmack:
Merged into `main` as part of #3089 (bundled with @ChenglinWei97's other server PRs; conflicts resolved against the `ServerBeta*` → `Server*` rename). Closing in favor of that PR — thanks for the work! 🙏

Issue #3064, opened 2026-06-26
Server-beta-only tools (e.g. `observation_search`) advertised in tools/list under `worker` runtime — agents call a non-functional tool; under Codex the call hangs the client


frarredondo:
## Before submitting

- [x] I searched [existing issues](https://github.com/thedotmack/claude-mem/issues) and confirmed this is not a duplicate

Closest existing issues, and why this is distinct:
- #2705 (closed → consolidated into #2782 / plan-09): same `Server beta error (transport): … requires CLAUDE_MEM_RUNTIME=server-beta` mechanism, but scoped to the **write** tools (`observation_add`/`memory_add`) and observed under Claude Code, where the error is *returned cleanly*. The planned fix (re-enable worker-runtime write tools) does not address read tools being advertised, nor a client that hangs.
- #2916 (open): server-beta `observation_add` param stripping — explicitly states "read tools (search, observation_search) work fine", but only under `server-beta` runtime. Different failure.

This report is specifically about **server-beta-only tools being advertised in `tools/list` while the runtime is `worker`**, and the **client-side hang** that results under the Codex MCP host.

### Bug Description

When claude-mem runs in `worker` runtime, the MCP server still advertises server-beta-only tools in `tools/list` — including `observation_search` (and `observation_add`, `memory_add`, `observation_context`, etc.). An LLM agent that is told to consult memory will frequently pick `observation_search` (it reads like the primary search tool). In `worker` runtime that tool is non-functional and returns an `isError` result telling the caller to use `search`/`timeline`/`get_observations` instead.

Under **Claude Code** this is merely annoying (the error is surfaced and the agent can recover). Under **Codex CLI** as the MCP host, the call to `observation_search` **never returns into the agent loop** — the whole session freezes. We observed two Codex `review` runs hang ~19 minutes at exactly this call before being killed.

Root cause (claude-mem side): a non-runtime-aware `tools/list`. Server-beta-only tools should not be advertised when `CLAUDE_MEM_RUNTIME=worker`, so agents can't select a tool that cannot work in the active runtime.

### Steps to Reproduce

1. Run claude-mem in `worker` runtime (default; `~/.claude-mem/worker.pid` present, no `.server-beta.port`/`.pid`).
2. Speak MCP to the server directly over stdio (`node .../plugin/scripts/mcp-server.cjs`): `initialize` → `notifications/initialized` → `tools/list`.
   - `tools/list` includes server-beta-only tools, e.g.:
[...]

PR #3082, opened 2026-06-29
fix(mcp): make the `search` tool work in server-beta runtime (route to /v1/search)

alessandropcostabr:
## Problem

In **server-beta** runtime the `search` MCP tool returns prompts but **0 observations**, even though the observations exist in Postgres. `observation_search` (which calls `/v1/search`) returns them fine, so the data and the server are healthy — only `search` is blind.

## Root cause

The `search` handler always routes to the local worker `/api/search`, which reads the local SQLite via the Chroma-backed `SearchOrchestrator`. In server-beta installs the local SQLite is **frozen** (generated observations live in PG, not SQLite) and, when Chroma isn't configured, observation **text** queries return empty — so `search` silently yields 0 observations. Prompts still return via a separate FTS path, which is why it looks like *"prompts work, observations don't."*

This is the server-beta-side mirror of #3064 (server-beta-only tools advertised under `worker` runtime): the tool surface isn't runtime-aware.

## Fix

When the runtime is server-beta **and** the query has text, route to the PG-backed `/v1/search` (the same path `observation_search` uses); otherwise fall back to the worker `/api/search` for worker-mode or filter-only (no text) queries. Reuses existing helpers (`resolveServerBetaToolContext`, `formatJsonResult`, `ServerBetaSearchObservationsRequest`).

`timeline` and `get_observations` are intentionally left unchanged — they're the SQLite-era numeric-id workflow and don't map to server-beta's UUID-keyed observations.

## Testing

- `tsc --noEmit` passes.
- Built the MCP server, spawned it with `CLAUDE_MEM_RUNTIME=server-beta`, and called `tools/call search` → returned Postgres observations (previously 0).

Related: #3064

### DIFF
--- a/src/servers/mcp-server.ts
+++ b/src/servers/mcp-server.ts
@@ -485,11 +485,23 @@ NEVER fetch full details without filtering first. 10x token savings.`,
       // the Chroma-backed SearchOrchestrator. When the install runs server-beta (where
       // generated observations live in Postgres, not local SQLite) and Chroma is not
       // configured, observation text queries return empty — so `search` silently yields
-      // 0 observations even though the data is in PG. Route text queries to the PG-backed
-      // /v1/search (same path as observation_search) when server-beta is available; fall
-      // back to the worker for worker-mode or filter-only (no text) queries.
+      // 0 observations even though the data is in PG.
+      //
+      // Route to the PG-backed /v1/search (same path as observation_search) ONLY when it
+      // can serve the request faithfully: server-beta is available, there is a text query,
+      // the result type is observations (or unspecified), and no filter /v1/search cannot
+      // honor is set. /v1/search is observations-only and takes only { projectId, query,
+      // limit } — so prompt/session-typed queries and platformSource/project/obs_type/date/
+      // offset/orderBy filters must keep the worker path (which applies them), otherwise we
+      // would silently drop the filter or mis-route the query.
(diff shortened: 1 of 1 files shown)


alessandropcostabr:
Thanks @chatgpt-codex-connector and @greptile-apps — both flagged the same real over-reach, and you're right.

The first commit routed **every** text query to `/v1/search` forwarding only `{ projectId, query, limit }`, which (a) silently dropped `platformSource` / `project` / `obs_type` / `dateStart` / `dateEnd` / `offset` / `orderBy`, and (b) mis-routed `type: 'prompts'` / `type: 'sessions'` queries to an observations-only endpoint.

Fixed in b23101c2: the server-beta branch is now taken **only** when `/v1/search` can serve the request faithfully — server-beta available, a text query, `type` is `observations` (or unset), and none of those filters is present. Everything else (prompt/session types, any of those filters, filter-only or worker-mode) keeps the worker `/api/search` path, which applies them. So `platformSource` scoping and the prompt/session result types are preserved.

Verified by spawning the built MCP server in server-beta mode: a plain text query returns PG observations, while `{ query, type: 'prompts' }` falls back to the worker and returns prompts (not PG observations).

Known limitation (called out in the commit): observation searches that **do** carry those filters still fall back to the worker in server-beta (0 obs there today) until `/v1/search` gains the params — a server-side follow-up, out of scope for this client-side routing fix.

Issue #2791, opened 2026-06-05
bun-runner.js SyntaxError on optional chaining (Node 22 ESM)


ven-kotha:
## Bug

Stop hook fails every session with:

```
file:///Users/.../.claude/plugins/cache/thedotmack/claude-mem/13.4.0/scripts/bun-runner.js:69
    return settings?.enabledPlugins?.['claude-mem@thedotmack'] === false;
                    ^

SyntaxError: Unexpected token '.'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:140:18)
    at async link (internal/modules/esm/module_job.js:42:21)
```

## Environment

- Plugin version: 13.4.0
- Node: v22.12.0
- Platform: macOS (Darwin 25.5.0)
- `package.json` has `"type": "module"` set correctly

## Analysis

The optional chaining syntax (`?.`) is valid ES2020+ and Node 22 supports it. The error stack trace references `internal/modules/esm/translators.js:140` which is from Node 12-13 era internals — suggesting Claude Code's hook runner may be invoking this script with a different/older Node binary than the user's shell Node.

The error is non-blocking but fires on every session stop, creating noise.

PR #3099, opened 2026-07-01
fix(bun-runner): bound hook invocations so a wedged worker can't block the prompt

rwjk16:
## Problem

On Windows, every prompt began hanging: slash commands and normal input appeared frozen. Root cause is the `UserPromptSubmit` → `session-init` hook. A hook payload invocation inherits the bun worker's runtime and blocks the Claude Code event that spawned it. When the worker wedges (observed as a stuck daemon connection after an auto-update), the prompt blocks on `bun-runner.js` until the host's 60s hook timeout fires — unusable.

Because the Windows child is a `cmd.exe` wrapper (`shell:true`), `child.kill()` only kills the wrapper, leaving the real `bun` grandchild alive. These orphans stacked up **one per prompt** (observed ~8 wedged `worker-service.cjs hook` processes).

## Fix

Add a watchdog around **non-lifecycle** (hook payload) invocations in `bun-runner.js`:

- If the worker doesn't exit within `CLAUDE_MEM_HOOK_TIMEOUT_MS` (default `8000`), tree-kill it and `exit 0`, appending a line to `runner-errors.log` as the durable signal (consistent with the existing exit-0 / marker-file IO discipline).
- On Windows the kill uses `taskkill /pid <pid> /T /F` so the whole tree dies (not just the `cmd.exe` wrapper) — this also stops the orphan pileup.
- Lifecycle commands (`start`/`stop`/`restart`/`status`) are **exempt** — the daemon they manage is meant to outlive the launcher.

The launcher parser-compat constraint (#2791, no `?.`/`??`) and the `fixBrokenScriptPath` safety net are preserved.

## Tests

Added source-pattern assertions to `tests/bun-runner.test.ts` (matching the file's existing test style). `bun test tests/bun-runner.test.ts` → 7 pass. `node --check` passes; no `?.`/`??` introduced.

### DIFF
--- a/plugin/scripts/bun-runner.js
+++ b/plugin/scripts/bun-runner.js
@@ -93,6 +93,28 @@ if (args.length === 0) {
 
 args[0] = fixBrokenScriptPath(args[0]);
 
+// Lifecycle commands manage the long-lived worker daemon; every other invocation
+// is a hook payload whose runtime blocks the Claude Code event that spawned it
+// (UserPromptSubmit, PostToolUse, ...). A wedged worker on one of those must never
+// hang the user's prompt — see the watchdog below.
+const LIFECYCLE_COMMANDS = ['start', 'stop', 'restart', 'status'];
+const isLifecycle = LIFECYCLE_COMMANDS.some(cmd => args.includes(cmd));
+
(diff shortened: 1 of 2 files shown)


thedotmack:

Issue #3092, opened 2026-06-30
v13.9.1: two more missing-bundle modules — SessionStore.js (backfill fails at startup) + .install-version marker ("runtime not yet set up")


Robin528919:
## Summary

Beyond the observation sync failure already documented in #3091, v13.9.1 ships with **two additional missing files** that break independent features.

---

### Symptom

On every worker startup, the Chroma backfill pipeline logs:

```
[ERROR] [CHROMA_SYNC] Failed to initialize backfill resources
  ResolveMessage: Cannot find module '../sqlite/SessionStore.js'
  from '...13.9.1/scripts/worker-service.cjs'
```

### Root cause

In `worker-service.cjs`:

```js
var vC;
function c2() {
  return vC || (vC = l2()("../sqlite/SessionStore.js").SessionStore), vC;
}

// called at worker startup:
static async backfillAllProjects(e) {
  let r, n;
  try {
    const s = c2();      // ← throws ResolveMessage — SessionStore.js missing from bundle
    r = e ?? new s;
    n = new Ta("claude-mem");
  } catch (s) {
    g.error("CHROMA_SYNC", "Failed to initialize backfill resources", {}, s);
    // ...
  }
}
```

`sqlite/SessionStore.js` exists as bundled code in 13.8.1 but was moved to an external module in 13.9.1 without being included in the bundle output.

**Note:** in practice `backfillAllProjects` is always called with the existing `dbManager.getSessionStore()` as argument `e`, so `new s` is never reached — but `c2()` itself throws before `r = e ?? new s` even evaluates, causing the entire backfill to be skipped.

### Workaround (until 13.9.2)

Create `sqlite/SessionStore.js` at both plugin cache paths:

```js
// sqlite/SessionStore.js
import { Database } from 'bun:sqlite';
import { join } from 'path';
import { homedir } from 'os';
import { mkdirSync, readFileSync } from 'fs';

function resolveDataDir() {
  if (process.env.CLAUDE_MEM_DATA_DIR) return process.env.CLAUDE_MEM_DATA_DIR;
  const base = join(homedir(), '.claude-mem');
  try {
    const cfg = JSON.parse(readFileSync(join(base, 'settings.json'), 'utf-8'));
    const env = cfg.env ?? cfg;
    if (env.CLAUDE_MEM_DATA_DIR) return env.CLAUDE_MEM_DATA_DIR;
  } catch {}
  return base;
}

export class SessionStore {
  db;
  constructor(dbPath) {
[...]

PR #3102, opened 2026-07-01
fix: worker-service missing sqlite/SessionStore.js + observations/files.js (#3092)

bionicbutterfly13:
## Summary

Fixes Bug 1 from #3092: the Chroma backfill pipeline failed on every worker startup with `Cannot find module '../sqlite/SessionStore.js'`.

`ChromaSync.ts` loads `SessionStore` and `parseFileList` through a runtime `createRequire(import.meta.url)(...)` call instead of a static `import`, intentionally, so the cmem-sdk (tsup) build doesn't follow them and drag `bun:sqlite` into SDK consumers. But esbuild's worker-service bundle in `scripts/build-hooks.js` doesn't follow that same indirection either — it only inlines modules reached through a literal `require(...)` call, not one made via a locally-created `require` reference — so `worker-service.cjs` shipped without either module. Every worker startup logged `Backfill failed (non-blocking)` from the resulting `Cannot find module` error, and Chroma vector-search backfill silently never ran.

- `scripts/build-hooks.js`: after the existing `worker-service.cjs` build step, add a build pass that emits both modules as loose files next to the bundle (`plugin/sqlite/SessionStore.js`, `plugin/sqlite/observations/files.js`), matching the relative path `ChromaSync.ts`'s lazy `require()` expects at runtime.
- `package.json`: add `plugin/sqlite` to the npm `files` whitelist — without this the fix builds and tests clean locally but the published package still wouldn't ship the emitted files.
- `tests/worker-service-lazy-sqlite-modules.test.ts`: regression test that resolves both modules the same way `ChromaSync.ts` does at runtime. Verified locally that it reproduces the exact `#3092` error when the fix is reverted, and passes with it applied.
- `CHANGELOG.md`: `[Unreleased]` entry per the contribution guidelines in `docs/public/development.mdx`.

Out of scope: Bug 2 from #3092 (the `.install-version` marker / "runtime not yet set up" message) is a separate, architecturally distinct issue — it's about the Claude Code marketplace install path never invoking the npx installer that writes that marker, not a build-pipeline gap. Left for a follow-up so this PR stays scoped to one concern.

## Test plan

- [x] `bun test tests/worker-service-lazy-sqlite-modules.test.ts` — 4/4 pass with the fix applied
- [x] Confirmed the test fails with the exact `#3092` error text when `plugin/sqlite/` is removed (proves it's a real regression guard, not a tautology)
- [x] Verified via isolated `npm pack --dry-run --json` that the new `plugin/sqlite` `files` entry recursively packs the nested `plugin/sqlite/observations/files.js`
- [ ] Full `npm run build && npm run typecheck && bun test` — not run locally (heavy toolchain: tree-sitter-cli, onnxruntime-node, Chroma); relying on this repo's own CI (`.github/workflows/ci.yml`) to run these on the PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+## Bug Fix
+
+**Worker's Chroma backfill failed on every startup — `sqlite/SessionStore.js` and `sqlite/observations/files.js` missing from the bundle (#3092).**
+
+`ChromaSync.ts` loads `SessionStore` and `parseFileList` through a runtime `createRequire(import.meta.url)(...)` call instead of a static `import`, so the cmem-sdk (tsup) build doesn't follow them and drag `bun:sqlite` into SDK consumers. But the worker's esbuild bundle doesn't follow that same indirection either, so `worker-service.cjs` shipped without these two modules — every worker startup logged `Backfill failed (non-blocking)` from a `Cannot find module '../sqlite/SessionStore.js'` error, and Chroma vector-search backfill silently never ran.
+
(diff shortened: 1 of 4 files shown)


thedotmack:
Closing in favor of the newer static-import/bundling direction in #3116 and #3146. Emitting loose SQLite sidecars preserves an obsolete runtime loading pattern.

Issue #3107, opened 2026-07-01
Chroma vector sync broken on 13.9.x: worker-service.cjs requires ../sqlite/*.js which the build never emits


KaizenPrompt:
### TL;DR

On **13.9.0/13.9.1**, the built `plugin/scripts/worker-service.cjs` lazy-`require()`s two sibling modules that are **not produced by the build and not shipped in the package**:

- `../sqlite/SessionStore.js` → `plugin/sqlite/SessionStore.js`
- `../sqlite/observations/files.js` → `plugin/sqlite/observations/files.js`

Because `plugin/sqlite/` does not exist, the worker throws `Cannot find module '../sqlite/SessionStore.js'` and **Chroma vector sync degrades on every observation** (observations still persist to SQLite, but `[CHROMA] SDK chroma sync failed, continuing without vector search` fires for each one, so semantic search is effectively broken). `13.8.1` is not affected — it never referenced those modules. This looks like a **regression introduced with the 13.9.0 `claude-mem/sdk` rework**.

### Environment

- `claude-mem` plugin **13.9.1** (installed via the Claude Code marketplace, `thedotmack`; `plugin/package.json` → `"version": "13.9.1"`).
- Worker runtime: **Bun 1.3.13**, launched as `bun worker-service.cjs --daemon`.
- OS: Windows 11. MCP client: Claude Code (CLI v2.1.116).
- Backend: SQLite (`~/.claude-mem/claude-mem.db`) + Chroma via `uvx chroma-mcp`.

### Symptom (from `~/.claude-mem/logs/claude-mem-YYYY-MM-DD.log`)

At worker startup (once):

```
[ERROR] [CHROMA_SYNC] Failed to initialize backfill resources ResolveMessage:
  Cannot find module '../sqlite/SessionStore.js' from
  'C:\...\plugins\marketplaces\thedotmack\plugin\scripts\worker-service.cjs'
[ERROR] [CHROMA_SYNC] Backfill failed (non-blocking) {}
```

Then, on **every** observation:

```
[ERROR] [CHROMA] SDK chroma sync failed, continuing without vector search {obsId=…, type=discovery, title=…}
```

Net effect: memory is written to SQLite, but nothing is embedded/indexed into Chroma → semantic search returns nothing / degrades silently.

### Root cause

`worker-service.cjs` loads the two modules through a runtime `createRequire`, deliberately kept **out** of the worker bundle (dead-weight stripping, see #2584). In the minified output (13.9.1):

[...]

PR #3108, opened 2026-07-01
fix(build): emit plugin/sqlite runtime modules required by worker-service

KaizenPrompt:
## What

`scripts/build-hooks.js` now builds the two runtime modules that `worker-service.cjs` lazy-requires — `SessionStore.ts` and `observations/files.ts` — into `plugin/sqlite/`, using the same esbuild options and `external` list as the worker.

## Why

`worker-service.cjs` loads these via `createRequire("../sqlite/SessionStore.js")` and `createRequire("../sqlite/observations/files.js")`, kept external from the worker bundle (#2584). But no build step emitted them, and `plugin/sqlite/` isn't tracked in git or shipped in the npm tarball. On a shipped 13.9.x install the worker therefore throws `Cannot find module '../sqlite/SessionStore.js'` on first embed, and Chroma vector sync degrades on **every** observation (`[CHROMA] SDK chroma sync failed, continuing without vector search`). SQLite persistence still works, which masks the regression. Full details, repro and logs in #3107. `13.8.1` was unaffected (it never referenced these modules); this is a regression from the 13.9.0 `claude-mem/sdk` rework, and it's still present on `main` (13.9.2).

## How it's safe / self-contained

The dependency closure of `SessionStore.ts` (and `observations/files.ts`) only imports Node built-ins (`fs`, `path`, `crypto`, `os`, `url`, `child_process`) and `bun:sqlite`; no third-party runtime package. With `bun:sqlite` kept external (resolved natively at runtime, since the worker runs under Bun), esbuild bundles each module self-contained — no `plugin/node_modules` needed.

## Verification

```
$ node scripts/build-hooks.js
  …
  🔧 Building sqlite runtime modules...
  ✓ plugin/sqlite/SessionStore.js built (76.88 KB)
  ✓ plugin/sqlite/observations/files.js built (0.69 KB)

# contract check, loaded exactly as the worker does (createRequire, under bun):
$ bun -e '
  const { createRequire } = require("node:module");
  const r = createRequire(require("path").resolve("plugin/scripts/worker-service.cjs"));
  console.log(typeof r("../sqlite/SessionStore.js").SessionStore);        // function
  console.log(typeof r("../sqlite/observations/files.js").parseFileList); // function
'
function
function
```

Before this change: `[CHROMA_SYNC] Failed to initialize backfill resources: Cannot find module '../sqlite/SessionStore.js'`.
After (verified on a live 13.9.x worker with these artifacts in place): `[CHROMA_MCP] Connected to chroma-mcp successfully` / `[CHROMA_SYNC] Smart backfill complete`, and zero `SDK chroma sync failed` on subsequent observations.

## Note on committed artifacts

This branch commits the two generated files under `plugin/sqlite/` (consistent with `plugin/scripts/*.cjs` being tracked). If you'd rather not track built `plugin/` artifacts, the `scripts/build-hooks.js` change alone is sufficient — it regenerates them on every build. Happy to drop the artifacts from the PR if you prefer.

## Follow-up (not in this PR)

Consider a build-time guard that parses the built `worker-service.cjs` for `require("../sqlite/…")` targets and asserts each output exists, to prevent a future lazy sibling from silently shipping broken (same spirit as `scripts/check-sdk-bundle.cjs`).


thedotmack:

PR #3110, opened 2026-07-01
fix: preserve install markers for Codex plugin roots

dongyeop-cmyk:
## Summary
- resolve install markers against nested `plugin/` roots so marketplace installs and Codex-installed plugin roots share the same marker path
- write the install marker into the durable marketplace plugin root after successful install and repair flows
- cover the marketplace-root marker behavior with setup-runtime and install wiring regressions

## Verification
- bun test tests/setup-runtime.test.ts tests/install-non-tty.test.ts tests/plugin-version-check.test.ts

### DIFF
--- a/src/npx-cli/commands/install.ts
+++ b/src/npx-cli/commands/install.ts
@@ -1603,7 +1603,6 @@ async function runInstallCommandInner(options: InstallOptions, summary: InstallS
             } finally {
               stopHeartbeat();
             }
-            writeInstallMarker(cacheDir, version, bunVersion, uvVersion);
           }
           return `Runtime ready (Bun ${bunVersion}, uv ${uvVersion}) ${pc.green('OK')}`;
         },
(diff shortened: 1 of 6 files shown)


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3604 ([plan-16] Canonical Install Identity & Bundle Integrity)** as **needs rework**: the diagnosis is right, but the item noted below blocks it. If you'd like to keep it moving, address that against the design in the plan doc; otherwise the fix will be absorbed into the plan-16 bundle.

> step 3 (`.install-version` written after success at the resolved root); P1: `cacheDir` out of scope → TS build fails

thedotmack:
Second pass — routing holds at **#3604 (plan-16, Canonical Install Identity & Bundle Integrity)**: resolving the install marker against the nested `plugin/` root so marketplace and Codex-installed roots agree on one marker path is the "one resolvable plugin root / doctor and status disagreeing" defect the master owns. Not a duplicate of #3113 — that one changes what `version-check.js` does when the marker is missing, this one changes where the marker is written and read. The `cacheDir` P1 flagged in review is still open and breaks the TS build, so that needs fixing before it can be considered.

Issue #1447, opened 2026-03-22
Worker startup race condition causes 2 SessionStart hook errors on every first session


Pepsi1978:
## Bug Description

On every first Claude Code session start (after a reboot or when no worker daemon is running), the claude-mem SessionStart hooks produce 2 errors. Immediately closing and reopening Claude Code (second session) produces 0 errors because the worker daemon is already running.

## Root Cause

There is a race condition between the MCP server startup and the SessionStart hooks:

1. Claude Code starts → MCP servers initialize → claude-mem MCP server (`mcp-server.cjs`) starts
2. MCP server detects worker not running → starts worker daemon on port 37777
3. SessionStart hooks run → `worker-service.cjs start` also tries to start the worker → **fails with "port 37777 in use"**
4. `worker-service.cjs hook claude-code context` may also fail because the worker is still stabilizing

## Evidence from logs

```
[09:41:45.312] [INFO ] Claude-mem search server started
[09:41:45.357] [ERROR] Worker not available {"workerUrl":"http://127.0.0.1:37777"}
[09:41:45.540] [INFO ] Starting worker daemon    ← TWO parallel starts!
[09:41:45.540] [INFO ] Starting worker daemon
[09:41:45.666] [INFO ] HTTP server started {host=127.0.0.1, port=37777, pid=1619}
[09:41:45.667] [INFO ] Worker started
[09:41:45.669] [ERROR] ✗ Worker failed to start Failed to start server. Is port 37777 in use?
```

On second session start (worker already running):
```
[09:41:56.979] [INFO ] Worker already running and healthy  ← No errors
```

## Expected Behavior

The `worker-service.cjs start` command should:
1. Check if the worker is already running/starting on the target port
2. If another process already bound the port, verify it's actually the worker (health check)
3. If health check passes, treat it as success (exit 0) instead of error (exit 1)

## Suggested Fix

In the worker start logic, when the port is already in use:
```javascript
// Instead of immediately failing:
// throw new Error("Failed to start server. Is port 37777 in use?")

// Check if the existing process IS the worker:
const isWorkerHealthy = await checkHealth("http://127.0.0.1:37777/api/health");
if (isWorkerHealthy) {
  console.log("Worker already running and healthy (started by MCP server)");
  process.exit(0);  // Success, not error
}
```

## Workaround

[...]

PR #3112, opened 2026-07-01
fix(worker): detect stale/orphaned port socket on startup and fail with actionable guidance

KaizenPrompt:
## Problem

When a worker exits uncleanly it can leave a **stale, orphaned `LISTEN` socket** holding its port — especially on Windows, where the socket can outlive its now-dead PID.

The daemon's duplicate gate uses `isPortInUse()`, which **on Windows is health-based** (`fetch(/api/health)`) and therefore blind to a *bound-but-dead* socket. So the gate passes → `server.listen()` rejects with `EADDRINUSE` → the daemon dies with a bare `Failed to start server. Is port <p> in use?` and no path to recovery. **Memory capture then stops silently** until the user manually edits `CLAUDE_MEM_WORKER_PORT` or reboots.

### Observed on a real install (claude-mem 13.9.x, Windows 11)

A worker died uncleanly and its socket stayed held by the now-dead PID:

```
[ERROR] [SYSTEM] ✗ Worker failed to start Failed to start server. Is port 37777 in use?
```

`Get-NetTCPConnection` confirmed the port was in `Listen` state owned by a PID that no longer existed (orphaned socket), while `isPortInUse()` reported the port free (its `/api/health` probe got no response). The port stayed blocked for hours; recovery required manually setting `CLAUDE_MEM_WORKER_PORT` to a different port. Meanwhile the dashboard kept serving stale data with no indication that ingestion had stopped.

## Root cause

`isPortInUse()` is the daemon's "ground truth" duplicate gate, but on Windows it only asks whether a *healthy* worker answers `/api/health`. That cannot distinguish:

1. **port free** (no listener) — proceed,
2. **healthy duplicate** (a live worker answers) — defer, and
3. **stale socket** (port bound at the OS level, nothing healthy behind it).

Case 3 slips through the gate; `server.listen()` then throws `EADDRINUSE` and the daemon exits with an unactionable error and no recovery.

## Fix

Add `isPortBindable()` — a real bind probe (`net.createServer`) that returns `false` whenever `server.listen()` *would* hit `EADDRINUSE`, regardless of health. Use it as a **third startup gate**:

- If the port isn't bindable and no healthy worker answers → **retry briefly** (orphaned sockets often clear on their own).
- If a healthy worker claims the port meanwhile → **defer to it** (`exit 0`, same as the existing duplicate gate).
- Otherwise → `exit 1` with **actionable guidance** (reboot to release the socket, or set `CLAUDE_MEM_WORKER_PORT` to a free port) instead of a cryptic `EADDRINUSE`.

This respects the existing "port = single-owner mutex" design — **no port failover, no new invariants** — it only makes the third case observable and self-explanatory.

## Tests

- Behavioral coverage for `isPortBindable` (bindable / `EADDRINUSE` / other error) in `tests/infrastructure/health-monitor.test.ts`.
- Source-level assertions for the new daemon stale-socket guard in `tests/services/worker-daemon-port-race.test.ts`, matching the existing `#1447` port-race test style.
- `tsc --noEmit`: clean.

> Note: `health-monitor.test.ts` has 5 pre-existing failures on Windows — the `isPortInUse`/`waitForPortFree` mocks assume the non-Windows `net` code path, so they fail on the `win32` `fetch` branch. These reproduce on unmodified `main` and are unrelated to this change.

### DIFF
--- a/src/services/infrastructure/HealthMonitor.ts
+++ b/src/services/infrastructure/HealthMonitor.ts
@@ -51,6 +51,29 @@ export async function isPortInUse(port: number): Promise<boolean> {
   });
 }
 
+/**
+ * True bind-availability probe: attempts an actual bind (net.createServer) on
+ * the given host and reports whether it succeeded.
+ *
+ * This differs from isPortInUse() in an important way. On Windows isPortInUse()
+ * is health-based — it only asks whether a *healthy* worker answers /api/health
+ * — so it cannot see a port that is bound at the OS level with nothing serving
+ * behind it. That happens when a worker exits uncleanly and leaves a stale,
+ * orphaned LISTEN socket that can outlive its now-dead PID. isPortBindable()
+ * catches exactly that case: it returns false whenever server.listen() would
+ * reject with EADDRINUSE, regardless of whether anything healthy is listening.
+ */
+export async function isPortBindable(port: number, host: string = '127.0.0.1'): Promise<boolean> {
+  return new Promise((resolve) => {
+    const probe = net.createServer();
+    probe.once('error', () => resolve(false));
+    probe.once('listening', () => {
+      probe.close(() => resolve(true));
+    });
+    probe.listen(port, host);
+  });
+}
+
 async function pollEndpointUntilOk(
   port: number,
   endpointPath: string,
--- a/src/services/infrastructure/HealthMonitor.ts
+++ b/src/services/infrastructure/HealthMonitor.ts
@@ -52,23 +52,28 @@ export async function isPortInUse(port: number): Promise<boolean> {
 }
 
 /**
- * True bind-availability probe: attempts an actual bind (net.createServer) on
- * the given host and reports whether it succeeded.
+ * Probe whether `port` can actually be bound on `host`. Resolves `null` on a
+ * successful bind, otherwise the failing error code (e.g. `'EADDRINUSE'`,
+ * `'EADDRNOTAVAIL'`, `'EACCES'`).
  *
  * This differs from isPortInUse() in an important way. On Windows isPortInUse()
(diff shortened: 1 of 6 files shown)


KaizenPrompt:
Thanks for the reviews — both findings were valid and are addressed in `5774351`.

**Greptile (P1) — preserve non-conflict bind errors.** The bind probe collapsed every `listen()` failure to a boolean, so `EADDRNOTAVAIL` (invalid `CLAUDE_MEM_WORKER_HOST`) or `EACCES` were misreported as a stale socket. Replaced `isPortBindable()` with `probePortBind()`, which resolves the **failing error code** (or `null` on success). Only `EADDRINUSE` is treated as a stale socket; any other code now surfaces as itself.

**Codex (P2) — surface the failure to the spawner.** The guard previously ran inside the detached daemon (`stdio: 'ignore'`), so its actionable error was invisible: the spawner's post-spawn health check timed out and returned `'warming'`, masking the failure. **Moved the check into the user-facing parent** (`ensureWorkerStarted`), before `spawnDaemon`:

- stale socket (`EADDRINUSE`, no healthy worker) → retry briefly, defer to any worker that becomes healthy meanwhile, otherwise return **`'dead'`** with actionable guidance;
- genuine bind error (`EADDRNOTAVAIL`/`EACCES`) → return `'dead'` surfacing that error.

The daemon-side gate is removed — `worker-service.ts` reverts to upstream.

Tests: `probePortBind` (`null` / `EADDRINUSE` / `EADDRNOTAVAIL`+`EACCES`) and source-level assertions for the parent-side guard (probe runs before `spawnDaemon`, `'dead'` on stale socket). `tsc --noEmit` clean.

thedotmack:

PR #3113, opened 2026-07-01
fix: force reinstall on missing .install-version marker instead of nagging (#3092)

bionicbutterfly13:
## Summary

Fixes Bug 2 from #3092: the Setup hook printed "runtime not yet set up - run: npx claude-mem@latest install" on every single Setup run for installs made purely through the Claude Code plugin marketplace, since `.install-version` is only ever written by the npx CLI installer, which those installs never invoke. The suggested fix didn't even work — `npx claude-mem@latest` resolves to a stale, version-mismatched npm release.

This PR went through two rounds of review. The first (committed as [9607a33](https://github.com/thedotmack/claude-mem/commit/9607a33)) simply self-healed the marker by writing it whenever missing. Review caught a real regression: `ContextBuilder.ts` deliberately deletes `.install-version` after an `ERR_DLOPEN_FAILED` to signal "native module needs rebuilding" — but `ensurePluginDependencies()` only ever reinstalled when `node_modules` was entirely missing, never when it existed but contained a broken native binary. The naive self-heal would have silently erased that recovery signal on the very next Setup run without actually fixing anything.

The corrected fix (this PR's current state):

- `version-check.js` now checks marker presence **before** calling `ensurePluginDependencies()`, and passes `force: true` when the marker is missing.
- `ensurePluginDependencies(pluginRoot, { force })` now clears any existing `node_modules` first and always reinstalls when forced — instead of trusting `node_modules`'s mere existence — and returns an outcome string instead of void.
- The marker is only self-healed with the current version when that reinstall actually reports `'installed'`. Any other outcome (bun not found, install failed) falls through to the original hint, so a genuinely broken install still has a visible, actionable signal.

This preserves the `ContextBuilder.ts` native-rebuild recovery path while fixing the false nag for the common marketplace-only-install case.

## Test plan

- [x] Three new cases in `tests/plugin-version-check-ensure-deps.test.ts` (reusing the file's existing fake-`bun` infrastructure): forces a reinstall and clears stale `node_modules` when the marker is missing, self-heals the marker after a successful forced reinstall, does not self-heal (keeps the hint) if the forced reinstall fails
- [x] Reverted two tests from an earlier commit in `tests/plugin-version-check.test.ts` — that file's `beforeEach` pre-seeds `node_modules` specifically to keep dependency-install logic out of scope for marker-format tests, which no longer holds once a missing marker can trigger `force:true`; real coverage now lives in the ensure-deps file instead
- [x] Confirmed via `git stash` on just `plugin/scripts/version-check.js` that the new tests correctly fail against the old code (empty stderr where an install diagnostic was expected)
- [x] `npm run typecheck` passes
- [ ] Full `npm run build` and complete `bun test` suite — not run locally (relying on this repo's CI, same as #3102)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 **Setup hook always printed "runtime not yet set up" for marketplace-only installs (#3092).**
 
-`.install-version` is only ever written by `npx claude-mem@latest install`. Installs made purely through the Claude Code plugin marketplace never run that installer, so the marker never existed and `version-check.js` printed the same misleading "runtime not yet set up - run: npx claude-mem@latest install" hint on every Setup hook — even though `ensurePluginDependencies()` earlier in the same script had already materialized the real runtime deps for that exact install. Worse, `npx claude-mem@latest` resolves to a stale, version-mismatched npm release, so the suggested fix didn't even work.
+`.install-version` is only ever written by `npx claude-mem@latest install`. Installs made purely through the Claude Code plugin marketplace never run that installer, so the marker never existed and `version-check.js` printed the same misleading "runtime not yet set up - run: npx claude-mem@latest install" hint on every Setup hook. Worse, `npx claude-mem@latest` resolves to a stale, version-mismatched npm release, so the suggested fix didn't even work.
 
-`version-check.js` now self-heals a missing marker by stamping it with the currently-installed version instead of emitting that hint, since by the time this code runs the runtime genuinely is set up.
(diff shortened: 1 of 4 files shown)


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3604 ([plan-16] Canonical Install Identity & Bundle Integrity)** as a **bundle candidate**: the fix is sound and is slated to be folded into the plan-16 bundle PR with author credit (`Co-authored-by` + `Closes` reference), sequenced by the step noted below. It stays open until that bundle lands.

> step 3 (dependencies materialize with the copy; marker only after success)

thedotmack:
Second pass — routing holds at **#3604 (plan-16, Canonical Install Identity & Bundle Integrity)**. Heads up that half of this has since landed: #3972 (`d959572bd`, "guard plugin deps on completeness, not node_modules existence") merged the completeness-based `ensurePluginDependencies` guard and took `tests/plugin-version-check-ensure-deps.test.ts` with it, and main now explicitly preserves the marker rather than depending on its deletion for retry. What is still unfixed on main is the original Bug 2: `version-check.js` line ~344 still emits "runtime not yet set up - run: npx claude-mem@latest install" whenever `.install-version` is absent, which is every marketplace install. Keeping this open, but it needs a rebase down to just that nag/force-reinstall change.

PR #3114, opened 2026-07-01
feat(observations): reversible dismiss to hide observations from surfacing

20alexl:
## What

Activates the reserved `observation_feedback` table (`signal_type='dismissed'`) so a user can hide an observation from **proactive surfacing** — the file-context banner, search, and session-start injection — **without deleting it**. It stays fully retrievable by id via `get_observations`. Fully reversible (undismiss).

## How

- **SessionStore**: `dismissObservation` (idempotent guarded insert — the table has no `UNIQUE(observation_id, signal_type)`), `undismissObservation` (DELETE), `isDismissed`.
- **Read filter** (shared `NOT_DISMISSED_SQL`): spliced into `getObservationsByFilePath`, `searchObservations` (FTS + filter-only), and `queryObservations`/`queryObservationsMulti`. **Unconditional but a no-op until a dismiss row exists** (byte-identical otherwise). Id-addressed reads (`getObservationById`/`getObservationsByIds`) are intentionally left unfiltered — dismiss means hide-from-surfacing, not delete.
- **Worker**: `POST` / `DELETE /api/observations/:id/dismiss`, WRITE-gated behind new `CLAUDE_MEM_ALLOW_DISMISS` (default `false`; clear 403 when off). Dismiss verifies existence first (404), avoiding an FK 500.
- **MCP**: `observation_dismiss(id, reason?)` / `observation_undismiss(id)` relaying the worker.
- **Schema**: `schema.sql` *declared* `observation_feedback` but no migration ever created it (schema.sql reads as the reference doc; table creation lives in `SessionStore`), so `createObservationFeedbackTable` (schema v36) adds it idempotently in the additive `createPendingMessagesTable` style.

## Notes / honest caveats

- Activates a **reserved-but-unwired** table; **default-off** write; **no new runtime dependencies**; pure TypeScript, `tsc --noEmit` clean.
- The table wasn't actually created at runtime, so this necessarily wires its creation — additive and idempotent (`IF NOT EXISTS` + version guard), not a data migration.
- When `CLAUDE_MEM_CHROMA_ENABLED`, `/api/search`'s Chroma strategy hydrates via the (intentionally unfiltered) `getObservationsByIds`, so a dismissed observation can still surface through Chroma-ranked search; the SQLite/FTS strategy is filtered. This follows the "don't filter by-id reads" design — happy to also post-filter Chroma results or skip syncing dismissed docs if you'd prefer.
- If `observation_feedback` is reserved for something specific (tier routing?), say the word and I'll rework or close this.

## Testing

- `tsc --noEmit`: 0 errors
- `bun test` on the touched areas (sqlite, context, worker routes, new suites): 251 pass / 0 fail
- New tests cover: hidden from file-context/search/session-start; still returned by id; undismiss restores; dismiss idempotent; write no-op + 403 when the setting is off; read filter byte-identical with zero dismiss rows.


20alexl:
good catch — fixed in 24f0a07. DELETE now resolves getObservationById(id, platformSource) first and 404s when the observation is not visible in the requesting platform scope, mirroring the POST path. added a regression test for the cross-platform undismiss case (18/18 on the dismiss suites, tsc clean).

thedotmack:
Closing in the PR consolidation pass — superseded: landed on community-edge as 9aac37ba0 (reversible observation dismiss). The analysis here fed the plan-21 design (#3609); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

Issue #2389, opened 2026-05-09
Bug: `/api/search` ignores `platformSource`, causing Codex MCP search to return Claude/Cursor/null-source memories


ThorsHammer666369:
## Before submitting

- [x] I searched existing issues and pull requests and confirmed this is not a
      duplicate.
- [x] I verified this against the official upstream code, not only a local fork.
- [x] I confirmed the database contains distinct `platform_source` values.
- [x] I reproduced the issue through the worker HTTP API.

---

## Summary

`/api/search` accepts a `platformSource` query parameter, but the search layer does
not apply it. As a result, calls like:

```text
/api/search?query=ManualCodexSessionImport&platformSource=codex
/api/search?query=ManualCodexSessionImport&platformSource=claude-code

return the same result set.

This causes Codex MCP memory search to return Claude / Claude Code / Cursor /
null-source memories when the caller expects Codex-only memory.

Anfield88:
Confirming this is still present on **v13.2.0**, and adding a few related gaps found while wiring up a shared Claude Code ⇄ Codex CLI setup (one `~/.claude-mem` DB, both agents ingesting):

**1. The MCP `search` tool doesn't even expose the param.** Beyond `/api/search` ignoring `platformSource`, the `search` MCP tool's input schema only offers `query, limit, project, type, obs_type, dateStart, dateEnd, offset, orderBy` — there's no `source`/`platformSource` argument at all. So even once the API honors it, the MCP tool would need the parameter plumbed through for callers (and the model) to use it.

**2. Passive context injection has no source scoping either.** The SessionStart context block (Claude Code) and the `AGENTS.md` injection (Codex) pull recent observations regardless of `platform_source`, and there's no env knob to scope/exclude — e.g. a `CLAUDE_MEM_CONTEXT_SOURCES=claude` (allowlist) or `CLAUDE_MEM_CONTEXT_EXCLUDE_SOURCES=codex` (denylist). For a "Claude Code is primary, Codex contributes but shouldn't dilute Claude's recall" setup, there's currently no lever.

**3. Provenance isn't surfaced.** The `search` result table doesn't show `platform_source`, so users can't even *see* which results came from which agent without going to SQLite directly. Adding a source column (or a `📍source` marker) would help a lot.

[...]

PR #3116, opened 2026-07-02
Fix empty folder CLAUDE.md and Chroma observation sync

Garcia6l20:
## Symptoms

Two user-visible failures this fixes:

1. Generated folder CLAUDE.md files always contained an empty
   `<claude-mem-context>` wrapper (or no useful content).
2. Every observation vector sync logged and failed:

   ```
   [🔮 CHROMA ] SDK chroma sync failed, continuing without vector search {obsId=1526, type=discovery, title=...} {}
   ```

## Root causes

**Empty folder CLAUDE.md** — `updateFolderClaudeMdFiles` queried the by-file
search with the *absolute* folder path, but observations persist
*project-relative* file paths. The SQL `LIKE %path%` never matched, so every
folder returned zero results and produced empty/skeleton content. Two
follow-on bugs compounded it: `isFolder` was dropped before reaching
`SessionSearch` (folder queries degraded to exact-file lookups), and the
metadata fallback was gated on `platformSource`, so a by-file/folder lookup
returned nothing whenever Chroma was empty, stale, or mid-backfill.

**Chroma sync failures** — `syncObservation` loaded `parseFileList` via a lazy
`createRequire(import.meta.url)('../sqlite/observations/files.js')`. In the
esbuild worker bundle `import.meta.url` is the bundle path, so the relative
require resolved to a nonexistent location and threw "Cannot find module"
before any sync ran. `parseFileList` is pure (imports only `logger`, no
`bun:sqlite`), so it is now a static import — inlined into the worker bundle,
still absent from the SDK bundle. The backfill path had the same shape: it
called `loadSessionStoreCtor()` unconditionally even when a store was
injected, so it now only falls back to the lazy require when none is given.

## Changes

- Query folder observations by project-relative path (commit 1).
- Forward `isFolder` through `HybridSearchStrategy.findByFile`.
- Return the complete metadata set on file/folder lookups regardless of
  Chroma state (Chroma only reorders, never drops exact matches).
- Static `parseFileList` import; conditional `SessionStore` lazy load.
- Never write an empty/skeleton folder CLAUDE.md (commit 2).

## Verification

- `build:sdk` + `check:sdk-bundle` green — SDK bundle still free of
  `bun:sqlite` and other shell-only deps.
- `tests/utils/claude-md-utils.test.ts`: 69 pass, added regression coverage
  for no-observation folders and empty-content-leaves-file-untouched.

### DIFF
--- a/src/services/sync/ChromaSync.ts
+++ b/src/services/sync/ChromaSync.ts
@@ -2,18 +2,18 @@
 import { ChromaMcpManager } from './ChromaMcpManager.js';
 import { ChromaSyncState, ProjectWatermarks } from './ChromaSyncState.js';
 import { ParsedObservation, ParsedSummary } from '../../sdk/parser.js';
-// cmem-sdk: keep SessionStore + parseFileList off the SDK's import graph.
-// Both come from the SQLite layer (`bun:sqlite`). The SDK only uses the
-// constructor + ensureCollectionExists + close() surface of ChromaSync,
-// so a TYPE-ONLY import is sufficient — value-level uses (`new
-// SessionStore()` / parseFileList(...)) are loaded lazily inside the
-// SQLite-only methods that need them. Plan §3 anti-pattern: do NOT add
-// `bun:sqlite` to the SDK bundle externals — fix the import chain.
+// cmem-sdk: keep SessionStore off the SDK's import graph. It comes from the
+// SQLite layer (`bun:sqlite`). The SDK only uses the constructor +
+// ensureCollectionExists + close() surface of ChromaSync, so a TYPE-ONLY
+// import is sufficient — the value-level use (`new SessionStore()`) is loaded
+// lazily inside the SQLite-only backfill methods that need it. Plan §3
+// anti-pattern: do NOT add `bun:sqlite` to the SDK bundle externals — fix the
+// import chain.
 import type { SessionStore as SessionStoreType } from '../sqlite/SessionStore.js';
 import { logger } from '../../utils/logger.js';
 import { ChromaUnavailableError } from '../worker/search/errors.js';
 import { normalizePlatformSource } from '../../shared/platform-source.js';
-import type * as SqliteFilesModule from '../sqlite/observations/files.js';
+import { parseFileList } from '../sqlite/observations/files.js';
 
 type SessionStore = SessionStoreType;
 type SessionStoreCtor = new () => SessionStoreType;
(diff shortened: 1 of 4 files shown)


Garcia6l20:
Re: **P2 Unscoped Metadata Fallback**

The "instead of only the active source" framing does not hold for the folder CLAUDE.md path. That path is already cross-platform on the healthy (chroma-up) path, and the folder caller never passes a `platformSource`:

- Metadata fetch: `SessionSearch.findByFile` runs with `platformSource` undefined, so the #2389 source-scoping clause is skipped and observations from every platform source match.
- Chroma-up path: `buildObservationWhereFilter(project, undefined)` adds no `platform_source` filter, and `getObservationsByIds` with an undefined `platformSource` does not filter either.

So when Chroma is populated, a folder CLAUDE.md already aggregates observations across all platform sources for the project. The fallback change only makes the degraded path (Chroma empty/stale/mid-backfill) return that same cross-platform set instead of an empty result.

Before this change, the down-chroma + no-`platformSource` case returned `[]` — that is exactly the empty `<claude-mem-context>` bug this PR fixes. Re-gating the fallback on `platformSource` would reintroduce that bug and leave the degraded path inconsistent with the chroma-up path.

If platform-scoping folder CLAUDE.md is desired, the correct place is the caller: have `updateFolderClaudeMdFiles` pass the active `platformSource` on the by-file request, which scopes both paths consistently. That is a pre-existing gap independent of this fix, so I've left it out of scope here.

carterwayneskhizeine:
Thanks for picking this up — the `parseFileList` static import is the right call.

One latent issue worth flagging on the `loadSessionStoreCtor` side of this fix: the lazy `createRequire(import.meta.url)('../sqlite/SessionStore.js')` path is now only dead-code-suppressed by the `storeOverride ?? new (loadSessionStoreCtor())()` short-circuit in `ensureBackfilled` / `backfillAllProjects`. It still exists, still references a path that doesn't exist on disk in the installed worker bundle, and will throw `MODULE_NOT_FOUND` again the moment anyone calls either method without passing `storeOverride` (e.g. a future refactor of `worker-service.ts`, or an SDK consumer exercising the backfill surface).

Two ways to make this airtight — both very small:

1. **Delete the lazy path entirely** if the contract is "caller must always provide a store". Just throw a clear `Error('storeOverride is required')` instead of the silent rethrow. `lazyCreateRequire` and `loadSessionStoreCtor` both go away.

2. **Make the lazy path actually work** by switching to a dynamic `import()` with a computed path (so esbuild/tsup can't statically follow it into the SDK bundle, but Bun resolves it at runtime in the worker). Then no caller dependency required.

Happy to send a patch for either if useful. Otherwise just a heads-up in case the maintainer wants to clean up the now-dead code in the same release.

thedotmack:

> folder-CLAUDE.md half → step 4 (writer never creates empty wrapper; relative-path query); Chroma half superseded by 38fc189b5

thedotmack:
Second pass — confirming this one sits under **#3608 (plan-20, Project Identity & Injection Scope)**, on the strength of the folder CLAUDE.md half: `updateFolderClaudeMdFiles` querying with an absolute path against project-relative stored paths, plus the dropped `isFolder` flag, is exactly the "folder CLAUDE.md stubs" symptom the master covers. The ChromaSync half (the `createRequire` resolve failing inside the esbuild bundle) is already fixed on main by 38fc189b5, so that part of the diff is dead weight and will conflict. Leaving open, but it wants a rebase down to just the path-resolution and `isFolder` changes before it can land.

Issue #3126, opened 2026-07-04
Backfill crashes on missing module '../sqlite/observations/files.js'; search query param ignored (v13.10.0, v13.9.3)


socialwealth:
## Summary

Chroma vector-sync backfill crashes on a missing module, and separately, the `search` MCP tool ignores its `query` parameter entirely. Both confirmed in **v13.10.0** (current `latest` on npm) and reproduced in **v13.9.3**. Raw observation/summary/prompt capture to SQLite is unaffected — this only breaks retrieval.

## Bug 1: Backfill crashes — `Cannot find module '../sqlite/observations/files.js'`

**Repro:** Have a project with observations already stored where the Chroma watermark is behind the total count (e.g. after a worker restart). On next session init, the worker attempts a backfill and throws immediately:

```
[ERROR] [CHROMA_SYNC] Backfill failed {project=<name>} ResolveMessage: Cannot find module '../sqlite/observations/files.js' from '.../plugin/scripts/worker-service.cjs'
[ERROR] [CHROMA_SYNC] Backfill failed for project: <name> Backfill failed: ResolveMessage: Cannot find module '../sqlite/observations/files.js' from '.../plugin/scripts/worker-service.cjs'
```

**Root cause confirmed, not a local install issue:** I pulled the pristine package directly —

```
npm pack claude-mem@13.10.0
tar -tzf claude-mem-13.10.0.tgz | grep -i observ   # zero results
tar -tzf claude-mem-13.10.0.tgz | grep -iE "sqlite|\.cjs$|\.js$"
```

The published tarball ships `plugin/scripts/worker-service.cjs` (which contains the `require('../sqlite/observations/files.js')` call in the backfill path) but **no `sqlite/` directory at all** anywhere in the package. This is a build/bundling gap in the published artifact itself, not a corrupted local copy — I checked three separate local install locations (plugin marketplace cache, and both the 13.9.3 and 13.10.0 versioned caches) and none contain the file either.

Before a restart clears the queue, individual (non-backfill) real-time sync attempts to Chroma also fail, logged as:
```
[ERROR] [CHROMA] SDK chroma sync failed, continuing without vector search {obsId=N, type=..., title=...} {}
```
[...]

Issue #3138, opened 2026-07-05
[plan-root-cause] Holistic worker/runtime reliability fixes for lifecycle, Chroma, Windows, packaging, migrations, observer loops, project scoping, and data integrity


thedotmack:
## Problem

The current backlog has many symptom issues that keep reappearing because the runtime contracts are split across several partial implementations:

- Worker liveness is inferred from a mix of PID files, port probes, health/version endpoints, spawn cooldowns, marketplace/cache paths, and hook exit behavior.
- Chroma is a long-lived sidecar with its own uvx spawn, dependency install, timeout, process-tree, and metadata contract, but it can outlive the worker or write to the same data dir from multiple writers.
- Windows hosts still have multiple command-generation paths, some host configs run POSIX shell strings through PowerShell, and several spawn sites leak console windows or inherited handles.
- Install and runtime packaging can disagree about the executable root, marker root, cache root, marketplace root, and bundled dependency set.
- SQLite migrations and data-pipeline invariants are not tested as direct upgrades across real historical schema states.
- Observer no-op/error behavior can be treated as poison, causing respawn loops, transcript accumulation, and dropped observations.
- Project identity is inconsistent across git roots, cwd-relative subprojects, worktrees, submodules, Chroma metadata, and distilled memory namespaces.

This plan is for branch `plan/root-cause-holistic-fixes`. It is intentionally a root-cause plan-master, not another batch of one-off fixes. Child issues should be redirected here only after this issue exists and each child is mapped to a concrete phase and verification row.

Snapshot source: `/tmp/claude-mem-three-track/backlog-full.json`, `/tmp/claude-mem-three-track/backlog-summary.json`, `/tmp/claude-mem-three-track/open-prs.json`.

Snapshot metadata: `thedotmack/claude-mem`, 116 open issues, 134 open PRs, captured at `2026-07-05T04:56:22Z`.

Related existing plan masters:

- #2779 - existing spawn-contract templating plan. Keep as source material for host shell contracts and close only when this plan explicitly supersedes its child coverage.
- #2782 - existing data-pipeline integrity plan. Keep as source material for content/session/data transparency and close only when this plan explicitly supersedes its child coverage.
- #2785 - provider/extensibility roadmap. Mostly out of scope here except where provider/auth/env bugs break the runtime contract.

### 1. Worker lifecycle and liveness need one source of truth

[...]

PR #3146, opened 2026-07-05
fix(chroma): static-import parseFileList so worker backfill stops crashing on missing sqlite/observations/files.js (#3126, #3107)

joao-oliveira-softtor:
## Summary

Fixes the Chroma **backfill crash** reported in #3126 (Bug 1) and the packaging gap diagnosed in #3107: at runtime the worker throws

```
[CHROMA_SYNC] Backfill failed: Cannot find module '../sqlite/observations/files.js'
  from '.../scripts/worker-service.cjs'
[CHROMA] SDK chroma sync failed, continuing without vector search
```

Observations keep landing in SQLite, but never reach Chroma — semantic/vector search is silently dead for any project with a watermark gap.

## Root cause

`ChromaSync` loaded `parseFileList` through a lazy `createRequire('../sqlite/observations/files.js')`:

```ts
import type * as SqliteFilesModule from '../sqlite/observations/files.js'; // type-only, erased at build
...
_filesHelper = req('../sqlite/observations/files.js'); // dynamic createRequire
```

esbuild (which builds `worker-service.cjs`) **does not follow dynamic `createRequire` calls**. Since the only value-level reference was that dynamic require, esbuild emitted **neither** an inlined copy of the helper **nor** a sidecar `scripts/sqlite/observations/files.js` file. The published bundle ships `worker-service.cjs` with a `require('../sqlite/observations/files.js')` in the backfill path but no `sqlite/` directory anywhere in the package — exactly as #3126/#3107 observed.

The lazy require was justified by a comment stating `files.ts` is SQLite-coupled (*"`bun:sqlite` in the import chain"*). That's not accurate: `files.ts` only imports `logger`, and its full transitive chain (`files.ts → logger → paths/hook-io`) is **`bun:sqlite`-free**. So the dynamic require protected against a coupling that doesn't exist.

## Fix

Replace the type-only import + lazy `createRequire` loader with a **plain static import**:

```ts
import { parseFileList } from '../sqlite/observations/files.js';
```

Because the chain is `bun:sqlite`-free, this is safe for **both** bundles — the worker (esbuild) and the SDK (tsup) — and lets esbuild inline the helper directly into `worker-service.cjs`. `SessionStore` stays a type-only import (it genuinely pulls in `bun:sqlite`).

Net: **−32 / +13**, one file.

## Verification

Bundled `ChromaSync.ts` with the worker build's exact esbuild options (`--bundle --platform=node --format=cjs --target=node18` + the same externals as `scripts/build-hooks.js`) and inspected the output:

- ✅ **No** `require()` of `sqlite/observations/files` remains (only esbuild's inlined-module header comment).
- ✅ `createRequire` shim: gone.
- ✅ `parseFileList` body inlined into the bundle (the `File list is not JSON` branch is present).
- ✅ esbuild build succeeds (the `import.meta` warning is the same benign one the official build already suppresses via `logLevel: 'error'`).

## Not covered here

This PR fixes **only Bug 1** (backfill crash / missing module). The two other items in #3126 are out of scope and left for follow-ups:

1. The `[CHROMA]` real-time error logging serializing `Error` to `{}` (should log `message`/`stack`).
2. Bug 2 — the `search` MCP tool ignoring its `query` parameter.

### DIFF
--- a/src/services/sync/ChromaSync.ts
+++ b/src/services/sync/ChromaSync.ts
@@ -2,38 +2,23 @@
 import { ChromaMcpManager } from './ChromaMcpManager.js';
 import { ChromaSyncState, ProjectWatermarks } from './ChromaSyncState.js';
 import { ParsedObservation, ParsedSummary } from '../../sdk/parser.js';
-// cmem-sdk: keep SessionStore + parseFileList off the SDK's import graph.
-// Both come from the SQLite layer (`bun:sqlite`). The SDK never calls the
-// SQLite-only methods of ChromaSync, so a TYPE-ONLY import is sufficient —
-// the value-level use (parseFileList(...)) is loaded lazily inside the
-// methods that need it. Plan §3 anti-pattern: do NOT add `bun:sqlite` to
-// the SDK bundle externals — fix the import chain.
+// cmem-sdk: SessionStore stays a TYPE-ONLY import — it pulls in `bun:sqlite`
+// and the SDK never calls the SQLite-only methods of ChromaSync.
 import type { SessionStore as SessionStoreType } from '../sqlite/SessionStore.js';
 import { logger } from '../../utils/logger.js';
 import { ChromaUnavailableError } from '../worker/search/errors.js';
 import { normalizePlatformSource } from '../../shared/platform-source.js';
-import type * as SqliteFilesModule from '../sqlite/observations/files.js';
+// parseFileList is a pure helper (JSON.parse over a text column) — its import
+// chain (files.ts -> logger -> paths/hook-io) is `bun:sqlite`-free, so a plain
+// static import is safe for BOTH bundles. It was previously loaded via a lazy
+// createRequire('../sqlite/observations/files.js') on the mistaken assumption
+// that files.ts was SQLite-coupled; esbuild does not follow that dynamic
+// require, so the worker build emitted neither the inlined helper nor the
+// sidecar file, and the backfill path crashed at runtime with
+// `Cannot find module '../sqlite/observations/files.js'`. See #3126 / #3107.
+import { parseFileList } from '../sqlite/observations/files.js';
 
 type SessionStore = SessionStoreType;
-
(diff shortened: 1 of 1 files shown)


joao-oliveira-softtor:
### Related PRs / issues

Cross-referencing prior work on this bug so a maintainer can pick an approach:

- **#3102** (`fix: worker-service missing sqlite/SessionStore.js + observations/files.js`, closes #3092) — targets the same missing-module crash but takes the **opposite approach**: it keeps the runtime `createRequire(...)` and instead makes the build **emit both modules as loose sidecar files** (`plugin/sqlite/observations/files.js` + `plugin/sqlite/SessionStore.js`) and whitelists `plugin/sqlite` in `package.json`'s `files`. It also covers `SessionStore.js`, which this PR does not (on `main`/13.10.1 `SessionStore` is imported statically elsewhere, so esbuild already bundles it; the `SessionStore.js` gap was a 13.9.x state).
- **#3116** — separately touches Chroma observation sync.
- Issues: #3126 (backfill crash + `search` query bug), #3107 (build never emits `sqlite/*.js`), #3092 (`SessionStore.js` variant), #3138 (maintainer's umbrella plan).

**Trade-off vs #3102:** this PR is the smaller diff (−32/+13, one file) and removes the dynamic require entirely so esbuild inlines the helper — no build-step or `package.json` `files` change needed, nothing extra to keep in sync. #3102 is more defensive (sidecar emit) and also fixes the `SessionStore.js` case. The two are mutually exclusive on `ChromaSync.ts`; happy to rebase/adapt if you prefer #3102's shape or want this extended to cover `SessionStore` the same way.

thedotmack:
Closing in the PR consolidation pass — superseded: landed as 38fc189b5 (ship sqlite runtime modules as sidecar). The analysis here fed the plan-16 design (#3604); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

PR #3143, opened 2026-07-05
[plan-root-cause] Holistic worker/runtime reliability fixes

thedotmack:
Refs #3138

## Phase summary
- Phase 0: recorded root-cause execution intake and working constraints.
- Phase 1: routed worker lifecycle checks through the shared health oracle and configured worker endpoint.
- Phase 2: guarded Chroma persistent writer ownership to prevent duplicate sidecar writers.
- Phase 3: centralized Windows process/spawn environment shims.
- Phase 4: repaired marketplace install/runtime root behavior.
- Phase 5: hardened historical SQLite migrations and settings writes.
- Phase 6: bounded observer provider context to stop poison-loop amplification.
- Phase 7: used canonical project identity for observation paths and metadata.
- Phase 8: completed available release gates; external platform smoke gates skipped where unavailable.

## Key commits
- 089e54e78 docs: record root-cause execution intake
- 3695d1160 fix: use health oracle and configured worker endpoint
- 49e5dadfd fix: guard chroma persistent writer ownership
- 92997115e fix: centralize windows spawn shims
- 339d3eca4 fix: repair marketplace runtime root
- b20024b77 fix: harden sqlite migrations and settings writes
- 016668338 fix: bound observer provider context
- 107ef946e fix: use canonical project identity

## Verification
- npm run typecheck: passed
- npm run build: passed
- npm test: passed; 2249 pass, 15 skip, 0 fail, 5769 expect() calls
- npm run lint:spawn-env: passed
- npm run lint:hook-io: passed
- git diff --check: passed
- npm run e2e:server:docker: passed; Docker server beta E2E passed for run=1783239819-650

## Skipped external smoke gates
- Windows native process behavior smoke: skipped because this environment is macOS-only.
- Linux host smoke: skipped because no separate Linux host was available.
- WSL2 smoke: skipped because WSL2 is unavailable in this environment.

PR #3151, opened 2026-07-06
feat: bound observer conversation history for OpenAI-compatible providers

ruddro-roy:
## Problem

`OpenAICompatibleProvider` (and `GeminiProvider` through it) re-sends the whole `session.conversationHistory` on every `query()`. Each observation turn appends a user message carrying the full tool payload, up to ~32k chars of `<parameters>` + `<outcome>`. Nothing is ever trimmed, so a session with N tool uses re-transmits every prior payload on every later turn:

- O(N^2) input tokens on per-token billing (OpenRouter, Gemini, custom gateways). A 50-observation burst at ~3k tokens per payload re-sends roughly 4M input tokens where ~1M would do.
- Requests grow until they cross the model's context window. `classifyGeminiError` / OpenRouter classification treat that as unrecoverable, so the memory session dies mid-workflow.

## Why pruning is safe

A raw payload is only needed for the single turn that converts it into `<observation>` blocks. After `processAgentResponse()` persists those to SQLite, the payload is dead weight: the assistant messages already in the history are the compressed record of that work, and the Claude Code transcript remains the durable source of truth for replay (same reasoning as the SessionMessageBuffer design note).

## Change

New `pruneProcessedObservationPayloads()` runs before each observation and summary query. It replaces observation payloads outside the recent window with a compact stub that preserves role, position, tool name, and timestamp:

```
<observed_from_primary_session pruned="true">
  <what_happened>Read</what_happened>
  <occurred_at>2026-07-06T...</occurred_at>
  <note>Tool payload removed after it was recorded as observations. Do not re-describe this event.</note>
</observed_from_primary_session>
```

Never touched: the init/continuation prompt (observer instructions), summary-mode prompts, all assistant responses, messages at or under 600 chars, and the trailing 8 messages (~4 exchanges, enough verbatim context for batching and dedup). Role alternation and chronology are preserved because messages are shrunk in place, never removed. The function is idempotent; stubs are skipped on later passes.

Per-request input becomes bounded (init + stubs + recent window + current payload) instead of growing without limit, so long sessions keep working instead of hitting the context ceiling.

`ClaudeProvider` is untouched: its history lives inside the SDK subprocess, so this array is only bookkeeping there.

## Tests

`tests/history-pruning.test.ts`: stubbing preserves tool name/timestamp/role; init prompt, assistant messages, and recent window are never modified; summary prompts and small messages are skipped; idempotent second pass; 40-exchange history shrinks by >5x with message count and order unchanged; short histories untouched.

Full suite: 2208 pass / 0 fail, `typecheck:root` clean. Logger standards satisfied (debug log on prune).

## Relationship to #3096 and the core-dev revert (066826a5)

I am aware client-side context management has history here: #3096 removed `truncateHistory()` because a hardcoded 20-message cap dropped messages at 12k tokens and mislabeled it as cost prevention, and the later `boundObserverHistory` attempt from #3143 was reverted on core-dev as a regression of #3096 (it reintroduced a hardcoded 12-message cap plus content sanitization of live payloads).

This change shares neither mechanism:

- No message is ever dropped or reordered. Count, chronology, and role alternation are preserved; messages are shrunk in place.
- No size heuristic, token estimate, or message-count cap decides anything. The only trigger is a state transition: the payload was already converted into observations and persisted by `processAgentResponse()`. Until that happens, a payload is never touched (the recent window covers in-flight exchanges).
- Nothing the model still needs is removed. The assistant messages that summarize each payload stay verbatim, the observations live in SQLite, and the transcript remains the durable source, which is the same recovery contract `SessionMessageBuffer` already documents.
- The init prompt with the observer's instructions is never modified.

The failure #3096 fixed was "a second system guessing at the model's context window". This is not a guess about capacity; it is removing redundant copies of data that the system has, by its own design, already compressed and stored.

### DIFF
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -2,6 +2,7 @@ import { DatabaseManager } from './DatabaseManager.js';
 import { SessionManager } from './SessionManager.js';
 import { logger } from '../../utils/logger.js';
 import { buildInitPrompt, buildObservationPrompt, buildSummaryPrompt, buildContinuationPrompt } from '../../sdk/prompts.js';
+import { pruneProcessedObservationPayloads } from './history-pruning.js';
 import type { ActiveSession, ConversationMessage } from '../worker-types.js';
 import { ModeManager } from '../domain/ModeManager.js';
 import type { ModeConfig } from '../domain/types.js';
@@ -206,6 +207,12 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
     });
 
     session.conversationHistory.push({ role: 'user', content: obsPrompt });
+
+    // Stub out payloads already converted to stored observations so the
+    // request below stays bounded instead of re-sending every prior tool
+    // dump (see history-pruning.ts).
+    pruneProcessedObservationPayloads(session.conversationHistory);
+
     session.lastPromptSentAt = Date.now();
     session.lastGeneratorSource = 'ingest';
     const obsResponse = await this.query(session.conversationHistory, config);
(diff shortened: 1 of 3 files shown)


thedotmack:

> step 3 (history trimmed, never emptied; compact older payloads)

thedotmack:
Second pass over the open PR list — routing confirmed: this stays under **#3606 (plan-18, Observer Response Pipeline)**. It bounds `session.conversationHistory` for the OpenAI-compatible path by stubbing already-processed payloads, which is squarely the "conversationHistory unbounded / double-appended" defect the master tracks. Not a duplicate of #2957 — that one caps the Agent SDK path, which has no resendable history to slice, so the two are complementary halves of the same step. Nothing on main supersedes it; still the cleanest existing implementation of the history bound.

PR #3154, opened 2026-07-06
docs: add Ollama Setup instructions and example .env

dane11stanisic-a11y:
Adds a new Ollama Setup section to configuration docs, including example ~/.claude-mem/.env values and test commands for local Ollama usage.


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#2785 ([plan-12] Provider & Extensibility Roadmap)** as **needs rework**: the diagnosis is right, but the item noted below blocks it. If you'd like to keep it moving, address that against the design in the plan doc; otherwise the fix will be absorbed into the plan-12 bundle.

> Ollama docs; P1: references a helper script that does not exist

thedotmack:

PR #3156, opened 2026-07-06
Add GenericCliProvider for CLI-based LLM backends (Kimi Code)

Dandi007:
## What

Adds `GenericCliProvider`, a new AI provider that routes claude-mem's observation/summary compression to any CLI-based LLM agent (flagship: **Kimi Code**), instead of `ClaudeProvider` + Claude OAuth. This isolates claude-mem's background compression cost from the user's Claude subscription.

## Why

By default, every `PostToolUse` observation and session summary consumes Claude subscription quota — mixed into the same bill as Claude Code, with no attribution. Users who also pay for a separate CLI agent (Kimi Code) can now route compression there, keeping Claude quota for interactive work.

## How

- New `GenericCliProvider` mirrors `OpenAICompatibleProvider`'s session lifecycle (init / observation / summary loop, history maintenance, token accounting).
- `query()` spawns `<cmd> -p "<prompt>" --output-format text`, parses `content` + `session_<uuid>` from stdout/stderr.
- **Transient retry** (3×, exponential backoff) for spawn failures; **content-layer retry** (2×) for non-XML responses, with `<skip_summary>` fallback so the worker never poisons the session.
- Session resume via `<cmd> -r <session> -p ...` keeps context cheap across observations.
- Registered through `CLAUDE_MEM_PROVIDER=generic-cli` (selection guards + `validProviders` union + settings routes + logger component).

## Notable fix included

kimi `--output-format text` intermittently writes the resume marker to **stderr** (~11% of invocations observed). `queryOnce` parses `sessionId` from stdout first, then falls back to stderr. Without this, those sessions lose the real session id, keep the synthetic placeholder stamped by `startSession`, and hit `Session "kimi-..." not found` on every subsequent resume. (Added test: "init: captures sessionId from STDERR when kimi emits resume marker there".)

## Verification

- `bun test tests/worker/` — **245 pass / 0 fail** (no regressions); `tsc --noEmit` clean.
- New tests: `GenericCliProvider.test.ts` (44 tests), `provider-registration.test.ts`.
- End-to-end on NUC: real `kimi -p` spawn with marker on stderr now captures `session_<uuid>` (before-fix: `null`); observation/summary resume succeeds instead of `exit 1`.

## Files

- `src/services/worker/GenericCliProvider.ts` — provider (config, spawn, parse, transient retry, content retry, session lifecycle)
- `src/services/worker-service.ts`, `worker-types.ts`, `SettingsRoutes.ts`, `SessionRoutes.ts`, `SettingsDefaultsManager.ts`, `logger.ts` — registration + wiring
- `tests/worker/GenericCliProvider.test.ts`, `provider-registration.test.ts`
- `README.md`, `docs/providers/generic-cli.md` — docs

## Backward compatibility

Default behavior unchanged. `GenericCliProvider` only activates when `CLAUDE_MEM_PROVIDER=generic-cli`; existing `claude` / `gemini` / `openrouter` users are untouched.

## Out of scope

- Real token accounting (kimi text mode returns no usage; `chars/4` estimate only, feeds history truncation, not telemetry).
- Tool-subcommand-level `CLAUDE_MEM_SKIP_TOOLS` (still exact-name match).


Dandi007:
Closing — opening this against my own fork instead. Apologies for the noise.

PR #3165, opened 2026-07-07
feat(worker,server): track advisor tool calls across sessions

s546126:
## Summary
- Adds a durable, viewable record of every Claude Code `advisor` tool call: which session called it, the model that served it (`advisorModel`), a lightweight pointer to the forwarded context (turn's user message + transcript path:line — not a full transcript copy), and the full advice text verbatim.
- **Capture happens in the Stop hook via a transcript scan** (`shared/advisor-transcript.ts`), because the advisor is a *server-side* tool: transcripts record it as a `server_tool_use` block with the paired `advisor_tool_result` embedded in the assistant message, so PostToolUse hooks never fire for it. (Verified empirically on two machines: 23 sessions containing advisor calls, zero ever reached the hook pipeline.) The scan is single-pass, per-turn (`currentTurnOnly`), skips error results and sidechain entries, and dedupes on the `srvtoolu_...` id.
- **Classic worker + SQLite**: new `advisor_calls` table with `UNIQUE(tool_use_id)` (replayed scans are no-ops), written via `POST /api/advisor-calls`, read via `GET /api/advisor-calls` / `GET /api/advisor-call/:id`, and surfaced as a new card type in the React viewer (with an advisor-model badge), wired into the existing feed pagination + SSE live-update plumbing.
- **Server runtime (SQLite or Postgres)**: recorded as `tool_use` agent_events (`generate:false` — advice is already model output, no compression job), listed via `GET /v1/advisor-calls`. Both backends share one serializer (`shared/advisor-call-view.ts`); the SQLite repo dedups by payload `toolUseId` at read time since its `create()` has no idempotency key.

## Test plan
- [x] `tsc --noEmit` (root + viewer) — clean
- [x] `bun test` — full suite passes (2 unrelated pre-existing timeout flakes reproduce identically on `main` with this branch stashed: `plugin-distribution` npm-pack >5s, `spawn-contract-windows`)
- [x] New unit tests: transcript parser (pairing, error/sidechain/malformed-line skipping, per-turn attribution, `currentTurnOnly`, system-reminder stripping), `advisor_calls` persistence + `UNIQUE(tool_use_id)` dedup, `stringifyAdvice` shapes
- [x] **End-to-end against a real transcript** containing 7 advisor invocations (4 succeeded, 2 errored, 1 interrupted): parser extracts exactly the 4 successful calls with model/line/timestamp/turn-context; all 4 store; full replay dedupes to 0; pagination orders by `occurred_at_epoch`
- [x] `npm run build` — plugin bundles regenerate cleanly


s546126:
Redesigned the capture layer in b97c0a9 after post-review verification disproved the original approach.

**What was wrong:** the first commit captured advisor calls in the PostToolUse hook. But the advisor is a server-side tool — transcripts record it as a `server_tool_use` block (id `srvtoolu_...`) with the result embedded in the assistant message as an `advisor_tool_result` block, so it never goes through the client tool-execution loop and PostToolUse never fires. Checked two long-running claude-mem installations: 23 sessions contained advisor calls in their transcripts, and the hook pipeline had captured zero of them. The feature as first written would never have recorded anything.

**The fix:** capture moved to the Stop hook, which scans the just-ended turn's slice of the transcript (`shared/advisor-transcript.ts`) and posts extracted calls to the worker (or records them as agent_events on the server runtime). Dedup on the `srvtoolu` id makes replays no-ops at every layer. Bonus from reading the real block shapes: we now also record `advisorModel` — which model actually served each advice.

Verified end-to-end against a real transcript with 7 advisor invocations: the parser extracts exactly the 4 that succeeded (2 errored, 1 was interrupted), stores them, and a full replay inserts 0 duplicates.

thedotmack:

> advisor-call capture (new table/routes/UI); product decision

thedotmack:
Routing this to **#2785 (plan-12, Provider & Extensibility Roadmap)** — this is a net-new capture surface plus viewer UI, not a defect fix, which is that cluster's remit. The empirical finding is the interesting part: the advisor is a server-side tool, so transcripts record it as a `server_tool_use` block and PostToolUse hooks never fire (23 sessions, zero reaching the hook pipeline), which is why the Stop-hook transcript scan is the only capture point. Leaving open; no duplicate. At ~2k lines across SQLite, Postgres, routes and the viewer it will want splitting before it can land.

PR #3166, opened 2026-07-07
fix(env): preserve CLAUDE_CODE_USE_FOUNDRY through the SDK env sanitizer

jimweller:
## Summary

`sanitizeEnv` (`src/supervisor/env-sanitizer.ts`) strips every `CLAUDE_CODE_*` / `CLAUDECODE_*` variable before the SDK subprocess is spawned, unless the name is in `ENV_PRESERVE`. The provider selectors `CLAUDE_CODE_USE_BEDROCK` and `CLAUDE_CODE_USE_VERTEX` are already preserved; `CLAUDE_CODE_USE_FOUNDRY` (Microsoft Foundry) is not, so it gets stripped and the subprocess loses provider selection on Foundry-backed installs.

This adds `CLAUDE_CODE_USE_FOUNDRY` to `ENV_PRESERVE`, alongside the sibling selectors.

## Why

Foundry users set `CLAUDE_CODE_USE_FOUNDRY=1` the same way Bedrock/Vertex users set their selector. Without preservation, the observer/worker SDK subprocess cannot reach the Foundry endpoint and observation capture breaks. This is parity with the existing Bedrock/Vertex handling, not a new mechanism.

## Changes

- `src/supervisor/env-sanitizer.ts`: add `CLAUDE_CODE_USE_FOUNDRY` to `ENV_PRESERVE`
- `tests/supervisor/env-sanitizer.test.ts`: dedicated preserve test plus an assertion in the selective-preserve test

## Safety

- Layer 1 (`BLOCKED_ENV_VARS` in `EnvManager.ts`) is unchanged; no credential or effort var is affected.
- The `CLAUDE_CODE_EFFORT_LEVEL` / `CLAUDE_CODE_ALWAYS_ENABLE_EFFORT` strip is untouched.

## Test

`bun test tests/supervisor/env-sanitizer.test.ts` -> 12 pass / 0 fail.


thedotmack:
Closing in the PR consolidation pass — superseded: landed as 7508de5d6 (preserve Azure AI Foundry env). The analysis here fed the plan-19 design (#3607); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

PR #3173, opened 2026-07-08
fix(search): supplement Chroma-empty categories from SQLite FTS5 (CJK queries return 0 observations)

JunichiAJISAKA:
## Problem

Searches with CJK (e.g. Japanese) query text return **0 observations** from the unified `/api/search` path and the MCP `search` tool, even when SQLite FTS5 has exact matches for the query.

Reproduction (real-world DB, ~45k observations, Chroma configured):

- `GET /api/search?query=<Japanese term>` → `(0 obs, 1 sessions, 99 prompts)` (term redacted — project-specific; any Japanese term that appears in observation text reproduces it)
- Direct FTS5 check on the same DB: `SELECT COUNT(*) FROM observations_fts WHERE observations_fts MATCH '"<Japanese term>"'` → 20 rows
- English queries (e.g. `query=estimate`) return observations normally

Reproduced on both v13.2.0 and v13.10.2.

## Root cause

For a Japanese query, Chroma's top-100 nearest neighbors are dominated by Japanese **user prompts** (observations are written in English by the SDK), so the `observations` bucket of the categorized Chroma result comes back empty while `prompts` does not.

The existing SQLite fallback never fires in this situation, in **both** duplicated pipelines:

- `SearchOrchestrator.executeWithFallback` only fell back when `platformSource` was set **and** all three categories were empty (`isEmptyResult`).
- `SearchManager.performChromaSemanticSearch` (the inline pipeline that `/api/search` actually uses) only has the platform-scoped zero-ids fallback.

Since prompts still match, "all categories empty" is never true and the observations category stays empty forever.

## Fix

After a successful Chroma search, any **requested category that came back empty is supplemented from SQLite FTS5** (per category). If *every* requested category is empty, the whole search falls back to SQLite — now regardless of `platformSource`.

- Implemented once in `SearchOrchestrator.supplementEmptyCategories` and reused from `SearchManager.search`, so the two pipelines share one fallback policy.
- Supplement queries default to the same **90-day recency window** the Chroma paths enforce via `filterByRecency` (an explicit `dateRange` is preserved), so supplemented categories keep the same time semantics.
- Telemetry: `search_performed` now reports `search_strategy: 'hybrid'` (per-category supplement) and `fallback_reason: 'chroma_zero_results'` (any zero-match fallback, including the previously mislabeled platform-scoped one, which used to report `chroma_error`). Enums stay closed; docs (`docs/public/telemetry.mdx`), the npx CLI transparency listing, and the scrub.ts comment are updated.
- Cleanups: shared `isCategoryRequested`/`SEARCH_CATEGORIES` helper replaces three hand-rolled copies of the category predicate; typed `copyCategory` helper replaces an `as any` merge cast; a stale debug log ("final result, no FTS5 fallback") is corrected.

## Deliberate behavior changes

1. **Unscoped Chroma zero matches no longer stay final.** The tests `keeps unscoped Chroma zero matches final` (added in 348d9ee4) pinned zero-result finality for unscoped searches; this PR inverts them. Rationale: with CJK queries the zero is a retrieval artifact, not ground truth — FTS5 provably has matches. The platform-scoped fallback introduced in 348d9ee4 is generalized rather than special-cased.
2. **Empty categories are filled with keyword matches on any query** (not just CJK). E.g. an English query whose sessions category was empty in Chroma may now return FTS5-matched sessions. This is a recall improvement but changes result composition.
3. **`build_corpus` with a query filter** goes through `orchestrator.search`, so a corpus that previously came back empty on zero semantic matches can now contain FTS5 keyword matches. If corpora should stay strictly semantic, I'm happy to gate the supplement behind an option (e.g. `strategyHint`).

## Known limitations (out of scope here)

- **Not a full CJK fix**: the FTS5 tables use the default `unicode61` tokenizer, so only delimited CJK tokens match; substrings inside unspaced Japanese prose still return nothing. A trigram tokenizer or per-doc_type Chroma queries would be the deeper fix.
- Comma-separated `type` values (`type=observations,sessions`) still return 0 results — pre-existing on main (`normalizeParams` array-splits and every category flag goes false); the supplement intentionally no-ops on that shape.
- The inline platform-scoped fallback in `performChromaSemanticSearch` still coexists with the orchestrator supplement (guarded so they never double-run). Consolidating them changes limit/order defaults for platform-scoped queries, so I left it for a follow-up.
- With explicit `orderBy=date_desc` + text format, supplemented rows participate in the cross-category date sort and can displace semantic rows within the limit window.

## Verification

- `bun test`: 2268 pass / 0 fail (new unit tests for both pipelines: supplement, full fallback, untouched-when-full, recency-window default + explicit dateRange, telemetry values)
- `npm run typecheck`: clean
- End-to-end on a real ~45k-observation DB: the failing Japanese query went from 0 to 20 observations; English queries unchanged (observations identical; sessions gained FTS5 supplement rows)

Bundles were regenerated with `scripts/build-hooks.js` (bun 1.3.11) per repo convention — happy to drop or regenerate them if you prefer building on your side.


thedotmack:

> step 5 (honest fallback on emptiness); overlaps #3399/#3285 — reconcile into one fallback policy; inverts pinned tests deliberately

thedotmack:
Routing this to **#3982 (plan-25, Search Read Path / FTS-CJK)** — this PR is the cluster's founding repro: Chroma's top-100 for a Japanese query is dominated by prompts, the `observations` bucket comes back empty, and neither `SearchOrchestrator.executeWithFallback` nor `SearchManager` supplements it from FTS5 even though the rows are sitting there.

Overlap worth resolving: #3285 fixes the same fallback in `SearchManager` for a different trigger (Chroma hits that all fall outside the recency window). Leaving both open; whichever becomes the plan-25 architectural fix should cover both conditions in one place rather than in two duplicated pipelines. Note the telemetry/scrub files in this diff look unrelated to the search fix and would be cleaner split out.

Issue #3163, opened 2026-07-07
Capture-side dedup: title-level boilerplate observations bypass content_hash and flood the recall window


di3str0:
## Summary

`content_hash` dedup on `observations` is effective for byte-identical rows, but it structurally **misses title-level boilerplate** — the same restated fact re-observed across many sessions with slightly different narrative wording (so the hash differs). These accumulate and dominate the recency-based recall window that gets injected at session start.

## Evidence (one real DB, ~45k observations)

- `content_hash`: 100% populated, **0 exact duplicates** — the hash path works perfectly.
- But **535 observations share a title with ≥1 other** (158 distinct repeated titles). Top offenders are session-opener / restated-fact observations re-emitted every session:
  - `"IQlex Project Uses Forgejo for Source Control"` ×18
  - `"DXVPN Feature Gap Analysis Initiated"` ×16
  - `"IQLex Comprehensive Upgrade Plan Initiated"` ×14
  - `"DB-masking Phase 2(b) FK-discovery deployed"` ×13
- `type=discovery` is ~59% of all observations; a large slice of the repeated titles are `discovery`-type restated facts.
- 212 observations have an **empty title** (mostly `type=bugfix`) — malformed/low-signal.

Net effect: with a recency-ordered recall window (`CLAUDE_MEM_CONTEXT_OBSERVATIONS`, semantic inject off), a meaningful fraction of the injected slots are spent on "project X uses Forgejo" boilerplate instead of live state — every session pays for it in context tokens.

## Proposed fixes (capture-side, where hash-dedup can't reach)

1. **(title, project) near-dup suppression at capture** — before persisting a `discovery` observation, if an observation with the same (or ≥ cosine-threshold similar) title already exists for that project, skip or coalesce it (bump a `relevance_count` instead of inserting a new row).
2. **Boilerplate/restated-fact suppression** — heuristic to drop "session initiated / project uses X / analysis request initiated" openers at capture (title matches a small set of restated-fact patterns, no new facts vs. the last summary).
3. **Reject empty-title observations** (or backfill a title) — 212 title-less rows are pure noise.

Optionally: a maintenance/prune command in the CLI that keeps the N newest per (project, title) group for `discovery`-type boilerplate, so existing DBs can be cleaned without hand-written SQL.

## Workarounds users have today

- Lowering `CLAUDE_MEM_CONTEXT_OBSERVATIONS` trims the noisy window but also drops signal.
[...]

PR #3176, opened 2026-07-08
fix(sqlite): skip empty-title observations at capture

winklemad:
## What & why

`content_hash` dedup only catches byte-identical observations, so it can't stop malformed rows from piling up. #3163 measured 212 empty-title observations in a real ~45k-row DB. They're pure noise that still take slots in the recency-ordered recall window injected at session start.

This is the smallest, no-ambiguity slice of #3163: stop empty-title observations at capture.

## Change

- `SessionStore.storeObservations` (the live capture path the summarizer uses) skips any observation whose title is null, empty, or whitespace-only, so the row is never written.
- The single-observation `storeObservation` wrapper requires a non-empty title (throws rather than returning an undefined id), keeping its `{ id: number }` contract.
- Its one caller, `POST /api/memory/save` (`MemoryRoutes`), now falls back to the text-derived title for empty **and** whitespace-only titles. Previously `title || fallback` let a whitespace-only title through, which would have hit the guard; `text` is validated non-empty, so the fallback is always valid.
- `importObservation` is untouched, so restoring an existing DB preserves historical rows.

## Testing

- `tests/sqlite/session-store-empty-title.test.ts`: batch skips empty/whitespace/null titles; the single-observation wrapper throws on an empty title and stores a titled one.
- `tests/worker/http/routes/memory-routes.test.ts`: a whitespace-only title falls back to the text-derived title, and a provided title is used (trimmed).
- SQLite suite: 72 pass. Route suite: 9 pass. `tsc --noEmit`: clean.

## Follow-ups

The bigger part of #3163 (coalescing repeated `(project, title)` boilerplate onto a counter) has real design choices: exact vs fuzzy title match, which types to coalesce, and the coalesce semantics. I left it out to keep this bounded. Happy to do it next if you let me know how you'd want the dedup to behave.

Refs #3163

### DIFF
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -1890,6 +1890,13 @@ export class SessionStore {
     overrideTimestampEpoch?: number,
     generatedByModel?: string
   ): { id: number; createdAtEpoch: number } {
+    // storeObservations skips empty-title rows, which would leave no id to return here.
+    // This wrapper stores exactly one observation, so require a title up front rather than
+    // returning an undefined id.
+    if (!observation.title || observation.title.trim() === '') {
+      throw new Error('storeObservation requires a non-empty title');
+    }
+
     const result = this.storeObservations(
       memorySessionId,
       project,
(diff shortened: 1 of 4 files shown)


thedotmack:

> step 5 (missing title rejected at write time)

winklemad:
Thanks @thedotmack — CI is green here (the only check is the skipped Mintlify docs deploy), so this is ready to land whenever it fits the plan-18 sequence. Appreciate the consolidation write-up.

thedotmack:

`MERGEABLE` / `CLEAN`, but CLEAN is only the skipped Mintlify check — not product CI. Skipping empty-title rows at capture is an opinionated filter (drop data from the recency window), not a crash/corruption RCA. The July hunks also sit on a SessionStore that has moved a long way since. Leaving open; not rebasing in this batch.

PR #3178, opened 2026-07-08
fix: reap children leaked by non-graceful worker deaths at boot

### LINKED ISSUE #3175: chroma-mcp processes orphan on session exit → accumulate, hold uv cache lock, bloat ~/.cache/uv
## Summary

claude-mem spawns a `uvx chroma-mcp` vector-DB server per session, but the plugin's `hooks.json` has **no `SessionEnd` teardown**. When a Claude Code session exits abnormally (killed, crashed, or force-quit), the `chroma-mcp` child process is reparented to launchd (`ppid=1`) and lingers indefinitely. Over time these accumulate.

## Evidence (real install, v13.10.2, macOS 15.3)

On one long-running install I found **13 orphaned processes** like:

```
uv tool uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 \
  chroma-mcp==0.2.6 --client-type persistent --data-dir ~/.claude-mem/chroma
```

- All had **`ppid=1`** (parent session long dead) — ages ranged from 2 to **55 days**.
- They each held **`~/.cache/uv/.lock`** open, which **blocks `uv cache prune`** entirely (`error: Timeout … waiting for lock … is another uv process running?`).
- Combined with never being reaped, they contributed to a **26 GB `~/.cache/uv`** on a disk that reached 100% full.

A live session's chroma-mcp correctly has the `claude` process as its parent — only the abandoned ones show `ppid=1`, which makes them easy to identify.

## Root cause

`plugin/hooks/hooks.json` defines `Setup`, `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, and `Stop` — but **no `SessionEnd`**. The `Stop` hook fires when the assistant finishes a turn, not when the session ends, so a session that dies never tears down the chroma-mcp server it started.

## Proposed fix

Either (or both):

1. **Add a `SessionEnd` hook** routing to a `worker-service` `session-end` handler that terminates the chroma-mcp associated with that session id.
2. **Reap on startup** — in the `SessionStart`/worker-startup path, kill stale `ppid=1` `chroma-mcp` processes (belt-and-suspenders for sessions that died without firing `SessionEnd`).

Since `plugin/hooks/hooks.json` and `plugin/scripts/*.cjs` are build outputs of `scripts/build-hooks.js` from `src/services/`, the change belongs in the TypeScript source + hook template rather than the generated files.

Happy to help test a patch on the affected install.

tiagogbarbosa:
## Problem

Fixes #3175 — orphaned `chroma-mcp` process trees accumulating indefinitely.

Field data from two Linux hosts (v13.x, worker runtime):

- Host A: 15 leaked `uvx chroma-mcp` + python pairs, oldest 21 days, ~4.2 GB RSS combined, one pair per day of use.
- Host B: 31 leaked pairs at the moment of cleanup.
- Each leaked tree also holds `~/.cache/uv/.lock`, which blocks `uv cache prune` (the cache-bloat side effect reported in the issue).

## Root cause (not the missing SessionEnd hook)

`chroma-mcp` is spawned by the **worker** (`ChromaMcpManager` singleton), not per Claude session, so a `SessionEnd` hook would not help. The graceful path already tears the tree down correctly (`performGracefulShutdown` → `chromaMcpManager.stop()` → tree-kill).

The leak happens when a worker dies **outside** the graceful path — SIGKILL, OOM, crash, or the `runShutdownSequence` hard deadline expiring before chroma teardown:

1. `ProcessRegistry` persists to `supervisor.json` across runs.
2. On the next boot, `initialize()` calls `pruneDeadEntries()`, which removes only entries whose pid is **dead**. The predecessor's still-alive chroma entry survives the prune.
3. The successor's first `registerProcess('chroma-mcp', ...)` overwrites that entry — the old pid is forgotten while the process keeps running. Permanent orphan, one tree per non-graceful death.

## Fix

New `reapLeakedProcesses(registry)` in `src/supervisor/shutdown.ts`, called from worker `start()` after `startSupervisor()` and **before** `server.listen`:

- At that point this run has registered nothing (no hook traffic yet), so every live registry record other than our own pid is a leak from a dead predecessor.
- Reuses the existing `signalProcess` / `waitForExit` primitives: SIGTERM → wait 5 s → SIGKILL → unregister. pgid-based group signaling reaches the re-parented `uv → python → chroma-mcp` grandchildren, same as the shutdown cascade.
- **PID-recycling guard**: the registry file survives system reboots, so a recorded pid may now belong to an unrelated process. Marker-typed entries (`chroma` → `chroma-mcp`) are verified against the live process cmdline first; on mismatch the entry is dropped **without** signaling anyone.
- Unlike `runShutdownCascade`, this never touches the PID file — the boot path owns it.
- Best-effort: failures log a warning and never block worker boot.

## Tests

Three new cases in `tests/supervisor/shutdown.test.ts`:

- prunes dead entries without signaling and reaps live leaked ones
- never touches the current process entry
- drops a recycled pid without killing it (real `sleep` bystander, cmdline mismatch)

`bun test tests/supervisor tests/services/worker-shutdown-sequence.test.ts tests/infrastructure/graceful-shutdown.test.ts`: 79 pass, 0 fail. `tsc --noEmit` clean.

### DIFF
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -17,7 +17,7 @@ import { logger } from '../utils/logger.js';
 import { ChromaMcpManager } from './sync/ChromaMcpManager.js';
 import { ChromaSync } from './sync/ChromaSync.js';
 import { openConfiguredSqliteDatabase } from './sqlite/connection.js';
-import { configureSupervisorSignalHandlers, getSupervisor, startSupervisor } from '../supervisor/index.js';
+import { configureSupervisorSignalHandlers, getSupervisor, reapLeakedProcesses, startSupervisor } from '../supervisor/index.js';
 import { sanitizeEnv } from '../supervisor/env-sanitizer.js';
 
 import { ensureWorkerStarted as ensureWorkerStartedShared, type WorkerStartResult } from './worker-spawner.js';
@@ -412,6 +412,24 @@ export class WorkerService implements WorkerRef {
     this.detectPreviousShutdown();
 
     await startSupervisor();
+
+    // Reap children leaked by a previous run that died outside the graceful-
+    // shutdown path (crash, SIGKILL, or shutdown-deadline expiry). Must run
+    // before server.listen: with no hook traffic yet this run has registered
+    // nothing, so every live registry record except our own pid is a leak
+    // from a dead predecessor (issue #3175).
+    try {
+      const reaped = await reapLeakedProcesses(getSupervisor().getRegistry());
+      if (reaped > 0) {
+        logger.info('SYSTEM', 'Reaped processes leaked by previous worker run', { reaped });
+      }
+    } catch (error: unknown) {
+      if (error instanceof Error) {
+        logger.warn('SYSTEM', 'Leaked-process reap failed (non-blocking)', {}, error);
+      } else {
+        logger.warn('SYSTEM', 'Leaked-process reap failed (non-blocking, non-Error)', { error: String(error) });
+      }
+    }
 
     await this.server.listen(port, host);
 
--- a/src/supervisor/index.ts
+++ b/src/supervisor/index.ts
@@ -8,6 +8,7 @@ import {
   type ProcessRegistry
 } from './process-registry.js';
 import { runShutdownCascade } from './shutdown.js';
+export { reapLeakedProcesses } from './shutdown.js';
 import { startHealthChecker, stopHealthChecker } from './health-checker.js';
 import { paths } from '../shared/paths.js';
 
--- a/src/supervisor/shutdown.ts
+++ b/src/supervisor/shutdown.ts
@@ -148,6 +148,135 @@ export function removeOwnedPidFile(pidFilePath: string, currentPid: number | nul
   }
 }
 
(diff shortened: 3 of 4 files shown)


thedotmack:

> step 1 (registry keyed by PID; kill predecessor before spawn); P1: pgid fallback with non-detached child

thedotmack:
Routing this to **#3602 (plan-14, Child Process Ownership)** — boot-time reaping of trees leaked by non-graceful worker death is one of that cluster's named gaps, and your field data (31 leaked pairs on one host, `~/.cache/uv/.lock` blocking `uv cache prune`) is the best evidence anyone attached to it.

Closing as superseded: this landed on `main` in 22879ef5, `fix(supervisor): reap chroma-mcp trees no worker owns at boot (refs #3905)`, which adds `src/supervisor/orphan-chroma-sweep.ts` wired into `src/supervisor/index.ts` at the same initialize point this PR targets, plus a `doctor` surface. Same fix, same insertion site. Thanks for the diagnosis — the "it isn't the missing SessionEnd hook, chroma-mcp is worker-scoped" call was correct and shaped the merged version.

PR #3184, opened 2026-07-10
fix: capture last assistant message when session ends mid-tool-call

girish-kanjiyani7:
## Problem

When a Claude Code session ends while the assistant is mid-tool-call, the Stop hook sends `lastAssistantMessage` as an **empty string** (not `undefined`). The handler's `!== undefined` check passed, so the empty value was used directly and the entire session summary was silently skipped — no summary, no observations for that session.

## Fix

1. **`src/cli/handlers/summarize.ts`** — change the `!== undefined` check to a truthiness check, so empty strings fall through to transcript parsing instead of being used as-is.
2. **`src/shared/transcript-parser.ts`** — add `synthesizeToolDescription()` as a true last resort: only used when the backward scan finds no assistant text anywhere in the transcript. Sessions with real prior text still return that text unchanged.

## Test plan

- [x] `npm run typecheck` passes (both root and viewer tsconfigs)
- [x] Verified sessions ending mid-tool-call now fall through to transcript parsing and produce a summary
- [x] Verified sessions with prior assistant text are unaffected (backward scan returns existing text before the synthesized fallback is reached)


girish-kanjiyani7:
Good catch — the checked-in worker bundle still had the old `!== undefined` gate. Regenerated `plugin/scripts/worker-service.cjs` via `npm run build` in ce176161; the bundle now carries the truthiness gate and the synthesized tool-use fallback. (`server-service.cjs` rebuild produced only nondeterministic churn in vendored dep code with no logic change, so it's left untouched.)

thedotmack:

> Stop-hook empty `lastAssistantMessage` truthiness fix; drop the committed .cjs and rebase

thedotmack:
Routing this to **#3606 (plan-18, Observer Response Pipeline)** — a session whose summary is silently skipped and never retried is the same "never drop work quietly" invariant that cluster enforces at the other end of the pipe.

Verified still live on current `main`: `src/cli/handlers/summarize.ts` still branches on `input.lastAssistantMessage !== undefined` (line 88), so an empty string takes the direct path and then trips the `!lastAssistantMessage` guard at line 108 and returns without ever falling back to the transcript. Leaving open; no duplicate found.

PR #3197, opened 2026-07-11
Observe silent worker failures: log disabled-refusal + wire dead ai.lastInteraction field

danscMax:
## Problem

Two silent failure modes make "memory mysteriously stopped recording" effectively undiagnosable:

**1. A disabled plugin exits silently.** When `enabledPlugins["claude-mem@thedotmack"] === false`, `worker-service.ts` `main()` refuses every `start` / `hook` / `restart` / `--daemon` with a bare `process.exit(0)` and no log line — indistinguishable from a crash. The only way to find the cause today is to reverse-engineer the shipped bundle.

**2. Auth-failure prose is dropped indistinguishably, and `ai.lastInteraction` is a dead field.** When the SDK is not authenticated it returns assistant prose ("Not logged in · Please run /login", or a `401 / Invalid authentication credentials`). `classifyObserverOutput` labels that `'prose'`, and `ResponseProcessor` confirms-and-drops it exactly like a benign `idle` ("nothing to observe"). Meanwhile `/api/health` → `ai.lastInteraction` is declared (worker-service.ts) but **never assigned anywhere in `src/`**, so it is always `null`. Net effect: observations silently stop landing while `/api/health` keeps returning `status: ok` and `ai.lastInteraction: null` — no signal that memory has stopped recording.

## Changes (source-only; no drop-semantics or HTTP-status change)

- **`worker-service.ts`** — one `logger.info` before the disabled-plugin `exit(0)`, naming the cause and the fix.
- **`sdk/output-classifier.ts`** — `isAuthFailureObserverOutput()`: a deliberately narrow predicate (mirrors the existing `isQuotaLimitedObserverOutput`) that recognizes auth-failure prose, distinct from benign `idle`.
- **`worker-service.ts`** — implement `recordAiInteraction()`, wiring the previously-dead `lastAiInteraction` field (deriving `provider` the same way `getAiStatus` already does).
- **`worker/agents/ResponseProcessor.ts`** — record `{ success: false, error: 'unauthenticated' }` when auth-failure prose is dropped, and `{ success: true }` after a successful store. `WorkerRef` gains an optional `recordAiInteraction?`.

Result: `/api/health` → `ai.lastInteraction` now reflects whether the SDK is actually authenticating (and when the last successful extraction happened), and recovers automatically on the next success. HTTP status codes and the existing drop behavior are unchanged.

## Tests

- `tests/sdk/output-classifier.test.ts` — `isAuthFailureObserverOutput` is true for the auth strings; false for empty/whitespace/non-string, benign observer prose, a real `<observation>` string, and a non-auth sentence containing "401".
- `tests/worker/agents/response-processor.test.ts` — records failure on auth-failure prose, nothing on ordinary prose, success on a valid store.
- `bun run typecheck:root` is clean; `bun test` on the touched files → **49 pass / 0 fail**.

### DIFF
--- a/src/sdk/output-classifier.ts
+++ b/src/sdk/output-classifier.ts
@@ -70,3 +70,30 @@ export function isQuotaLimitedObserverOutput(raw: unknown): boolean {
     /\b(rate limit|quota)\b.*\b(subscription|weekly|claude usage)\b.*\b(reached|exceeded|exhausted|reset|resets|try again)\b/.test(text)
   );
 }
+
+/**
+ * Detect authentication-failure prose returned as an assistant message instead of
+ * a structured error — e.g. the Claude CLI's "Not logged in · Please run /login"
+ * or an API "401 / Invalid authentication credentials". Distinct from a benign
+ * `idle` "nothing to observe": auth prose means the extraction silently produced
+ * NOTHING and will keep doing so until re-auth, so the pipeline should record a
+ * durable signal (see ResponseProcessor) rather than drop it indistinguishably.
+ * Kept intentionally narrow so ordinary prose about auth/HTTP does not trip it.
+ */
+export function isAuthFailureObserverOutput(raw: unknown): boolean {
+  if (typeof raw !== 'string' || raw.trim() === '') {
+    return false;
+  }
+
+  const text = raw.toLowerCase().replace(/\s+/g, ' ').trim();
+
+  return (
+    /\bnot logged in\b/.test(text) ||
+    /\brun\b[^.]{0,30}\/login\b/.test(text) ||
+    /invalid authentication/.test(text) ||
+    /authentication (failed|error|credentials)/.test(text) ||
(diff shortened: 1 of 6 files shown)


thedotmack:

> auth-classifier half landed via #3207; remaining: `/api/health` ai.lastInteraction wiring + disabled-plugin exit log

thedotmack:
Routing this to **#3606 (plan-18, Observer Response Pipeline)** — "classify every output, never confirm a batch on failure" is the cluster, and an auth-failure prose reply being classified `prose` and then confirmed-and-dropped exactly like a benign idle is the canonical instance of it. The dead `ai.lastInteraction` field is the reason `/api/health` keeps saying `ok` while nothing lands, so wiring it belongs in the same change. The disabled-plugin silent `process.exit(0)` is a smaller diagnosability win riding along. Leaving open; no duplicate.

PR #3208, opened 2026-07-12
fix: complete OpenCode memory lifecycle integration

SamuelZ12:
## Summary
- inject directory-scoped claude-mem context into OpenCode system prompts with bounded, deletion-safe request caching
- capture real user prompts, tool arguments, completed assistant replies, and summaries with awaited non-fatal writes and per-stage retries
- preserve OpenCode platform attribution, default-only bundle exports, explicit search behavior, and valid empty observer initialization logging

## Verification
- `bun test tests/integrations/opencode-plugin-contract.test.ts tests/integration/opencode-installer.test.ts tests/worker/openai-compatible-provider.test.ts` (36 passed)
- `npm run typecheck:root`
- production build and default-only OpenCode export check
- built and installed bundle SHA-256: `5bda8d4be0d0149fbdc8f81e55495359006dd2ab32fec03a1cc38f7cc961e201`
- live OpenCode 1.17.18 lifecycle verified exact prompt/platform/session IDs, assistant observation, summary, queue drain, and vector-backed retrieval


thedotmack:

> OpenCode capture half overtaken by main's plan-08 rework; startup-context injection novel; P1s: no fetch timeout, latest-only reply, no idempotency key

thedotmack:
Routing this to **#3611 (plan-23, Host Integration Contracts)** — OpenCode plugin export shape, event wiring and platform attribution are named in that cluster, and the live 1.17.18 lifecycle verification is exactly the kind of contract evidence it wants.

Overlap to sort out rather than a duplicate: #3803 also fixes the OpenCode export contract and session attribution, and the bundle PR #3614 carries #2985 for opencode events. This PR is the broadest of the three (prompt capture, tool args, assistant replies, summaries, bounded request caching). Leaving open, but it and #3803 should not both land as-is.

PR #3215, opened 2026-07-12
chore: bump version to 13.10.4

### LINKED ISSUE #3159: Feature request: native fire-and-forget Stop hook (still synchronously blocks in 13.9.3)
## Summary

The default `hooks.json` Stop hook still directly invokes `worker-service.cjs hook claude-code summarize` with `timeout: 120`, and the hook process awaits that subprocess's full completion before returning. This is the same blocking-hot-path symptom described in #2721, #1762, #1895, #1870, #1716 (all closed/consolidated) — confirmed still present in the currently-installed **13.9.3** default hooks config, not just older versions.

## Evidence

Installed plugin's `hooks/hooks.json` (both `plugins/marketplaces/thedotmack/plugin/` and `plugins/cache/thedotmack/claude-mem/13.9.3/`), Stop hook entry:

```json
{
  "type": "command",
  "shell": "bash",
  "command": "... node \"$_P/scripts/bun-runner.js\" \"$_P/scripts/worker-service.cjs\" hook claude-code summarize",
  "timeout": 120
}
```

This synchronously awaits the summarize subprocess. Issue #726 (in our own downstream repo) observed `running stop hooks 2/3 · 1m 18s` on real turns from this exact path.

## Workaround we've been running since May 2026

We patch the Stop hook command in-place to background the worker call and return immediately (~30ms):

```bash
bash $HOME/.claude/hooks/claude-mem-stop-async.sh "${CLAUDE_PLUGIN_ROOT:-}"
```

`claude-mem-stop-async.sh` captures stdin, backgrounds the original worker invocation (lockfile-protected against pile-up, 10-minute stale-lock recovery), and returns. This works, but it's a locally-maintained patch to a vendored file that gets silently overwritten by every plugin update — we've now had to reapply it at least twice (May 2026 initial patch, and again just now in July 2026 after it was wiped, likely by an update or reinstall).

## Ask

Would the maintainer consider making the Stop hook's worker call fire-and-forget natively (spawn detached + return, same shape as our wrapper) rather than requiring downstream users to maintain their own patch indefinitely? Happy to share the exact wrapper script if useful as a reference implementation.

### LINKED ISSUE #3206: PostToolUse observation hook blocks main loop for ~2.4s per tool call — should be async
## Environment
- claude-mem: 13.2.0 (verified same hook config on main branch as of 2026-07-12)
- Claude Code: 2.1.207
- Install: npm global via nvm (node v24.15.0)
- OS: macOS 26.5

## Problem

Every PostToolUse observation hook invocation blocks the Claude Code main loop for **~2.4 seconds** because none of the hooks in `plugin/hooks/hooks.json` are marked `"async": true`.

## Hard data (from 20 recent session transcripts, ~25,000 hook entries)

| Hook | Count | Avg | P50 | P95 | Max | Total time wasted |
|---|---|---|---|---|---|---|
| PostToolUse:Bash (observation) | 2,483 | 2,418ms | 2,099ms | 3,688ms | 94,891ms | ~100 minutes |
| PostToolUse:Read (observation) | 832 | 2,280ms | 2,060ms | 3,533ms | 6,421ms | ~32 minutes |
| PreToolUse:Read (file-context) | 626 | 2,189ms | 2,047ms | 2,955ms | 15,267ms | ~23 minutes |
| Stop (summarize) | 492 | 2,385ms | 2,155ms | 4,001ms | 8,700ms | ~20 minutes |
| PostToolUse:Edit (observation) | 344 | 2,357ms | 2,158ms | 3,488ms | 10,222ms | ~14 minutes |

**~175 minutes cumulative blocking time** in a ~48-hour observation window. The matcher `"*"` on PostToolUse means this fires on EVERY tool call — Bash, Read, Edit, Write, WebSearch, WebFetch, Skill, Agent, Workflow, everything.

## Root cause

`plugin/hooks/hooks.json` defines PostToolUse with `"matcher": "*"` and no `"async": true`:

```json
{
  "matcher": "*",
  "hooks": [{
    "type": "command",
    "command": "export PATH=...; node .../bun-runner.js .../worker-service.cjs hook claude-code observation",
    "timeout": 120
  }]
}
```

Observation recording is fire-and-forget logging — it should never block the tool loop. The fix is one line:

```json
"async": true
```

## Additional concern

On the main branch, `hooks.json` now includes a `PreToolUse matcher=Read` entry (the file-context hook previously only in `codex-hooks.json`). This means upgrading from 13.2.0 to current will add ANOTHER blocking hook before every Read — making the problem worse, not better.

## Expected behavior

All observation/logging hooks (`PostToolUse observation`, `PreToolUse file-context`, `Stop summarize`) should run with `"async": true` so they don't block the Claude Code tool loop. The user should never wait 2+ seconds after every tool call for background logging to complete.

## Proposed fix

Add `"async": true` to these hooks in `plugin/hooks/hooks.json`:
- `PostToolUse` matcher `*` → observation
- `PreToolUse` matcher `Read` → file-context
- `Stop` → summarize

Optionally reduce the `PostToolUse` matcher from `*` to `Read|Write|Edit|Bash` since observation on `WebSearch`/`WebFetch`/`Skill`/`Agent` adds latency without meaningful observation data.

## Claude Code reference

Per Claude Code hooks documentation, `"async": true` runs the hook in the background without blocking: https://code.claude.com/docs/en/hooks

thedotmack:
## v13.10.4

Patch release for non-blocking Claude Code memory hooks.

## Changes

- run PostToolUse observation recording asynchronously
- run PreToolUse file-context lookup asynchronously
- run Stop summarization asynchronously
- prevent memory bookkeeping from delaying tool calls and turn completion
- lock the non-blocking invariant with distribution coverage

## Issues

- resolves #3206
- resolves #3159

## Validation

- all eight version-bearing manifests set to 13.10.4
- old 13.10.3 version absent from those manifests
- full build completed successfully, including worker, server, MCP, transcript watcher, NPX CLI, OpenClaw, and OpenCode artifacts
- generated distribution files and shell-template parity verified
- marketplace sync/worker restart skipped because this worktree cannot write outside its sandbox


thedotmack:
Verified generated-artifact determinism under the release worktree toolchain (Node 24.5.0, Bun 1.3.9): SHA-256 hashes for mcp-server, server-service, transcript-watcher, worker-service, and viewer-bundle were identical before and after a second full `npm run build`. CI and Windows workflows also passed. The reported drift under Node 24.18.0/Bun 1.3.10 is therefore toolchain-dependent and broader than this version-only release; the committed artifacts are reproducible under the toolchain used to prepare v13.10.4.

PR #3225, opened 2026-07-13
fix(hooks): memory-worker outage must never block the editor (fail-open) + build guard for the worker zod closure

remten341:
## Summary

Two independent reliability fixes for the memory worker. Together they make a memory-worker outage a *degraded* experience instead of a hard editor lockout, and prevent the specific packaging drift that caused it from shipping again.

## The problem — a memory outage could lock the editor

When the worker becomes unreachable, `recordWorkerUnreachable()` counts consecutive failures and, at the fail-loud threshold, called `emitBlockingError()` → `process.exit(2)`.

On the **`UserPromptSubmit`** hook, exit code 2 **blocks prompt submission**. So a *benign, recoverable* memory-worker outage escalated into a **total editor lockout** — the user could no longer submit anything and had to relaunch in `--safe-mode` from an external terminal to recover.

A concrete trigger: the worker bundle externalizes `zod` (the shipped `worker-service.cjs` contains bare `require("zod/v3")` resolved at runtime from the plugin's installed `node_modules`). If a shipped `bun.lock` drifts from `plugin/package.json`, the runtime `bun install --frozen-lockfile` produces no resolvable `zod` and the worker crashes at startup with `Cannot find module 'zod/v3'` → unreachable → the fail-loud path above → lockout.

## Fix A — hook fail-open (an auxiliary service must never block the critical path)

- New non-blocking primitive **`emitDegradedNotice()`** in `src/shared/hook-io.ts`: flushes the buffered stderr and writes a loud warning to real stderr, but **never calls `process.exit`**.
- `recordWorkerUnreachable()` now escalates through `emitDegradedNotice()` instead of `emitBlockingError()`. The hook still exits `0` via `exitGraceful`, so the prompt is never blocked. The failure counter and `hook_failed` telemetry still fire for diagnosis.
- The message is explicit that the memory features are degraded **but the prompt was not blocked**, with a recovery hint.
- The generic blocking-error path (real handler bugs → exit 2) is **unchanged**. Only the auxiliary-service (worker-unreachable) escalation is made non-blocking.

Rationale: the memory worker is an auxiliary service; it should degrade gracefully, not take down the critical path it augments (SRE cascading-failures / graceful degradation; blocking user input to force attention is a DevEx anti-pattern).

## Fix B — build/CI gate for the worker's externalized zod closure

`mcp-server.cjs` is already guarded (the build fails if it externalizes `zod`, because Claude Desktop can launch it without plugin `node_modules`). The worker had the exact same exposure but **no equivalent guard**.

New `scripts/check-worker-runtime-deps.cjs`, wired into `npm run build` after `gen-plugin-lockfile.cjs`:

1. Reproduces a clean `bun install --frozen-lockfile --ignore-scripts` of the shipped `plugin/package.json` + `plugin/bun.lock` in a temp dir.
2. Asserts the worker's always-required externals (`zod`, `zod/v3`, `zod/v4`, `zod/v4-mini`) resolve under `bun` from that clean closure.
3. **Fails the build** on lockfile drift or an unresolvable `zod` — i.e. before a broken artifact can ship.
4. Opt-in `CLAUDE_MEM_VERIFY_WORKER_BOOT=1` additionally loads the built worker bundle against the clean closure.

## Proofs

- **Reproduced the lockout**: the old escalation `process.exit(2)`-ed at the threshold, killing the test runner mid-run (`exit_code=2`) — the literal lockout.
- **Fail-open verified**: `recordWorkerUnreachable()` no longer exits at/over the threshold, keeps counting, and emits a loud non-blocking notice.
- **Build guard verified**: passes on the real shipped closure, and fails with a clear message on a drifted `bun.lock` (the reproduction of the original breakage).
- Full suite parity with the pre-change baseline (the only remaining failures are pre-existing, isolation-dependent flakes that pass when run alone); `typecheck` clean.

## Tests added / updated

- `tests/cli/worker-failopen.test.ts` — no-exit at/over threshold + loud non-blocking notice + source contract.
- `tests/cli/hook-io.test.ts` — `emitDegradedNotice` (writes + never exits, flushes buffer first).
- `tests/build/worker-runtime-deps-guard.test.ts` — passes on real closure, fails on drift, wired into `build`.
- `tests/cli/hook-stream-discipline.test.ts` — updated to the new non-blocking contract.

## Notes

- No behavior change to the generic blocking-error path or to successful hooks.
- Draft: opening for review of the approach before finalizing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
   },
   "scripts": {
     "dev": "npm run build-and-sync",
-    "build": "node scripts/sync-plugin-manifests.js && node scripts/build-hooks.js && node scripts/gen-plugin-lockfile.cjs",
+    "build": "node scripts/sync-plugin-manifests.js && node scripts/build-hooks.js && node scripts/gen-plugin-lockfile.cjs && node scripts/check-worker-runtime-deps.cjs",
     "build-and-sync": "npm run build && npm run sync-marketplace && (cd ~/.claude/plugins/marketplaces/thedotmack && npm run worker:restart)",
     "sync-marketplace": "node scripts/sync-marketplace.cjs",
     "sync-marketplace:force": "node scripts/sync-marketplace.cjs --force",
--- a/scripts/check-worker-runtime-deps.cjs
+++ b/scripts/check-worker-runtime-deps.cjs
@@ -109,7 +109,9 @@ function main() {
     // Off by default because it spawns bun on the real bundle; the maintainers
     // keep daemon-spawning out of the default test/build path (flaky in CI, see
     // tests/cli/hook-stream-discipline.test.ts). `status` loads the bundle then
-    // reports "not running" without leaving a daemon behind.
+    // reports "not running" and exits 0 without leaving a daemon behind — so a
+    // top-level require("zod/v3") failure surfaces as a NON-ZERO exit / spawn
+    // error / timeout signal, which is exactly what we gate on below.
     if (process.env.CLAUDE_MEM_VERIFY_WORKER_BOOT === '1') {
       const workerCopy = path.join(tmp, 'worker-service.cjs');
       copyFileSync(worker, workerCopy);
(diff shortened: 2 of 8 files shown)


thedotmack:

> step 2 (`emitDegradedNotice`, never exits) — fail-open half; the zod-closure gate is covered by main's clean-room smoke

thedotmack:
Routing this to **#3605 (plan-17, Hook Wrapper Contract)** for the fail-open half — same `recordWorkerUnreachable` → exit 2 → `UserPromptSubmit` lockout as #3269 and #3243.

Not closing it, because the second half is a different cluster and nobody else has it: the `check-worker-runtime-deps` build guard against `bun.lock` drifting from `plugin/package.json` and leaving the worker's externalized `zod/v3` unresolvable belongs to **#3604 (plan-16, bundle integrity)** and is worth keeping on its own. On the fail-open half, note #4033 (bf25c897) already landed a `thresholdTripped` latch on `main` so the outage can no longer lock the editor permanently — rebase onto that rather than around it. Splitting this into two PRs would help it land.

PR #3226, opened 2026-07-13
fix(codex): preserve runtime marker across marketplace refresh

Steaeavean:
## Summary

- fall back to the versioned Claude plugin cache marker when a rematerialized plugin root no longer contains its local install marker
- preserve the local marker as the authoritative first choice when it exists
- add regression coverage for a marketplace refresh followed by Codex caching the nested plugin root

## Root cause

PR #3210 writes the runtime marker into both the marketplace root and its nested plugin directory. That fixes the initial Codex copy, but the marketplace is also an auto-updating Git checkout. A subsequent Claude Code marketplace refresh can replace that checkout and remove the ignored, untracked install marker.

The versioned Claude plugin cache still retains its marker because the installer writes it only after the runtime dependencies are ready. SessionStart can therefore use that version-matched cache marker as a durable fallback when the local marker is absent, instead of emitting a false runtime not yet set up warning.

Observed lifecycle:

1. the installer completed the versioned cache runtime and wrote its marker
2. Claude Code refreshed the Git-backed marketplace afterward
3. the refreshed nested plugin root no longer contained the ignored marker
4. Codex cached that markerless plugin root and SessionStart emitted the false warning

## Verification

- bun test tests/plugin-version-check.test.ts: 5 pass, 0 fail
- npm run typecheck:root: pass
- bun test --only-failures: 2288 pass, 15 skip, 0 fail

Follow-up to #3210 and the marker failure class reported in #3092.

### DIFF
--- a/plugin/scripts/version-check.js
+++ b/plugin/scripts/version-check.js
@@ -174,9 +174,26 @@ function readInstallMarkerVersion(markerPath) {
   }
 }
 
+function resolveInstallMarkerPath(root, version) {
+  const localMarkerPath = join(root, '.install-version');
+  if (existsSync(localMarkerPath)) return localMarkerPath;
+
+  const claudeConfigDir = process.env.CLAUDE_CONFIG_DIR || join(homedir(), '.claude');
+  const cacheMarkerPath = join(
+    claudeConfigDir,
+    'plugins',
+    'cache',
+    'thedotmack',
+    'claude-mem',
+    version,
+    '.install-version',
+  );
+  return existsSync(cacheMarkerPath) ? cacheMarkerPath : localMarkerPath;
+}
+
 try {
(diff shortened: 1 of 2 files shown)


thedotmack:

> step 3/5 (marker resolution) — small; note the deeper fix is marker at the resolved root

thedotmack:
Routing this to **#3604 (plan-16, Canonical Install Identity & Bundle Integrity)** — "which marker proves this plugin root is provisioned" is an install-identity question, and an auto-updating marketplace checkout wiping the untracked local marker is the same class of defect as the mtime-based root resolution that cluster covers. Builds on merged #3210 rather than duplicating it, and the versioned cache marker is a reasonable durable fallback since the installer only writes it after deps are ready. Leaving open.

PR #3227, opened 2026-07-13
fix(server): skip local-worker startup in server runtime for SessionStart and hook commands

### LINKED ISSUE #2991: server-beta: SessionStart context injection is not runtime-aware (injects stale/empty context)
## Summary

When `CLAUDE_MEM_RUNTIME=server-beta`, new Claude Code sessions inject **stale or empty** context. The SessionStart context-injection hook is not runtime-aware: it unconditionally targets the local worker's `GET /api/context/inject` route (backed by local SQLite) and never queries server-beta's PostgreSQL.

## Impact

- In `server-beta` mode the worker is typically not running, so the hook hits the worker-unreachable fallback and injects an **empty** context string → sessions start with no memory.
- If a stale worker *is* running, it injects from **stale local SQLite** instead of server-beta's Postgres → stale context.
- Bonus symptom: each miss calls `recordWorkerUnreachable()`, which after `CLAUDE_MEM_HOOK_FAIL_LOUD_THRESHOLD` (3) consecutive misses emits a blocking "worker unreachable" error — in a configuration where the worker is intentionally absent.

## Root cause

`src/cli/handlers/context.ts` (the `context` hook handler, dispatched from `src/cli/handlers/index.ts`) contains **no** runtime selection. Every sibling handler is runtime-aware (`session-init.ts:14`, `observation.ts`, `summarize.ts` import `resolveRuntimeContext` and branch on `runtime.runtime === 'server-beta'`), but the context handler was never given that branch.

Code path:
- `plugin/hooks/hooks.json` SessionStart → `worker-service.cjs hook claude-code context`
- `src/cli/handlers/index.ts:23` maps `'context' → contextHandler`
- `src/cli/handlers/context.ts:29` hardcodes `/api/context/inject?projects=…`
- `src/cli/handlers/context.ts:37` `executeWithWorkerFallback<string>(apiPath, 'GET')`
- `src/cli/handlers/context.ts:38-40` returns empty `additionalContext` on worker fallback
- worker route `src/services/worker/http/routes/SearchRoutes.ts:153` reads local SQLite

The gap is already acknowledged in-tree:
- `src/cli/handlers/session-init.ts:73-76`: *"Server-beta does not currently support the same context-injection protocol as the worker. Skip semantic injection in server-beta mode until the server-beta context endpoint exists."*
- `docs/server-beta-parity-map.md:68`: `GET /api/context/inject` → *(none yet)* → **unsupported**

## Server-beta status

- `POST /v1/context` exists (`src/server/routes/v1/ServerV1PostgresRoutes.ts:895`, Postgres-backed, returns `{observations, context}`), with client method `contextObservations({projectId, query, limit})` (`src/services/hooks/server-beta-client.ts:255`).
- **However** `/v1/context` is query-driven FTS — `query: z.string().min(1)` is required. The worker inject is query-**less** (recent observations for the project). So server-beta currently has no equivalent of the unconditional SessionStart "recent timeline" inject.

## Proposed fix

1. Add a query-less recent-context server route (e.g. `GET /v1/context/recent?projectId=&limit=`) that returns the same `{observations, context}` shape from `PostgresObservationRepository`, mirroring the worker's recent-observations semantics.
2. Add a corresponding client method on `ServerBetaClient`.
3. Make `contextHandler` runtime-aware (mirror `session-init.ts`): on `server-beta`, call the new endpoint; on worker, keep existing behavior. Preserve the fallback-to-worker path on eligible server-beta client errors.
4. Update `docs/server-beta-parity-map.md` to reflect the new endpoint.

This fills the exact parity-map gap and preserves injection semantics (recent timeline, not FTS relevance).

## Environment

- `CLAUDE_MEM_RUNTIME=server-beta`
- Affects every new/cleared/compacted session's SessionStart context injection.

derrickchwong:
## Summary

Two related fixes for the `server`/`server-beta` runtime, found while running claude-mem in production against a real deployment:

1. **SessionStart context injection was never implemented for server runtime** (closes #2991). `src/cli/handlers/context.ts`'s SessionStart hook only knew how to talk to a local worker (`executeWithWorkerFallback`) — a server-runtime deployment has no local worker, so every session silently got empty context. Adds a server-runtime branch that calls `/v1/context` (query-less recency mode) directly via the existing server HTTP client, and a matching query-less recency mode on the server route/repository so it's not forced to require a search query just to get "what happened recently."

2. **`hooks.json`'s `start` hook and every `hook <platform> <event>` invocation unconditionally call `ensureWorkerStarted()`**, regardless of `CLAUDE_MEM_RUNTIME`. In server runtime there's no local worker to start — and worse, the local MCP search server already occupies the worker's port in server runtime (it correctly skips its own auto-start), so `ensureWorkerStarted()` sees the port "in use" by a non-worker process and waits for it to answer worker health checks it will never answer. This stalls the **entire SessionStart hook group** — including the fix in (1) above — for up to 180s (Claude Code's own hook timeout) on every single session start.

## Verification

End-to-end against a live server-runtime stack (Postgres + Valkey + claude-mem-server + claude-mem-worker, real Claude Code sessions):

- Before: both SessionStart hooks (`start` and `hook claude-code context`) timed out at ~180,068ms / ~180,074ms every session.
- After: both complete in ~500-750ms, and `additionalContext` correctly returns real prior observations.

Full test suite passes (`bun test`): 2309/2309.

## Test plan

- [x] `bun test` — full suite green
- [x] `npx tsc --noEmit` — clean
- [x] Manual E2E: brought up a real server-runtime stack, ran real Claude Code sessions before/after, confirmed hook duration and context content via session transcripts


derrickchwong:
Thanks for the detailed review — all four points were valid and fixed:

1. **Codex fallback**: `isServerRuntime` is now resolved independent of platform, so Codex's own MCP context fetch (worker-backed, unaware of server runtime) correctly falls back to the server-context path in server runtime instead of the unreachable worker-HTTP path. Still correctly returns empty (not worker-fallback) when server runtime is selected but misconfigured — added a regression test for both.
2. **Recency default limit**: the CLI now passes `CLAUDE_MEM_CONTEXT_OBSERVATIONS` (default 50) explicitly; the route's own bare default for query-less mode is raised from 10 to 50 (query-mode keeps its existing default of 10 — no change there).
3. **Recency ordering**: switched to `created_at DESC`, matching `listByProject`'s existing convention in the same file. Added a Postgres-gated regression test that bumps an old observation's `updated_at` and confirms it doesn't jump to the top.
4. **Viewer URL**: now uses the actual server base URL when the server-runtime path was used, instead of the local worker port.

Full suite green, typecheck clean.

thedotmack:

> step 4 (runtime-aware hooks/context: no `ensureWorkerStarted` in server runtime; `/v1/context` branch); drop committed bundles + plans file, rebase

thedotmack:
Routing this to **#3618 (plan-24, Server Runtime & Cloud Sync Contract)** — both halves are server-runtime contract gaps that cluster names directly: SessionStart injection never implemented for `CLAUDE_MEM_RUNTIME=server` (so every server-runtime session got empty context), and hook commands unconditionally calling `ensureWorkerStarted()` where there is no local worker and the MCP search server already holds the port. Leaving open; no duplicate. The query-less recency mode on `/v1/context` is the piece most likely to need a second look, since it widens the route contract.

PR #3228, opened 2026-07-13
feat(server): support a dynamically-loaded custom generation provider

derrickchwong:
## Summary

The built-in server generation providers (Claude/Gemini/OpenRouter) are each constructed once at process startup from a single set of env vars, so `claude-mem-worker` can only ever use one fixed credential/model for the whole process's lifetime. That doesn't fit every deployment's credential model — e.g. a host application that stores its own per-project API keys and wants each generation job to use the right one, not one static key for every project.

Every `ServerGenerationJob` already carries `project_id`, and it's passed into `provider.generate(context)` **per call**, not just at construction time (see `ServerGenerationContext` in `providers/shared/types.ts`). This PR adds a small, generic extension point that takes advantage of that instead of hardcoding any particular external credential source:

`CLAUDE_MEM_SERVER_PROVIDER=custom` + `CLAUDE_MEM_CUSTOM_PROVIDER_MODULE=<path>` dynamically `import()`s an operator-supplied module at server startup (resolved from the filesystem, not bundled at build time) and calls a `createProvider(helpers)` factory it exports, handing it:

- `buildServerGenerationPrompt` — this codebase's own prompt construction
- `ClaudeObservationProvider` — this codebase's own Anthropic HTTP client
- `ServerClassifiedProviderError` — needed for correct retry classification; the retry pipeline (`ProviderObservationGenerator`) only treats a thrown error as retryable when it's `instanceof ServerClassifiedProviderError` with kind `transient`/`rate_limit` — a plain `Error`, even with a `.kind` property attached, is always treated as non-retryable

None of these three are exported/published anywhere else, so a custom provider module has no other way to reuse them without reimplementing prompt construction and Anthropic's HTTP protocol from scratch. The loader handles the three common CJS/ESM export shapes a factory module might use (named export, default function export, default object export with a method), since a module author outside this codebase could reasonably write any of them.

Also gives `ClaudeObservationProvider` an optional `baseUrl` (previously hardcoded to `api.anthropic.com`) so a custom provider — or any deployment routing through an Anthropic-API-compatible gateway — can reuse it as-is instead of reimplementing the HTTP call.

## Test plan

- [x] `bun test` — full suite green (2308/2308, plus new coverage: `tests/server/runtime/create-server-service-custom-provider.test.ts` with fixture modules for each export shape)
- [x] `npx tsc --noEmit` — clean
- [x] Manual E2E: implemented a real custom-provider module against this mechanism, wired into a live `claude-mem-worker` deployment, confirmed dynamic per-job credential resolution and correct `transient`/`unrecoverable` error classification during a real outage window


derrickchwong:
Fixed — `baseUrl` is now trimmed before use (matching the pattern already in `OpenRouterObservationProvider`), with added test coverage for the default URL, a custom gateway URL, and the whitespace/trailing-slash case. Full suite still green (2311/2311).

thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3618 ([plan-24] Server Runtime & Cloud Sync Contract)** as a **capability / product decision** — tracked on the plan master but not part of the defect fix sequence.

> custom server provider module (extension point for self-hosters); decide separately

thedotmack:
Routing this to **#2785 (plan-12, Provider & Extensibility Roadmap)** — this is net-new capability rather than a defect fix, and "new providers / extensibility knobs" is exactly that cluster. Nice that it leans on the fact that `ServerGenerationContext` already carries `project_id` per call instead of hardcoding a credential source. Leaving open; no duplicate found.

PR #3229, opened 2026-07-13
fix: detect occupied ports on Windows

ShiroKSH:
## Summary

- Detect Windows worker-port conflicts with a socket bind instead of an HTTP health response.
- Add a regression test for an occupied port whose health endpoint returns a non-success response.
- Document the fix in the Unreleased changelog.

## Root cause

On Windows, a non-success response from `/api/health` was treated as evidence that the port was free. The worker spawner could then launch a second daemon into an already-bound port and hit `EADDRINUSE`.

## Validation

- `bun test tests/infrastructure/health-monitor.test.ts`
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+## Fixes
+
+- **Windows worker lifecycle:** an occupied worker port is now detected by binding the configured address instead of treating a non-200 health response as a free port, preventing duplicate worker spawns and `EADDRINUSE` failures.
+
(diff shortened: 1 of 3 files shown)


ShiroKSH:
@thedotmack We'd appreciate it if you could take a look at this pull request when you have a chance.

ShiroKSH:
Hi maintainers, this PR has been open for over two weeks. When you have a moment, I would appreciate a review. Thank you.

thedotmack:
Closing in the PR consolidation pass — superseded: landed as 11518bc09 (#3356 win32 bind fallback); only a redundant test remains. The analysis here fed the plan-15 design (#3603); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

Issue #3218, opened 2026-07-13
chroma-mcp orphan leak still present on 13.10.3 — OOMs an 8GB machine in ~20 min (regression of #1063 / #1077)


mglmnz:
Orphaned `chroma-mcp` processes still accumulate until the machine is out of memory, on **13.10.3**.

This looks like a regression (or an incomplete fix) of #1063 (closed via #1065) and #1077 (closed). The changelog for **v13.0.1** states *"enforce a single `chroma-mcp` subprocess per worker (singleton via `disposeCurrentSubprocess()` on every code path; tree-kill of orphans on dispose)"* — but the leak reproduces on 13.10.3, several minors later.

### Environment

- claude-mem **13.10.3** (plugin, `thedotmack` marketplace)
- Fedora Asahi Remix 43, aarch64 (Apple M2), **8 GB RAM**, 16 KB pages
- Backend: Chroma, `CLAUDE_MEM_DATA_DIR=~/.claude-mem`, index ~1.7 GB
- Typically 2–3 concurrent Claude Code sessions, left running for days

### Symptom

Hard system freeze every ~20–30 min under load; only a force power-cycle recovers. Root cause turned out to be orphaned `chroma-mcp` processes exhausting RAM and swap.

### Measurements

Orphan accumulation with 2 sessions active:

```
54 chroma-mcp procs,  998 MB RSS   (26 min uptime)
44 chroma-mcp procs,  ~2 GB RSS    (25 min after a manual sweep)
```

State immediately before intervention:

```
Mem:  7.3Gi total, 6.9Gi used, 389Mi available
Swap:  15Gi total,  12Gi used
```

Killing **only the orphans** (nothing else changed) immediately recovered 8.5 GB of swap:

| | before | after |
|---|---|---|
| RAM available | 389 MB | 2.5 GB |
| Swap used | 12 GB | 3.5 GB |
| chroma-mcp procs | 57 | 4 |

### Spawn characteristics

- A new orphan pair (`uv` wrapper + its python child) appears roughly **every 20–60 s** while a session is active.
- **The trigger is tool-call activity, not a timer.** Controlled test — idle 120 s with zero tool calls:

  ```
  t=0    chroma procs: 18
  t=60   chroma procs: 18
  t=120  chroma procs: 18   -> flat while idle
  ```

  The count only grows while tool calls are being made. That is consistent with the PostToolUse → `syncObservation()` → `ensureConnection()` path described in #1063.

[...]

Issue #3216, opened 2026-07-12
Stale-version recycle loop leaks chroma-mcp processes (759 procs / ~71 GB RSS) until the machine OOMs


jhh3:
## Summary

After a plugin auto-update created a hooks/worker version skew (`pluginVersion=13.10.3` hooks vs a running worker reporting `13.10.2`), the stale-version recycle path looped for ~2 hours, spawning a fresh `uvx` + `chroma-mcp` process pair roughly every 14–24 seconds and orphaning the previous pair each time. Result: **759 leaked `chroma-mcp` processes totaling ~71 GB RSS**, which drove the machine (128 GB RAM, no swap) into OOM — systemd-oomd and the kernel OOM killer killed Chrome (twice), Spotify, and Slack before the box recovered.

## Environment

- claude-mem `13.10.3` (community-edge), installed via the `thedotmack` marketplace, user scope
- Claude Code CLI 2.1.207, Bun runtime
- Ubuntu (kernel 7.0.11), 128 GB RAM, no swap
- Several long-running Claude Code sessions (each generating hook events → each hook event is a recycle attempt)

## Timeline (from `~/.claude-mem/logs/claude-mem-2026-07-12.log`)

- **14:18:36** — plugin auto-update lands; first `Worker version mismatch — recycling stale worker {pluginVersion=13.10.3, workerVersion=13.10.2}`
- **14:18 → 16:19** — the mismatch line repeats **882×**; worker background init runs **447×**; each init spawns a chroma pair; prior pairs orphan to PID 1
- **16:19:14** — last mismatch; a 13.10.3 worker finally claims the port and the loop ends (in our case only because the stuck worker's chroma transport was externally killed)

## Log evidence (counts over the 2-hour window)

```
882  Worker version mismatch — recycling stale worker {pluginVersion=13.10.3, workerVersion=13.10.2}
447  [WORKER] Background initialization starting...
446  [SYSTEM] Shutdown initiated
446  [ERROR] [SYSTEM] Graceful shutdown failed — proceeding {reason=restart} Server is not running.
439  [SYSTEM] Worker is ready but still reports a stale version; not recycling again in this hook invocation (one recycle per hook event)
431  [SYSTEM] Port already in use, refusing to start duplicate
449  [CHROMA_MCP] Prewarming chroma-mcp uvx environment
398  [CHROMA_MCP] Connecting to chroma-mcp via MCP stdio
379  [CHROMA_MCP] Connected to chroma-mcp successfully
[...]

Issue #2959, opened 2026-06-16
tracking: chroma-mcp subprocess leak — zombie uv+python process pairs accumulate due to silent kill failures and orphan socket inheritance


thedotmack:
## Root Cause

Multiple overlapping bugs cause chroma-mcp Python subprocess pairs to accumulate into the hundreds on macOS and Windows: `killProcessTree` errors are silently swallowed so orphaned children survive; the `onclose` handler nulls the transport before `connectInternal()` can close it; the supervisor registry uses a fixed key so old PIDs are overwritten and become invisible; and on Windows the hook-parent's exit leaves chroma children holding the inherited listening socket, blocking future port binds. These combine to produce unbounded process accumulation consuming gigabytes of RAM.

### Issues
- Closes #2950 — https://github.com/thedotmack/claude-mem/issues/2950
- Closes #2896 — https://github.com/thedotmack/claude-mem/issues/2896
- Closes #2907 — https://github.com/thedotmack/claude-mem/issues/2907
- Closes #2897 — https://github.com/thedotmack/claude-mem/issues/2897

### PRs
- Related PR #2536 — https://github.com/thedotmack/claude-mem/pull/2536
- Related PR #2920 — https://github.com/thedotmack/claude-mem/pull/2920

## Resolution

This is the canonical tracking issue for this root cause. Individual issues and PRs listed above will be closed in favor of this one.

arjunghosh:
### Additional repro mode: multiple `CLAUDE_CONFIG_DIR` accounts under **one unix UID** deterministically defeats the fixed-key supervisor

Adding a data point that fits this tracker's root cause — specifically the *"supervisor registry uses a fixed key so old PIDs are overwritten and become invisible"* part. On a single machine running **several Claude Code accounts via distinct `CLAUDE_CONFIG_DIR` but the same unix user**, that fixed key collides by construction, so the orphan accumulation is not probabilistic (racing concurrent callers) — it's guaranteed on every cross-account (re)start.

**Environment (PROVEN — `ps` / reading the on-disk JSON):**
- macOS (Darwin 23.6.0), one unix user (`uid 501`).
- 3 Claude Code accounts, separate config dirs: `~/.claude`, `~/.claude-ac2`, `~/.claude-ac3`.
[...]

PR #3232, opened 2026-07-13
fix: eliminate the chroma-mcp PID-1/systemd orphan leak (version-skew recycle loop)

### LINKED ISSUE #3205: chroma-mcp process leak: stale marketplace version triggers an unbounded worker-recycle loop, and recycled workers orphan chroma-mcp to PID 1
## Summary

On a machine with **more than one `CLAUDE_CONFIG_DIR`** (or any config dir whose
marketplace checkout lags the installed worker build), claude-mem leaks
`uvx … chroma-mcp --client-type persistent` process trees at a rate of roughly
**one every 15–30 s** during active work. Each orphan holds **60–330 MB RSS**
(RES grows the longer it lives; observed up to ~580 MB). They reparent to PID 1
and never exit on their own.

There are **two independent defects** stacked on top of each other:

1. **Trigger — permanent worker-recycle loop on a version mismatch that can
   never resolve.** `wv()` recycles the worker whenever the *plugin* version
   (read from the marketplace checkout's `package.json`) ≠ the *worker's*
   baked-in build version, then waits for a successor **reporting the plugin
   version** — which a freshly spawned worker can never report. So the wait
   always times out, it lazy-spawns anyway, and the next hook repeats it. Result:
   the shared worker daemon is killed and respawned on essentially every hook.

2. **Leak mechanism — the `chroma-mcp` child is only killed on the graceful
   shutdown path.** The child is spawned non-detached by the MCP
   `StdioClientTransport`. On every non-graceful exit (SIGKILL after a restart
   timeout, the `EADDRINUSE` duplicate-daemon `process.exit(0)`, or a crash) the
   `uv → python chroma-mcp` tree is not killed and reparents to PID 1.

Either defect alone is survivable; together they leak continuously.

## Environment

- claude-mem `v13.10.2` (Claude Code plugin, `thedotmack/claude-mem`)
- macOS (Darwin 25.5.0), Apple Silicon
- Bun runner; `uv`/`uvx` at `~/.local/bin`; Python 3.13 (Homebrew)
- chroma launched as: `uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 --from chroma-mcp==0.2.6 chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma`
- Three config dirs in use: `~/.claude`, `~/.claude-personal` (both marketplace
  checkout at an old commit reporting **`13.4.0`**), and `~/.claude-work`
  (marketplace up to date at **`13.10.2`**). All three share one worker on port
  `37701`.

## Reproduction

1. Install claude-mem 13.10.x so the cache + worker bundle report `13.10.2`.
2. Arrange for at least one `CLAUDE_CONFIG_DIR` whose
   `plugins/marketplaces/thedotmack/package.json` `version` is **older** than the
   worker build (e.g. a stale marketplace checkout, or a second profile that was
   never re-synced). Here `~/.claude` / `~/.claude-personal` were pinned at
   `13.4.0`.
3. Run any Claude Code session under that config dir and do normal work
   (tool calls → PostToolUse hooks, etc.).
4. Watch `~/.claude-mem/logs/claude-mem-YYYY-MM-DD.log`: you get a steady stream
   of

   ```
   [SYSTEM] Worker version mismatch — recycling stale worker {pluginVersion=13.4.0, workerVersion=13.10.2}
   [SYSTEM] No successor worker appeared after recycle; falling through to lazy-spawn {pluginVersion=13.4.0, workerVersion=13.10.2}
   [SYSTEM] Worker not running — lazy-spawning {…}
   [SYSTEM] Worker is ready but still reports a stale version; not recycling again in this hook invocation {pluginVersion=13.4.0, workerVersion=13.10.2}
   ```

   (86–100 `recycling stale worker` events in a single day's log in our case.)
5. `ps -eo pid,ppid,etime,rss,command | grep '[c]hroma-mcp' | awk '$2==1'` shows
   a growing set of PID-1-owned chroma trees whose `etime` is much larger than
   the current worker's age — i.e. each was spawned by a prior, now-dead worker.

## Root cause (with code references — offsets are into the minified
`scripts/worker-service.cjs` of 13.10.2)

### Defect 1 — recycle loop

- `jle()` reads the **plugin** version from the marketplace checkout:
  `Vs = join(CLAUDE_CONFIG_DIR, "plugins","marketplaces","thedotmack")`, then
  `JSON.parse(readFileSync(join(Vs,"package.json"))).version`.
  (`Vs` def near offset 190065; `jle` near 232303.)
- `mU(port)` returns `matches: pluginVersion === workerVersion`
  (`workerVersion` = worker `/api/health` `.version`, i.e. the bundle's own
  version). (near 232806.)
- `wv()` (near 239058) — invoked by `Jle()` (near 241107), which is the
  reachability gate at the top of **`dn()`**, the HTTP helper every hook uses
  (near 242731). So `wv()` runs on essentially every hook. On mismatch:

  ```
  h.info("SYSTEM","Worker version mismatch — recycling stale worker",{pluginVersion:o,workerVersion:a});
  await In("/api/admin/restart", {method:"POST", …});           // kill current worker
  if (await Yle(o)) { … }                                        // WAIT FOR SUCCESSOR REPORTING plugin version `o`
  else h.warn("SYSTEM","No successor worker appeared after recycle; falling through to lazy-spawn", …);
  // lazy-spawn a fresh --daemon (detached:true)
  ```

- `Yle(t)` (near 234-…, "Worker health-version fetch") polls until a worker's
  `/api/health` version === `t` (the **plugin** version). A freshly spawned
  worker always reports **its own bundle version** (13.10.2), never the stale
  plugin version (13.4.0), so `Yle` **always times out** → lazy-spawn → next
  hook mismatches again → unbounded recycle loop.

The bug is that the successor-readiness check keys on the *plugin manifest*
version, which a newly spawned worker cannot satisfy when the manifest is stale.
There is no semver "is plugin newer than worker?" gate and no circuit breaker, so
a stale/older manifest loops forever instead of no-op'ing.

### Defect 2 — chroma child leaks on non-graceful exit

- The chroma child is spawned by the MCP `StdioClientTransport` (`ra.start`,
  near 184746):

  ```
  this._process = spawn(command, args, {
    env:{…}, stdio:["pipe","pipe", serverParams.stderr ?? "inherit"],
    shell:false, windowsHide:…, cwd
  });
  ```

  No `detached`, so the child shares the worker's process group but is **not**
  auto-killed when the worker dies; on parent death `uv` reparents to PID 1 and
  keeps its `python chroma-mcp` grandchild alive.

- The only teardown is the **graceful** path `hU()` (near 232-…):
  `… mcpClient.close() … chromaMcpManager.stop() …` where
  `ChromaMcpClient.stop()/disposeCurrentSubprocess()` calls `killProcessTree`
  (recursive `pgrep -P`, near 1245183). This runs only on clean
  SIGTERM/SIGINT/shutdown.

- Non-graceful exits that skip `hU()` and therefore leak:
  - restart via `/api/admin/restart` that ends in a SIGKILL after the graceful
    timeout;
  - **duplicate-daemon race**: `new g0().start().catch(… EADDRINUSE … process.exit(0))`
    (near 2452616) — a daemon that loses the port race calls `process.exit(0)`
    directly, with no chroma cleanup, after it may already have begun a chroma
    connection;
  - any uncaught crash.

## Observed impact

- Leak rate ≈ 1 chroma tree / 15–30 s of active work (tracks the per-hook
  recycle rate × number of concurrent config dirs/sessions).
- 60–330 MB RSS per orphan (observed one at ~580 MB).
- The worker PID changed 4× in ~4 minutes while observing (14072 → 15973 →
  16267 → 18580 → 19144), confirming the daemon is flapping, not stable.
- Users currently have to run an external watchdog (a launchd job reaping
  PID-1-owned chroma trees older than N seconds) to keep memory bounded.

### Fix A (stops the trigger — recommended, small)

In `wv()`/`mU()`:
- Only recycle when the plugin version is **strictly newer** than the worker
  version (semver compare), never when it is equal or older. A stale/older
  marketplace manifest should be a no-op, not a recycle.
- Make the successor check (`Yle`) wait for **readiness**, not for the worker to
  report the *plugin manifest* version (which a fresh worker can't produce when
  the manifest is stale). Compare against the version a freshly spawned worker
  will actually report, or just require readiness + `matches` becoming true.
- Add a recycle circuit breaker (e.g. max N recycles / M minutes per host) so no
  mismatch can loop unboundedly.

Config workaround (what we applied on the affected machine): correct the stale
`version` in each stale `CLAUDE_CONFIG_DIR/plugins/marketplaces/thedotmack/package.json`
to match the installed build (`13.4.0` → `13.10.2`). This is the value `jle()`
reads; once it matches the worker build, `wv()` stops recycling. (Equivalently:
`git pull` / re-sync the marketplace checkout.)

### Fix B (defense in depth — kill chroma on all exit paths)

- Spawn the chroma child **detached** (its own process group) and register, once,
  handlers that synchronously kill the whole group on any worker exit:

  ```js
  const child = spawn(cmd, args, { detached: true, stdio: […] });
  const killGroup = () => { try { process.kill(-child.pid, "SIGKILL"); } catch {} };
  process.once("exit", killGroup);
  process.once("SIGTERM", () => { killGroup(); process.exit(0); });
  process.once("SIGINT",  () => { killGroup(); process.exit(0); });
  process.once("uncaughtException", (e) => { killGroup(); throw e; });
  ```

  (macOS has no `PR_SET_PDEATHSIG`; a detached group + an `exit` handler is the
  portable way to guarantee the `uv → python` tree dies with the worker.)
- Run chroma cleanup before the `EADDRINUSE` duplicate-daemon `process.exit(0)`
  (near 2452616) and on the restart/SIGKILL path.

Fix A removes the storm; Fix B ensures that even legitimate restarts/crashes
never orphan chroma.

percy-raskova:
## Problem

`uvx … chroma-mcp --client-type persistent` process pairs leak, reparent to **PID 1 / `systemd --user`**, and never exit — accumulating one tree per worker recycle until the host runs out of memory (field reports of 227 procs / ~71 GB → OOM; on a swap machine the pressure got a whole desktop session killed by earlyoom).

## Root cause — two stacked defects

**Defect A — resolve by mtime, not semver.** The plugin hooks *and* the runtime worker-script resolver select the most-recently-**modified** cache dir (`ls -dt` / `mtimeMs` sort), not the highest semver. With several cached plugin versions this picks a stale build, manufacturing a permanent `pluginVersion != workerVersion` skew.

**Defect B — the skew drives a recycle loop whose non-graceful shutdown orphans chroma.** Every hook event recycles the worker; `performGracefulShutdown` fails **closed** (`closeHttpServer` throws `Server is not running` before reaching `chromaMcpManager.stop()`), so the non-detached chroma child reparents away. The supervisor tracked chroma under a **fixed** registry key, so each recycle's `register()` (a plain `Map.set`) overwrote the prior pid — hiding earlier orphans from the boot reaper.

## Fix — defense in depth

Built on **#3178** (@tiagogb's boot-time `reapLeakedProcesses`), which this branch cherry-picks and completes.

1. **Resolve by semver, not mtime** — all four resolver sites (runtime `worker-utils.ts` + the three generators in `hook-shell-template.ts`) via a shared `version-key` comparator; the three JSON launchers are regenerated and byte-verified. Eliminates the skew trigger.
2. **Last-resort child tree-kill on every shutdown path** — a `lastResortChildTreeKill` seam in `runShutdownSequence`, wired to the idempotent supervisor stop, runs for every reason (stop/restart/signal) and every graceful outcome. Stops the per-cycle leak even if a skew slips through.
3. **Per-pid supervisor keys + system-wide boot sweep** — chroma registers under `chroma-mcp:<pid>` (mirroring the `sdk:<session>:<pid>` pattern) so the reaper sees *every* orphan; `reapOrphanedChromaProcesses` scans `ps` for `chroma-mcp --data-dir <ours>` reparented to init **or** `systemd --user` (a PID-1-only check reaps nothing on systemd — this is why #3218 recurred) and tree-kills them, reclaiming already-leaked trees. `killProcessTree` is extracted to `src/supervisor/tree-kill.ts`.
4. **Cache reconciliation** — the worker warns at boot about stale cache versions; `claude-mem doctor --fix` prunes them (opt-in only — it never auto-deletes Claude Code's cache namespace).

## Why this holds where prior fixes didn't (#3218)

#3218 is a regression of the v13.0.1 singleton fix (#1063/#1065/#1077), which only enforced cleanup on the dispose path. This branch closes the **non-graceful exit path** (layer 2) and the **mtime trigger** (layer 1), and reaps orphans re-parented to `systemd --user`, not just PID 1 — the cases the dispose-path enforcement missed.

## Testing — durable regression contracts

- **Semver resolver**: comparator unit test + behavioral tests that run the *real* runtime, POSIX-shell, and Node launchers against a multi-version temp cache where mtime-newest ≠ semver-highest.
- **Shutdown tree-kill**: forces graceful failure/deadline, asserts the last-resort kill fires exactly once per cycle for every reason.
- **Orphan sweep**: a deterministic classifier test (CI-gating, synthetic `ps`) **plus** a hermetic real-process test (`tests/integration/orphan-sweep-real-process.test.ts`) that spawns a genuinely orphaned, re-parented tree and asserts the sweep reaps it — coverage the synthetic `FakeChildProcess` tests structurally cannot provide.

All new tests gate in CI on POSIX runners (no `uvx` needed). `npm run build` byte-verify passes.

## Reproduction

Put ≥2 versions in the plugin cache where the mtime-newest dir ≠ the marketplace `package.json` version; normal work recycles the worker each hook, leaking one chroma tree per event (`pgrep -fc 'client-type persistent'` climbs, never falls; PID-1/`systemd --user`-owned `chroma-mcp` trees accumulate). On this branch the resolver picks the highest semver (no skew) and any orphan is tree-killed at boot. Verified on a live host: after installing this branch the worker logs the stale-cache warning and reaps orphans on boot; 0 PID-1 chroma remained.

## Notes

- Built bundles (`plugin/scripts/*.cjs`) are intentionally **not** committed here: with the lockfile un-pinned a local rebuild isn't byte-reproducible (e.g. the React `viewer-bundle.js` churns with none of this branch's code). They regenerate via `npm run build` (byte-check passes) — please rebuild in the canonical release env.
- Spawn/kill code self-reviewed against `SECURITY.md`'s command-execution checklist (array args, no `shell:true`, no string-interpolated `execSync`).
- Does not touch the version-manifest files or `CHANGELOG.md` (release-time only).

Closes #3205, #3216, #3218, #3175. Builds on #3178 (credit @tiagogb). Refs #2959.


percy-raskova:
Thanks to @copilot and the greptile/T-Rex bot for the review. Triaged all four findings — three were real and are fixed in `5cd64b5f`, `5cfdaf1d`, `db057faa`; one is a false positive. Details below.

### ✅ Fixed — greptile P1: absolute-path `systemd --user` missed (`orphan-sweep.ts`)

Confirmed, and it defeated the core purpose of the sweep. On most systemd distros `ps -o args` renders the user manager with its argv[0] — on the Debian box I reproduced this on it's `/usr/lib/systemd/systemd --user`, which `startsWith('systemd --user')` never matches, so every chroma tree re-parented to it was left running (exactly the #3218 case). Now matched with `/(?:^|\/)systemd --user(?:\s|$)/` — anchored on the `systemd` basename with an exact `--user` word, so `.../notsystemd` and `--userland`-style flags still don't match. Added classifier tests for both abs-path variants plus an over-match guard.

### ✅ Fixed — greptile P1: descendants survive when `pgrep` is absent (`tree-kill.ts`)

Real. On images that ship `ps` but not `pgrep`, `collectDescendantPids` returned `[]` and only the wrapper was killed. Now the missing binary is detected specifically (`ENOENT`, distinct from `pgrep` exit 1 = "no children") and the walk falls back to a single `ps -eo pid,ppid` snapshot fed through a pure, cycle-safe, depth-bounded `descendantsFromProcTable`. Covered by unit tests plus a hermetic real-process test that the fallback finds a genuine live child from the real `ps` table.

### ✅ Fixed — Copilot: Windows PID-reuse in the boot reaper (`shutdown.ts`)

Real. `processCommandLine` returned `null` unconditionally on Windows, disabling the marker verification that stops a pid recycled across a reboot from being killed. Now reads `CommandLine` from CIM on Windows, mirroring the existing `queryWindowsCreationDate` pattern in `process-registry.ts` (wmic is removed on Win11). Fails safe — any error / missing PowerShell / null still yields `null` ("cannot verify"), identical to prior behavior, so it can only make the reaper *more* conservative. Note: the CIM path mirrors shipped code but couldn't be exercised on the Linux dev/CI runner here.

### ❌ False positive — Copilot: space separator in the shell resolver (`hook-shell-template.ts`)

The pipeline emits `<version> <path>` and reads it back with `IFS=' ' read -r _v _d`. Since `_d` is the *last* `read` variable it absorbs the entire remainder of the line — embedded spaces included — so a parent path with spaces is preserved. Verified against the exact generated pipeline with a cache under `.../cache test dir/...` where mtime-newest ≠ semver-highest: the path round-trips intact and `13.11.0` is still selected over the mtime-newer `13.9.0`. The first field is a `[0-9]*` glob dir name, so it can't contain a space either. No change needed.

Full suite: **2323 pass / 0 fail**; `npm run build` byte-verify passes.

percy-raskova:
i rebooted my pc (debian 13) and the fix still seems to hold, many running processes for a claude workflow but theres only one spawned instance of chroma

thedotmack:

> step 1/3/5 (per-pid registry keys, last-resort tree-kill on every shutdown path, boot sweep incl. systemd --user); the mtime→semver half landed as 7d3f1879e

thedotmack:

PR #3234, opened 2026-07-13
perf(hooks): skip login shell when node already resolves (closes #3186)

### LINKED ISSUE #3186: Every hook spawns a login shell ($SHELL -lc) to probe PATH — fixed ~250-300ms tax per hook event; skippable when node is already resolvable
Every hook command in `hooks/hooks.json` (13.10.2: SessionStart x2, UserPromptSubmit, PreToolUse, PostToolUse, Stop) begins with:

```sh
export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH"
```

This spawns a full login shell on every event to discover PATH. It makes sense for macOS/nvm setups, but when `node` is already on the system PATH (typical on Windows, where it lives in `C:\Program Files\nodejs`) it is pure overhead — and it runs on every single tool call via PreToolUse/PostToolUse.

**Environment**: Windows 11, Git Bash, claude-mem 13.10.2, Claude Code 2.1.205

**Measurements** (3 runs each, Git Bash):

| wrapper prefix | timing |
|---|---|
| with login-shell probe | 337 / 373 / 435 ms |
| static PATH prepend | 35 / 90 / 168 ms |

→ ~250-300ms saved per hook invocation. Over a 6-day window (50 sessions, 8,339 hook events aggregated from Claude Code transcripts) that is a meaningful share of the ~1.0-1.1s average PostToolUse duration we observe per tool call.

**Suggested fix** (cheap, behavior-preserving):

```sh
command -v node >/dev/null 2>&1 || export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH"
```

i.e. only pay for the login-shell probe when `node` is not already resolvable — or cache the probe result in a file under `~/.claude-mem/` and reuse it.

We patched our local `hooks.json` this way (static prepend) and the full chain (plugin-root discovery + `bun-runner.js` + worker round-trip) keeps working.

rodboev:
## Summary

Every non-Setup Claude Code hook launched a login shell to rebuild PATH before checking whether inherited PATH already resolved `node`. The canonical hook template now skips that probe when `command -v node` succeeds and retains the existing additive login-shell fallback when Node is absent.

## Why

`CLAUDE_CODE_PATH_PRELUDE` at `src/build/hook-shell-template.ts:61` ran `$SHELL -lc 'echo $PATH'` for six generated hook commands, including high-frequency PreToolUse and PostToolUse events. The trailing commands require `node`, so checking that runtime directly removes one process from the already-resolvable path while preserving recovery for sessions whose Node path exists only in login-shell startup files.

## Scope

- Changes the canonical Claude Code template and regenerates `plugin/hooks/hooks.json` through the build.
- Leaves Setup, Codex CLI, MCP, plugin-root discovery, cygpath conversion, trailing commands, and hook metadata unchanged.
- Preserves login-shell recovery whenever inherited PATH does not resolve Node.
- Limits performance claims to skipping one login-shell process; elapsed time varies by environment.

## Risk

The guard affects six Claude Code hook commands. Behavioral tests cover both inherited-Node and fallback states, while generator parity verifies every consumer still routes through the canonical template.

## Verification

- [ ] `bun test tests/infrastructure/plugin-distribution.test.ts`
- [ ] `npm run build`
- [ ] `npm run lint:hook-io && npm run lint:spawn-env && npm run strip-comments:check`
- [ ] Behavioral shell proof: inherited Node skips the sentinel login shell; missing Node invokes it and still reaches the trailing command

Closes #3186


rodboev:
Addressed in `9839287a`. The login-shell fast path now requires both `node -e` and `bun --version` to succeed before it skips the login shell, so a broken inherited Node or missing inherited Bun no longer suppresses the working login-shell PATH. Added regression coverage for both cases.

PR #3237, opened 2026-07-13
fix(worker): attribute observations to the current prompt project (closes #3217)

### LINKED ISSUE #3217: Observations inherit the session's launch-time project, not the per-prompt derived one (silently misattributes memories across repos/worktrees)
## Summary

`sdk_sessions.project` is bound **once, at session creation**, from the launch cwd, and every observation inherits it. But claude-mem already re-derives the correct (worktree-aware) project on **every prompt** and logs it as `INIT_COMPLETE` — it just discards that value at write time.

The result: any session that outlives its launch directory files all of its memories under the wrong project. On my machine this silently misattributed **71% of the last three months of observations**, and the affected project's memory bucket appeared frozen for 12 weeks while the plugin looked perfectly healthy.

## Evidence

The read path resolves correctly, per prompt (`~/.claude-mem/logs/`):

```
[session-233] INIT_COMPLETE | promptNumber=100 | project=dummy-project/feature-1
[session-233] INIT_COMPLETE | promptNumber=101 | project=dummy-project
[session-233] INIT_COMPLETE | promptNumber=102 | project=dummy-project/feature-3
```

The write path ignores all of it. For that same session:

```sql
sqlite> SELECT o.project, COUNT(*) FROM observations o
        JOIN sdk_sessions s ON s.memory_session_id = o.memory_session_id
        WHERE s.id = 233 GROUP BY 1;

dummy-project-related | 1173      -- the launch cwd, from 10 days earlier
```

Corroborating: **zero** observations exist under any worktree-style name despite ~120 prompts deriving them, and `merged_into_project` is NULL across all 7,617 rows, so the worktree roll-up path never fires either.

```sql
sqlite> SELECT COUNT(*) FROM observations WHERE project LIKE '%/%';   -- 0
```

## Reproduce

1. `cd repo-a && claude` — session row is stamped `project=repo-a`.
2. In that same session, work in `repo-b` (or a git worktree of `repo-a`).
3. Watch the logs: `INIT_COMPLETE` correctly reports `project=repo-b` / `repo-a/worktree`.
4. Query the DB: every observation is still stamped `repo-a`.

Two workflows hit this constantly:
- **Long-lived sessions** that move across repos (mine ran 10 days).
- **git worktrees** — `Sr()` computes the nice `parent/worktree` name and nothing ever persists it.

A session launched from a non-git scratch dir is the worst case: mine was launched from `~/code/tmp/random`, so months of real project work landed in a bucket literally named `random`.

## Suggested fix

The correct value is already in hand at write time — `INIT_COMPLETE` proves it. Either:

1. **Stamp observations with the per-prompt derived project** rather than inheriting `sdk_sessions.project` (observations already carry `prompt_number`, so the association is clean); or
2. **Update `sdk_sessions.project` when the derived project changes**, at minimum so new observations follow the session's actual location.

(1) is strictly better — it keeps a multi-repo session's memories attributed to the repo each one actually came from, which (2) cannot do.

## Recovery, for anyone else who hits this

The logs preserve exactly what the writer dropped. Since `observations.prompt_number` is populated and `INIT_COMPLETE` carries `sessionDbId` + `promptNumber` + the correct project, historical rows can be relabeled by joining the two — no heuristics needed. I recovered 1,708 observations this way (plus the mirrored `project` metadata in `chroma.sqlite3`, which goes stale the same way and must be synced separately, keyed on `sqlite_id` / `doc_type='observation'`).

**Version:** 13.10.3
**Platform:** macOS (darwin 25.5.0), Claude Code

rodboev:
## Summary

Long-lived worker sessions keep the project recorded at launch, so observations produced after moving to another repo or worktree are stored in the wrong project bucket. Carry the project already derived for each prompt into the active session before generation; the existing observation, Chroma, summary, and broadcast paths then consume the current value.

## Why

`sessionInitHandler` derives the current project from the prompt cwd and sends it to `/api/sessions/init`. The worker receives the value but initializes the active session without it, so `SessionManager` uses the launch-time `sdk_sessions.project`; `ResponseProcessor` later passes that stale value to `storeObservations` and the downstream projections. Passing the existing prompt project across that missing handoff fixes attribution without changing project-name derivation or durable session identity.

## Scope

This change updates attribution for new worker-runtime observations. It does not redesign project names, relabel historical rows, repair historical Chroma metadata, rewrite `sdk_sessions.project`, alter configured server-runtime project IDs, or change worktree adoption through `merged_into_project`.

## Risk

The active write project now changes when a prompt supplies a different valid derived project. Missing, empty, and `unknown` values retain the database fallback; focused tests now cover the uncached new-session branch, cached project-switch continuations, cached same-project continuations, worktree composite keys, unchanged durable metadata, and existing fallback behavior.

## Verification

- [x] Focused SessionManager and response-processor tests — 22 pass, 0 fail; proof now includes explicit uncached cold-start and same-project continuation assertions in `claude-mem-PR-TARGET-3217-PROOF.md`
- [x] `npm run build` — all build targets compiled and Rule A verification passed
- [x] `npm run lint:hook-io && npm run lint:spawn-env` — both policy checks passed
- [ ] `npm run strip-comments:check` — baseline reports 316 changed files in check mode; no target source file was written
- [x] Regression: a prompt-2 observation is stored and projected under `repo-b/worktree` while `sdk_sessions.project` keeps its launch value

Closes #3217


rodboev:
Addressed in `4ad2383f`. The worker now snapshots project and prompt attribution when each prompt is dispatched and carries that immutable context through storage, Chroma sync, and SSE broadcast. That keeps a deferred prompt-1 response on its original project even if prompt 2 updates the live session first. Added regression coverage in `tests/worker/agents/response-processor.test.ts` and `tests/gemini_provider.test.ts`.

PR #3238, opened 2026-07-13
fix(worker): wait through cold and concurrent startup readiness (closes #3231)

### LINKED ISSUE #3231: Worker frequently "did not become healthy before timeout" on any respawn (cold start / crash / concurrent launch), causing 60s UserPromptSubmit hook timeouts (not just upgrades)
## Before submitting

### Bug Description
The claude-mem background worker frequently fails to become healthy when it (re)spawns, and callers wait out their timeout. This shows up two ways:

- **`UserPromptSubmit` `session-init` hook** — killed at Claude Code's 60s hook ceiling, adding a ~60s stall at session start (and skipping memory injection that time).
- **MCP-client worker auto-start** — not bounded by the 60s hook limit, so it can hang far longer (observed up to ~35 minutes), during which worker-backed MCP tools fail.

The common log signature is:
```
Live PID detected but worker did not become healthy before timeout
```
A `worker.pid` exists, but the worker isn't actually answering on its port, so launchers block instead of quickly detecting a dead/slow worker.

Over ~3 months of logs (69 daily files), 18 discrete timeout incidents occurred. **Only 5 were near a plugin version change; 13 (72%) had no version change at all** — they were cold starts, worker crashes, or concurrent spawn races. So this is not upgrade-specific.

A contributing factor on Windows: the worker port is derived from `process.getuid()`, which is **undefined on Windows** (POSIX-only API — Windows uses SIDs, not numeric UIDs), so it falls back to a constant and **every profile computes the same port, 37777**, sharing one worker and one data dir (`~/.claude-mem`). This maximizes contention — multiple profiles' hooks and MCP servers all race to auto-start/reach the single shared worker.

### Steps to Reproduce
Any of these reliably produce a timeout (none require an upgrade):
1. **Concurrent spawn race** — open multiple Claude Code profiles/windows (all share port 37777 on Windows); when the worker isn't already up, their hooks and MCP servers race to auto-start it and lose the handoff. (Seen repeatedly, e.g. an MCP client reconnect `{reason=stdio-closed}` → auto-start → `did not become healthy before timeout`.)
2. **Cold start** — first session of the day / after reboot, worker spawns but doesn't open its port within the ~15s cold-boot budget.
3. **Crash / stale worker** — a `worker.pid` points at a dead or unresponsive worker; auto-start returns false and the condition persists for many minutes.
4. **Version upgrade** — plugin updates while an older worker is running; all profiles detect the mismatch and race to recycle/respawn.

### Expected Behavior
Worker (re)spawn should be robust: a stale/unresponsive PID is detected quickly, a single launcher wins and the others poll its actual health endpoint (not a fixed timer), and cold starts complete without multi-minute or 60s stalls.

### Environment
- **Claude-mem version**: 13.11.0 (issue observed across multiple versions back to 13.4.x)
- **Claude Code version**: 2.1.207
- **OS**: Windows 11 (10.0.26200)
- **Platform**: win32

### Logs
From `~/.claude-mem/logs/`. Representative incidents across categories:

Concurrent spawn race, no upgrade (2026-06-22, one of six that morning):
```
05:43:58  Claude-mem search server started
05:43:59  Worker not available, attempting auto-start for MCP client
05:44:05  Live PID detected but worker did not become healthy before timeout
(no version-mismatch line anywhere in that day's log)
```

Crash / stale worker, no upgrade (2026-05-18, persisted ~35 min):
```
04:44:14  Live PID detected but worker did not become healthy before timeout
          ERROR Worker auto-start returned false — MCP tools that require the worker will fail until the worker is running
```

Version-upgrade case (2026-07-13):
```
06:05:04  Worker version mismatch — recycling stale worker {pluginVersion=13.11.0, workerVersion=13.10.2}
06:15:11  ERROR ✗ Worker failed to start — Failed to start server. Is port 37777 in use?
```

### Additional Context
**Proof `process.getuid` is undefined on Windows** (Node v22.16.0):
```
> node -e "console.log('getuid type:', typeof process.getuid)"
getuid type: undefined
> node -e "console.log(37700 + ((process.getuid?.() ?? 77) % 100))"
37777
```
`~/.claude-mem/worker.pid` reports `"port": 37777`, and `127.0.0.1:37777` is the only listening worker socket, shared across all profiles.

**Suggested fixes**
1. Make the spawn-lock handoff poll the winning worker's actual health endpoint instead of a fixed cold-boot timer, and detect a stale/unresponsive PID quickly so callers fail fast instead of waiting out 60s (or minutes).
2. Isolate the worker per profile — derive the port and lock from a hash of `CLAUDE_CONFIG_DIR` rather than the `getuid`-based constant (which collapses to one port on Windows, and isn't per-profile even on POSIX). This removes the cross-profile contention that drives the concurrent-spawn-race case.
3. Recover cleanly from a crashed/stale worker rather than leaving the "auto-start returned false" condition persisting for tens of minutes.

rodboev:
## Summary

Worker startup used timing-dependent deadlines. A launcher that saw a live PID could return `warming` after the 3-second occupied-port window, while a launcher that spawned the worker could stack a 15-second health wait and a 30-second readiness wait. This change polls the terminal readiness endpoint through one bounded deadline for both known and attempted starts.

## Why

`/api/health` can succeed before core initialization finishes; `/api/readiness` succeeds only after initialization completes. `ensureWorkerStarted()` used health and readiness as sequential phases and selected a much shorter path when another launcher observed the PID early. Polling readiness directly removes that branch-dependent budget without changing endpoint semantics or timeout constants.

## Scope

This change covers deadline selection in `worker-spawner`. It does not terminate stale processes, add a spawn lock, isolate worker ports by profile, change hook-side startup logic, or claim to resolve duplicate spawning before a PID file exists.

## Risk

A validated live PID can keep an MCP or CLI startup task open longer than the old 3-second path, but the wait is bounded by one platform-scaled readiness deadline. Unknown occupied ports retain the short fail-fast path. Timeout results remain `warming`, and targeted tests still show PID touching only after readiness succeeds.

## Verification

- [x] `bun test tests/services/worker-spawner.test.ts` — 8 passing, 0 failing, 30 expectations
- [x] `npm run typecheck` — clean
- [x] `npm run build` — bundles regenerated successfully
- [ ] `npm run lint:hook-io && npm run lint:spawn-env && npm run strip-comments:check` — hook-IO and spawn-env clean; strip-comments baseline check reports 316 existing changes
- [ ] Manual Windows cold/concurrent launch: readiness completes before the 60-second hook ceiling without the live-PID timeout log

Closes #3231


rodboev:
Addressed in `1fe9bea6`. The stale-lock cutoff now stays behind the 90 s readiness window, so a second launcher will not reclaim the lock exactly at the boundary while the first worker is still legitimately starting. Added boundary coverage in `tests/shared/worker-spawn-gate.test.ts`.

rodboev:
Follow-up pushed in 0b76b741.

The stale-startup fail-closed guard now handles four readiness-timeout cases on this branch:

- a live pid that disappears before readiness comes up,
- a freshly spawned worker that never becomes healthy,
- a spawn-lock loser waiting on a winner that never materializes,
- a slow spawned child that is still alive by its returned spawn PID even before `worker.pid` exists.

That last case now stays `warming` instead of flipping to `dead` just because `/api/health` and `worker.pid` are both still absent at the timeout edge.

Local validation on this branch passed:

- `bun test tests/services/worker-spawner.test.ts`
- `npm run build`
- `npm run lint:hook-io`
- `npm run lint:spawn-env`

Issue #2926, opened 2026-06-13
Worker never starts on Windows — port 37777 held by orphaned process; UserPromptSubmit hook fails in a loop and blocks prompts


cateim:
## Before submitting
- [x] I searched existing issues and confirmed this is not a duplicate

### Bug Description
On Windows, the claude-mem worker fails to start on **every** launch with `✗ Worker failed to start — Failed to start server. Is port 37777 in use?`. The configured worker port (37777) is held in `LISTEN` by an **orphaned process from a previous session** (a PID that isn't even enumerable from the current user session). The lazy-spawn/auto-start logic never detects or recovers from this port conflict — it just keeps retrying and failing.

Because the worker never comes up, the **`UserPromptSubmit` hook fails on every prompt** and the failure counter escalates indefinitely (`hook-failures.json → "consecutiveFailures": 40`). In Claude Code this manifests as: **I type a command and the turn ends immediately returning nothing** — the prompt never gets processed. I had to disable claude-mem to use Claude Code at all.

Important: disabling the MCP server (`mcp__plugin:claude-mem:mcp-search`) does **not** fix it, because the hook is separate from the MCP server — the hook keeps firing and failing after the MCP plugin is turned off.

### Steps to Reproduce
1. On Windows, run a Claude Code session with claude-mem; let the worker get into a bad state (e.g. a previous session's worker process is left orphaned holding port 37777).
2. Start a new Claude Code session — the new worker tries to bind 37777 and fails (`Failed to start server. Is port 37777 in use?`).
3. Send any prompt. The `UserPromptSubmit` hook reports `claude-mem worker unreachable for N consecutive hooks` (N keeps growing), and the prompt returns nothing.

### Expected Behavior
- If port 37777 is already held by a **stale/orphaned** claude-mem worker, the launcher should detect it (health-check the existing listener) and either reuse it or kill the stale process and rebind — instead of failing forever.
- A failing `UserPromptSubmit` hook should **fail open** (never block/abort prompt processing), and the consecutive-failure counter should trigger a self-heal (or at least stop blocking the user) instead of escalating indefinitely.

### Environment
- **Claude-mem version**: 13.6.0
- **Claude Code version**: 2.1.177
- **OS**: Windows 11 Pro (10.0.28000)
- **Platform**: win32 (x64), runtime: Bun (C:\\Users\\<USER>\\.bun\\bin\\bun.exe)

### Logs
[...]

PR #3243, opened 2026-07-14
feat: add CLAUDE_MEM_HOOK_FAIL_MODE=warn to fail open when worker is unreachable

SaadSharif4:
# feat: add `CLAUDE_MEM_HOOK_FAIL_MODE=warn` to fail open when the worker is unreachable

Addresses #2996 and #2926 (both open reports of this exact lockout on Windows); related to the reliability plan in #3138.

## Problem

When the worker dies, the fail-loud path (`recordWorkerUnreachable` → `emitBlockingError` → exit 2 past `CLAUDE_MEM_HOOK_FAIL_LOUD_THRESHOLD`) **blocks every `UserPromptSubmit` in every session on the machine**. The failure counter (`~/.claude-mem/state/hook-failures.json`) is global and the hooks are machine-wide, so the user cannot reach the model *at all* — in any project, in any session — until the worker is manually repaired. The intent of #2292 (surface failures instead of swallowing them) is sound, but for a memory plugin, "memory unavailable" should degrade to "memory skipped," not "model unreachable."

### Windows makes this easy to trigger (reproduced twice on Win 11)

1. The worker spawns a `chroma-mcp` child chain (`uv.exe` → `chroma-mcp.exe` → 2× `python.exe`).
2. When the worker dies, those children survive and **inherit the worker's listening socket handle** on the worker port. `netstat` shows the port LISTENING, owned by a dead PID.
3. Every respawn attempt fails `EADDRINUSE` ("Failed to start server. Is port 37777 in use?"), so the failure streak never breaks, the threshold (default 3) trips, and the user is locked out of Claude Code entirely.

A second aggravator: if the plugin is disabled in `~/.claude/settings.json` while sessions with loaded hooks are still open, worker starts exit 0 silently (the `enabledPlugins` gate) while the hooks keep firing and blocking — same lockout, no diagnostic anywhere.

## Fix

New setting `CLAUDE_MEM_HOOK_FAIL_MODE` (default `block` — **existing behavior unchanged**):

- `block` — current behavior: exit 2 past the threshold.
- `warn` — identical counter, threshold, telemetry (`hook_failed`), and message, surfaced on real stderr via `emitDiagnostic` (bypassing the hook buffer), but the process exits 0. The handler falls through to its worker-fallback result (`{ continue: true }`), so the prompt is delivered and memory is skipped for the affected hook.

## Verification

- New seam-level tests in `tests/shared/worker-utils-fail-mode.test.ts` (same no-daemon-spawn rationale as `hook-stream-discipline.test.ts`): block mode escalates via `emitBlockingError` at the threshold, unset mode falls back to block, warn mode routes to `emitDiagnostic` and never blocks, warn keeps the same per-failure cadence past the threshold, and the failure counter still increments/persists in warn mode — 5 pass.
- `bun test tests/cli/hook-stream-discipline.test.ts tests/shared/settings-defaults-manager.test.ts tests/shared/worker-spawn-gate.test.ts tests/shared/worker-utils-recycle-successor.test.ts` — 48 pass, 0 fail.
- `CLAUDE_MEM_HOOK_FAIL_MODE` documented in `docs/production-guide.md` (recommended `warn` for production).
- End-to-end against a dead worker (port occupied by a 500-returning server, failure counter pre-seeded past threshold):
  - `warn`: exit 0, stdout `{}`, stderr `claude-mem worker unreachable for 6 consecutive hooks. Memory is not being recorded (CLAUDE_MEM_HOOK_FAIL_MODE=warn). Run: npx claude-mem doctor`, counter incremented.
  - `block`: exit 2, stderr `claude-mem worker unreachable for 6 consecutive hooks.` — byte-identical to current behavior.

## Possible follow-ups (not in this PR)

- Job-object / process-group cleanup of chroma children on Windows so a dead worker can't pin its own port.
- Have the worker log (not silently exit 0) when refusing to start because `enabledPlugins` marks it disabled.


SaadSharif4:
Pushed an additional commit (268f404) to this branch — a separate but related Windows worker-startup fix found while debugging a recurring "Failed to start worker" on my machine.

**Root cause:** `isPortInUse()` on win32 decided "port free" purely from an HTTP health probe. When a listener is dead/orphaned (e.g. a crashed worker's leftover `chroma-mcp` child chain still holding the socket) but no longer answers HTTP, the probe was misread as "port free." That let `worker-spawner.ts` race a brand-new worker spawn straight into a real `EADDRINUSE`, which died immediately — producing the generic, undiagnosed "Failed to start worker" with no indication of what actually went wrong. Reproduced this twice locally with the exact same signature (TCP connects, HTTP never answers, `isPortInUse` returns false).

**Fix:** on win32, fall back to the same real OS-level bind test POSIX already used whenever the HTTP probe doesn't succeed, instead of assuming the port is free. This correctly routes into the existing (previously dead-on-Windows) diagnostic branch in `worker-spawner.ts` that logs *"port in use but worker not responding to health checks"* — an actionable message instead of a silent misdiagnosis.

Added 3 regression tests in `tests/infrastructure/health-monitor.test.ts` covering: the fast-path health probe succeeding (no bind test needed), the fallback correctly identifying a stale/unresponsive listener as in-use, and the fallback correctly identifying a genuinely free port as free.

Happy to split this into its own PR if you'd rather keep this one scoped to just the fail-open hook mode setting — let me know.

thedotmack:

> step 2 — but the plan makes non-blocking unconditional (no `HOOK_FAIL_MODE` flag); Windows half landed via #3356

thedotmack:
Routing this to **#3605 (plan-17, Hook Wrapper Contract)** — same root cause as the rest of that cluster: `recordWorkerUnreachable` → `emitBlockingError` → exit 2 turning a memory outage into an editor lockout.

Closing this one, though. It's the third open PR on the identical path (`src/shared/worker-utils.ts`) alongside #3269 and #3225, and the lockout itself was fixed on `main` by #4033 (bf25c897), which added a `thresholdTripped` latch so the threshold blocks exactly once per outage and every later hook falls open. The remaining delta here is a new `CLAUDE_MEM_HOOK_FAIL_MODE` env-var mode, which `docs/merge-rubric.md` §3 rules out. The Windows chroma-child socket-inheritance analysis in your writeup is the valuable part and belongs on #3602 (plan-14) / #3603 (plan-15) — please do carry it over there.

PR #3254, opened 2026-07-15
fix(installer): provision tree-sitter-cli past script suppression and resolve tree-sitter.exe on Windows (refs #2910)

rodboev:
## Summary

Clean installs can leave `tree-sitter-cli` without its downloaded executable, and Windows can ignore an existing `tree-sitter.exe` when resolving the package-local CLI. This change keeps bulk dependency scripts suppressed, runs the trusted target-local `tree-sitter-cli/install.js` after cache installation when the CLI is unusable, and resolves `tree-sitter.exe` before PATH fallback on Windows through the shared package-local selector.

The version-cache path now awaits provisioning before critical-module verification. Initial install and repair write their cache marker only after provisioning succeeds. Marketplace provisioning remains best-effort and records the existing warning when it fails.

Refs #2910
Refs #2964

## Why

`runNpmInstallInMarketplace()` still uses `--ignore-scripts`, and `installPluginDependencies()` still uses `bun install --frozen-lockfile --ignore-scripts`. The repository trusts `tree-sitter-cli`, but those bulk installs suppress its package-owned download step. Bin-only verification does not detect the missing executable, so the cache path must invoke the existing package-local provisioner explicitly.

The parser has a separate Windows resolution defect. Its package-local probe checks only `tree-sitter`, while Windows releases can provide `tree-sitter.exe`. One shared selector now covers both installer and parser paths.

## Scope

This PR fixes the trusted CLI provisioning step, cache marker boundary, bounded failure reporting, and Windows package-local binary resolution. It does not change native `tree-sitter` runtime packaging, grammar dependency reconciliation, bundle contents, or WASM migration. The external release download and native runtime remain unverified here.

Attribution: https://github.com/thedotmack/claude-mem/pull/2918 and https://github.com/thedotmack/claude-mem/pull/3095 explored the same installer and resolver surface and closed unmerged.

## Risk

Cache provisioning failure now aborts `npx claude-mem install` and `npx claude-mem repair` before either cache marker is written, with process details retained in `last-install-error.json` and timeout remediation naming `CLAUDE_MEM_INSTALL_TIMEOUT_MS`. Marketplace provisioning still warns and continues. Bulk lifecycle scripts remain suppressed, and the bounded package-local subprocess runs only for the trusted top-level package.

## Verification

- [x] `bun test tests/setup-runtime.test.ts`, 26 pass, 1 Windows-only skip, 0 fail
- [x] `bun test tests/install-error-matrix.test.ts`, 61 pass, 0 fail
- [x] `bun test tests/install-non-tty.test.ts`, 45 pass, 0 fail
- [x] `bun test tests/services/smart-file-read/parser-bin-resolution.test.ts`, 4 pass, 0 fail
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] `npm run lint:hook-io`
- [ ] `npm run strip-comments:check`, repository baseline reports `Changed: 375`
- [ ] Verified on a real Windows install

Refs #2910
Refs #2964

### DIFF
--- a/src/npx-cli/commands/install.ts
+++ b/src/npx-cli/commands/install.ts
@@ -16,6 +16,7 @@ import { ensureWorkerStarted, type WorkerStartResult } from '../../services/work
 import { formatHostForUrl } from '../../shared/worker-utils.js';
 import {
   ensureBun,
+  ensureTreeSitterCliBinary,
   ensureUv,
   installPluginDependencies,
   writeInstallMarker,
@@ -715,7 +716,10 @@ async function runNpmInstallInMarketplace(summary: InstallSummary): Promise<void
 
   const baseFlags = ['install', '--omit=dev', '--ignore-scripts'];
   const strictResult = await runNpmStrict(marketplaceDir, baseFlags);
-  if (strictResult.code === 0) return;
+  if (strictResult.code === 0) {
+    await warnMarketplaceTreeSitterCliIfUnavailable(summary, marketplaceDir);
+    return;
+  }
 
   if (strictResult.timedOut) {
     installerError(ErrorSeverity.ABORT, {
@@ -747,6 +751,7 @@ async function runNpmInstallInMarketplace(summary: InstallSummary): Promise<void
       message: 'tree-sitter peer-dep ERESOLVE was resolved with the --legacy-peer-deps fallback. Benign for the marketplace install; re-evaluate when tree-sitter peer ranges change.',
       remediation: 'No action required.',
     });
+    await warnMarketplaceTreeSitterCliIfUnavailable(summary, marketplaceDir);
     return;
   }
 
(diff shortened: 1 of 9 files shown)


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3604 ([plan-16] Canonical Install Identity & Bundle Integrity)** as **needs rework**: the diagnosis is right, but the item noted below blocks it. If you'd like to keep it moving, address that against the design in the plan doc; otherwise the fix will be absorbed into the plan-16 bundle.

> step 3 (tree-sitter binary fetched by explicit step, .exe resolution); P1: cache install path never invokes provisioner; CI red

rodboev:
Thanks for the precise pointer. The version-cache path now uses the same package-local provisioning step as the marketplace path:

1. In `src/npx-cli/install/setup-runtime.ts`, `installPluginDependencies()` awaits `ensureTreeSitterCliBinary(targetDir)` immediately after the script-suppressed Bun install and before `verifyCriticalModules(targetDir)`. The initial cache setup and repair paths already call this shared function, and both write `.install-version` only after it succeeds. The bounded probe and package installer close child stdin, and cache failures abort with bounded process diagnostics before marker writes.
2. The cache-path regression covers a missing binary, the explicit `tree-sitter-cli/install.js` step, retained `--ignore-scripts`, package-layout and process failures, timeout handling, diagnostic output, and the failure path that prevents the cache marker from being written.

The package-local version probe is bounded and keeps the install process off the event loop. Marketplace warning behavior and the Windows `.exe` resolver stay unchanged. This remains limited to the cache dependency-install path; it does not re-enable lifecycle scripts or change bundle contents, grammar dependencies, or plugin-root selection.

thedotmack:
Routing this to **#3604 (plan-16, Canonical Install Identity & Bundle Integrity)** — a plugin root whose `tree-sitter-cli` has no downloaded executable is the "dependency-complete install root" failure the cluster exists to close, alongside the missing externalized deps (zod, ajv) cases. Running the trusted target-local `install.js` after the `--ignore-scripts` bulk install, and awaiting provisioning before the cache marker is written, is the right ordering for that. Leaving open; no duplicate found for the tree-sitter provisioning path.

PR #3284, opened 2026-07-16
feat(sqlite): periodically optimize FTS5 indexes to reclaim delete-marker bloat (refs #2793)

laihenyi:
## Problem

FTS5 external-content indexes delete **logically** — removing a row appends a delete-marker segment rather than reclaiming the original postings. Neither `VACUUM` nor `auto_vacuum` can compact this; only an FTS `'optimize'` (or `'rebuild'`) merges the segments and drops the deleted-row content. claude-mem never runs `'optimize'` outside migrations, so a heavy-delete install grows an **unbounded shadow index that survives every other reclaim path**.

This is a **distinct root cause** from the incremental-`auto_vacuum` work already on `main`:

- On a real install (relevant to #2793), `user_prompts_fts_data` reached **~12.3 GB against an 18 MB `user_prompts` source table**. The source rows had already been cleaned up; the bloat was entirely stale FTS postings.
- `auto_vacuum = INCREMENTAL` and `incremental_vacuum` could only reclaim the ~1 GB freelist — they can't touch FTS segment bloat.
- After `INSERT INTO user_prompts_fts(user_prompts_fts) VALUES('optimize')` + `VACUUM`, the same DB dropped **~14 GB → ~140 MB** with `observations`/`session_summaries`/`user_prompts` fully intact and `PRAGMA quick_check` clean.

The rebuild recipe is currently only documented manually in `troubleshooting.mdx`; this makes the reclaim automatic.

## Change

`runPeriodicFtsOptimize(db, opts)` in `src/services/infrastructure/FtsMaintenance.ts`:

- Runs `'optimize'` on each **existing** FTS index (`observations_fts`, `session_summaries_fts`, `user_prompts_fts`); missing tables are skipped.
- **Throttled to once/day** via a `.fts-optimize-applied` marker file, mirroring the `CleanupV12_4_3` marker pattern — cheap to call on every startup.
- A corrupt/unreadable marker is treated as "never optimized" so maintenance can't wedge itself.
- Table names are a constant array (never user input), so the interpolated statements are not dynamic SQL.

Wired into `worker-service.ts` startup right after `runOneTimeV12_4_3Cleanup()`, using the shared connection (`dbManager.getConnection()`), wrapped in try/catch so a maintenance failure never blocks startup.

## Tests

`tests/services/infrastructure/fts-maintenance.test.ts` (TDD, 4 cases):
- **Reclaim**: seed an external-content FTS, insert 400 rows → delete all → `SUM(LENGTH(block))` shrinks after optimize; marker written.
- **Throttle**: a second run inside the interval is a no-op; runs again once the interval elapses.
- **No-FTS-tables**: skips gracefully and still writes the marker.

`npm run typecheck` clean; `npm run build` OK; local `bun test` across infrastructure + worker + sqlite suites: **300 pass / 0 fail**. Source + test only (built artifacts excluded; CI rebuilds them).

### DIFF
--- a/src/services/infrastructure/FtsMaintenance.ts
+++ b/src/services/infrastructure/FtsMaintenance.ts
@@ -1,0 +1,104 @@
+import path from 'path';
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
+import { Database } from 'bun:sqlite';
+import { DATA_DIR } from '../../shared/paths.js';
+import { logger } from '../../utils/logger.js';
+
+// FTS5 external-content indexes delete logically: removing a row appends a
+// delete-marker segment rather than reclaiming the original postings. Neither
+// VACUUM nor auto_vacuum can compact this — only an FTS 'optimize' (or
+// 'rebuild') merges the segments and drops the deleted-row content. Without it,
+// a heavy-delete install grows an unbounded shadow index: #2793 reported a
+// user_prompts_fts data table reaching ~12GB against an 18MB source table,
+// reclaimable only via the manual troubleshooting recipe. Running a throttled
+// optimize on startup makes that reclaim automatic.
+
+const MARKER_FILENAME = '.fts-optimize-applied';
+const DEFAULT_INTERVAL_MS = 24 * 60 * 60 * 1000; // once per day
+
+// Constant identifiers (never user input), so interpolating them into the
+// maintenance statements below is not dynamic SQL.
+const FTS_TABLES = ['observations_fts', 'session_summaries_fts', 'user_prompts_fts'] as const;
+
+interface MarkerPayload {
+  optimizedAt: string;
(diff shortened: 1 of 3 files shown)


thedotmack:

> step 5 (retention & reclaim: FTS5 `optimize` pass)

thedotmack:
Routing this to **#3609 (plan-21, SQLite Schema Evolution & Queue State Integrity)**, which owns unbounded growth / retention / VACUUM. Confirmed still needed: there is no `VALUES('optimize')` call anywhere under `src/` on current `main`, so FTS delete-marker bloat really is unreclaimable outside a migration today, and a 12.3 GB `user_prompts_fts_data` against an 18 MB source table is a convincing repro. Not a duplicate of the incremental-`auto_vacuum` work — different reclaim path entirely. Leaving open.

PR #3285, opened 2026-07-16
fix(search): fall back to FTS5 when Chroma has no usable matches

jamesdsizemore:
## Summary

\`SearchManager.performChromaSemanticSearch\` had two related gaps that left SQLite/FTS5 — which has correct, complete data — never consulted as a safety net for Chroma's incomplete coverage:

1. **Zero raw Chroma matches**: fallback to FTS5 only fired when `options.platformSource` was set. The common unscoped case (no `platformSource`) returned empty instead of falling back.
2. **Non-zero raw Chroma matches, but all filtered out by the recency window** (explicit `dateRange`, or the 90-day default): no fallback existed at all. In practice this is the *dominant* path — Chroma's nearest-neighbor search rarely returns literally zero candidates for any query across the full collection, so gap (1) alone doesn't cover most real cases.

Both surface whenever a project's Chroma coverage is incomplete — e.g. via the backfill crash in #3126 (partially fixed by #3146). Any observation that depended on backfill to reach Chroma stays invisible to `search` regardless of `platformSource` or date scoping, even though it's sitting correctly in SQLite.

Relates to #3126 ("Bug 2" in that report — `search` appearing to ignore its `query` parameter — was left as a follow-up with no dedicated issue).

## Verification

Reproduced against a real project: 13k+ SQLite observations, but only ~350 Chroma-embedded documents (incomplete backfill coverage). Historical queries returned empty even after fixing gap (1) alone — confirmed via direct SQLite comparison and via the live worker — until gap (2) was patched too. After both fixes, date-scoped and unscoped historical queries correctly surface real data.

- Existing test that asserted gap (1)'s behavior as *intended* (`searchObservations` should not be called on unscoped zero-match Chroma) rewritten to assert the corrected behavior.
- New test added reproducing gap (2): Chroma returns a non-empty but entirely out-of-range candidate set; asserts the FTS5 fallback now fires.
- `bun test tests/worker/search-manager.test.ts`: 7 pass, 0 fail.
- `bun test tests/worker/search/`: 41 pass, 0 fail.
- Full suite `bun test`: 2288 pass, 0 fail, 15 pre-existing unrelated skips.

## Test plan

- [x] Unit tests updated/added and passing
- [x] Full repo test suite passes with no regressions
- [x] Manually verified against a real installed copy with a live worker restart, confirming previously-missing historical observations now surface for both date-scoped and unscoped queries

### DIFF
--- a/src/services/worker/SearchManager.ts
+++ b/src/services/worker/SearchManager.ts
@@ -401,9 +401,9 @@ export class SearchManager {
           platformSource: options.platformSource
         });
       }
-    } else {
-      if (options.platformSource) {
-        logger.debug('SEARCH', 'Platform-scoped ChromaDB search found no matches; falling back to scoped FTS5 search', {});
+
+      if (obsIds.length === 0 && sessionIds.length === 0 && promptIds.length === 0) {
+        logger.debug('SEARCH', 'ChromaDB matches did not survive date filtering; falling back to FTS5 search', {});
         platformScopedChromaZeroFallback = true;
 
         if (searchObservations) {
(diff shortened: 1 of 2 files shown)


thedotmack:

> step 5 (honest fallback; SQLite reachable when Chroma coverage is partial); rebase over #3325 ranking rework

thedotmack:
Routing this to **#3982 (plan-25, Search Read Path / FTS)** — "Chroma returned nothing usable, so consult SQLite FTS5" is the whole point of that cluster.

Heads up on overlap: #3173 rewrites the same fallback in `SearchManager.performChromaSemanticSearch` *and* in `SearchOrchestrator.executeWithFallback`, driven by the CJK repro. Your gap (2) — non-zero Chroma hits that are all filtered out by the recency window — looks distinct from the Chroma-empty-category case #3173 handles, so I'm leaving both open rather than picking one; whichever becomes the plan-25 architectural fix should absorb the other's condition. Worth the two of you comparing notes.

PR #3289, opened 2026-07-16
fix(worker): search(type=<custom>) drops custom observation types

### LINKED ISSUE #3279: search(type=<custom>) returns no results for custom observation types (worker runtime, 13.11.0)
## Summary

On the worker runtime (v13.11.0), the MCP `search` tool's `type` parameter returns zero results for **any user-defined observation type** (one added via a custom mode, not in the built-in set), even when matching observations exist and are found by free-text search.

## Repro

Let `mytype` stand for any custom observation type id that is **not** one of the built-in types (`bugfix`, `feature`, `refactor`, `change`, `discovery`, `decision`, …).

1. Configure a custom mode (via `~/.claude-mem/modes/` + `CLAUDE_MEM_MODES_DIR`) that adds an observation type with id `mytype`.
2. Have the observer capture at least one observation of that type (verified present: queryable by ID, its stored `type` field == `"mytype"`).
3. Call `search(query="…", type="mytype")` → **0 results**.
4. Call `search(query="<free text matching the observation>")` (no `type` filter) → the same observation **is** returned.

## Expected

`search(type="mytype")` returns observations whose stored `type` is `mytype`.

## Actual

The typed query returns nothing; only untyped free-text search finds the row. This suggests the `type` parameter is validated against a hardcoded allowlist of built-in types and silently drops unknown values, rather than passing the string through to the query.

## Impact

Storage of custom types works (the `type` field is a plain string), but they cannot be retrieved by `type=`; callers must fall back to free-text search and filter client-side. Related feature context: #1640 / #2009.

## Environment

claude-mem 13.11.0, worker runtime (`CLAUDE_MEM_RUNTIME=worker`).

povesma:
## The bug

On the **worker runtime**, the MCP `search` tool's `type` parameter is documented as *"Filter by observation type"*, but filtering by a **custom** observation type (one added via a custom mode, e.g. `correction`, rather than a built-in like `bugfix`/`feature`/`refactor`/…) returns **zero results** — even though the same observations are found by free-text search.

Closes #3279.

## Root cause

`type` is overloaded. In `SearchManager.search()` it is consumed only as a **doc-type selector** with three reserved values — `observations` / `sessions` / `prompts`:

```ts
const searchObservations = !type || type === 'observations';
const searchSessions     = !type || type === 'sessions';
const searchPrompts      = !type || type === 'prompts';
```

When a caller follows the tool's documented meaning and passes `type="correction"` (any value that is not one of those three selectors), **all three** flags evaluate to `false`, so no corpus is searched and the call silently returns nothing. The actual observation-type filter (`o.type = ?` in `SessionSearch.buildFilterClause`) is only reachable via the separate, less-obvious `obs_type` param — so custom types are effectively unfilterable via the documented `type` argument.

This is a dropped-parameter passthrough bug: a non-selector `type` never reaches the observation-type WHERE clause.

## The fix

In `SearchManager.search()`, distinguish the reserved doc-type **selectors** from an observation-type **filter**. A `type` value that is not `observations`/`sessions`/`prompts` is treated as an observation-type filter: it is routed into `obs_type` (which already threads into the SQL `o.type = ?` clause) and the search is scoped to observations. Reserved selectors keep their existing meaning unchanged.

Minimal, one-path change — no refactor, no new params, no schema change.

## Scope note (worker vs server)

There are two runtimes. The reporter uses the **worker** path (`/api/search` → `handleUnifiedSearch` → `SearchManager.search`), which is where the defect lives and where this fix applies.

The **server** path (`observation_search` → `/v1/search` → `buildSearchPayload`) does **not** expose a `type` parameter anywhere in its chain — not in the MCP tool schema, the request type, the client payload builder, the zod route schema, or the Postgres repo. Adding type filtering there would be a new multi-layer feature rather than a passthrough fix, so it is intentionally out of scope for this bug.

## Verification

- Added two focused tests in `tests/worker/search-manager.test.ts`:
  - `search(query, type="correction")` now searches observations with the obs-type filter set to `correction`, and does **not** search sessions/prompts. Confirmed this test **fails on the pre-fix code** (0 calls to `searchObservations`, reproducing the zero-results bug) and **passes** with the fix.
  - Reserved selector `type="observations"` still behaves as a selector (searches observations, obs-type filter unset).
- `bun test tests/worker/` — 244 pass, 0 fail.
- `tsc --noEmit` — clean (0 errors project-wide).

### DIFF
--- a/src/services/worker/SearchManager.ts
+++ b/src/services/worker/SearchManager.ts
@@ -425,7 +425,7 @@ export class SearchManager {
 
   async search(args: any, telemetryOut?: SearchTelemetryEnvelope): Promise<any> {
     const normalized = this.normalizeParams(args);
-    const { query, type, obs_type, concepts, files, format, ...options } = normalized;
+    const { query, type: rawType, obs_type: rawObsType, concepts, files, format, ...options } = normalized;
     let observations: ObservationSearchResult[] = [];
     let sessions: SessionSummarySearchResult[] = [];
     let prompts: UserPromptSearchResult[] = [];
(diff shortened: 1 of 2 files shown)


thedotmack:
Closing in the PR consolidation pass — superseded: landed as ecf9a61ac (#3281 non-category `type` → obs_type alias). The analysis here fed the plan-23 design (#3611); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

PR #3292, opened 2026-07-17
fix(windows): recover hung workers and orphaned Chroma trees

manoi-bms:
## Summary

This adds bounded, ownership-checked recovery for the Windows failure mode where the worker stops answering health checks but its port remains occupied, often by an orphaned Chroma/UV process tree.

The change:

- detects port occupancy with a real TCP bind probe instead of treating a failed health request as a free port;
- applies deadlines to worker health requests so startup and restart cannot hang indefinitely;
- records a real Windows worker PID plus a process-creation token for PID-reuse protection;
- reclaims only positively identified claude-mem worker/Chroma process trees;
- proves the worker port is bindable before spawning a replacement; and
- wires the same recovery path into normal startup, hook lazy-spawn, and CLI restart fallback.

No new dependency is introduced, and non-Windows behavior is unchanged.

## Problem

On Windows, a worker crash or hang can leave the configured worker port in `LISTENING` even though `/api/health` is unreachable. Once that happens, the existing startup paths cannot recover reliably:

1. `HealthMonitor.isPortInUse()` used the HTTP health endpoint as a proxy for socket occupancy on Windows. A hung or orphan-owned listener therefore looked like a free port.
2. Health requests did not have their own abort deadline, so a half-open TCP peer could stall recovery.
3. The detached PowerShell launcher returned `0` as a success sentinel instead of the worker's PID, which prevented reliable ownership checks and targeted cleanup.
4. The `uvx.exe` shim can exit after launching `uv.exe tool uvx`, leaving the long-lived Chroma process outside the recorded tree.
5. Startup could attempt another worker without first proving that the existing listener had been released.

The result is a persistent locked-port state that may otherwise require manually killing several processes or rebooting Windows.

### 1. Bounded health and port checks

- `HealthMonitor` now uses an actual TCP bind probe on every platform.
- Health requests use `AbortController` with an explicit deadline.
- Polling caps each request to the remaining overall timeout.
- A listener that accepts TCP but never responds is therefore reported as occupied/unhealthy within a bounded time.

### 2. Ownership-safe Windows recovery

`WorkerRecovery` is a new Windows-only recovery boundary. It queries a narrow process set through PowerShell/CIM and classifies candidates before terminating anything.

Worker identity is accepted only when:

- the stored PID is still live; and
- its process-creation token matches the token captured at registration.

If CIM temporarily cannot return a creation token, recovery permits a fallback only for an exact claude-mem worker command line/daemon path. A positive token mismatch is treated as PID reuse and is never killed.

Chroma/UV ownership requires an exact claude-mem Chroma data-directory marker or the exact remote UV launcher marker. Recovery selects only roots among the matched process tree, terminates them with `taskkill /PID ... /T /F`, removes the stale PID record, and waits until a bind probe proves the worker port is reusable.

If the port cannot be released, replacement startup is refused.

### 3. Stable process identity

- The Windows PowerShell launcher now uses `Start-Process -PassThru` and returns the real worker PID.
- Invalid and non-positive PID output is rejected.
- Managed process records now include a process-creation token.
- Registry pruning detects PID reuse, while transient token lookup failures do not discard an otherwise live record.
- Platform-specific command-path comparisons use `path.win32`/`path.posix` explicitly.

### 4. Stable Chroma process root

When `uvx.exe` has a sibling `uv.exe`, `ChromaMcpManager` invokes the stable executable directly as:

```text
uv.exe tool uvx ...
```

This avoids recording the short-lived shim as the managed root and makes descendant cleanup reliable.

### 5. Recovery integration

The recovery sequence is shared by:

- MCP/normal worker startup (`worker-spawner.ts`);
- hook lazy-spawn (`worker-utils.ts`); and
- CLI restart fallback (`worker-service.ts`).

Each path rechecks health under the existing spawn lock before cleanup. The Windows spawn cooldown is cleared only after the occupied unhealthy port has been reclaimed successfully.

## Safety boundaries

- Destructive recovery is Windows-only.
- A process-creation token mismatch blocks termination.
- Command-line fallback is limited to the exact worker daemon path.
- Chroma matching requires claude-mem-specific data-directory or launcher markers.
- Only owned process-tree roots are passed to `taskkill /T`.
- Health is rechecked after acquiring the spawn lock to avoid racing a recovering worker.
- A replacement worker is never spawned until the port passes a fresh bind probe.

## Tests

Added or extended coverage for:

- a TCP peer that accepts a connection but never sends HTTP data;
- timeout-bounded health polling;
- orphan root selection and descendant cleanup;
- creation-token match, unavailable-token fallback, and PID-reuse mismatch;
- recovery-before-spawn and refusal to spawn after failed recovery;
- the real PowerShell PID launcher/parser contract;
- process-registry token capture and pruning;
- Windows path normalization; and
- bypassing the short-lived `uvx.exe` shim.

Local verification from a clean worktree based on current `main`:

```text
bun test tests/services/worker-recovery.test.ts \
  tests/services/worker-spawner.test.ts \
  tests/supervisor/process-registry.test.ts \
  tests/infrastructure/health-monitor.test.ts \
  tests/infrastructure/process-manager.test.ts \
  tests/services/integrations/spawn-contract-windows.test.ts

116 pass, 5 platform-skipped, 0 fail

npm run typecheck
pass

npm run build
pass (including bundle-size guardrails)

bun test tests/infrastructure/version-consistency.test.ts \
  tests/infrastructure/worker-json-status.test.ts

34 pass, 0 fail

git diff --check
pass
```

### Live Windows fault injection

I also verified the failure and recovery against the packaged worker on Windows:

1. Start a healthy worker and Chroma process tree.
2. Terminate only the recorded worker PID, leaving its Chroma descendants alive.
3. Confirm that the worker port remains in `LISTENING` while health is unreachable.
4. Run the patched startup path.
5. Confirm that recovery identifies and terminates the owned orphan root and descendants.
6. Confirm that the port becomes bindable before a replacement is launched.
7. Confirm the new worker reaches `ready` and the search endpoint returns HTTP 200.

The same host subsequently encountered the ghost-listener state during a normal restart; the patched restart path logged automatic recovery, released the port, and brought the worker back to `ready` without a Windows reboot.

## Relationship to existing Windows work

This is complementary to #3286, which prevents one important orphaning path by binding the Chroma child tree to a Windows Job Object. This PR provides the recovery backstop for:

- orphaned state that already exists before startup;
- a live worker that is hung rather than exited;
- Job Object setup/assignment failure or degradation;
- bounded health requests and accurate occupied-port detection; and
- safe refusal to respawn when the listener cannot be reclaimed.

The stable Chroma-root work overlaps conceptually with #3286. If that PR lands first, this part can be narrowed during review while retaining the detection, identity checks, and recovery path.

Related context: #3243, #2996, #2926, and #2894.

## Generated artifacts and changelog

The three generated bundles that contain these recovery paths are included:

- `plugin/scripts/worker-service.cjs`
- `plugin/scripts/mcp-server.cjs`
- `plugin/scripts/transcript-watcher.cjs`

The build also produced unrelated viewer/server bundle churn; those files were intentionally excluded to keep this diff scoped. The changelog was not edited because it is generated by the release workflow.

## Validation scope

The targeted Windows recovery suites, typecheck, clean build, built-bundle consistency/status tests, and live fault injection all pass. The complete repository-wide test suite was not run locally; CI remains the source of truth for that broader matrix.


thedotmack:

> step 2/3 (owner record + bounded health, verified reclaim); P1: destructive recovery against a healthy worker when only the PID file is stale

thedotmack:

PR #3308, opened 2026-07-17
fix(worker): pick cache version by semver, not mtime (#3298)

### LINKED ISSUE #3298: cue() picks worker version by mtime, not semver — lands on stale build after installs that write dirs out of order, causing unbounded chroma-mcp/worker leak + OOM
## Summary

`worker-service.cjs`'s version-picker (`cue()`) selects the plugin's script directory by filesystem **mtime**, not semver. When an update writes two adjacent version directories within the same install transaction, whichever one happens to be written *last* wins — even if it's the **older** version. On my machine, `13.10.4/` was written ~48ms after `13.11.0/` during the same install, so `cue()` picked `13.10.4` as "current" despite `13.11.0` being the actual installed version per `installed_plugins.json`.

Because the stale `13.10.4` worker's version-mismatch/recycle logic can't recognize its `13.11.0` sibling as a valid successor (same failure mode as #3205), it falls through to lazy-spawn on **every hook call** — and each lazy-spawn is *also* resolved via the same buggy `cue()`, so it spawns another `13.10.4` daemon forever. Each daemon drags in a `uv`-wrapped `chroma-mcp` (onnxruntime + persistent Chroma DB). Over ~30 minutes of normal Claude Code use this produced 24+ orphaned `bun`/`uv`/`chroma-mcp` process trios and triggered the Linux OOM-killer four times, which also killed unrelated processes on the box (a `gunicorn` server, `systemd`, `chronyd`) as collateral.

This looks like a root cause underlying several of the open "version mismatch never resolves" leak reports (#3205, #3218, #3230, #3216), not just another duplicate of the leak symptom.

## Root cause (code-level)

`worker-service.cjs`, function `cue()`:

```js
function cue(){
  let t = Kn.default.dirname(Kn.default.dirname(Xs)),
      e = Kn.default.join(t,"cache","thedotmack","claude-mem");
  try {
    return (0,mr.readdirSync)(e)
      .filter(r=>/^\d/.test(r))
      .map(r=>Kn.default.join(e,r))
      .filter(r=>{try{return (0,mr.statSync)(r).isDirectory()}catch{return!1}})
      .sort((r,n)=>(0,mr.statSync)(n).mtimeMs-(0,mr.statSync)(r).mtimeMs)  // <-- sorts by mtime, not semver
      .map(PU)
  } catch { return [] }
}
```

`ky()` (the script-path resolver used both for initial spawn and lazy-respawn) calls `cue()` first, before falling back to the marketplace checkout path. Since directory `mtime` reflects extraction/write order during install — not release order — any install that happens to write an older-numbered version directory fractionally later than the newest one permanently pins the worker to the wrong build until the directory timestamps change again (e.g. another update, or a manual `touch`).

## Reproduction (Linux, 13.11.0 installed)

```
$ stat -c '%Y.%N %n' \
    ~/.claude/plugins/cache/thedotmack/claude-mem/13.10.4 \
    ~/.claude/plugins/cache/thedotmack/claude-mem/13.11.0
1783915826.357657505 .../claude-mem/13.10.4
1783915826.309720819 .../claude-mem/13.11.0
```

Both directories were written in the same install (same second), but `13.10.4` is ~48ms newer by mtime, so `cue()` ranks it first. `installed_plugins.json` correctly lists `13.11.0` as the installed version — `cue()` disagrees.

Live symptom, `ps` during an active Claude Code session (hooks firing every ~15–30s from normal tool use):

```
15386 1409 00:13 /home/.../.bun/bin/bun .../13.10.4/scripts/worker-service.cjs --daemon
15478 15386 00:12 uv tool uvx --python 3.13 --with onnxruntime>=1.20 ... chroma-mcp ...
15502 15478 00:12 .../chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma
```
...repeating with a new PID trio roughly every hook call, none reaped. `dmesg`/`journalctl -k` confirmed the OOM-killer firing 4x in ~30 minutes, at one point recording **94 python + 94 uv processes** from this chain (~8 GB RSS) in the kernel's OOM task table.

## Suggested fix

Sort `cue()`'s candidates by parsed semver (or read `version` out of each candidate's own `package.json`/`plugin.json`) instead of `statSync(...).mtimeMs`. Filesystem mtime has no defined relationship to release order and is actively misleading immediately after any install/update that writes multiple version directories close together.

This wouldn't by itself fix the recycle-loop-never-resolves defect described in #3205, but it removes the specific trigger that made the loop latch onto a stale build in the first place.

## Workaround (used to unblock myself)

```
touch ~/.claude/plugins/cache/thedotmack/claude-mem/<newest-version>/
find ~/.claude/plugins/cache/thedotmack/claude-mem/<newest-version>/ -exec touch {} +
```
This forces `cue()` to rank the correct version first again. Not durable — the next update that writes directories in an unlucky order will reproduce it.

## Environment

- claude-mem `13.11.0` (worker script itself running from cache dir `13.10.4` due to the bug above)
- Linux (WSL2, kernel 6.6.87.2-microsoft-standard-WSL2), Bun runtime, `uv`/`uvx` for chroma-mcp
- `installed_plugins.json` correctly pinned to `13.11.0`

Related: #3138 (root-cause plan, worker lifecycle cluster), #3205, #3218, #3230, #3216, #3134

stantheman0128:
Closes #3298

## What was broken

`cacheWorkerScriptCandidates()` (bundled as `cue()` in `worker-service.cjs`) ranked plugin cache version directories by filesystem mtime. During an install that writes two version directories in the same transaction, whichever directory is written last wins even when it is the older semver. On the reporter's machine `13.10.4/` was ~48ms newer than `13.11.0/`, so every spawn/lazy-respawn pinned the stale worker. That stale worker could not recycle to its newer sibling (same failure mode as #3205), so each hook call spawned another orphaned `bun`/`uv`/`chroma-mcp` trio until the OOM killer fired.

## What this change does

- Sort cache version directories by parsed `major.minor.patch` (highest first) in `src/shared/worker-utils.ts`.
- Apply the same semver ranking in Rule A shell/Node launchers (`src/build/hook-shell-template.ts` plus regenerated `plugin/hooks/*.json` and `plugin/.mcp.json`) so hooks and MCP discovery cannot reintroduce mtime ordering.
- Add targeted tests that seed `13.11.0` older-mtime and `13.10.4` newer-mtime and assert the highest semver is first.

No regenerated `worker-service.cjs` bundle in this PR (source + hook launchers only; the worker bundle picks this up on the next release build).

## Verification

Windows 11, bun 1.3.11.

```
$ bun test tests/shared/worker-utils-semver-cache.test.ts
(pass) parseCacheVersionKey reads major.minor.patch from a directory name
(pass) compareCacheVersionNamesDesc ranks higher semver first
(pass) prefers the highest semver even when an older dir has newer mtime
(pass) returns an empty list when the cache root is missing
 4 pass
 0 fail
```

```
$ bun run typecheck

$ bun run lint:hook-io

$ bun run lint:spawn-env

$ bun test tests/infrastructure/plugin-distribution.test.ts --test-name-pattern "Rule A generator parity"
 14 pass
 0 fail
```

Related worker-utils suites on this branch: `worker-utils-version-recycle.test.ts` + `worker-utils-timeout.test.ts` (5 pass).

Shell resolution matrix tests under `plugin-distribution.test.ts` still fail on this Windows box against unmodified `main` as well (empty `RESOLVED=` from the bash fixture); unrelated to this change.

This fix was developed with AI assistance (Cursor/Grok) and verified locally as shown above.


stantheman0128:
Greptile P1 on cache-root priority looks like a false positive (agrees with T-Rex).

Shipped MCP path is the Node launcher in `plugin/.mcp.json`, not the POSIX `allGlobs | sort` block in `candidateBlock`. For `host: 'mcp'`, `buildShellCommand` returns `buildMcpNodeLauncher` early. That launcher builds:

```
K = [E, $PWD/plugin, $PWD, ...L(codex-local), ...L(codex), ...L(claude-cache), marketplace]
```

`L(root)` sorts versions inside one root only. Roots stay in documented priority order, so a preferred Codex cache still wins even when `~/.claude` has a higher semver.

Local check (synthetic HOME, empty cwd, Codex `13.10.4` + Claude `13.11.0`, both with `mcp-server.cjs`): resolver picked the Codex `13.10.4` path. Same as T-Rex. No code change planned for this note.

stantheman0128:

Stan Shih (stantheman0128)

stantheman0128:

Stan Shih (stantheman0128)

thedotmack:
Closing in the PR consolidation pass — superseded: landed as 7d3f1879e (single version oracle, semver not mtime); residual test flaky per review. The analysis here fed the plan-16 design (#3604); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

PR #3309, opened 2026-07-17
fix(worker): stop Windows listen-socket inheritance and reclaim dead spawn locks (#3300)

### LINKED ISSUE #3300: Worker daemon leaks its listening socket into child workers → port never frees → infinite lazy-spawn thrash → multi-GB RAM exhaustion (Windows)
# Worker daemon leaks its listening socket into child workers → port never frees → infinite lazy-spawn thrash → multi-GB RAM exhaustion (Windows)

## Summary

On Windows, once the worker daemon dies uncleanly, port `37800` stays `LISTENING` under the **dead** daemon's PID because the listening socket handle was inherited by its child hook-workers. Nothing serves the health endpoint, so every subsequent hook:

1. sees a **stale `spawn.lock`** (holder PID long dead, never reclaimed),
2. waits out the ~15 s cold-boot timeout,
3. tries to spawn, hits `EADDRINUSE` ("Is port 37800 in use?"),
4. fails — **but the spawned `bun` worker does not exit**; it hangs on an `ESTABLISHED` connection to the dead listener.

This repeats on every hook event, across every concurrent Claude Code session, forever. Result: 30+ orphaned `worker-service.cjs` processes at ~140–185 MB each (~5–7 GB), free RAM driven to ~600 MB on a 16 GB machine, and unrelated apps (Claude Desktop / Electron) OOM-ing with `net::ERR_INSUFFICIENT_RESOURCES`.

## Environment

- **claude-mem**: 13.11.0 (latest; npm `dist-tags.latest` and GitHub both 13.11.0)
- **OS**: Windows 11 Pro (26200)
- **Runtime**: Bun + Node launcher (`bun-runner.js` → `worker-service.cjs`)
- **Host**: Claude Code plugin (`claude-mem@thedotmack`)
- Config: `CLAUDE_MEM_RUNTIME=worker`, fixed port (`37800`; earlier installs `37777`)

Likely regression window: 13.11.0 "retired the standalone `cloud-sync.mjs` daemon; the worker now syncs memories itself." Folding that background flusher into the worker process is a plausible origin for the socket-handle/init-timing change.

## Evidence

**Dead PID still owns the port** (`netstat -ano` + `tasklist`):

```
TCP  127.0.0.1:37800  0.0.0.0:0        LISTENING    20824   <- tasklist: "No tasks are running which match PID 20824"
TCP  127.0.0.1:37800  127.0.0.1:55530  ESTABLISHED  20824   <- 6 live conns pinned to the dead listener
TCP  127.0.0.1:55530  127.0.0.1:37800  ESTABLISHED  11272   <- live bun hook-worker, hung
TCP  127.0.0.1:55531  127.0.0.1:37800  ESTABLISHED  32260   <- live bun hook-worker, hung
TCP  127.0.0.1:37777  0.0.0.0:0        LISTENING    6900    <- older-version orphan, also a dead PID
```

**Stale spawn lock, holder dead** (`~/.claude-mem/spawn.lock`):

```json
{"pid":49576,"startedAt":"2026-07-17T07:29:50.988Z"}   // PID 49576 not alive; lock ~5h old, never reclaimed
```

**Failure counter runaway** (`~/.claude-mem/state/hook-failures.json`):

```json
{"consecutiveFailures":31,"lastFailureAt":1784273388316}
```

**The thrash loop** (`~/.claude-mem/logs/claude-mem-2026-07-17.log`):

```
[SYSTEM] Another launcher holds the spawn lock — skipping lazy-spawn and waiting for its worker
[ERROR ] ✗ Worker failed to start  Failed to start server. Is port 37800 in use?
[SYSTEM] Another launcher holds the spawn lock — skipping lazy-spawn and waiting for its worker
[ERROR ] ✗ Worker failed to start  Failed to start server. Is port 37800 in use?
   ... (repeats indefinitely)
```

**Process pileup**: ~30 `bun.exe` + ~25 `node.exe`, all `worker-service.cjs hook claude-code {observation|file-context|summarize}`, 140–185 MB each. System free RAM hit ~600 MB / 16 GB; a co-resident Electron app logged `net::ERR_INSUFFICIENT_RESOURCES` and multi-second event-loop stalls; even a fresh PowerShell shell threw `System.OutOfMemoryException` on launch.

## Root cause

Two compounding defects:

1. **Listening socket handle inheritance.** Child hook-workers are spawned without excluding the daemon's listening socket fd. On Windows the socket stays bound as long as *any* holder of the inherited handle is alive, so killing/crashing the daemon does **not** free the port — the orphaned children keep it `LISTENING` under the dead parent PID, with no health server behind it.

2. **Stale `spawn.lock` is not force-reclaimed on a dead holder.** The PID-file path has stale detection ("Removing stale PID file…"), but the `spawn.lock` path trusts the lock even when its `pid` is dead, so launchers wait on a holder that will never produce a healthy worker.

Together they form a permanent wedge: port never frees (defect 1) + lock never releases (defect 2) → the lazy-spawn retry never terminates → a worker leaks per hook.

## Reproduction

1. Windows, claude-mem 13.11.0, worker runtime.
2. Run one or more Claude Code sessions so hooks fire, let a worker daemon come up.
3. Kill the daemon uncleanly (OOM, `taskkill /T` on the wrong node, or a crash mid-init).
4. Keep triggering hooks. Observe: port stays `LISTENING` under the dead PID, `spawn.lock` retains the dead PID, `consecutiveFailures` climbs, worker processes accumulate without bound.

## Suggested fixes

- **Do not leak the listening socket into children.** Spawn hook-workers with the server fd excluded from inheritance (e.g. don't pass the server handle in `stdio`; create the listener with `{ ... }` after fork, or explicitly `unref`/close it in the child before it does work). Verify with `netstat -ano` that the port frees within seconds of the daemon exiting.
- **Reclaim a stale `spawn.lock` when its holder PID is dead** — apply the same liveness check already used for the PID file (`process.kill(pid, 0)` → `ESRCH` ⇒ `rmSync(lock)` and proceed), instead of unconditionally waiting out the cold-boot timeout.
- **On `EADDRINUSE`, probe the port's health before deferring.** If the port is bound but `/api/health` fails AND the recorded owner PID is dead, treat it as a stale bind: force-reclaim (kill orphan holders / bail to a fresh port) rather than looping "port in use" forever.
- **Cap the failure loop.** After N `consecutiveFailures`, stop lazy-spawning and surface a single actionable hint rather than leaking a worker per hook.

## Impact

- Uncapped memory growth → host-wide RAM exhaustion, OOM of unrelated apps.
- Zero memory capture while wedged (31 consecutive failures) — the plugin is pure cost until manually unwedged (kill orphans, delete `spawn.lock`, free the port).

stantheman0128:
## Summary

On Windows, Bun/Node listening sockets are inheritable by default. When the worker daemon spawns children (chroma-mcp, SDK agents, etc.) and then dies uncleanly, those children keep port 37800 `LISTENING` under the dead parent PID. Hooks then thrash on lazy-spawn forever.

This PR fixes both compounding defects from #3300:

1. Clear `HANDLE_FLAG_INHERIT` on the listen socket right after bind (via `SetHandleInformation`, with a getsockname HANDLE scan because Bun's `http.Server` does not expose `_handle.fd`).
2. Break `spawn.lock` when the holder PID is dead, even if mtime still looks fresh (Windows indexers can refresh mtime and defeat the 60s breaker alone).

## AI disclosure

This change was prepared with AI assistance (Cursor/Grok). I reproduced the socket inheritance leak on Windows 11, wrote the failing tests first, and verified the commands below myself.

## Verification / Evidence

\\\	ext
bun test tests/shared/worker-spawn-gate.test.ts tests/shared/windows-listen-socket.test.ts
9 pass, 0 fail

bun run typecheck
pass

npm run build
All build targets compiled successfully
\\\

Windows live repro (before clear): parent listens with `http.createServer`, spawns a detached child, `SIGKILL` parent -> port stays `LISTENING` under the dead PID while the child lives. After `clearWindowsListenSocketInherit`, the same sequence leaves the port bindable while the child is still alive.

## Test plan

- [x] Targeted bun tests for spawn-gate dead-PID reclaim and Windows listen-socket inherit clear
- [x] Typecheck
- [x] Build (worker-service.cjs + server-service.cjs include the fix; `bun:ffi` kept external)
- [ ] Maintainer smoke on Windows: kill worker mid-session, confirm port frees without orphan pileup

Closes #3300


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3603 ([plan-15] Worker Port & Liveness Authority)** as a **bundle candidate**: the fix is sound and is slated to be folded into the plan-15 bundle PR with author credit (`Co-authored-by` + `Closes` reference), sequenced by the step noted below. It stays open until that bundle lands.

> step 1 (non-inheritable listener) + step 4 (dead-PID lock reclaim) — exactly the plan's first two steps

thedotmack:
Routed to **#3603 (plan-15, Worker Port & Liveness Authority)** — inherited listen-socket handles and a `spawn.lock` with no holder-liveness check are both named in that master, and this is the only PR that attacks the ghost listener upstream by clearing `HANDLE_FLAG_INHERIT` at bind so it never forms.

It overlaps #3416 and #3408 at the symptom level (a Windows port stuck LISTENING under a dead PID) but not at the fix: plan-15 will likely want this plus one reclaim path, not all three reapers, so nothing is being closed. One practical blocker — the diff still carries regenerated `plugin/scripts/*.cjs` bundles, which will conflict forever; dropping those and keeping the source change would make it mergeable.

PR #3310, opened 2026-07-17
fix(context): exclude subagent observations from SessionStart injection (#3274)

### LINKED ISSUE #3274: Context injection should not include subagent observations (filter agent_id IS NULL, ideally configurable)
## Problem

Since 12.2.2, subagent Stop hooks correctly skip *summarization*, and observations are labeled with `agent_type` / `agent_id`. But subagent **observations** are still captured **and** still injected into a new session's SessionStart context. The context-injection query has no `agent_id` filter, so in subagent-heavy projects the carried-forward memory window is dominated by ephemeral subagent step-logging (test-passed, files-staged, commit-attempted, etc.), which evicts the durable main-agent memory (decisions, bugs, architecture) the injection exists to carry forward.

## Evidence (from a real DB)

The injection is a fixed recency window (`ORDER BY created_at_epoch DESC LIMIT ?`) per project. Composition of the most-recent 50 observations that get injected:

- Subagent-heavy project: **38 subagent / 12 main** → ~76% of the window is subagent process-noise.
- Non-subagent project: **0 subagent / 50 main** → unaffected.

So the harm scales with how much subagent-driven plan execution a project does. (Overall DB in that install: ~940 subagent rows / ~6,600 main.)

## Root cause

The per-project context query (in `context-generator.cjs`, also bundled into `worker-service.cjs`) is:

```sql
FROM observations o
LEFT JOIN sdk_sessions s ON o.memory_session_id = s.memory_session_id
WHERE (o.project IN (...) OR o.merged_into_project IN (...))
  AND (? IS NULL OR s.platform_source = ?)
  AND type IN (...)
  AND EXISTS (SELECT 1 FROM json_each(o.concepts) WHERE value IN (...))
ORDER BY o.created_at_epoch DESC
LIMIT ?
```

The `agent_id` column added in 12.2.2 is exactly the groundwork needed to filter here, but no consumer uses it.

## Proposed fix

Add `AND o.agent_id IS NULL` to the context-injection query (main-agent observations only). Subagent rows stay in the DB and remain searchable (mem-search / get_observations) and in the timeline — they're just no longer force-fed into every new session.

Ideally behind a config flag for people who want the old behavior, e.g. `CLAUDE_MEM_CONTEXT_MAIN_AGENT_ONLY` (default `true`), alongside the existing `CLAUDE_MEM_CONTEXT_*` knobs. A middle option (a `CLAUDE_MEM_CONTEXT_SUBAGENT_TYPES` allowlist) would let useful agent types through while dropping noisy ones, but a simple boolean covers the common case.

## Note

The same filter should **not** be applied to search / timeline / `getObservationsByIds` queries — those should keep returning subagent rows on demand. Only the SessionStart injection query should filter.

stantheman0128:
Fix #3274: exclude subagent observations from SessionStart context injection

## Problem

Since 12.2.2 subagent observations are labeled with `agent_id`, but the SessionStart context-injection query never filters on it. In subagent-heavy projects the fixed recency window (`ORDER BY created_at_epoch DESC LIMIT ?`) fills with ephemeral subagent step-logging (test-passed, files-staged, commit-attempted), which evicts the durable main-agent memory the injection exists to carry forward. The issue reports ~76% of a 50-row window taken by subagent noise in one real install.

## Fix

Add `AND o.agent_id IS NULL` to the SessionStart injection query only (main-agent observations are stored with a null `agent_id`; subagent rows set it). The filter is gated by a new `CLAUDE_MEM_CONTEXT_MAIN_AGENT_ONLY` knob (default `true`), matching the existing `CLAUDE_MEM_CONTEXT_*` settings, so anyone who wants the old behavior can set it to `false`.

The filter is scoped to the injection query. Search, timeline, and `getObservationsByIds` are untouched, so subagent observations stay in the database and remain retrievable on demand, exactly as the issue asks.

## Changes

- `src/services/context/ContextConfigLoader.ts`: parse `CLAUDE_MEM_CONTEXT_MAIN_AGENT_ONLY` (default true).
- `src/services/context/ObservationCompiler.ts`: apply `AND o.agent_id IS NULL` to the injection query when the flag is on.
- `src/services/worker/http/routes/SettingsRoutes.ts`: register the new knob alongside the other context settings.
- `plugin/scripts/context-generator.cjs`, `plugin/scripts/worker-service.cjs`: regenerated bundles.
- `tests/context/observation-compiler.test.ts`: new coverage.

## Verification

Targeted test file (the full `bun test` suite hangs on Windows, so only the touched file plus typecheck were run):

`bun test tests/context/observation-compiler.test.ts`

Before the source change (RED), the two new behavior tests fail:

```
(fail) excludes subagent observations from injection when mainAgentOnly is true (default)
  + "SUB_OBS",
    "MAIN_OBS",
(fail) maps CLAUDE_MEM_CONTEXT_MAIN_AGENT_ONLY to the mainAgentOnly flag
  Expected: false   Received: undefined
 10 pass, 2 fail
```

After the fix (GREEN):

```
 12 pass
 0 fail
 32 expect() calls
Ran 12 tests across 1 file. [1099.00ms]
```

The tests are falsifiable, not tautological: one seeds a main-agent row (`agent_id` null) and a subagent row (`agent_id` set) and asserts injection returns only the main row; its companion flips the flag off and asserts both come back; a third asserts `countObservationsByProjects` still returns both rows so on-demand retrieval is unaffected.

`bun run typecheck` (`tsc --noEmit` + viewer tsconfig): exit 0.

Regenerated bundles embed the flag-gated clause:

```
context-generator.cjs: agent_id IS NULL  (1)
worker-service.cjs:     agent_id IS NULL  (1)
```

## Note

This PR was written with AI assistance (Claude) and verified on a Windows 11 machine. An independent review pass confirmed the filter is applied to only the injection query (not count, summaries, search, timeline, or getObservationsByIds) and that the flag defaults to true. Human author: Stan Shih (stantheman0128).


stantheman0128:

Stan Shih (stantheman0128)

stantheman0128:

Stan Shih (stantheman0128)

stantheman0128:
Pushed test isolation fix for the CI failure on \maps CLAUDE_MEM_CONTEXT_MAIN_AGENT_ONLY\ (mock settings load so persisted runner settings.json cannot skew the default assertion).

thedotmack:

> step 3 (`ContextScope.mainAgentOnly` default true on the injection query)

thedotmack:
Routed to **#3608 (plan-20, Project Identity & Injection Scope)** — deciding what the SessionStart query is allowed to return is the injection-scope half of that master, and the ~76% subagent-noise measurement is a good justification for the default.

Two things to reconcile rather than duplicates: #3410 is what makes the `CLAUDE_MEM_CONTEXT_*` keys actually survive load and reach the query, and #2741 filters subagent observations from a different angle. Plan-20 wants one config path and one filter, so this should sequence behind #3410 (which is currently clean) instead of adding a second knob route. Conflicting against main now.

PR #3311, opened 2026-07-17
fix(project-name): stabilize non-git subdir project keys (closes #3194)

### LINKED ISSUE #3194: Context injection resolves project as basename(cwd) outside git repos while capture attributes to the parent project — subdir launches silently get the first-run banner instead of memory
## Environment

- claude-mem **13.10.2** (Claude Code plugin cache install), Linux
- Worker healthy (`/api/health` ok, `initialized: true`), DB intact (2200+ observations), `hook-failures.json` clean

## Symptom

A session launched from a **subdirectory of a directory that is not a git repo** gets the first-run banner ("This project has no memory yet…") instead of injected context — on session start **and** on `/compact` — even though the database holds months of history for the parent project. Capture keeps working the whole time, so nothing is lost; the read side just looks in an empty bucket. Because the failure is silent (the banner looks like a fresh install, not an error), it's easy to attribute to "memory randomly not working."

Common trigger: background/automation jobs that pin their cwd to a subfolder, or any `claude` launch from below a non-repo home/workspace directory.

## Root cause — read and write paths resolve the project name differently

**Read path (verified in 13.10.2 code):** the SessionStart context hook (`worker-service.cjs … hook claude-code context`, matcher `startup|clear|compact`) takes `cwd` from the hook's stdin payload and resolves the project as `basename(gitRoot(cwd) ?? cwd)`. Commit `daf6d9d` (#2663) made this stable across subdirs **of a git repo** via `rev-parse --show-toplevel` — but when no git root exists anywhere above cwd, the fallback is still `basename(cwd)`, so every subdirectory of a non-repo directory becomes its own brand-new project scope on read.

**Write path (verified empirically, not reverse-engineered):** observations captured from those same subdir-launched sessions are attributed to the **parent** project. Live check on this machine: sessions launched from `/home/lc/<subdir>` (home is not a git repo) write observations that land under project `lc`, and `/api/projects` returns `{"projects":["lc"]}` only — the subdir scopes the injector looks up never exist on the write side.

So for any cwd below a non-repo directory: capture writes to `lc`, inject reads from `basename(cwd)`. The two never meet.

## Repro / direct evidence

With an existing history under project `lc` (home dir `/home/lc`, not a git repo):

```
$ echo '{"session_id":"t1","hook_event_name":"SessionStart","source":"compact","cwd":"/home/lc/bin"}' \
    | node scripts/bun-runner.js scripts/worker-service.cjs hook claude-code context
# → first-run banner ("This project has no memory yet…")

$ echo '{"session_id":"t2","hook_event_name":"SessionStart","source":"compact","cwd":"/home/lc"}' \
    | node scripts/bun-runner.js scripts/worker-service.cjs hook claude-code context
# → full "[lc] recent context" injection, 50 observations
```

Identical payloads, one path component of difference, opposite outcomes. Sessions launched inside a real git repo resolve to the repo's own project on both paths and behave correctly — the mismatch is specific to the no-git-root fallback.

## Ruled out

- Worker down / capture broken / DB corruption — all disproven live (worker healthy, observations from the affected sessions present in the DB under the parent project).
- `CLAUDE_PROJECT_DIR` — unset in these sessions either way; the read-path resolution doesn't consult it.

## Workaround we're running

Local patch to the cached scripts mapping any non-git-repo cwd under the home directory to the home project name, re-applied by a SessionStart hook after plugin updates. Works, but it's machine-specific and shouldn't be needed — mentioning it mainly to confirm that unifying the two resolutions does fix the observed behavior.

Not prescribing which side should be normalized to the other — the read path and write path each have history (#1819, #2663), and you have the full picture of what project identity should mean here. Happy to test a fix or provide more traces.

stantheman0128:
Closes #3194

## What was broken

Outside a git repo, context injection resolved the project as `basename(cwd)`. Sessions launched from a subdirectory of a non-repo directory (common for automation that pins cwd to a subfolder under a home/workspace directory) therefore looked up an empty project bucket and showed the first-run banner, even when months of history already existed under the parent project name.

Capture often attributes those same sessions to the parent directory (Claude Code reports a parent cwd on write hooks, or the parent already holds the historical key), so read and write silently diverged. Git repos were already stable via `#2663` (`rev-parse --show-toplevel`); the non-git fallback was not.

## What this change does

- When `git rev-parse --show-toplevel` fails, walk up for a non-git project marker (`CLAUDE.md`, `package.json`, `pyproject.toml`, `Cargo.toml`, `go.mod`, `.claude-mem.json`) and use that directory's basename as the project key. Subdir launches under a marked parent (e.g. home with `CLAUDE.md`) now share the parent key on both read and write.
- When no marker exists, keep `basename(cwd)` as primary (preserves existing keys) but include the immediate parent basename in `allProjects` so SessionStart context injection can still find parent-scoped memory.
- Route the user-message context banner through `getProjectContext()` / `projects=` instead of raw `basename(cwd)`, matching SessionStart.

Does not treat `~/.claude` as a marker (global Claude Code config would collapse every home subdir into one key). Source only; no regenerated plugin bundles.

## AI assistance

Prepared with AI assistance (Cursor/Grok). Human author: Stan Shih (@stantheman0128).

## Evidence

Windows 11, bun 1.3.11.

New #3194 tests:

```
$ bun test tests/utils/project-name.test.ts --test-name-pattern "3194"
(pass) subdir under a CLAUDE.md parent resolves to the parent basename
(pass) context primary matches for parent and marked subdir
(pass) marker-less subdir keeps basename primary but includes parent in allProjects
(pass) package.json marker wins over cwd basename
 4 pass
 0 fail
```

Full file on this branch (2 pre-existing Windows tilde failures on main, A/B confirmed; unrelated):

```
$ bun test tests/utils/project-name.test.ts
 22 pass
 2 fail
```

The two failures are pre-existing on `main` on Windows (tilde tests split `homedir()` on `/` before `\`, so Expected is `C:\Users\stans` while Received is `stans`).

```
$ bun run typecheck

$ bun run lint:hook-io

$ bun run lint:spawn-env
```


stantheman0128:

Stan Shih (stantheman0128)

stantheman0128:

Stan Shih (stantheman0128)

thedotmack:

> step 1 (marker walk-up before basename fallback)

thedotmack:
Routed to **#3608 (plan-20, Project Identity & Injection Scope)** — a non-git subdirectory launch resolving to a different project key than capture used is precisely the read/write divergence that master exists to end.

I re-checked main after #4055 (`e2d309b37`, anchoring hook project names to `CLAUDE_PROJECT_DIR`): that closes the SDK/subagent temp-cwd case, but `getProjectName` still falls straight through to `basename(cwd)` when `git rev-parse --show-toplevel` fails, so your marker walk-up is still needed and nothing has superseded this. Rebase it onto the new `resolveHookProjectPath` seam and it should slot in cleanly.

stantheman0128:
Merged current main into this branch (no rebase, no force-push).

#4055 / `resolveHookProjectPath` still only runs at the Claude Code hook adapter. It prefers `CLAUDE_PROJECT_DIR` over an SDK temp cwd. It does not replace the #3194 non-git marker walk when `git rev-parse` fails.

This update keeps `findNonGitProjectRoot` on `getProjectName` / `getProjectContext`, keeps user-message on `getProjectContext` + `?projects=` (same as SessionStart) plus the current context-banner line, keeps both the #3194 and #3437 test blocks, and adds one case: `CLAUDE_PROJECT_DIR` points at a non-git `CLAUDE.md` tree, hook cwd is an SDK temp name, and `resolveHookProjectPath` then the marker walk still yields the parent project key. `~/.claude` is still not a marker.

Local: `bun test tests/utils/project-name.test.ts`. The #3194 and #3437 cases passed. Three older tilde-expansion cases fail on this Windows box because the test splits `homedir()` on `/` first (so `C:\Users\stans` becomes the expected name) and because `~/projects/my-app` case-folds onto a real `Projects` directory here. Those are not the #3194 path.

stantheman0128:
Same injection-scope face as the project-key fix.

#3511 closed after SessionStart honored `CLAUDE_MEM_EXCLUDED_PROJECTS`. Capture, file-context, and summarize already used `shouldTrackProject`. UserPromptSubmit (`user-message`) still fetched `/api/context/inject` and printed the banner for an excluded cwd.

`c5406e0a` puts that handler on the same gate. Excluded cwd: exit 0, no worker call, no banner. Tracked cwd: still `getProjectContext` + `?projects=`.

Also fixed three tilde project-name tests that assumed `~/projects/...` is marker-less. On Windows that path case-folds onto a real `Projects` directory, so the #3194 walk is correct and the old expected basename was not.

Local: `bun test tests/utils/project-name.test.ts tests/cli/handlers/user-message-excluded.test.ts` 38 pass. Same bun invocation with `context-session-start`, `should-track-project`, and `hook-stream-discipline` 18 pass.

PR #3316, opened 2026-07-17
feat(hooks): env opt-out for PreToolUse/PostToolUse (#3106)

### LINKED ISSUE #3106: Add an env var to disable the PreToolUse / PostToolUse hooks (Windows console-window focus theft)
**Repo:** thedotmack/claude-mem
**Affects:** 13.8.0 and current `main` (13.9.2) -- unchanged
**Environment:** Windows 11, Claude Code CLI 2.1.178, Git Bash

## Problem

On Windows, every `"shell": "bash"` command hook spawns a visible console window that grabs foreground focus for the moment it runs. Two of the claude-mem hooks fire on nearly every action:

- `PostToolUse` (matcher `*`) -- runs the `observation` hook after *every* tool call
- `PreToolUse` (matcher `Read`) -- runs the `file-context` hook before *every* Read

The result on Windows is a near-constant stream of console windows stealing focus while you work -- it makes the session unusable. The lower-frequency hooks (`SessionStart`, `UserPromptSubmit`, `Stop`) are tolerable; it's specifically the two tool-level hooks that fire too often.

## The old workaround no longer works

Previously you could disable a hook by renaming its event key (e.g. `PostToolUse` -> `_DISABLED_PostToolUse`). A newer Claude Code build tightened hook-config schema validation: it now hard-rejects any key that isn't a valid event name, and fails the **entire** `hooks` record when it sees one. So renaming a key now produces:

```
Plugin (claude-mem@thedotmack): Hook load failed:
  "path": ["hooks", "_DISABLED_PostToolUse_windows_popup_fix"],
  "message": "Invalid key in record"
```

...which takes down *all* claude-mem hooks, not just the one you tried to disable. The only remaining local workaround is to delete the two blocks from the plugin's cached `hooks.json` -- which gets overwritten on every plugin update.

## Request

Add an environment-variable gate so users can turn off the tool-level hooks without editing shipped files. Options, in order of preference:

1. **Single coarse flag:** `CLAUDE_MEM_DISABLE_TOOL_HOOKS=1` -- when set, the `observation` and `file-context` hook commands exit 0 immediately (no bash subprocess work, no window).
2. **Granular flags:** `CLAUDE_MEM_DISABLE_OBSERVATION=1` and `CLAUDE_MEM_DISABLE_FILE_CONTEXT=1`.

Ideally the check happens as early as possible inside `worker-service.cjs` (or the hook entry) so the process short-circuits before doing anything heavy. `SessionStart` / `UserPromptSubmit` / `Stop` capture should keep working, so cross-session memory is unaffected.

A `windowsHide`-style spawn option or routing the hooks through a hidden-window launcher on Windows would also solve the root cause, but an env gate is the simplest cross-platform fix and lets users opt out of the highest-frequency hooks specifically.

## Why this matters

Windows users currently have to hand-edit `plugins/cache/thedotmack/claude-mem/<version>/hooks/hooks.json` after every update, or disable the plugin entirely (losing memory capture). An env var makes the opt-out durable and update-proof.

stantheman0128:
Closes #3106

## Summary

On Windows, every PreToolUse (file-context) and PostToolUse (observation) hook can flash a console window and steal focus. Renaming those keys in hooks.json is no longer valid: Claude Code rejects unknown hook keys and drops the whole plugin hook set.

This adds an env opt-out that exits those two hooks with code 0 before `ensureWorkerStarted` / stdin work:

- `CLAUDE_MEM_DISABLE_TOOL_HOOKS=1` — both tool hooks
- `CLAUDE_MEM_DISABLE_OBSERVATION=1` / `CLAUDE_MEM_DISABLE_FILE_CONTEXT=1` — granular

SessionStart / UserPromptSubmit / Stop stay active. Documented in configuration + troubleshooting. Source-only (no regenerated `worker-service.cjs`); takes effect on the next release build that bundles worker-service.

## Verification

Windows 11, bun 1.3.11.

```
$ bun test tests/hook-constants.test.ts tests/hook-command.test.ts
 50 pass
 0 fail

$ bun run typecheck

$ bun run lint:hook-io

$ bun run lint:spawn-env
```

This fix was developed with AI assistance (Cursor/Grok) and verified locally as shown above.

### DIFF
--- a/docs/public/configuration.mdx
+++ b/docs/public/configuration.mdx
@@ -22,6 +22,11 @@ Settings are managed in `~/.claude-mem/settings.json`. The file is auto-created 
 | `CLAUDE_MEM_WORKER_HOST`      | `127.0.0.1`                     | Worker service host address           |
 | `CLAUDE_MEM_DATA_DIR`         | `~/.claude-mem`                 | Data root — every other path (database, chroma, logs, settings.json, worker.pid) derives from this |
 | `CLAUDE_MEM_SKIP_TOOLS`       | `ListMcpResourcesTool,SlashCommand,Skill,TodoWrite,AskUserQuestion` | Comma-separated tools to exclude from observations |
+| `CLAUDE_MEM_DISABLE_TOOL_HOOKS` | _(unset)_                         | Set to `1` to no-op PreToolUse (`file-context`) and PostToolUse (`observation`) hooks while keeping SessionStart / UserPromptSubmit / Stop active (useful on Windows when bash hook consoles steal focus) |
+| `CLAUDE_MEM_DISABLE_OBSERVATION` | _(unset)_                      | Set to `1` to no-op only the PostToolUse `observation` hook |
+| `CLAUDE_MEM_DISABLE_FILE_CONTEXT` | _(unset)_                     | Set to `1` to no-op only the PreToolUse `file-context` hook |
+
+These disable flags can also be set under the `env` block in `~/.claude/settings.json` so Claude Code inherits them for every hook spawn.
 
 ### Gemini Provider Settings
 
--- a/docs/public/troubleshooting.mdx
+++ b/docs/public/troubleshooting.mdx
@@ -150,6 +150,24 @@ PORT=${CLAUDE_MEM_WORKER_PORT:-$(jq -r .CLAUDE_MEM_WORKER_PORT ~/.claude-mem/set
 
 5. If Chroma continues to fail, hybrid search will gracefully degrade to SQLite FTS5 only
 
+### Windows: console windows steal focus on every tool call
+
+**Symptoms**: On Windows, Claude Code flashes a console window (and steals focus) on nearly every Read / tool call while claude-mem is enabled.
+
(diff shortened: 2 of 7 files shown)


stantheman0128:

Stan Shih (stantheman0128)

stantheman0128:

Stan Shih (stantheman0128)

thedotmack:

> step 2 explicitly adopts `CLAUDE_MEM_DISABLE_TOOL_HOOKS=1` short-circuit before any work

thedotmack:
Routed to **#3605 (plan-17, Hook Wrapper Contract)** — flashing console windows on every PreToolUse / PostToolUse hook is named in that master, and confirmed `CLAUDE_MEM_DISABLE_TOOL_HOOKS` is still absent from main, so the escape hatch does not exist today.

Saying it plainly so the sequencing is clear: plan-17's actual fix is to stop spawning a console at all (the exec-form launcher work in #3403 / #3564), so this should land as the opt-out for people who need relief now, not as the cure — and the docs wording should say so. No duplicate; needs a rebase.

PR #3321, opened 2026-07-18
fix(windows): resolve PATH binaries via where.exe argv

stantheman0128:
## Summary
Resolve PATH binaries with where.exe argv instead of shell `where ...` strings.

- ProcessManager.lookupBinaryInPath: execFileSync("where.exe"|"which", ...) + windowsHide
- findClaudeExecutable Windows discovery: same pattern, CRLF-safe split
- Unit tests updated for where.exe argv keys

## AI assistance
Prepared with Cursor (Grok). Human reviewed and verified on Windows 11.

## Evidence
```
bun test tests/shared/find-claude-executable.test.ts -t "on Windows"
1 pass (Windows where-discovery preference)
```
Pre-existing Windows-only failure remains in the darwin known-install-locations case (path.join separators). A/B against origin/main shows the same fail before this diff.

## Test plan
- [x] bun test tests/shared/find-claude-executable.test.ts -t "on Windows"
- [ ] CI

### DIFF
--- a/src/services/infrastructure/ProcessManager.ts
+++ b/src/services/infrastructure/ProcessManager.ts
@@ -2,7 +2,7 @@
 import path from 'path';
 import { homedir } from 'os';
 import { existsSync, writeFileSync, readFileSync, unlinkSync, mkdirSync, statSync, utimesSync, copyFileSync } from 'fs';
-import { execSync, spawnSync } from 'child_process';
+import { execFileSync, execSync, spawnSync } from 'child_process';
 import { spawnHidden } from '../../shared/spawn.js';
 import { logger } from '../../utils/logger.js';
 import { sanitizeEnv } from '../../supervisor/env-sanitizer.js';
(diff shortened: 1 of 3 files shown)


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3607 ([plan-19] Observer Subprocess Isolation)** as **standalone / merge-now**: it is small and self-contained enough to land on its own ahead of the plan-19 bundle once CI is green.

> step 2/5 (argv `execFileSync` binary discovery, no shell strings, CRLF-safe)

thedotmack:
Routed to **#3607 (plan-19, Observer Subprocess Isolation)** — `findClaudeExecutable` and `ProcessManager.lookupBinaryInPath` are the executable-resolution seam that master owns, and swapping shell `where ...` strings for `execFileSync('where.exe', …)` with `windowsHide` is the right fix for it.

Not a duplicate of #3291, which self-heals an unspawnable CLI *after* resolution — this fixes resolution itself. Note that main already uses `where.exe` in `src/npx-cli/commands/ide-detection.ts` and `CodexCliInstaller.ts`, so this brings the two remaining sites in line with an existing convention rather than introducing one. Rebase needed.

PR #2583, opened 2026-05-20
feat: add PreCompact hook to capture session state before context compaction

AmeliaLiu0803:
## Summary

- Add a **PreCompact** lifecycle hook that fires before Claude Code context compaction, capturing key decisions, findings, and session state so they persist through context window resets.
- The hook reuses the existing summarize infrastructure: PreCompact queues a summary task → AI generates a structured summary → stored in `session_summaries` → surfaced on next SessionStart via existing context injection.
- Added `CLAUDE_MEM_PRECOMPACT_ENABLED` environment variable for **opt-in control**.

## Who is this for

This is an **opt-in advanced feature** for users with unusually long or information-dense sessions (e.g., long-running monitoring loops, multi-step debugging, PR workflows) where mid-session compaction risks losing critical context. Most users will continue to rely on the existing Stop-hook summarization, which already captures session state at session end.

## Files Changed

| File | Change |
|------|--------|
| `plugin/hooks/hooks.json` | Added PreCompact hook entry |
| `src/cli/handlers/pre-compact.ts` | New handler (mirrors `observation.ts` pattern) |
| `src/cli/handlers/index.ts` | Mapped `'pre-compact'` event type |
| `src/services/worker-types.ts` | Widened `PendingMessage.type` union to include `'pre-compact'` |
| `src/services/worker/SessionManager.ts` | Added `queuePreCompact()` method |
| `src/services/worker/http/routes/SessionRoutes.ts` | Added `/api/sessions/pre-compact` route + tier routing support |
| `src/services/worker/ClaudeProvider.ts` | Handle `'pre-compact'` message type (same as `'summarize'`) |
| `src/services/worker/GeminiProvider.ts` | Handle `'pre-compact'` message type |
| `src/services/worker/OpenRouterProvider.ts` | Handle `'pre-compact'` message type + widened type signature |

## How It Works

```
PreCompact Hook (before compaction)
  → stdin JSON → pre-compact handler
    → POST /api/sessions/pre-compact
      → sessionManager.queuePreCompact()
        → ensureGeneratorRunning()
          → AI generates structured summary
            → stored in session_summaries table

SessionStart (source: "compact")
  → context handler calls /api/context/inject
    → queries session_summaries
      → surfaces most recent summary in additionalContext
```

## Known Risks

[...]

PR #3322, opened 2026-07-18
feat: Endless Mode v1 — post-compaction session continuity via message-in-a-bottle

thedotmack:
Endless Mode and the Message in a Bottle

Every long Claude Code session ends the same way. You build shared knowledge for hours — the plan, the weird bug you fixed together, how you like your commits:

The Perfect Collaboration

…and then the context window fills, compaction squeezes the whole conversation into a few paragraphs of summary, and your collaborator comes back *a little dumber*. Re-explains things you both knew. Suggests the approach you rejected an hour ago. Sometimes re-sends the exact message it just sent.

Then, Compaction Happens

**This PR is the cure.** Three characters do the work — claude-mem's hooks were already writing everything down; a new `BottleRenderer` service rebuilds the story after every wipe; and the post-compaction model just has to read one file:

Meet the Endless Mode Crew

The whole design rests on one observation: **compaction only destroys the context window.** The transcript JSONL and claude-mem's database sit on disk, untouched. There is nothing to rescue before the wipe — no PreCompact hook, no seal/drain machinery. After the wipe, we simply rebuild:

The Core Insight: Nothing is Actually Lost

The rebuild is called the **bottle**: your messages and Claude's messages verbatim (the 5–20% of tokens that matter), with the tool noise (the other 80–95%) replaced by the one-line observations claude-mem already wrote in real time. Result: a faithful session record 10–20× smaller than the context it replaces, at `~/.claude-mem/bottles/<session-id>.md`:

Crafting the Bottle

The new context doesn't get the bottle shoved into it. It gets a **~20-line pointer**: *read this file and continue from where it ends.* The model pulls the bottle in via its own Read tool — which means it arrives as a tool result, exactly the category the *next* compaction evicts first. The bottle never accumulates across legs. And one line in the pointer does disproportionate work:

The Magic Trick: A Pointer, Not a Payload

Why not just trust the summary? Because a summary is a paraphrase written about the conversation; the bottle *is* the conversation. The model continues from its own actual last words:

Summaries vs. The Bottle

Three rules keep it honest. When the transcript has been pruned (~30-day retention, so most old sessions), the bottle **degrades honestly** — prompts stay verbatim from the DB, summaries render clearly marked as generated, and Claude's old replies are simply absent, never faked. Every rendered line passes the `<private>` **tag-stripper**, verified by tests that read the actual file off disk. And if anything fails — worker down, nothing to render, timeout — you get **exactly today's behavior**. The safety floor is the status quo:

The Rules of the Realm

It ships on by default, fires automatically on every compaction and resume, and keeps a live session diary fresh after every turn. Power users can go further: two env vars make compaction fire at ~30% of the window instead of ~95% — short legs, frequent cheap handovers, a smooth rhythm instead of one traumatic wipe:

From Traumatic Wipes to an Endless Rhythm

Built July 17 in one orchestrated session: 6 commits, 81 new tests, full suite green, and an adversarial review gauntlet that caught real bugs before they shipped (subagent chatter masquerading as user turns, the harness compact-summary blob rendering as "turn 1", a fake-background render, a path-traversal hole — all fixed and locked with tests):

Built, Tested, Ready Today

---

### What's in the PR

| Piece | Where |
|---|---|
| `BottleRenderer` (full + reconstructed modes, save-time-mirrored turn numbering, `stripMemoryTags` on every rendered message) | `src/services/worker/BottleRenderer.ts` |
| `POST /api/sessions/render-bottle` (settings-gated; `wait:true` sync, queue-and-return default) | `src/services/worker/http/routes/SessionRoutes.ts` |
| SessionStart pointer injection, `resume` matcher, `/clear` archive | `src/cli/handlers/context.ts`, `plugin/hooks/hooks.json` |
| Render-on-Stop live diary (awaited, 5s, never affects the Stop result) | `src/cli/handlers/summarize.ts` |
| Injected-envelope classifier (`system-reminder` / command wrappers / task-notification / caveat) | `src/utils/tag-stripping.ts` |
| Foundations: `BOTTLES_DIR`, `writeTextFileAtomic`, 3 SessionStore read methods, `CLAUDE_MEM_ENDLESS_MODE_ENABLED` (default **on**) | `src/shared/`, `src/services/sqlite/` |
| Docs incl. optional 30%-trigger env vars | `docs/public/endless-mode.mdx` |
| 81 new tests (renderer, route, hooks, adapter, privacy, idempotency) | `tests/` — full suite 2377 pass / 0 fail |

### How to test

1. Check out this branch and `npm run build-and-sync` (syncs your live plugin + restarts the worker — this is the point: dogfood it).
2. Work a session past auto-compact. At the top of the new context you should see the `[claude-mem] Endless Mode — session continuation` pointer, the model should Read the bottle and continue **without repeating its last message or redoing finished work**.
3. Check `~/.claude-mem/bottles/<sessionId>.md` — conversational spine verbatim, observations in place of tool noise, **no `<private>` content**.
4. Resume an old session (`--resume`) whose transcript is pruned → bottle should say `mode: reconstructed`, with no fabricated assistant turns.
5. `/clear` → bottle moves to `~/.claude-mem/bottles/archive/`.
6. Optional early-handover throttle (see docs page): in `~/.claude/settings.json` set `"env": { "CLAUDE_AUTOCOMPACT_PCT_OVERRIDE": "30", "CLAUDE_CODE_AUTO_COMPACT_WINDOW": "200000" }` — short legs, frequent handovers, much easier to reproduce step 2.
7. Kill switch: `CLAUDE_MEM_ENDLESS_MODE_ENABLED: "false"` in `~/.claude-mem/settings.json` (no restart needed).

### Open questions for testers

- **`/clear` session-id semantics**: does SessionStart(clear) carry the old or a new `session_id`? Doc sources contradicted each other. Old id → archive works but pre-clear content can reappear in a later bottle (v2 needs a clear-boundary cut). New id → archive is a harmless no-op and forgetting is automatic. Either is safe in v1; report what you observe.
- **The experimental compaction note**: every injection now ends with a nudge telling the auto-compaction summarizer a minimal summary suffices. It's not obligated to listen. If you run multiple legs, compare summary lengths and tell us whether it does anything.
- **Turn alignment**: builtin commands (`/model`, `/effort`) shift full-mode observation grouping by one (they appear in transcripts but never fire UserPromptSubmit). Misgrouped observations land in a trailing bucket rather than disappearing. Report if you see worse.

### Known v1 non-goals (deliberate)

No PreCompact hook, no seal/drain, no bottle-state table, no byte cursors or render mutex (double-fire is idempotent), no `tool_use_id` threading, no leg roll-up — the bottle is a derived view, rendered whole, never maintained. Codex/other platforms are explicitly out: everything is gated to claude-code.


thedotmack:
The earlier consolidation pass missed this PR, so routing it now: **#2785 (plan-12, Provider & Extensibility Roadmap)**. Endless Mode is a net-new continuity capability rather than a fix for any of the plan-14…25 defect clusters, even though it lands in the Stop / context hook path that plan-17 and plan-20 also touch.

State of it: conflicting against main, no review since July, and it overlaps the compaction-adjacent work in #2583 (PreCompact capture). Worth an explicit keep-or-park call before it drifts further from the hook manifest, which has changed several times underneath it.

PR #3323, opened 2026-07-18
feat: honor CLAUDE_MEM_PUBLIC_URL for the live-view URL printed to users

gartz:
## What

When claude-mem runs in a remote sandbox / dev-container, the worker's live view is reached through a port-forward alias (e.g. `https://37700.host.<user>.<domain>`), not `localhost`. The three places that print the viewer URL hardcode `http://localhost:<port>`, which is unreachable from the user's browser:

- the "no memory yet" welcome hint — `SearchRoutes.ts`
- the SessionStart "View Observations Live @ …" line — `cli/handlers/context.ts`
- the per-message "📺 Watch live in browser …" banner — `cli/handlers/user-message.ts`

## How

- Register `CLAUDE_MEM_PUBLIC_URL` in `SettingsDefaultsManager.DEFAULTS` (and the `SettingsDefaults` interface). This is what makes `loadFromFile` (settings.json) and `applyEnvOverrides` (env) carry the value through to loaded settings — both drop keys not present in `DEFAULTS`.
- Add a `getViewerBaseUrl(port)` helper in `worker-utils.ts`. It prefers `process.env.CLAUDE_MEM_PUBLIC_URL`, then the settings value (trailing slash trimmed), and falls back to `http://localhost:<port>`.
- Use it at the three print sites.

Empty/unset preserves the exact current `http://localhost:<port>` output — **no behavior change for local users.**

## Test

`tests/shared/worker-utils-viewer-url.test.ts` — env precedence over settings, settings value used when env unset, trailing-slash trim, and the localhost fallback. `bun test` green; `tsc --noEmit` clean.

## Why (context)

We run claude-mem inside remote editor sandboxes where the worker binds `0.0.0.0` and is reached at a per-user proxy alias. An external carrier already writes the browser-reachable `CLAUDE_MEM_PUBLIC_URL` into `~/.claude-mem/settings.json` per editor; today the worker discards the key and prints `localhost`. This change lets the plugin honor it. It's fully opt-in and inert unless the value is set.

### DIFF
--- a/src/cli/handlers/context.ts
+++ b/src/cli/handlers/context.ts
@@ -9,6 +9,7 @@ import {
   executeWithWorkerFallback,
   isWorkerFallback,
   getWorkerPort,
+  getViewerBaseUrl,
 } from '../../shared/worker-utils.js';
 import { getProjectContext } from '../../utils/project-name.js';
 import { HOOK_EXIT_CODES } from '../../shared/hook-constants.js';
@@ -141,7 +142,7 @@ export const contextHandler: EventHandler = {
     const displayContent = coloredTimeline || (platform === 'antigravity-cli' ? additionalContext : '');
 
     const systemMessage = showTerminalOutput && displayContent
-      ? `${displayContent}\n\nView Observations Live @ http://localhost:${port}`
+      ? `${displayContent}\n\nView Observations Live @ ${getViewerBaseUrl(port)}`
       : undefined;
 
     return {
--- a/src/cli/handlers/user-message.ts
+++ b/src/cli/handlers/user-message.ts
@@ -5,6 +5,7 @@ import {
   executeWithWorkerFallback,
   isWorkerFallback,
   getWorkerPort,
+  getViewerBaseUrl,
 } from '../../shared/worker-utils.js';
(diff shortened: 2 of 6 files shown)


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#2785 ([plan-12] Provider & Extensibility Roadmap)** as **standalone / merge-now**: it is small and self-contained enough to land on its own ahead of the plan-12 bundle once CI is green.

> `CLAUDE_MEM_PUBLIC_URL` for remote-sandbox viewer links; small opt-in; rebase

thedotmack:
Still routed to **#2785 (plan-12, Provider & Extensibility Roadmap)** as a small standalone capability.

Checked main for this pass: `CLAUDE_MEM_PUBLIC_URL` is honoured only by the server runtime (`src/server/routes/v1/ServerV1PostgresRoutes.ts`), there is no `getViewerBaseUrl` anywhere, and the three worker-side call sites still hardcode `http://localhost:<port>` — so nothing has superseded this and the sandbox/dev-container case is still broken. No duplicate in the queue; it just needs a rebase to land.

PR #3339, opened 2026-07-20
fix: cap OpenRouter context history

desmond-rai:
## What changed

- Add `CLAUDE_MEM_OPENROUTER_MAX_CONTEXT_MESSAGES` with a default of 20 and a safe 2-200 range.
- Keep the initial observer prompt plus the newest messages when an OpenRouter session exceeds the cap.
- Preserve the full in-memory session history while bounding only the outbound API request.
- Ship the corresponding worker bundle and regression coverage.

## Why

OpenRouter observation requests reused the complete multi-turn history. Long Claude Mem sessions therefore resent an ever-growing context for every tool event, causing extreme token usage even when the configured context-message cap was present in `settings.json`.

## Impact

OpenRouter requests remain bounded by default while retaining the observer instructions and latest context. Existing users can tune the cap through settings without changing providers or models.

## Validation

- `bun test tests/openrouter_provider.test.ts tests/shared/settings-defaults-manager.test.ts` - 37 passed, 0 failed
- `bun run typecheck:root`
- `node --check plugin/scripts/worker-service.cjs`
- Local worker probe: 43 post-fix calls, maximum 20 messages per request, no runtime errors


thedotmack:

> step 3 (budget object shared by all providers) — apply at OpenAICompatibleProvider level, not OpenRouter-only; drop committed bundle

thedotmack:
Routed to **#3606 (plan-18, Observer Response Pipeline)** — unbounded `conversationHistory` on the OpenRouter path is named directly in that master, and the diagnosis here is correct.

Re-reading the queue for this pass, #3868 fixes the same defect in the same provider as a strict superset: it bounds the outbound history *and* re-anchors the observation schema in a system message, backed by logs showing 99.1% of idle discards sitting exactly at the context ceiling, and it is current against main where this branch conflicts. Closing as superseded by #3868 — the bounded-history behaviour you argued for is what plan-18 will ship.

PR #3346, opened 2026-07-21
fix(deploy): pin container identity/ports, fix worker queue-prefix mismatch

lukaj99:
## Summary

- `docker-compose.yml`'s `claude-mem-server` published port 37877 as an unqualified `"37877:37877"`, which Docker Compose resolves to `0.0.0.0` — exposing the server on every interface, including the public one, on any host without a hardened default-deny firewall. Changed to explicit `127.0.0.1` (local hooks/healthcheck) + a configurable Tailscale-style IP (`CLAUDE_MEM_TAILSCALE_IP`, cross-machine access) bind only.
- `claude-mem-worker` never set `CLAUDE_MEM_SERVER_PORT`/`CLAUDE_MEM_WORKER_PORT`. The BullMQ queue prefix is derived from `CLAUDE_MEM_WORKER_PORT` (falling back to a UID-derived port when unset — see `SettingsDefaultsManager.ts`), so with these unset the worker silently consumed a different, permanently-empty queue than the one the server enqueued jobs under. Generation looked "disabled" but was actually just listening to the wrong queue, indefinitely, with no error. Pinned both to `37877` to match the server.
- `CLAUDE_MEM_OPENROUTER_BASE_URL` was missing from the worker service's environment block entirely, so a bare `docker compose up` using an OpenRouter-compatible custom endpoint would silently fall back to the default instead.
- Pinned `image:` on both server/worker services (alongside the existing `build:`) so a bare `docker compose up` reuses the last-built image instead of implicitly rebuilding and potentially drifting from a known-good state.
- Added `container_name` to all four services so they match what ops tooling / runbooks typically reference by fixed name.
- Marked the three named volumes `external: true` so redeploying via compose reuses existing data (Postgres tables, Valkey queue state, claude-mem data dir) instead of silently creating fresh empty volumes on next `up`.

## Context

Found and fixed while debugging a production deployment where the worker had been silently stuck on an empty queue for a week — `docker compose up` from this file reproduces the exact same silent failure from a clean state, since the missing port env vars and the unqualified port bind are present in the tracked file itself, not just a runtime drift issue.

## Test plan

- [x] `docker compose config --quiet` validates cleanly
- [x] Verified port bindings resolve to explicit `127.0.0.1` + configured IP (not `0.0.0.0`) via `docker compose config`
- [x] Confirmed via BullMQ Redis keys that server and worker now share the same `claude_mem_37877` queue prefix
- [x] Full stack (`postgres`, `valkey`, `claude-mem-server`, `claude-mem-worker`) brought up via `docker compose up -d` against pre-existing external volumes — all four report healthy, worker resumed draining a real backlog (observed `completed` counters incrementing, `waiting` decreasing) with zero new errors

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -54,6 +54,7 @@
 services:
   postgres:
     image: postgres:17-alpine
+    container_name: claude-mem-postgres
     restart: unless-stopped
     environment:
       POSTGRES_USER: ${POSTGRES_USER:?POSTGRES_USER is required}
@@ -70,6 +71,7 @@ services:
 
   valkey:
     image: valkey/valkey:8-alpine
+    container_name: claude-mem-valkey
     restart: unless-stopped
     # BullMQ requires noeviction; AOF gives durability across restarts.
     command:
@@ -89,9 +91,11 @@ services:
       retries: 12
 
   claude-mem-server:
+    image: claude-mem-server:${CLAUDE_MEM_IMAGE_TAG:-v13.11.0}
     build:
       context: .
       dockerfile: docker/claude-mem/Dockerfile
+    container_name: claude-mem-server
     restart: unless-stopped
     depends_on:
       postgres:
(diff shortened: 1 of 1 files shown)


lukaj99:
Closing — self-review caught that this bundled two genuinely universal upstream bugs (worker missing CLAUDE_MEM_SERVER_PORT/WORKER_PORT causing a queue-prefix mismatch, and a missing CLAUDE_MEM_OPENROUTER_BASE_URL passthrough on the worker) with changes that were specific to my own deployment and wrong as general defaults:

- hardcoded my personal Tailscale IP as the port-bind default
- added `container_name` to claude-mem-worker, which breaks the `--scale claude-mem-worker=N` workflow this file's own header documents as supported
- marked volumes `external: true`, which breaks a fresh clone's first `docker compose up` (no pre-existing volumes to attach to)

Will reopen with a narrower PR containing just the two real fixes. Apologies for the noise.

PR #3350, opened 2026-07-21
fix(paths): expand ~ in CLAUDE_MEM_DATA_DIR (stray ~ dirs)

quinnmacro:
## Problem

A literal `~`-prefixed `CLAUDE_MEM_DATA_DIR` (e.g. `~/.claude-mem` written into `settings.json`, or set as an env var from a non-shell context) produced stray directories literally named `~` across the workspace.

`resolveDataDir()` returned the value verbatim from both the env var and `settings.json`, without tilde expansion. Node's `path.join` / `fs` do **not** expand `~` (only the shell does), so the value was treated as a *relative* path and resolved against the process cwd. claude-mem workers inherit the cwd of whatever spawned them — a subagent pinned to a subdirectory, a plugin-install dir, the repo root, an Obsidian plugin folder — so each worker created its own `cwd/~/.claude-mem/` tree (logs + a `settings.json` snapshot), e.g.:

```
<repo>/~/.claude-mem/logs/...
<repo>/<subdir>/~/.claude-mem/logs/...
<repo>/<subdir>/<nested-dir>/~/.claude-mem/logs/...
<repo>/<plugin-dir>/~/.claude-mem/...
```

The default value (`join(homedir(), '.claude-mem')`) was already absolute, so this only bit users whose `settings.json` literally contained `"~/.claude-mem"` (hand-edited configs, or configs written by tooling that didn't expand `~`).

## Fix

Add `expandHome()` in `src/shared/paths.ts` and route both return paths of `resolveDataDir()` through it:

- `~/foo` → `join(homedir(), 'foo')`
- bare `~` → `homedir()`
- absolute paths → untouched
- relative paths without `~` → untouched
- `~user/...` → intentionally untouched (resolving another user's home is out of scope and platform-dependent)

This keeps every downstream path (`LOGS_DIR`, `DB_PATH`, `paths.*`) absolute regardless of how the value was authored, without changing behavior for the (already-absolute) default.

## Verification

- `bun test tests/shared/paths.test.ts` → 12 pass (6 new `expandHome` cases + 3 `resolveDataDir` tilde cases + 3 existing)
- `bun test tests/shared/` → 130 pass, 0 fail
- `tsc --noEmit` → 0 errors

The `resolveDataDir` tests exercise the env-var branch (consulted first) so they don't touch the real `~/.claude-mem/settings.json` on disk.

## Workaround already in place locally

Beyond this source fix, the equivalent defense without a code change is to set `CLAUDE_MEM_DATA_DIR` to an absolute path in the Claude Code `env` block (`~/.claude/settings.json`) so `te()`/`resolveDataDir()` hits the env-var branch first — useful for anyone who can't upgrade immediately.

### DIFF
--- a/src/shared/paths.ts
+++ b/src/shared/paths.ts
@@ -14,9 +14,30 @@ function getDirname(): string {
 
 const _dirname = getDirname();
 
+/**
+ * Expand a leading `~/` (or a bare `~`) to the user's home directory.
+ *
+ * Node's `path.join` / `fs` do NOT expand `~` — only the shell does. So a
+ * literal `~/.claude-mem` read from `settings.json` or an env var is treated
+ * as a *relative* path, creating a directory literally named `~` in the
+ * process cwd. claude-mem workers inherit the cwd of whatever spawned them
+ * (subagents pinned to a subdirectory, a plugin-install dir, etc.), so a
+ * `~`-prefixed DATA_DIR scattered stray `~/.claude-mem/` trees across the
+ * workspace. Expanding here keeps every downstream path absolute regardless
+ * of how the value was written.
+ */
+export function expandHome(p: string): string {
+  if (typeof p !== 'string' || p.length === 0) return p;
+  if (p === '~') return homedir();
+  if (p.startsWith('~/')) return join(homedir(), p.slice(2));
+  // A `~user/...` form is intentionally left untouched — resolving another
+  // user's home is out of scope and platform-dependent.
+  return p;
+}
+
 export function resolveDataDir(): string {
   if (process.env.CLAUDE_MEM_DATA_DIR) {
-    return process.env.CLAUDE_MEM_DATA_DIR;
+    return expandHome(process.env.CLAUDE_MEM_DATA_DIR);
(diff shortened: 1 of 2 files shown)


quinnmacro:
Hi 👋 — gentle follow-up on this one.

This came from a real production hit: a `~`-prefixed `CLAUDE_MEM_DATA_DIR` (written literally into `settings.json`) produced stray directories named `~` scattered across the workspace wherever a worker's cwd happened to be (subagent-pinned subdirs, plugin-install dirs, the repo root). The fix is narrow — route both `resolveDataDir()` return paths through a new `expandHome()` that only touches leading `~/` and bare `~`; absolute and non-tilde paths are byte-for-byte unchanged.

Greptile's T‑Rex already cloned the branch and ran the tests green (focused path tests, the shared suite, and `tsc --noEmit` — all exit 0), so the verification is already on the PR.

Happy to squash, rebase onto a newer `main`, split the test additions into a separate commit, or adjust the `~user/` scope decision if you'd prefer different behavior. Just let me know.

PR #3352, opened 2026-07-21
fix: strengthen anti-repetition guidance in observer prompt

ryanoboril:
## Summary

Small, generation-side complement to #3038 and #3163. Neither of those issues' proposed fixes touch the observer's own prompt, so I'm filing this separately rather than duplicating the write-time dedup work they already scoped.

`content_hash` only catches byte-identical observations. When the primary session re-confirms the same fact across an iterative loop (PR review rounds, retries, polling), the observer re-summarizes it in different words each time, and the hash never matches, so it gets stored as a "new" observation.

## Evidence

Reproduced against a real `claude-mem.db`: one single user prompt (constant `prompt_number` across all rows, confirming this happened within one turn, not across repeated user messages) produced 9 separate observation batches over about 10 minutes, several of them near-identical paraphrases of a single underlying bug fix. All 9 batches were variations on the same three facts (a validation-guard change, an error-registry update, and a decision to reuse existing error-handling infrastructure instead of adding a new module), each phrased differently, each with a distinct `content_hash` despite describing the same fact.

This matches the shape both #3038 and #3163 already documented from their own DBs.

## Why a prompt fix helps here

The observer's own `conversationHistory` already includes everything it previously wrote earlier in the session (see `di()` in `worker-service.cjs`, which pushes each generation's output into history before the next call). So it has the data needed to recognize a repeat, the prompt just never instructed it to check. `skip_guidance` already had one vague bullet ("Repetitive operations you've already documented") but nothing calling out the actual failure mode (iterative loops re-confirming the same outcome), and nothing telling it to check its own prior output before writing.

## What this does NOT fix

This is instruction-following on a (by default) small model, not a hard constraint, so it won't eliminate the problem. The write-time fuzzy/semantic dedup guard proposed in #3038 and the capture-side suppression in #3163 are still the right backstop for cases where the model doesn't comply. This PR just reduces how often the paraphrase gets generated in the first place, cheaply, with no schema or runtime changes.

## Change

Two prompt-text edits in `plugin/modes/code.json`:
- `recording_focus`: added an explicit instruction to check prior messages in the conversation before writing, and skip if already recorded.
- `skip_guidance`: added a specific bullet naming iterative-loop re-confirmation as a skip case, ahead of the existing generic bullets.

No code, schema, or config changes.


thedotmack:

> step 5 adopts exactly this ("check your prior output" line in the observation prompt)

thedotmack:
Routed to **#3606 (plan-18, Observer Response Pipeline)** under the observation-output-quality line of that master.

Deliberately not a duplicate of the write-time dedup work in #3038 / #3063: a content hash cannot catch a re-paraphrase and a prompt cannot catch a byte-identical repeat, so plan-18 wants both halves. Two lines in `plugin/modes/code.json` with real evidence behind it (9 batches from one prompt) — cheapest item in the whole cluster, it only needs a rebase.

Issue #2424, opened 2026-05-11
Worker daemon spawns from wrong version after update; MCP server and worker use separate version resolution paths


zer0kt:
## Bug Description

After updating the plugin via `/plugin update` and running `/reload-plugins`, the MCP server correctly resolves to 13.1.0 but the worker daemon spawns from 12.7.5. The two components use separate version resolution code paths that independently walk the cache directory, and they disagree on which version to pick.

## Steps to Reproduce

1. Have claude-mem installed and running (e.g. 12.7.5)
2. Run `/plugin update` from within Claude Code to update to 13.1.0
3. Run `/reload-plugins`
4. Check running processes: `ps ugxaw | grep claude-mem`

## Expected Behavior

Both the MCP server and the worker daemon load from the same version directory (13.1.0).

## Actual Behavior

The MCP server resolves 13.1.0 correctly, but the worker daemon spawns from 12.7.5:

```
[11:36:04.006] Cached SKILL.md at boot {path=~/.claude/plugins/cache/thedotmack/claude-mem/13.1.0/skills/mem-search/SKILL.md}
[11:36:04.420] Port in use, waiting for worker to become healthy
[11:36:04.421] Starting worker daemon {workerScriptPath=~/.claude/plugins/cache/thedotmack/claude-mem/12.7.5/scripts/worker-service.cjs}
```

The `/plugins` UI correctly reports 13.1.0 as the installed version, and the MCP server resolves it correctly, but the worker spawn path uses a separate directory resolution call that lands on the older version. Both version directories exist simultaneously in cache after the update.

Confirmed via `ps`:

```
node ~/.claude/plugins/cache/thedotmack/claude-mem/12.7.5/scripts/mcp-server.cjs
```

## Root Cause (suspected)

The MCP startup script resolves the version directory via `ls -dt` sorted by mtime. After an in-place update, both version directories have identical mtimes, making sort order undefined. More importantly, the worker spawn path appears to use a completely separate resolution call rather than reusing the path already resolved by the MCP server process. These two independent resolutions can and do disagree.

A fix would be either: unify into a single resolution call whose result is passed to the worker spawn, or use semver-aware sorting (`ls -d .../[0-9]*/ | sort -V | tail -1`) so both paths consistently pick the highest version.

## Environment

- claude-mem version: 13.1.0 (installed), 12.7.5 (running worker)
- Claude Code version: 2.1.138
- OS: Linux
- Platform: VS Code extension

## Logs

```
[...]

Issue #499, opened 2025-12-30
Claude Haiku subagent processes never terminate, causing massive memory leak (42GB+ RAM)


yungweng:
## Environment
- **claude-mem version:** 8.5.1
- **OS:** macOS Darwin 25.2.0
- **Platform:** darwin (Apple Silicon)

## Bug Description

The worker-service daemon spawns Claude Haiku subagent processes but never terminates them. Over a ~5.5 hour period, **236 orphaned `claude` processes** accumulated, consuming approximately **42 GB of RAM** (each process uses ~180-330 MB).

## Evidence

All haiku subagents share the same parent - the worker-service daemon:

```bash
$ ps -eo pid,ppid,command | grep "claude.*haiku" | head -3
256 83776 /Users/yonnock/.local/bin/claude --output-format stream-json --model claude-haiku-4-5 ...
1035 83776 /Users/yonnock/.local/bin/claude --output-format stream-json --model claude-haiku-4-5 ...
```

Parent is the worker-service daemon:

```bash
$ ps -p 83776 -o command
/Users/yonnock/.bun/bin/bun /.../worker-service.cjs --daemon
```

236 orphaned processes consuming 42GB:

```bash
$ ps aux | grep "claude.*haiku" | wc -l
237

$ ps aux | grep "claude.*haiku" | awk '{sum+=$6} END {print sum/1024 " MB"}'
43260 MB
```

Before/after killing orphaned processes:

| Metric | Before | After |
|--------|--------|-------|
| RAM Used | 62 GB | 27 GB |
| RAM Free | 721 MB | 36 GB |
| Compressor | 32 GB | 8.5 GB |
| Haiku processes | 236 | 4 |

## Root Cause (suspected)

The worker-service.cjs spawns Claude Haiku subagents via the Claude SDK for memory summarization/processing tasks, but the cleanup handlers are not properly terminating the child processes when the work completes.

Looking at the bundled code, when a process closes:

```javascript
this._process.on("close",c=>{var u;this._process=void 0,(u=this.onclose)===null||u===void 0||u.call(this)})
```

The `this._process = undefined` clears the reference but doesn't actually `kill()` the child process if it's still running.

## Impact

- **Severe:** System runs out of RAM within hours of normal usage
- Affects all users running the claude-mem worker for extended periods
- macOS compressor kicks in (32GB compressed in my case) before free RAM is exhausted
- New worker daemons continue spawning haiku processes that don't terminate

## Workaround

[...]

PR #3354, opened 2026-07-21
fix(worker): resolve cache version by semver, not mtime

jamincollins:
## Problem

`cacheWorkerScriptCandidates()` (src/shared/worker-utils.ts) picks which
`cache/thedotmack/claude-mem/<version>/scripts/worker-service.cjs` to spawn
by sorting version directories on `mtimeMs`. An in-place plugin update
extracts the old and new version directories in the same pass, so they
frequently end up with **identical mtimes** — at that point
`Array.prototype.sort`'s comparator returns `0` for every pair, and the
resulting order is whatever `readdirSync` happened to return, not version
order.

This is exactly the scenario in #2424, which was closed as not planned. I'm
reopening it as a PR because it reproduced live rather than theoretically:

## What I observed

On a machine with both `13.10.2` and `13.11.0` cached (mtimes tied to the
second), the resolver kept picking `13.10.2`. Every worker restart re-ran
the same resolution, so:

1. Worker spawns from the stale `13.10.2` script.
2. `checkVersionMatch()` notices the plugin (`13.11.0`) and running worker
   (`13.10.2`) disagree and requests a recycle.
3. The recycle's own spawn hits the identical tied-mtime resolution and
   picks `13.10.2` again.
4. Repeat — the worker crash-loop-restarted every few seconds.

Each cycle left behind an orphaned chroma-mcp/uvx process pair (the old
worker's `ChromaMcpManager` never got a chance to tear its subprocess down
cleanly before being replaced). Over roughly 10 minutes this produced ~75
orphaned pairs (~150 processes, several GB RSS) before I killed the loop
manually — a variant of the accumulation described in #499/#789/#803/#1090,
but caused by the *resolver*, not by `ChromaMcpManager`'s own cleanup paths.

## Fix

Sort candidates by the version directory **name** itself (semver
descending) rather than by mtime. The directory name *is* the version
identifier, so this is deterministic regardless of extraction order or
mtime ties — no filesystem timestamp involved at all.

## Testing

- Added `tests/shared/worker-utils-cache-version-resolution.test.ts`:
  unit tests for the new comparator (major/minor/patch ordering, equal
  versions, a realistic multi-version directory listing).
- `bun test tests/shared/` — 126 pass, 0 fail (no regressions).
- `tsc --noEmit` clean.

I didn't touch the equivalent mtime-sort in `src/build/hook-shell-template.ts`
(the generated shell/Node launcher embedded in `hooks.json`/`.mcp.json`/
codex hooks) — that path has its own byte-matching generator test and is a
separate, larger surface; happy to follow up there if this approach looks
right to you.

### DIFF
--- a/src/shared/worker-utils.ts
+++ b/src/shared/worker-utils.ts
@@ -216,21 +216,44 @@ function candidateWorkerScriptPath(root: string): string {
   return path.join(pluginRoot, 'scripts', 'worker-service.cjs');
 }
 
+/**
+ * Compare two `<major>.<minor>.<patch>` version strings, descending (newest
+ * first). Non-numeric or missing components compare as 0. Exported for unit
+ * testing (see tests/shared/worker-utils-cache-version-resolution.test.ts).
+ */
+export function compareVersionDescending(a: string, b: string): number {
+  const partsA = a.split('.').map(part => Number.parseInt(part, 10));
+  const partsB = b.split('.').map(part => Number.parseInt(part, 10));
+  const length = Math.max(partsA.length, partsB.length);
+  for (let i = 0; i < length; i += 1) {
+    const numA = Number.isFinite(partsA[i]) ? partsA[i] : 0;
+    const numB = Number.isFinite(partsB[i]) ? partsB[i] : 0;
+    if (numA !== numB) {
+      return numB - numA;
+    }
+  }
+  return 0;
+}
+
 function cacheWorkerScriptCandidates(): string[] {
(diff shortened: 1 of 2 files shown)


thedotmack:
Closing in the PR consolidation pass — superseded: landed as 7d3f1879e (single version oracle, semver not mtime). The analysis here fed the plan-16 design (#3604); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

PR #3364, opened 2026-07-22
fix(chroma): persist watermark gaps across bootstrap and live sync (closes #3360)

### LINKED ISSUE #3360: #2487 recurred on 13.11.0: watermark still leaps unwritten rows (max() bootstrap + live-sync bump over a gap) — 32% of observations unembedded while state claims caught-up
## Summary

**#2487 has recurred on v13.11.0.** 32% of observations (7,643 of 24,003) are absent from Chroma while `chroma-sync-state.json` claims caught-up. #2487's *option 2* (verify-on-write watermark bumps) did land — I can see the gap-latch in `backfillKind`. Its *option 1* (a reconciliation pass that distrusts the watermark) did not, and without it two remaining paths still walk the watermark straight over unwritten rows. One of them I captured live while writing this issue.

Same install as #2487, ~2 months later, after the store was rebuilt to recover from the #2534 segfault class.

## Measured gap (v13.11.0)

Set difference between the Chroma `embeddings` table and `claude-mem.db`, by `sqlite_id`, not by counting metadata rows:

| Kind | In source DB | Embedded | Missing |
|---|---|---|---|
| observations | 24,003 | 16,360 | **7,643 (32%)** |
| session_summaries | 6,959 | 5,612 | **1,347 (19%)** |
| user_prompts | 13,203 | 11,749 | **1,454 (11%)** |

Per project, the missing observations: `bootle-v3` 4,758 · `shared` 2,874 · three others 11. Every affected project's watermark sat at or above the max source id, i.e. every one of them reported "caught up". By age, 7,008 of the 7,643 are from the *current month* — the loss concentrates on recent work, which is the opposite of what a "backfill still catching up" reading would predict, and it is why the hole is easy to miss: recall of old material still works.

## Path 1 — `bootstrapWatermarksFromChroma` takes `max()` over a set with holes

```js
async bootstrapWatermarksFromChroma(e){
  let r = await this.getExistingChromaIds(e),
      n = s => s.size ? Math.max(...s) : 0;
  Ai.replace(e, {observations:n(r.observations), summaries:n(r.summaries), prompts:n(r.prompts)});
}
```

Called from `backfillAllProjects` whenever `!Ai.exists()` — i.e. any time the state file is absent: fresh install against an existing store, or a rebuild where the state file was moved aside (which is what the docs and prior threads advise doing).

`getExistingChromaIds` returns the **full ID set**. The bootstrap reduces it to `max()`. Every hole below that max becomes permanently unreachable, because the only reader of the watermark is

```sql
SELECT ... FROM observations WHERE project = ? AND id > ?  -- watermark
```

If the store is sparse at bootstrap time — mid-repopulation, or partially written — the watermark is seeded to "complete" over a store that is not. The set needed to do this correctly is already in the same function, one line up.

## Path 2 — live sync bumps the same watermark the backfill is protecting

`backfillKind` correctly refuses to advance past a failed batch, and latches so later batches in that run don't advance either. But that latch is per-run and in-memory. `syncObservation` bumps the very same persisted watermark on every successful single write:

```js
u === l.length ? Ai.bump(n, "observations", e) : h.warn(..., "watermark bump skipped — partial write", ...)
```

So a batch that fails at 14:00 leaves the watermark behind, and the next live observation to arrive — a much higher id — bumps it clean over the hole. The persisted watermark carries no notion of "there is an unwritten gap below me", so the protection is undone by the next ordinary write.

**Captured live.** While remediating today I rewound `bootle-v3`'s observation watermark to 3846 and restarted the worker. The backfill correctly picked up all 8,362 rows:

```
[CHROMA_SYNC] Backfilling observations {project=bootle-v3, missing=8362, watermark=3846, total=8362}
```

Within 90 seconds the persisted watermark read **26962** — the current max — while only ~250 of the 4,758 genuinely-missing rows had actually been embedded. The backfill was still running; live syncs had jumped the watermark past ~8,000 rows the backfill had not yet reached. Had the worker restarted in that window, the whole gap would have been re-orphaned and the state file would once again have claimed completion.

## Why the symptom is silent

Keyword/FTS search reads the source DB, so it stays complete and correct. Only vector recall is holed. A project-scoped semantic query for something that lives only in an unembedded row returns an empty result with `usedChroma: true` and no fallback (filed separately), so the caller reads "nothing was ever recorded about this" instead of "this is not indexed". There is no divergence signal anywhere in the product: `Smart backfill complete` is logged for every project on every boot regardless.

## Suggested fix

1. **Bootstrap from the ID set, not its max.** `getExistingChromaIds` already returns it; seed the watermark to the highest id with no holes beneath it (or persist the missing-id set and drain it), and never to a bare `max()`.
2. **A count reconciliation at boot** — #2487's option 1, still the safety net that would have caught both paths: compare per-project source counts against embedded counts and re-scan on divergence rather than trusting the watermark. Cheap: two `COUNT(*)`s per project.
3. **Do not let a live-sync bump cross a known gap.** Persist the gap (a low-water mark, or the pending set) so an unrelated later write cannot paper over it.
4. **Say so out loud.** `Smart backfill complete` should carry embedded-vs-source counts, and diverge loudly rather than logging completion at DEBUG-level progress.

Also worth flagging for anyone hitting this: **deleting `chroma-sync-state.json` to force a re-scan makes it worse**, since the missing file is exactly what triggers the `max()` bootstrap and re-derives the same false watermarks. The workaround that does work is rewinding each `(project, kind)` watermark to `min(missing id) - 1` and restarting — with the caveat that the run is not restart-safe for the reason in Path 2, and that re-adding rows that are already present goes through the delete+add duplicate path in #3266.

## Environment

claude-mem v13.11.0, macOS 15 (Apple Silicon), `chroma-mcp==0.2.6` via uvx, persistent local store (551 MB), ~24k observations across 17 projects, several concurrent Claude Code sessions against one shared SQLite. Same install as #2487.

rodboev:
## Summary

`bootstrapWatermarksFromChroma` currently seeds each watermark from `max(existing ids)`, which marks sparse Chroma stores as fully caught up and strands any lower missing ids forever. A later live sync can then bump the same scalar watermark past an older gap even when the backfill path already knows that gap exists.

This persists per-kind pending ids alongside the scalar watermark, seeds those pending ids during bootstrap, and replays them during backfill. Backfill now treats each source row as complete only after every Chroma document for that row lands, so higher live ids can advance the watermark without erasing lower unwritten rows and split rows cannot clear themselves halfway through.

## Why

The bug has two paths through the same state model in `src/services/sync/ChromaSync.ts` and `src/services/sync/ChromaSyncState.ts`.

First, `getExistingChromaIds` already returns the full embedded id set, but `bootstrapWatermarksFromChroma` reduced that set to `max()`. Any hole below that max became unreachable because the normal backfill query only reads `id > watermark`.

Second, `backfillKind` only tracked a gap in memory for the current run. After any failed or partial batch, the next successful live write in `syncObservation`, `syncSummary`, or `syncUserPrompt` could move the persisted watermark past the gap. Persisting a pending-id set fixes both paths: bootstrap can remember holes below the max, later live writes can advance the high-water mark without orphaning those pending rows, and the row-atomic backfill loop keeps a multi-document row pending until every batch for that row succeeds.

## Scope

This fixes the watermark-gap model for observations, summaries, and prompts across bootstrap, backfill, and live sync.

It does not add count-based reconciliation, and it does not change the duplicate-write behavior for rows that are already present in Chroma.

## Risk

The on-disk state now includes optional `pending` arrays under each project's watermarks. Older state files still normalize to the legacy all-scalar shape, and backfill clears stale pending ids when the source row no longer exists.

The backfill path may revisit pending ids until a successful row clears them, which is intentional. The tradeoff is extra replay work instead of silently orphaning gaps after a restart, a later live write, or a split-row interruption between Chroma batches.

## Verification / Test plan

- [x] `npm run typecheck` — clean
- [x] `bun test tests/services/sync/chroma-sync-watermarks.test.ts tests/services/sync/chroma-sync-unavailable.test.ts` — 5 passing
- [x] `npm run build` — bundle within guardrail
- `npm run strip-comments:check` currently reports the existing repo baseline: `Changed: 316 (check mode, no writes)`
- `bun test tests/services/sync/` still shows the existing Windows `chroma-mcp-manager-singleton.test.ts` kill-tree failures, reproduced unchanged on the sibling 3351 worktree

Closes #3360


rodboev:
Pushed follow-up fixes on the current head for the TypeScript nullability regression and the split-row backfill case Greptile called out. The PR body now matches the latest verification and row-atomic backfill behavior.

PR #3368, opened 2026-07-23
fix: fail-fast guard for empty assistant responses instead of retry fan-out

Wasabi-221:
## Problem

An empty assistant response (safety block, MAX_TOKENS truncation, gateway hiccup) is mishandled on both observer output paths:

**Worker path** (`OpenAICompatibleProvider` → Gemini/OpenRouter): an empty observation/summary response left the claimed batch in the queue unconditionally. Because `getMessageIterator` resets claimed-but-unconfirmed messages back to pending on every generator start, a model persistently returning empty content re-queried the same batch on every restart — an unbounded retry loop.

**Server path** (BullMQ → `ProviderObservationGenerator`): all three server adapters log a warning and pass empty `rawText` downstream. `parseAgentXml('')` rejects it, so the failure was misattributed as `parse_error` ("parser rejected response") instead of a provider failure. The plain `Error` thrown afterwards then burned all 3 BullMQ retry attempts re-dispatching a job whose outbox row was already terminally `failed` — pure retry fan-out where every attempt was a no-op. The catch block also re-marked the failure a second time under a bogus `unknown` classification.

## Fix

**Worker path** — track consecutive empty responses per session; after 3 in a row, drop the claimed batch loudly instead of leaving it queued. Any non-empty response resets the counter.

**Server path** —
- `processGeneratedResponse` / `processSessionSummaryResponse` detect empty `rawText` **before** parsing and return a distinct `empty_response` outcome without touching the database.
- The generator throws `ServerGenerationTerminalOutcomeError extends UnrecoverableError` (bullmq) for terminal outcomes (`empty_response`, `parse_error`), so BullMQ fails the job immediately instead of consuming its remaining attempts. The error name is pinned to `UnrecoverableError` because BullMQ also detects these by name (`job.js`: `err.name == 'UnrecoverableError'`), which survives duplicated bullmq installs where `instanceof` breaks.
- `process()`'s catch no longer re-marks already-terminal failures, removing the duplicate lifecycle event.

## Tests

- `tests/worker/empty-response-guard.test.ts` — worker-path counter/drop behavior.
- `tests/server/generation/empty-response-guard.test.ts` — runs without Postgres; a poisoned pool proves the empty-response path never touches the DB. Also pins: whitespace-only counts as empty, malformed-but-non-empty text still classifies as `parse_error`, and the terminal error is BullMQ-unrecoverable by both `instanceof` and name.
- `tests/server/generation/provider-observation-generator.test.ts` — PG-gated end-to-end: empty response → exactly 1 provider call, job `failed`, `lastError.classification === 'empty_response'`, rejection named `UnrecoverableError`.

`bun test tests/server/ tests/worker/`: **412 pass, 0 fail** (13 PG-gated skips locally). `tsc --noEmit` clean.

### DIFF
--- a/src/server/generation/ProviderObservationGenerator.ts
+++ b/src/server/generation/ProviderObservationGenerator.ts
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import type { Job } from 'bullmq';
+import { UnrecoverableError, type Job } from 'bullmq';
 import { logger } from '../../utils/logger.js';
 import { PostgresAgentEventsRepository } from '../../storage/postgres/agent-events.js';
 import { PostgresObservationGenerationJobRepository } from '../../storage/postgres/generation-jobs.js';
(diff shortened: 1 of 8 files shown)


thedotmack:

> step 1/2 (`empty_request` preserves batch; generator never sends with queueDepth=0); P1: whitespace-only content bypasses guard; server half → plan-24

thedotmack:
Routed to **#3606 (plan-18, Observer Response Pipeline)** — classifying an empty assistant turn as a provider failure instead of leaving a claimed batch to be re-queried on every generator restart is the master's core rule ("classify every output, never confirm a batch on failure"), and this is one of the few PRs that fixes it on both the worker and the server path.

Related but not duplicate, and all of these need one shared classifier rather than four local guards: #3467 names the cause of an empty observer turn, #3460 preserves a batch when transport failure arrives as text, #3635 adds a server-side empty-response retry, #3624 preserves rejected batches. Needs a rebase.

Issue #2313, opened 2026-05-05
Worker spawns new chroma-mcp instance per task on Linux — 20+ accumulate per session


NewAiCoder:
## Summary

On Linux, the worker daemon (`scripts/worker-service.cjs --daemon`) spawns a fresh `chroma-mcp` subprocess for what appears to be every task or query, and never closes the previous one. They accumulate indefinitely. After ~2.5 hours of moderate usage across 2 active Claude Code sessions, my system had **23 concurrent `chroma-mcp` processes** alive, all parented to the single worker daemon and all pointing at the same `~/.claude-mem/chroma` data directory.

Per `docs/architecture-overview.md`, the design is one worker + one persistent `chroma-mcp` accessed over stdio. Reality on my install is many.

## Environment

- **claude-mem version:** 12.6.0
- **OS:** Linux (Ubuntu, kernel 6.8.0)
- **Claude Code:** running 2 active sessions concurrently (under tmux)
- **Install method:** plugin via Claude Code marketplace
- **uv invocation:** `uv tool uvx --python 3.13 chroma-mcp==0.2.6`
- **chroma-mcp:** 0.2.6

## Repro / observed state

```
$ pgrep -fa chroma-mcp | wc -l
23

$ pgrep -fa chroma-mcp | head -6
PID1  uv tool uvx --python 3.13 chroma-mcp==0.2.6 --client-type persistent --data-dir ~/.claude-mem/chroma
PID2  ~/.cache/uv/archive-v0/<hash>/bin/chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma
PID3  uv tool uvx --python 3.13 chroma-mcp==0.2.6 --client-type persistent --data-dir ~/.claude-mem/chroma
PID4  ~/.cache/uv/archive-v0/<hash>/bin/chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma
PID5  uv tool uvx --python 3.13 chroma-mcp==0.2.6 --client-type persistent --data-dir ~/.claude-mem/chroma
PID6  ~/.cache/uv/archive-v0/<hash>/bin/chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma
```

All 11 uvx wrappers shared the same parent — the single worker daemon. Process ages ranged from ~45 minutes to 2h34m.

[...]

PR #3369, opened 2026-07-23
fix(chroma): contain chroma-mcp orphans from ungraceful worker deaths

mic2112:
## The gap

#2313 hardened every **in-process** leak path (reconnect, failed connect, unexpected close, `stop()`), but none of those can run when the **worker itself** dies without cleanup:

- SIGKILL / crash - nothing in the dying process runs at all
- a restart whose graceful shutdown **failed or hit its hard deadline** and proceeded by design (`worker-shutdown.ts`'s documented "proceed on error, never abort the handoff" policy)

`chroma-mcp` does not exit on stdin EOF, so on those paths the `uv -> python` pair re-parents to init/launchd and lives forever.

## Real-world impact (how I found this)

On my machine (macOS, 13.11.0 -> 13.12.0 upgrade window), a version-mismatch recycle storm repeatedly killed and respawned the worker. One day of logs showed **128 "Prewarming chroma-mcp" / 116 "Connecting to chroma-mcp" events and zero "Stopping chroma-mcp" events** - every death was ungraceful. Result: **221 orphaned uv+python pairs totalling ~3 GB RSS**, all parented to launchd, some alive for days:

```
pid=1829  ppid=1  .../bin/uv tool uvx --python 3.13 --with onnxruntime>=1.20 ... chroma-mcp --client-type persistent --data-dir ~/.claude-mem/chroma
pid=1982  ppid=1829  .../Python  (chroma-mcp)
... x221
```

## The fix - two complementary defenses in `ChromaMcpManager`

1. **Exit-time safety net**: a `process.on('exit')` handler that synchronously tree-kills the live chroma child (pid tracked at connect, cleared on dispose/unexpected close). Covers every path where the worker exits under its own power - including the failed-graceful restart path, which today deliberately proceeds without chroma cleanup.

2. **Boot-time orphan reaper**: at connect time (fire-and-forget, once per process), kill chroma-mcp processes **already re-parented to init** that **reference our data dir**. Covers SIGKILL/crash, which no in-process handler can. Selection is deliberately narrow - `ppid == 1` AND command contains our normalized chroma data dir - so live workers' children (ppid = worker) and other installs' processes are never touched. POSIX-only: init re-parenting is a POSIX phenomenon, and ppid matching is unsafe on Windows due to pid reuse (win32 keeps its existing `taskkill /T` handling).

Both reuse the existing `killProcessTree` layering (descendants first, then root). The reaper's selection logic is extracted as a pure static (`findOrphanedChromaRoots`) with unit coverage (orphan uv + orphan python selected; live worker's child, other install's orphan, malformed rows, empty output all rejected).

## Verification

- `bun test tests/services/sync/` - 30 pass (existing #2313 lifecycle tests + 5 new)
- End-to-end on macOS against the built worker:
  - `SIGKILL` the worker -> pair re-parents to `ppid 1` -> next worker boot reaps both within seconds
  - `SIGTERM` (graceful) -> exit handler leaves zero chroma processes
  - Steady state: exactly one properly-parented pair per worker

## Related observation (not addressed here)

The leak *amplifier* on my machine was a recycle storm: with two cached plugin versions present, hook-side plugin-root resolution (`ls -dt` = directory mtime) can keep spawning the older version's worker while the version check demands the newer one, producing an endless kill/respawn loop (135 worker starts in a day). This PR contains the damage (each cycle no longer leaks a pair); resolving versions by semver rather than dir mtime would kill the storm itself - happy to file that separately if useful.

### DIFF
--- a/src/services/sync/ChromaMcpManager.ts
+++ b/src/services/sync/ChromaMcpManager.ts
@@ -220,6 +220,15 @@ export class ChromaMcpManager {
       await this.disposeCurrentSubprocess();
       throw error;
     }
+
+    // Track the child from the moment it exists, not only after the handshake
+    // succeeds: connect() calls transport.start(), which spawns synchronously
+    // before awaiting, so a worker exiting during the pending/timed-out
+    // connection window must still be able to kill the already-running child
+    // (PR #3369 review). Failure paths route through
+    // disposeCurrentSubprocess(), which clears the tracked pid.
+    ChromaMcpManager.liveChildPid = (this.transport as unknown as { _process?: ChildProcess })._process?.pid;
+    ChromaMcpManager.registerExitSafetyNet();
     let timeoutId: ReturnType<typeof setTimeout>;
     const timeoutPromise = new Promise<never>((_, reject) => {
       timeoutId = setTimeout(
@@ -263,16 +272,6 @@ export class ChromaMcpManager {
 
     const currentTransport = this.transport;
     const currentTrackedPid = (this.transport as unknown as { _process?: ChildProcess })._process?.pid;
-
-    // Exit-time safety net: if this process exits while the chroma child is
-    // still alive (graceful shutdown failed or hit its deadline and
-    // worker-shutdown.ts proceeded anyway, or any stray process.exit), the
-    // 'exit' handler below sync-kills the child tree as the final act. This
-    // is the last in-process line of defense; SIGKILL deaths are covered by
-    // reapOrphanedChromaProcesses on the next boot.
-    ChromaMcpManager.liveChildPid = currentTrackedPid;
-    ChromaMcpManager.registerExitSafetyNet();
-
     this.transport.onclose = () => {
       if (this.transport !== currentTransport) {
         logger.debug('CHROMA_MCP', 'Ignoring stale onclose from previous transport');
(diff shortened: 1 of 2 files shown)


mic2112:
Addressed all four review findings in 0c0ff3e - each one was a genuine gap:

1. **PID tracked after connect** - the child (and the prewarm child, whose cold uvx window is even longer) is now tracked at spawn time, immediately after `client.connect()` is invoked, so a worker exiting during the pending/timed-out handshake can still kill the running tree. Failure paths clear via `disposeCurrentSubprocess` as before.
2. **PID cleared before descendant cleanup** - the unexpected-close handler no longer clears the tracked pid; `cleanupUnexpectedCloseSubprocess` clears it in its `finally` after the tree-kill completes, so a worker exit mid-sweep can still finish the job.
3. **Prefix match kills peers** - `findOrphanedChromaRoots` now requires a boundary-anchored `--data-dir <dir>` argument match (whitespace/end-of-command terminated), with a unit case proving a `chroma-backup` sibling is never selected.
4. **Failed enumeration disables reaping** - `orphanReapDone` latches only after a successful `ps` enumeration; transient failures retry on the next connect (in-flight guard prevents overlap).

`bun test tests/services/sync/` - 31 pass (30 existing + the new prefix-boundary case).

thedotmack:

> step 3/5 (exit-time tree-kill + boot reaper); reaper must match by cmdline+data-dir, not `PPID==1`

thedotmack:

PR #3383, opened 2026-07-23
fix(installer): tolerate non-string OpenCode context bodies (refs #3330)

rodboev:
## Summary

OpenCode installation assumes the worker context response body is a string and calls `.trim()` immediately. A non-string value returned by the response's `text()` method can throw at that boundary and be misclassified as an unavailable worker. Validate the body type there so valid text remains unchanged and every non-string body follows the existing placeholder-context path safely.

## Why

`fetchRealContextFromWorker` already owns readiness checks, context retrieval, and blank-body rejection. Its declared result is `string | null`, but the runtime value returned by `text()` was never checked before `.trim()`. Keeping the check there prevents an object from throwing or being stringified into `AGENTS.md` while preserving the installer's established fallback.

## Scope

This is the standalone fetch-boundary guard routed under https://github.com/thedotmack/claude-mem/issues/3611, and it can merge independently once CI is green. The broader OpenCode plugin contract remains with that plan. This change only hardens installer context retrieval and does not alter plugin exports or npx cache refresh behavior.

## Risk

The accepted path becomes stricter at runtime: only primitive strings can be injected as existing context. Unknown objects use the same placeholder path as blank or unavailable context, and no wrapper-specific fields are guessed.

## Verification

- [x] `bun test tests/integration/opencode-installer.test.ts`
- [ ] `npm run build`
- [x] `npm run lint:hook-io`

Reported by @kai1949 in https://github.com/thedotmack/claude-mem/issues/3330

Refs #3330

### DIFF
--- a/src/services/integrations/OpenCodeInstaller.ts
+++ b/src/services/integrations/OpenCodeInstaller.ts
@@ -191,7 +191,7 @@ async function fetchRealContextFromWorker(): Promise<string | null> {
   if (!contextResponse.ok) return null;
 
   const realContext = await contextResponse.text();
-  return realContext && realContext.trim() ? realContext : null;
+  return typeof realContext === 'string' && realContext.trim() ? realContext : null;
 }
 
 function writeOrRemoveCleanedAgentsMd(agentsMdPath: string, trimmedContent: string): void {
--- a/tests/integration/opencode-installer.test.ts
+++ b/tests/integration/opencode-installer.test.ts
@@ -6,9 +6,12 @@ import {
   addOpenCodePluginReference,
   deregisterOpenCodePluginFromConfig,
   getOpenCodeConfigPath,
+  getOpenCodeAgentsMdPath,
+  installOpenCodeIntegration,
   removeOpenCodePluginReference,
   registerOpenCodePluginInConfig,
 } from '../../src/services/integrations/OpenCodeInstaller.js';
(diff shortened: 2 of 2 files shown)


thedotmack:

> OpenCode fetch-boundary type guard; symptom patch, fine to land

rodboev:
Thanks for the routing note. I refreshed this as the standalone OpenCode fetch-boundary guard:

1. The branch is rebased onto current `main` without expanding the two-file scope.
2. The production change remains the primitive-string check in `fetchRealContextFromWorker`; non-string values returned by the response’s `text()` method and blank bodies use the existing placeholder path, while valid text is preserved exactly.
3. The installer coverage covers the non-string, valid-string, and blank-string paths. The PR body now records the plan-23 standalone route and current verification.

thedotmack:
**All-night ship gate: PASS**

1. RCA logged — plan-23 standalone fetch-boundary type guard (refs #3330 / #3611).
2. No second-system — one `typeof` check on the existing installer `text()` path; non-strings use the established placeholder fallback.
3. Narrow / not opinionated — 1 production line + focused tests.

Rebased onto `2714e869` locally (fork branch update blocked: GitHub App `workflows` permission). Focused tests: `bun test tests/integration/opencode-installer.test.ts` — 10 pass. CI on this head was green. Merging independently; no version bump.

Issue #3362, opened 2026-07-22
chroma-mcp is pinned but chromadb floats: an engine bump (1.5.9) segfaulted a persisted store for ~2.5 months with zero signal — pin chromadb + surface child SIGSEGV


lexverro:
## Summary

The chroma-mcp spawn pins `chroma-mcp` exactly and lets its heaviest transitive dependency float:

```
uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 --from chroma-mcp==0.2.6 chroma-mcp \
    --client-type persistent --data-dir ~/.claude-mem/chroma
```

`chromadb` — the package that carries the Rust extension that actually segfaults (#2534, consolidated into #2378) — is unpinned, so an upstream release lands under an existing persisted store without any version change on the claude-mem side. On this install `uvx` resolved **chromadb 1.5.9**, and the resulting `EXC_BAD_ACCESS at 0x0` inside `chromadb_rust_bindings` killed every data operation for roughly two and a half months while the worker retried in a loop and reported nothing. 187 crash reports accumulated in `~/Library/Logs/DiagnosticReports` before anyone thought to look there; nothing in claude-mem's own logs said the engine was dying.

Two separate asks, both small:

**1. Pin `chromadb` alongside `chroma-mcp`.** An exact `chroma-mcp==0.2.6` gives a reproducibility guarantee it cannot actually honour while the engine underneath it floats. `--with chromadb==<known-good>` in the same spawn line would close it, and would let a bad engine release be diagnosed as "we moved" rather than "it started crashing".

**2. Surface the crash.** A child that dies on SIGSEGV is currently indistinguishable from one that is slow: the reconnect path retries and the session continues with semantic search quietly broken. Checking the child's exit signal and logging `chroma-mcp died on SIGSEGV (chromadb <version>) — semantic search is unavailable` at WARN, plus a non-zero crash count in `doctor`, would turn two months of silence into a first-session symptom. Related silence in the same subsystem: #3360 (backfill logs `Smart backfill complete` for a project it has left 32% unembedded).

## Environment

claude-mem v13.11.0, macOS 15 (Apple Silicon), chroma-mcp 0.2.6 via uvx, chromadb 1.5.9 resolved from the uv cache, persistent local store.

thedotmack:
Consolidating into #3610 (plan-22). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

PR #3384, opened 2026-07-23
fix(chroma): pin chromadb in the uvx launcher (refs #3362)

rodboev:
## Summary

The Chroma uvx launcher pinned `chroma-mcp` while allowing its `chromadb` engine to change with registry releases and cache refreshes. The launcher now pins `chromadb==1.0.16`, the dependency floor declared by `chroma-mcp 0.2.6`, alongside the existing ONNX Runtime and protobuf overrides.

## Why

`ChromaMcpManager.buildLauncherPrefix()` feeds both prewarm and the long-running MCP transport. Its shared override list constrained two transitive dependencies but left `chromadb>=1.0.16` open-ended, so the same claude-mem release could resolve the reporter's 1.5.9 engine without any claude-mem version change. Completing the existing override tuple makes both launcher paths deterministic.

## Scope

This change controls dependency resolution for new uvx environments. It does not repair stores already opened by another chromadb version, change reconnect behavior, add crash diagnostics, or address embedding backfill completeness.

This pull request is a plan-22 step-1 bundle candidate and remains open until the bundle lands. The broader full-closure and settings redesign is tracked at https://github.com/thedotmack/claude-mem/issues/3610.

## Risk

The pin applies to local and remote Chroma modes because they share one launcher prefix. Existing Python, ONNX Runtime, protobuf, PATH, certificate, writer-lock, and shell-free spawn behavior remains unchanged.

## Verification

- [x] `bun test tests/services/sync/chroma-mcp-manager-ssl.test.ts`
- [ ] `bun test tests/services/sync/chroma-mcp-manager-ssl.test.ts tests/services/sync/chroma-mcp-manager-singleton.test.ts tests/services/integrations/spawn-contract-windows.test.ts`
- [x] `uvx --refresh --no-cache --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 --with chromadb==1.0.16 --from chroma-mcp==0.2.6 chroma-mcp --help`
- [x] `npm run build`
- [x] `npm run lint:hook-io`

Refs #3362


thedotmack:

> step 1 (exact-pin the whole uvx closure incl. chromadb)

rodboev:
Thanks for routing this into plan-22. The existing `ChromaMcpManager` change contributes the `chromadb` pin to step 1, so I kept this PR focused on that handoff:

1. I rebased the exact `chromadb==1.0.16` override and its local and remote launcher assertions onto current `main`.
2. I regenerated `plugin/scripts/worker-service.cjs` so the installed worker carries the same pin as `src/services/sync/ChromaMcpManager.ts`.
3. I dropped the unrelated historical test-isolation commit and refreshed the PR verification against the rebased head.

The broader all-dependency pins, settings override, sanitized environment, and shared spawn-spec redesign are part of the future plan-22 bundle tracked by https://github.com/thedotmack/claude-mem/issues/3610. I’ll leave https://github.com/thedotmack/claude-mem/pull/3384 open until that bundle lands.

thedotmack:
Routed to **#3610 (plan-22, Chroma Sidecar Contract)** as the dependency-pin step.

Confirmed against main today: `chromadb==` appears nowhere under `src/` (only in `tests/supervisor/orphan-chroma-sweep.test.ts`), so `buildLauncherPrefix()` still leaves `chromadb>=1.0.16` open-ended and the same claude-mem release can still resolve a different engine on a cache refresh. Live defect, no duplicate, and the cleanest single change in the plan-22 set — it is the natural first commit of that bundle. Rebase needed, `ChromaMcpManager` has moved under it.

PR #3385, opened 2026-07-23
fix(context): include prefixed observation concepts in injection (closes #3379)

### LINKED ISSUE #3379: Observations with prefixed concepts are silently dropped from context injection
## Summary

Observations whose `concepts` entries carry a description after the tag (e.g. `"gotcha: WASM libraries in Workers have different URL resolution semantics"` instead of `"gotcha"`) are **silently excluded from context injection**. They are stored, searchable, and visible in `/api/observations` — they just never reach the injected block, with no error, warning, or log line anywhere.

On my vault only 10 of 25 observations were being injected. The 15 missing ones were the most useful of the set (the whole `raw/` layer of a knowledge vault, full of hard-won gotchas). I only noticed because I happened to compare the `Stats: N obs` line against the project's real observation count.

## Root cause

The context query requires an **exact** match against the active mode's `observation_concepts`:

```sql
AND EXISTS (
  SELECT 1 FROM json_each(o.concepts)
  WHERE value IN (...)
)
```

The compression agent doesn't always emit bare tags. In the same database, from the same mode (`code`), both formats coexist:

```
#5  concepts=["how-it-works","pattern","why-it-exists"]                  -> injected
#29 concepts=["gotcha: WASM libraries in Workers have different URL ..."] -> never injected
```

Since `"gotcha: ..."` never equals `"gotcha"`, `EXISTS` fails and the row is dropped. Note that a single valid concept is enough to pass, so partially-malformed rows still make it through — which makes the failure even harder to spot, because it's intermittent rather than total.

This was with `CLAUDE_MEM_MODEL=claude-haiku-4-5`. Switching the compression model to Sonnet appears to reduce (but not structurally prevent) the malformed output.

## Impact

Silent data loss at the exact moment the feature is supposed to pay off. The observation is written, costs tokens to generate, shows up in search — and then quietly never informs a future session. There's no signal that anything is wrong; the injected block just looks shorter than it should.

For a `/learn-codebase` run this is particularly bad: the whole point is front-loading a codebase into memory, and half of it can end up unreachable by injection.

## Repro

1. Have a project with observations.
2. Set one observation's concepts to a prefixed form:
   ```sql
   UPDATE observations SET concepts = '["gotcha: some long description"]' WHERE id = <id>;
   ```
3. `GET /api/context/inject?projects=<project>` — the observation is absent, and `Stats: N obs` silently drops by one.
4. `GET /api/search?query=...` still returns it, confirming it's only the injection path.

## Suggested fixes

Any of these would have saved the debugging, roughly in order of preference:

1. **Normalize on write.** Truncate at the first `:` when persisting `concepts`, so malformed output can't enter the database at all.
2. **Match on the prefix in the query** rather than requiring exact equality — tolerant of whatever the model emits.
3. **At minimum, log it.** A warning when a stored observation matches zero valid concepts would turn a silent failure into a one-line diagnosis. Same for a mismatch between a project's observation count and the number actually injected.

I worked around it locally with a script that truncates at `:` and rewrites only rows where no concept matches (rows with at least one valid concept are left alone, since they already pass the filter). Happy to open a PR for whichever direction you'd prefer.

## Environment

- claude-mem 13.12.0
- Claude Code 2.1.218
- Windows 11, mode `code`
- `CLAUDE_MEM_MODEL=claude-haiku-4-5-20251001` when the malformed rows were generated

rodboev:
## Summary

Context injection requires stored observation concepts to equal the active mode's IDs exactly. Concepts such as `gotcha: description` remain stored and searchable but disappear from injected context. Normalize the concept token inside the injection query so existing and future prefixed values match their canonical active ID.

## Why

`ContextConfigLoader` supplies canonical IDs such as `gotcha`, while `queryObservationsMulti` compares those IDs against raw `json_each(o.concepts)` values. Matching the substring before the first colon repairs historical rows at read time and keeps storage, search, API output, ordering, and project adoption behavior unchanged.

## Scope

This changes only recent-context injection. It does not rewrite stored concepts, alter compression prompts, or broaden search behavior.

## Risk

The query accepts one additional stored form: an exact active concept ID followed by a colon and description. Exact IDs remain valid, unrelated prefixes remain excluded, and project scope still matches `project` or `merged_into_project`.

## Verification

- [x] `bun test tests/context/observation-compiler.test.ts` — 10 pass, 0 fail, 28 expect() calls
- [x] `npm run build` — all build targets compiled successfully; `plugin/scripts/context-generator.cjs` and the live `plugin/scripts/worker-service.cjs` inject bundle both reflect the source query, and unrelated generated bundle diffs were discarded
- [x] `npm run lint:hook-io` — hook-io discipline OK
- [x] `npm run lint:spawn-env` — spawn-env discipline OK
- [ ] `npm run strip-comments:check` — fails with `Changed: 375 (check mode, no writes)` on the existing repository baseline

Changed files: `src/services/context/ObservationCompiler.ts`, `tests/context/observation-compiler.test.ts`, generated `plugin/scripts/context-generator.cjs`, and generated `plugin/scripts/worker-service.cjs`.

Reported with a database reproduction by @maugarciasa in https://github.com/thedotmack/claude-mem/issues/3379

Closes #3379


rodboev:
Superseded by https://github.com/thedotmack/claude-mem/pull/3389 , which merged on July 23, 2026 and closes #3379 from current main.

PR #3389, opened 2026-07-23
fix(concepts): normalize concept tags at write time + v49 backfill (#3379)

### LINKED ISSUE #3379: Observations with prefixed concepts are silently dropped from context injection

thedotmack:
## Defect

The context-injection query (`ObservationCompiler.ts`, `AND EXISTS (SELECT 1 FROM json_each(o.concepts) WHERE value IN (...))`) matches concept tags **exactly**. But the `concept_guidance` prompt in each mode lists its keywords in `keyword: description` format (e.g. `gotcha: traps or edge cases`), inviting the observer model to echo that exact shape into `<concept>` elements. Tags stored as `"gotcha: WASM quirk"` therefore never matched `'gotcha'` and were silently excluded from context injection — the prompt itself invited the malformed data the SQL then dropped.

## Fix (three parts, no query-side tolerance)

1. **Writer normalization — `src/sdk/parser.ts`**: each extracted concept is truncated at the first `:` and trimmed, then the existing `!== finalType` filter applies and empties are dropped. Novel tags remain storable (no mode-list filtering), matching current behavior.
2. **Producer prompt rule — `plugin/modes/code.json`, `email-investigation.json`, `law-study.json`, `meme-tokens.json`**: `prompts.concept_guidance` gains an explicit OUTPUT RULE that each `<concept>` element must contain ONLY the bare keyword — never a description, sentence, or colon. Keyword lists unchanged.
3. **Backfill — migration v49 in `src/services/sqlite/SessionStore.ts`**: version-gated one-shot `UPDATE` rewriting each `concepts` array element via `json_group_array`, truncating at the first `:`; recorded in `schema_versions`, idempotent, wired at the end of the constructor migration chain. The FTS5 `AFTER UPDATE` trigger on observations auto-syncs the backfill in existing DBs; fresh DBs have no rows to rewrite.

The injection query is intentionally **unchanged** — no `LIKE`/prefix tolerance that would accept malformed data forever.

## Tests

- Parser: prefixed concept truncates to bare keyword; bare concept unchanged; concept equal to observation type still dropped (bare and prefixed); empty-after-truncation dropped.
- Compiler: a directly-seeded `'["gotcha: x"]'` row is excluded by `queryObservationsMulti` (documents the exact-match defect).
- Migration: v49 rewrites `["how-it-works","gotcha: x"]` → `["how-it-works","gotcha"]`, leaves colon-free rows untouched byte-for-byte, is idempotent on a second construction, and the healed row is returned by the injection query afterward.
- `bun test tests/sdk/ tests/context/ tests/sqlite/`: 212 pass, 0 fail. `npx tsc --noEmit` clean.

## Known caveat (out of scope)

Inbound cloud sync (`SyncApply.ts` / `CloudSync.ts`) can re-import malformed concepts from other clients that predate this fix; per the plan this is noted but not solved here.

Closes #3379

https://claude.ai/code/session_01RSNZmfi4vSYjTeEWtwqKrW


thedotmack:
Both P1 findings addressed in 77154ef9b: (1) v49 now requeues corrected NATIVE rows for sync — sync_rev bumped via incrementCanonicalDecimal and synced_at re-nulled, mirroring requeuePromptSync; replica rows are normalized locally only, their repair travels from their origin device. (2) The migration predicate now includes json_valid(concepts), so a non-JSON concepts value can no longer abort the constructor migration chain; invalid rows are equally unreadable before/after for every json_each reader, so skipping them changes no behavior. New migration test covers all four scenarios (invalid-JSON survival, native requeue, replica no-requeue, colon-free untouched).

PR #3393, opened 2026-07-24
fix(chroma): surface fatal child exits in doctor (refs #3362)

rodboev:
## Summary

Unexpected exits from the active Chroma MCP child were reduced to a generic reconnect warning, so the exit signal and repeated crash state were absent from claude-mem diagnostics. The Chroma lifecycle owner now records a worker-lifetime fatal-exit count and last process outcome, logs the exact detail, and exposes it through the existing worker doctor endpoint and `npx claude-mem doctor`.

The owner routed this fix to https://github.com/thedotmack/claude-mem/issues/3610 as the plan-22 bundle candidate for child exit code and signal capture plus the doctor crash counter. This PR stays open as the source and attribution record until that bundle lands.

## Why

`ChromaMcpManager` already distinguishes active, stale, and intentionally closed transports before applying reconnect backoff. It now captures the active child process before the MCP SDK clears its transport handle, then records `exitCode` and `signalCode` only for the current unexpected-close path. The worker diagnostics endpoint transports that snapshot, and the CLI renders a concrete warning instead of trying to reconstruct lifecycle state.

## Scope

The count lasts for the current worker process. This step does not alter the inherited launcher pin, persist crash history, ingest macOS crash reports, change reconnect or cleanup behavior, repair existing stores, or address embedding completeness. The broader plan-22 health-down state, backlog, immediate-death classification, and circuit breaker remain bundle work.

## Risk

Intentional stop, stale transport callbacks, prewarm failures, and successful reconnects keep their current lifecycle behavior. The new doctor row is warn-only; it does not change the command's required-check exit status.

## Verification

- [ ] Focused Chroma, endpoint, doctor, worker-spawner, and session-manager tests; 12 pass, while endpoint and Chroma files are blocked locally by missing `express` and MCP SDK packages in this worktree
- [ ] `npm run build`; blocked locally because `esbuild` is missing from this worktree
- [x] `npm run lint:hook-io`
- [ ] `npm run strip-comments:check`; blocked locally because `typescript` is missing from this worktree

Refs #3362

### DIFF
--- a/src/npx-cli/commands/doctor.ts
+++ b/src/npx-cli/commands/doctor.ts
@@ -23,6 +23,17 @@ interface CheckResult {
   required: boolean;
 }
 
+interface ChromaCrashState {
+  count: number;
+  lastExit: {
+    timestamp: string;
+    code: number | null;
+    signal: string | null;
+  } | null;
+  chromaMcpVersion: string;
+  dependencyOverrides: string[];
+}
+
 function probeVersion(bin: 'bun' | 'uv'): string | null {
   try {
     return bin === 'bun' ? getBunVersion() : getUvVersion();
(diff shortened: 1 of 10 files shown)


thedotmack:

> step 2 (child exit code/signal captured, crash counter in doctor)

rodboev:
Thanks for routing this into plan-22. I left the PR open as the source and attribution record for the bundle and kept its scope to the current worker-lifetime exit code and signal capture plus the warn-only doctor crash counter.

I limited the body correction to two points:

1. I replaced the old stacked-merge sequencing with the step-2 bundle route under https://github.com/thedotmack/claude-mem/issues/3610, covering the existing `ChromaMcpManager` snapshot through `worker-service.ts`, `Server.ts`, and `doctor.ts`.
2. I kept the broader health-down state, backlog, immediate-death classification, and circuit breaker with the plan-22 bundle rather than expanding this candidate.

The implementation head stays unchanged.

thedotmack:
Routed to **#3610 (plan-22, Chroma Sidecar Contract)**, matching the note already in the body — capturing the child's `exitCode` / `signalCode` and a worker-lifetime fatal-exit count is the "honest fallback, loud degradation" half of that master. It brushes #3602 (plan-14) too, since the thing dying is a spawned child, but the diagnostics surface makes plan-22 the right home.

Not a duplicate of #3384: that one pins the launcher's dependency closure, this one reports when the launcher's child dies. Both belong in the same bundle. Conflicting against main and wants a rebase first.

PR #3399, opened 2026-07-24
fix(search): fall back after empty semantic misses (closes #3361)

### LINKED ISSUE #3361: Empty semantic result returns confident-empty instead of falling back to SQLite: the fallback is gated on platformSource, not on emptiness
## Summary

`SearchOrchestrator.executeWithFallback` falls back to the SQLite/FTS strategy when the semantic strategy returns nothing — but only if `platformSource` happens to be set. A `project`-scoped search (or any search without `platformSource`) that returns zero from Chroma is handed back as an authoritative empty result, with no fallback and no signal that the vector index was the thing that came up empty.

The fallback is gated on which *filter* the caller used, not on whether the result was empty.

## Where — `scripts/worker-service.cjs`, v13.11.0

```js
let r = await this.chromaStrategy.search(e);
return e.platformSource && this.isEmptyResult(r)
  ? (h.debug("SEARCH","Orchestrator: platform-scoped Chroma search returned zero matches; falling back to SQLite",{}),
     await this.sqliteStrategy.search(e))
  : r;
```

`isEmptyResult(r)` is already the correct and complete predicate — observations, sessions and prompts all empty. The `e.platformSource &&` conjunct is what restricts the recovery to one filter shape. The debug string ("platform-scoped … falling back") reads as though the narrow scope were deliberate, so it is easy to walk past.

Note also that `executeChromaSearch` returns `usedChroma: true` on the zero-ids path, so the empty result is labelled as a successful semantic search rather than a miss.

## Why it matters

FTS and the vector index are not interchangeable views of the same data — the vector index is derived and can be incomplete. On this install 32% of observations are in SQLite but not in Chroma (filed separately as the #2487 recurrence). For every one of those rows, a `project`-scoped semantic search returns "no results" for material that is present, indexed by FTS, and one strategy away. The caller — an agent, in practice — reads that as "this was never recorded" and acts on it. A wrong-but-confident empty is considerably worse than a slow query, because nothing about the response invites a second look.

Any incompleteness in the vector store turns into silent, unfalsifiable recall loss at exactly the moment it matters. Our internal working rule right now is "verify any filtered miss with an unfiltered search," which is a convention standing in for a fallback the orchestrator could just do.

## Suggested fix

Drop the `platformSource` conjunct:

```js
return this.isEmptyResult(r) ? await this.sqliteStrategy.search(e) : r;
```

Optionally surface which strategy actually produced the returned rows (the `strategy` field already exists) so a caller can distinguish "semantic found nothing, keyword did" from "genuinely nothing".

## Environment

claude-mem v13.11.0, macOS 15 (Apple Silicon), `chroma-mcp==0.2.6` via uvx, persistent local store, ~24k observations across 17 projects.

rodboev:
## Summary
A zero-hit Chroma search could return an authoritative empty result when the caller omitted `platformSource`, even when the same query still had SQLite matches. This changes the fallback to key on emptiness instead of the filter shape, so empty semantic misses fall through to SQLite for the same search.

## Why
`SearchOrchestrator.executeWithFallback()` only retried SQLite when `options.platformSource && isEmptyResult(chromaResult)` was true. That made project-scoped and other no-`platformSource` searches treat an empty Chroma response as final. The existing test suite already covered the platform-scoped branch; this extends it to the unscoped zero-hit case and preserves non-empty Chroma results.

## Verification
- [x] `bun test tests/worker/search/search-orchestrator.test.ts`

Closes #3361

### DIFF
--- a/src/services/worker/search/SearchOrchestrator.ts
+++ b/src/services/worker/search/SearchOrchestrator.ts
@@ -58,8 +58,8 @@ export class SearchOrchestrator {
       logger.debug('SEARCH', 'Orchestrator: Using Chroma semantic search', {});
       try {
         const chromaResult = await this.chromaStrategy.search(options);
-        if (options.platformSource && this.isEmptyResult(chromaResult)) {
-          logger.debug('SEARCH', 'Orchestrator: platform-scoped Chroma search returned zero matches; falling back to SQLite', {});
+        if (this.isEmptyResult(chromaResult)) {
+          logger.debug('SEARCH', 'Orchestrator: Chroma search returned zero matches; falling back to SQLite', {});
           return await this.sqliteStrategy.search(options);
         }
         return chromaResult;
--- a/tests/worker/search/search-orchestrator.test.ts
+++ b/tests/worker/search/search-orchestrator.test.ts
@@ -20,7 +20,7 @@ const observation = {
   created_at_epoch: 1735689600000,
 };
 
-describe('SearchOrchestrator platform-scoped Chroma zero fallback', () => {
+describe('SearchOrchestrator Chroma zero fallback', () => {
   it('normalizes date_from/date_to filters into dateRange for SQLite search', async () => {
     const searchObservations = mock(() => [observation]);
     const orchestrator = new SearchOrchestrator(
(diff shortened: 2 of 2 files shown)


thedotmack:

> step 5 (fallback keyed on emptiness alone) — 2-line fix, green

thedotmack:
**All-night ship gate: PASS**

1. RCA logged — fallback was keyed on `platformSource` so unscoped empty Chroma misses stayed final (closes #3361 / #3610 step 5).
2. No second-system — same orchestrator fallback, condition keyed on emptiness alone.
3. Narrow / not opinionated — 2 production lines + tests.

Rebased onto `2714e869` locally (fork branch update blocked: GitHub App `workflows` permission). Focused tests: `bun test tests/worker/search/search-orchestrator.test.ts` — 4 pass. CI on this head was green. Merging independently; no version bump.

Issue #3392, opened 2026-07-23
CI: 'typecheck · build · test · bundle-size' job flakes with mass HealthMonitor/ProcessManager mock failures (order-dependent), plus CORS port-collision flake


thedotmack:
During the v13.12.4 release round the `typecheck · build · test · bundle-size` job failed nondeterministically on 4 of ~8 runs, on both main pushes and PR runs, with two distinct signatures — neither reproducible locally (the same suites pass 411/411 in isolation on the same commits).

**Signature A — mass mock failures (seen 3x: main run for fe1e02b40, main runs for the 13.12.3 bump/docs commits, PR #3389 attempts 2 and 4):** every `HealthMonitor` test (isPortInUse / waitForHealth / waitForPortFree) plus `ProcessManager > getPlatformTimeout` fails instantly (0.1–0.4 ms) with `expect(...).toBe` / `toHaveBeenCalled` mismatches — the mocks are simply not in effect. Identical code passes on re-run with zero changes. This is bun `mock.module` pollution that depends on test-file execution order: some other file in the full-suite run either clobbers or restores the modules these tests mock (`node:net`/`node:http`/platform), so whether they fail depends on which files ran before them.

**Signature B — CORS port collision (seen 1x: PR #3389 attempt 1):** `tests/worker/middleware/cors-restriction.test.ts` — one case's throwaway HTTP server fails to bind (`ConnectionRefused` on the fetch), then `afterEach`'s `server.close` throws raw `ERR_SERVER_NOT_RUNNING`. Other cases in the same file pass. Looks like a random-port collision between concurrently-listening test servers.

**Impact:** main shows red on good commits; PRs need repeated `gh run rerun --failed` roulette (PR #3389 needed 3 attempts twice). 

**Suggested direction:** for A, find the offending cross-file mock leak (run the job with bun's file list bisected, or audit `mock.module` users for missing restores) rather than serializing or retrying; for B, bind the test server to port 0 and read the assigned port from the handle instead of picking a random port. Evidence runs: 30044280601 (main, red), 30046958953 (PR #3389, red→red→green with no code change), 30048125985 (same, red→green).

thedotmack:
Consolidating into #3604 (plan-16). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

PR #3400, opened 2026-07-24
fix(worker): use collision-safe ports in CORS restriction tests

rodboev:
## Summary
The CORS restriction test suite picked a random fixed port for each server startup, so occasional collisions left no listener bound and turned the fetch assertion into a `ConnectionRefused` failure. This switches the suite to OS-assigned ports, captures the assigned port after bind, and only closes servers that actually started.

## Why
The previous setup used `41000 + Math.floor(Math.random() * 10000)` and assumed the listen call would always succeed. When another process won that port first, the test still fetched the guessed port and teardown hit `ERR_SERVER_NOT_RUNNING`. Binding on port `0` lets the OS choose an available listener and keeps setup and cleanup aligned with the server's real state.

## Scope
This PR only fixes the CORS port-collision slice from #3392. The separate order-dependent mock leakage in the same issue stays outside this diff.

## Verification
- [x] `bun test tests/worker/middleware/cors-restriction.test.ts`
- [x] Manual: repeated the focused test locally 25 times without a collision failure

Refs #3392

### DIFF
--- a/tests/worker/middleware/cors-restriction.test.ts
+++ b/tests/worker/middleware/cors-restriction.test.ts
@@ -68,14 +68,24 @@ describe('CORS Restriction', () => {
         res.json({ ok: true });
       });
 
-      testPort = 41000 + Math.floor(Math.random() * 10000);
-      await new Promise<void>((resolve) => {
-        server = app.listen(testPort, '127.0.0.1', resolve);
+      await new Promise<void>((resolve, reject) => {
+        const onError = (error: Error) => reject(error);
+        server = app.listen(0, '127.0.0.1', () => {
+          server.removeListener('error', onError);
+          const address = server.address();
+          if (!address || typeof address === 'string') {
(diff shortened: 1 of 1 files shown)


thedotmack:

> step 6 (flaky port test isolated: bind on port 0)

thedotmack:
Re-routing this one: the earlier pass filed it under #3604 (plan-16), but the root cause is a test picking `41000 + random()` and assuming the bind succeeded — that is the same "no verified answer about who owns this port" problem as **#3603 (plan-15, Worker Port & Liveness Authority)**, so it belongs there.

Binding port 0 and reading back the assigned port is the right shape, and I confirmed main still carries the random-port line at `tests/worker/middleware/cors-restriction.test.ts:71`. Test-only, 14 lines, still MERGEABLE / CLEAN, no duplicate — merge-now candidate.

Issue #3303, opened 2026-07-17
SessionStart/UserPromptSubmit hooks synchronous (async=false, timeout=60) through 13.11.0; version resolver picks by mtime not semver


LuigiNicaPRO:
## Summary

Two related issues found while auditing repeated hook-timeout warnings from claude-mem:

1. `SessionStart` and `UserPromptSubmit` hooks are declared synchronous (`async` key absent/false, `timeout: 60`) in every shipped version through 13.11.0, so they block the corresponding lifecycle event for up to 60s each.
2. The plugin's own version-resolution logic (embedded directly in `hooks/hooks.json`) picks the plugin root by `ls -dt <cache>/claude-mem/[0-9]*/ | head -1` (newest by directory **mtime**), not by semantic version. If an older version's cache directory is re-extracted/touched after a newer one is installed, the resolver silently pins the older (stale) version instead of the newest release.

## 1. Synchronous SessionStart / UserPromptSubmit hooks

In `hooks/hooks.json` (checked in versions 13.3.0, 13.9.1, 13.10.1, 13.10.2, 13.11.0 — all cached versions available in a local install), the async field per event is:

| Event | async | timeout |
|---|---|---|
| `SessionStart` (both hooks: worker-service `start` + `session-init` `context`) | **false / absent** | 60 |
| `UserPromptSubmit` (`session-init`) | **false / absent** | 60 |
| `PreToolUse` (Read matcher, `file-context`) | true | 60 |
| `PostToolUse` (`observation`) | true | 120 |
| `Stop` (`summarize`) | true | 120 |
| `Setup` (`version-check.js`) | n/a (setup hook) | 300 |

Only `PreToolUse`, `PostToolUse`, and `Stop` were made async in 13.11.0. `SessionStart` and `UserPromptSubmit` remain synchronous with a 60s timeout in every version we inspected, including the latest (13.11.0). Because these fire on every session start and every prompt submit respectively, a slow `bun-runner.js`/`worker-service.cjs` invocation (cold start, IPC hiccup, etc.) blocks the whole event for up to 60s. This produced repeated `SessionStart` and `UserPromptSubmit` hook-timeout warnings in normal usage.

**Ask:** consider making `SessionStart` and `UserPromptSubmit` hooks `async: true` (or otherwise non-blocking) the same way `PreToolUse`/`PostToolUse`/`Stop` were converted in 13.11.0, since they gate interactive turns directly.

[...]

PR #3401, opened 2026-07-24
fix(hooks): keep UserPromptSubmit synchronous while making SessionStart hooks async (refs #3303)

rodboev:
## Summary

The two Claude Code `SessionStart` hooks now run asynchronously, while `UserPromptSubmit` `session-init` remains synchronous so `/api/sessions/init` can persist the `user_prompts` row before downstream observation and summary hooks read it.

## Why

Startup worker launch and context loading don't need to hold the session-start lifecycle open. Prompt initialization does: `session-init` creates or resolves the session and writes the prompt row that downstream privacy and observation paths use. The manifest keeps those scheduling contracts separate.

## Scope

This PR covers the hook scheduling half of #3303. The version-resolver behavior remains unchanged.

## Risk

Startup context may arrive after the session-start event completes. `UserPromptSubmit` keeps its existing synchronous command and 60-second timeout, preserving prompt-row ordering for downstream hooks.

## Verification

- [x] `bun test tests/infrastructure/plugin-distribution.test.ts --test-name-pattern "runs SessionStart asynchronously and UserPromptSubmit synchronously"`
- [x] `npm run lint:hook-io`
- [x] `npm run build`

Refs #3303


thedotmack:

> step 3: SessionStart async yes, but UserPromptSubmit session-init stays synchronous (downstream hooks depend on the row)

rodboev:
Thanks, that ordering dependency makes sense. I’ll narrow the PR to the SessionStart side of the fix and align it with plan-17 step 3:

1. Keep `async: true` on both `SessionStart` entries in `plugin/hooks/hooks.json`.
2. Remove `async` from the `UserPromptSubmit` `session-init` entry, so `/api/sessions/init` finishes writing the `user_prompts` row before downstream observation and summary hooks can use it.
3. Update the focused distribution assertion and the PR title and body to describe that split contract.

The hook commands, timeouts, session-init implementation, and version-resolver scope will stay unchanged.

thedotmack:
Routed to **#3605 (plan-17, Hook Wrapper Contract)** — which hook events may run async and which must stay synchronous to preserve the `user_prompts` write ordering is part of the wrapper contract that master settles.

No duplicate in the queue; this is the only PR touching the SessionStart / UserPromptSubmit scheduling split. It is a two-file manifest change that now conflicts, so it needs a regenerate-and-rebase, and it should be sequenced after whichever exec-form rewrite plan-17 picks (#3403 / #3564) since both rewrite the same manifest.

PR #3402, opened 2026-07-24
fix(opencode): honor persisted worker port settings (closes #3365)

### LINKED ISSUE #3365: OpenCode plugin ignores CLAUDE_MEM_WORKER_PORT from settings.json, breaks capture
## Problem Statement

The OpenCode plugin cannot reach the claude-mem worker whenever `~/.claude-mem/settings.json` sets a non-default `CLAUDE_MEM_WORKER_PORT`. Every capture write (`/api/sessions/init`, `/api/sessions/observations`, `/api/sessions/summarize`) fails against an otherwise perfectly healthy worker.

Root cause: `resolveWorkerPort()` (`src/integrations/opencode-plugin/index.ts:93-103`) calls `SettingsDefaultsManager.get("CLAUDE_MEM_WORKER_PORT")`. That getter is implemented as `process.env[key] ?? this.DEFAULTS[key]` (`src/shared/SettingsDefaultsManager.ts:192-194`) — it **never reads `settings.json`** and falls back to the compiled UID-derived default `37700 + (uid % 100)` (`SettingsDefaultsManager.ts:102`).

The worker itself resolves its port through a different code path — `loadFromFile()`, reached via `src/shared/worker-utils.ts:132` — which **does** read `settings.json`. Because the plugin and the worker use two different resolution paths, they disagree for any non-default port.

On the reporter's machine (`uid=501`), the plugin computed port `37701` while `settings.json` pinned the worker to `37777`. Every POST failed with `Unable to connect`, even though `curl http://127.0.0.1:37777/api/health` returned 200 and Claude Code's own capture worked normally.

This is a distinct defect from:
- #3328 / #3330 (non-function exports break plugin load)
- #2986 / #2832 / #2968 (dead hook names, un-awaited POSTs)

It only becomes visible once the load bug (#3328/#3330) is worked around, and PR #3208 does not touch `resolveWorkerPort()`, so it would still ship broken for anyone with a custom port in `settings.json`.

**Why it's rarely seen:** it only bites users whose `settings.json` holds a non-default port. #3340 shows another user's worker on `37701` — the UID-derived default — where the plugin and worker agree by accident. The reporter's `37777` predates the UID-derived scheme introduced for multi-account isolation (#2101); the older hardcoded-`37777` issues (#1898, #2103) are closed, so anyone who upgraded from that era likely still has `37777` written in their settings file.

## Environment

- **Claude-mem**: 13.11.0
- **Claude Code**: 2.1.217
- **Node.js**: v24.3.0
- **Bun**: 1.3.11
- **OS**: macOS 26.5.2 (arm64)
- **Platform**: darwin
- **Plugin path**: `~/.claude/plugins/marketplaces/thedotmack`
- **Data directory**: `~/.claude-mem`
- **Dev Mode**: No
- **Worker**: running, PID 14557, port 37777, uptime 53 minutes, 3 active sessions
- **Database**: `~/.claude-mem/claude-mem.db` (exists, 124496.00 KB)
- **Settings file**: `~/.claude-mem/settings.json` (exists)
  - `CLAUDE_MEM_MODEL`: claude-sonnet-4-6
  - `CLAUDE_MEM_WORKER_PORT`: 37777
  - `CLAUDE_MEM_WORKER_HOST`: 127.0.0.1
  - `CLAUDE_MEM_LOG_LEVEL`: INFO
  - `CLAUDE_MEM_CONTEXT_OBSERVATIONS`: 50

## Steps to Reproduce

1. Put a `CLAUDE_MEM_WORKER_PORT` in `~/.claude-mem/settings.json` that differs from `37700 + (uid % 100)`. (Reporter used `37777` on `uid 501` — a legacy value predating the UID-derived scheme.)
2. Start the worker and confirm it is healthy on that port: `curl http://127.0.0.1:37777/api/health` returns 200.
3. Install the OpenCode plugin and work around #3328 so OpenCode will actually load it, by trimming the two non-function exports from the bundle.
4. Run any OpenCode session that executes a tool, e.g.:
   ```
   opencode run "Read the file sample.txt and tell me its contents."
   ```
5. Observe all three POSTs (`/api/sessions/init`, `/api/sessions/observations`, `/api/sessions/summarize`) report "Unable to connect", and no rows appear:
   ```sql
   SELECT COUNT(*) FROM sdk_sessions WHERE content_session_id LIKE 'opencode-%';
   -- returns 0
   ```
6. Re-run with `CLAUDE_MEM_WORKER_PORT=37777` exported in the environment — the session row is created successfully.

## Expected Behavior

A `CLAUDE_MEM_WORKER_PORT` set in `~/.claude-mem/settings.json` should be honoured by the OpenCode plugin, so capture works without needing an environment variable — matching how the worker itself resolves its port.

## Actual Behavior

The plugin ignores `settings.json` entirely for this value, silently falls back to the UID-derived default port (`37700 + (uid % 100)`), and every capture POST fails with "Unable to connect" against a healthy worker running on a different, `settings.json`-configured port.

## Relevant Logs

```
Worker POST /api/sessions/init failed: Unable to connect. Is the computer able to access the url?
```

Fetch-wrapper probe (before importing the installed bundle, hooking `tool.execute.after`):
```
http://127.0.0.1:37701/api/sessions/init
http://127.0.0.1:37701/api/sessions/observations
```
(Worker was actually listening on `37777`; confirmed healthy via `curl http://127.0.0.1:37777/api/health` → 200.)

With `CLAUDE_MEM_WORKER_PORT=37777` exported, the identical probe hits `37777`, logs no errors, and the session row is created.

## Suggested Fix

Either:
- `SettingsDefaultsManager.get()` should consult `settings.json`, or
- the plugin should resolve its port through the same `loadFromFile()` path the worker already uses.

The latter looks safer, since `get()` is used elsewhere in hot paths. Concretely, in `resolveWorkerPort()` / `resolveWorkerHost()` (`src/integrations/opencode-plugin/index.ts:93-103`): read the settings file (respecting `CLAUDE_MEM_DATA_DIR`), while keeping `process.env` as the highest-priority override so existing env-based setups are unaffected.

## Additional Context

Unrelated to the main defect, but noted during triage: `npm run bug-report` does not run from the path the bug-report template documents. The shipped marketplace copy at `~/.claude/plugins/marketplaces/thedotmack` has no `@anthropic-ai/claude-agent-sdk` in `node_modules`, even though `package.json` declares `^0.3.172`, so it exits with `Cannot find module @anthropic-ai/claude-agent-sdk`. The reporter had to install that package manually to produce this report and is willing to file it as a separate issue if useful.

rodboev:
## Summary
The OpenCode plugin resolved its worker port from env-or-default settings only, so a persisted `CLAUDE_MEM_WORKER_PORT` in `~/.claude-mem/settings.json` could point the plugin at the wrong worker. This switches the plugin to the shared settings-backed worker-port resolver and adds a regression test for a persisted port override.

## Why
`resolveWorkerPort()` called `SettingsDefaultsManager.get("CLAUDE_MEM_WORKER_PORT")`, which never reads `settings.json`. The worker itself already uses the shared settings-backed resolver, so the plugin and worker could disagree on the port as soon as the user pinned a non-default value on disk.

## Verification
- [x] `bun test tests/integrations/opencode-plugin-contract.test.ts`
- [x] `git diff --check`

Closes #3365

### DIFF
--- a/src/integrations/opencode-plugin/index.ts
+++ b/src/integrations/opencode-plugin/index.ts
@@ -1,4 +1,5 @@
 import { z } from "zod";
+import { join } from "node:path";
 import { SettingsDefaultsManager } from "../../shared/SettingsDefaultsManager.js";
 
 /**
@@ -91,9 +92,11 @@ interface BusEvent {
 }
 
 function resolveWorkerPort(): string {
-  // Canonical resolution: CLAUDE_MEM_WORKER_PORT env override, else the
-  // UID-derived default — identical to the rest of the codebase (#2406).
-  return SettingsDefaultsManager.get("CLAUDE_MEM_WORKER_PORT");
+  const settingsPath = join(
+    SettingsDefaultsManager.get("CLAUDE_MEM_DATA_DIR"),
+    "settings.json",
+  );
+  return SettingsDefaultsManager.loadFromFile(settingsPath).CLAUDE_MEM_WORKER_PORT;
 }
 
 function resolveWorkerHost(): string {
--- a/tests/integrations/opencode-plugin-contract.test.ts
+++ b/tests/integrations/opencode-plugin-contract.test.ts
@@ -1,4 +1,7 @@
 import { describe, it, expect } from "bun:test";
+import { mkdtempSync, readFileSync, writeFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import {
   ClaudeMemPlugin,
   parseSearchResponse,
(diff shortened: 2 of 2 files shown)


thedotmack:

> step 2 (adapters share `resolveWorkerEndpoint()`); OpenCode plugin was reading env-or-default

thedotmack:
**All-night ship gate: PASS**

1. RCA logged — OpenCode plugin used env-or-default port and ignored persisted `settings.json` (closes #3365 / #3611 step 2).
2. No second-system — same `SettingsDefaultsManager.loadFromFile` path the rest of the tree already uses; avoids importing `worker-utils` into the plugin bundle.
3. Narrow / not opinionated — 6 production lines + contract tests.

Rebased onto `2714e869` locally (fork branch update blocked: GitHub App `workflows` permission). Focused tests: `bun test tests/integrations/opencode-plugin-contract.test.ts` — 9 pass. CI on this head was green. Merging independently; no version bump.

PR #3403, opened 2026-07-24
fix(hooks): run Claude hooks as exec-form Node launchers

### LINKED ISSUE #3396: [Windows] 13.11.0 hooks format ("shell": "bash") reintroduces the console-flash-per-hook regression (third shape after #676/#681/#748)
**Version:** 13.11.0, Windows 11, Windows Terminal as default host

## Summary

The 13.11.0 hooks format reintroduces the visible-console-per-hook regression that #676 / #681 / #748 went through - in a new shape, so existing mitigations do not catch it.

`hooks/hooks.json` now declares hooks as **inline bash scripts with a `"shell": "bash"` key**:

```json
{ "type": "command", "command": "_HP=$(printenv PATH ...) ...", "shell": "bash", "timeout": 120, "async": true }
```

Claude Code spawns a visible bash window for each of these. And because 13.11.0 also made PreToolUse/PostToolUse/Stop `async` (great change otherwise - it removed the blocking latency from #3209), the spawns now fire on **every tool call**, so on Win11 + Windows Terminal you get a window flash per tool call in every active session.

Timeline connection: the bun-runner `shell:true` fix from 13.10.2 and the hooks.json fixes after #681 addressed the *previous* shapes; `"shell": "bash"` is a third shape that sidesteps both.

## Repro

1. Win11, Windows Terminal as default terminal, claude-mem 13.11.0 installed.
2. Start a fresh Claude Code session, run any tool a few times.
3. A Terminal window flashes per tool call. (Confirmed with an EnumWindows sampler - the new windows' owning processes are the bash spawns of these hooks.)

## Suggested fix

Same as the maintainer's earlier direction after #681, but the `-WindowStyle Hidden` wrapper is not sufficient on Win11 + Windows Terminal (the console is created visible before it hides - verified by isolation test). The construction that cannot flash is a GUI-subsystem (winexe) launcher that runs the script via Git Bash with `CreateNoWindow` and pipes stdio, preserving the hook contract (including `async`).

## Workaround for affected users (today)

I maintain a small doctor that detects this shape (plus the two earlier ones), wraps the hooks in exactly such a launcher (built locally from readable C# via the csc.exe present on stock Windows), keeps `async`/`timeout` intact, and re-heals after auto-updates: https://github.com/NiceLeader/agent-hooks-doctor

Happy to PR the launcher approach into the plugin's install path if there is interest - the pieces are Apache/MIT licensed.

rodboev:
## Summary

- Generate all seven Claude Code hooks as direct `node -e` exec-form entries from `src/build/hook-shell-template.ts`.
- Preserve plugin-root fallback order, cache version selection, orphan skipping, worker arguments, metadata, and child I/O behavior.
- Make Rule A verify `command`, `args`, and stale `shell` removal, then regenerate `plugin/hooks/hooks.json`.

## Scope

This rework covers the exec-form Claude launcher slice requested by #3396 and the direction shared with #3564. Codex and MCP launchers remain unchanged. Broader per-OS PATH recovery stays in #3605.

## Risk

Exec form requires Claude Code's inherited `PATH` to resolve the outer Node executable. An NVM-only GUI launch without Node on `PATH` can fail before the payload runs; adding a pre-Node shell here would recreate the console process this change removes.

## Validation

- [x] `bun test tests/infrastructure/plugin-distribution.test.ts`, 63 passed, 0 failed, 416 expectations.
- [x] `node scripts/build-hooks.js --write-shell-templates`.
- [x] `npm run build`, with generated bundles restored from the worktree afterward.
- [x] `npm run lint:hook-io`.
- [x] `npm run lint:spawn-env`.
- [x] Codex and MCP generated files remain byte-for-byte unchanged.
- [ ] `npm run strip-comments:check` remains the repository baseline failure, reporting 388 changed files.
- [ ] Windows 11 manual owner proof remains unavailable in this environment.

Refs #3396.


thedotmack:

> edits generated hooks.json; `shell: bash` removal must go through the template and be decided together with #3564 (exec-form hooks)

rodboev:
I reworked this through the canonical hook template. All seven Claude entries now use direct `node -e` exec form, and the generator verifies `command`, `args`, and stale `shell` removal before regenerating `hooks.json`.

The launcher preserves root fallback and worker arguments, uses `process.execPath` with inherited stdio and `windowsHide`, and forwards termination signals. Codex and MCP remain unchanged. This keeps the PR to the exec-form launcher slice; broader per-OS PATH recovery stays in #3605.

thedotmack:
Night-ship gate (Batch B, after #3453): **FAIL — skipped**.

1. RCA: YES — #3396 / PR body.
2. No second-system: **NO** — this replaces the bash NVM/common-bin PATH prelude just merged in #3453 with a parallel exec-form `node -e` launcher for all seven Claude hooks.
3. Narrow: **NO** — full launcher rewrite (generated `command`/`args`, stale `shell` removal), not a thin follow-up on #3453.

Rebase onto current main is CONFLICTING in `plugin/hooks/hooks.json`. Leaving this open for the #3605 plan-17 bundle to pick one launcher shape. Not merging.

thedotmack:
Routed to **#3605 (plan-17, Hook Wrapper Contract)** — exec-form `node -e` launchers instead of `shell: bash` is the Windows console-window and PATH-recovery half of that master.

Flagging the collision you already noted in the body: #3564 makes the same change to the same generator, and the two cannot both land, so plan-17 has to pick one. This branch is the more recently maintained of the pair and it also shares `src/build/hook-shell-template.ts` with #3428 and #3431 — realistically plan-17 wants a single template rewrite that carries all three. Leaving both open pending that call; the outstanding risk here is still the NVM-only GUI launch with no `node` on the inherited PATH.

PR #3407, opened 2026-07-25
fix(skills): timeline-report example SQL references non-existent observations.source_tool (closes #3332)

### LINKED ISSUE #3332: timeline-report skill: example SQL references non-existent observations.source_tool column
**Plugin version:** 13.10.4 (marketplace install, `plugins/cache/thedotmack/claude-mem/13.10.4`)

The Token Economics section of `skills/timeline-report/SKILL.md` tells the analysis agent that the `observations` table has columns `source_tool` and `source_input_summary`, and the "Explicit recall events" example query filters on `source_tool LIKE '%search%' …`.

Verified against a live 13.10.4 database (`PRAGMA table_info(observations)`): neither column exists. The schema instead carries `agent_type`, `generated_by_model`, `relevance_count` (alongside `discovery_tokens`, `created_at_epoch`, etc.).

**Effect:** an agent following the skill instructions either fails on the query or silently reports 0 explicit recall events. In our run the agent noticed the mismatch and flagged the metric as unverifiable, but the skill doc steers less careful runs into fabricating that section.

**Suggestion:** update the column enumeration and the example SQL in the skill to the current schema, and point the recall-event heuristic at whichever field now carries tool/provenance information (perhaps `agent_type` or narrative-text matching only).

JiataiWang:
## Symptom

An agent following the timeline-report skill hits a hard SQL error **today, on main**: the "Explicit recall events" example query fails with

```
Error: in prepare, no such column: source_tool
```

The skill's agent prompt also enumerates `source_tool, source_input_summary` as columns of the `observations` table, so less careful runs silently report 0 recall events or fabricate the Token Economics metric instead of erroring (#3332).

## Root cause

`observations.source_tool` / `source_input_summary` have never existed on main. They were only introduced on the unmerged `bugfix/security-hardening` branch (bb28c74e, 3c43cea7 — neither is an ancestor of main). The skill was authored against that phantom schema.

Current main schema (base `CREATE TABLE` at `src/services/sqlite/SessionStore.ts:952` plus every `ALTER TABLE observations ADD COLUMN` migration) is:

```
id, memory_session_id, project, text, type, created_at, created_at_epoch,
prompt_number, title, subtitle, facts, narrative, concepts, files_read,
files_modified, discovery_tokens, content_hash, generated_by_model,
relevance_count, merged_into_project, agent_type, agent_id, metadata
```

## Fix

3 lines in `plugin/skills/timeline-report/SKILL.md`:

- column enumeration → the actual current schema above
- recall-event heuristic → narrative matching only (no provenance column exists to filter on)
- example SQL → drop the three `source_tool LIKE` predicates, keep the narrative predicates

## Verification

Built a SQLite database from main's exact schema (base + all migrations), ran all six example queries from the skill — all succeed. The pre-fix query errors on that same database with `no such column: source_tool`.

Fixes #3332

### DIFF
--- a/plugin/skills/timeline-report/SKILL.md
+++ b/plugin/skills/timeline-report/SKILL.md
@@ -91,7 +91,7 @@ Deploy an Agent (using the Task tool) with the full timeline and the following a
 ```
 You are a technical historian analyzing a software project's complete development timeline from claude-mem's persistent memory system. The timeline below contains every observation, session boundary, and summary recorded across the project's entire history.
 
-You also have access to the claude-mem SQLite database at ~/.claude-mem/claude-mem.db. Use it to run queries for the Token Economics & Memory ROI section. The database has an "observations" table with columns: id, memory_session_id, project, text, type, title, subtitle, facts, narrative, concepts, files_read, files_modified, prompt_number, discovery_tokens, created_at, created_at_epoch, source_tool, source_input_summary.
+You also have access to the claude-mem SQLite database at ~/.claude-mem/claude-mem.db. Use it to run queries for the Token Economics & Memory ROI section. The database has an "observations" table with columns: id, memory_session_id, project, text, type, title, subtitle, facts, narrative, concepts, files_read, files_modified, prompt_number, discovery_tokens, created_at, created_at_epoch, content_hash, generated_by_model, relevance_count, merged_into_project, agent_type, agent_id, metadata.
 
 Write a comprehensive narrative report titled "Journey Into [PROJECT_NAME]" that covers:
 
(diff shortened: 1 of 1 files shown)


JiataiWang:
Quick follow-up: this remains a three-line, single-file fix, and Greptile’s executable check passed all six example queries. The workflow is currently `action_required`, apparently waiting for first-contributor approval; when convenient, could a maintainer approve the run and review? Happy to retarget if this packaged-skill change belongs on a different release branch.

thedotmack:

> step 6 (docs generated/checked against schema) — timeline-report SKILL.md column fix

thedotmack:
Routed to **#3609 (plan-21, SQLite Schema Evolution & Queue State Integrity)** — a bundled skill querying a phantom schema is the same "no single source of truth for the DDL" problem that master owns.

Verified on main today: `source_tool` is still referenced at `plugin/skills/timeline-report/SKILL.md` lines 94, 120 and 145 and still does not exist on the `observations` table, so the hard SQL error reproduces exactly as described. Three lines, MERGEABLE / CLEAN, no duplicate open. This one should just go in.

PR #3408, opened 2026-07-25
fix: improve Windows multi-window stability with zombie worker recovery (#2996)

quinnmacro:
Rebases #3009 onto v13.12.4 (PR #3009 closed to rebase; reopen refused after force-push). Addresses #2996. 15 Windows-stability fixes: 3s AbortController-bounded health probe, TCP zombie-port reaper + taskkill, 'auto' sentinel in getInt for fail-loud threshold, exact port-field parsing, ESM imports in worker-spawner, live-PID warmup preserved. Conflict resolutions: HealthMonitor kept 3s timeout + upstream formatHostForUrl host; SettingsDefaultsManager kept 'auto' sentinel + upstream parseJsonWithBom. Typechecks clean. Prior 6 Greptile review rounds on #3009.


thedotmack:

> step 2/3; P1s: taskkill can hit unrelated port holder, unbounded waitForHealth, hardcoded 127.0.0.1

thedotmack:
Routed to **#3603 (plan-15, Worker Port & Liveness Authority)**.

Being straight about the shape of it: this is a 15-fix Windows-stability bundle rather than one root cause, and its port half overlaps #3416 (the port-oracle disagreement) and #3309 (listen-socket inheritance at bind). Plan-15 has to land one port authority, not three reapers, so I am not closing any of the three — but they cannot merge independently either. The fastest way to get value out of this branch would be to split the AbortController-bounded health probe, the exact port-field parsing and the `'auto'` sentinel out from the zombie-port reaper, since those three stand on their own. Conflicting against main.

PR #3410, opened 2026-07-25
fix(context): honor CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES on session-start injection (closes #3409)

### LINKED ISSUE #3409: CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES / _CONCEPTS are documented and persisted but never reach the injection query
## Problem Statement

`CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES` and `CLAUDE_MEM_CONTEXT_OBSERVATION_CONCEPTS` are documented settings that do nothing. A user who sets them to narrow what gets injected at session start gets the active mode's full type/concept list injected anyway — no error, no warning, no effect.

The settings are advertised in three places and consumed in none.

## Root Cause

Two independent links in the chain are missing, so the value is lost twice.

**1. The value is dropped at load time.** `SettingsDefaultsManager` has no entry for either key — neither in the `SettingsDefaults` interface nor in `DEFAULTS` (`src/shared/SettingsDefaultsManager.ts`). `loadFromFile` copies only keys that exist in `DEFAULTS`:

```ts
const result: SettingsDefaults = { ...this.DEFAULTS };
for (const key of Object.keys(this.DEFAULTS) as Array<keyof SettingsDefaults>) {
  if (flatSettings[key] !== undefined) {
    result[key] = flatSettings[key];
  }
}
```

So a value written to `~/.claude-mem/settings.json` never survives the load. `applyEnvOverrides` iterates the same key list, so the env var form is dropped too.

**2. Nothing reads it anyway.** `loadContextConfig` (`src/services/context/ContextConfigLoader.ts:11-13`) builds both sets from the mode file alone:

```ts
const mode = ModeManager.getInstance().getActiveMode();
const observationTypes = new Set(mode.observation_types.map(t => t.id));
const observationConcepts = new Set(mode.observation_concepts.map(c => c.id));
```

Those sets are what `ObservationCompiler.queryObservationsMulti` filters on (`src/services/context/ObservationCompiler.ts:66`):

```sql
AND type IN (${typePlaceholders})
```

## Where it is advertised

- `docs/public/configuration.mdx:289` documents both keys in the `settings.json` example:
  ```json
  "CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES": "bugfix,decision,discovery",
  "CLAUDE_MEM_CONTEXT_OBSERVATION_CONCEPTS": "how-it-works,gotcha",
  ```
- `src/services/worker/http/routes/SettingsRoutes.ts:99-100` lists both in the allowed-keys array, so `POST /api/settings` accepts them and writes them to `settings.json`.
- Neither appears anywhere else in `src/`.

## Steps to Reproduce

1. Put a narrowing list in `~/.claude-mem/settings.json`:
   ```json
   { "CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES": "bugfix,decision,discovery" }
   ```
2. Start a session in a project that has observations of other types (`feature`, `change`, `refactor`, ...).
3. Observe the injected context still contains every type in the active mode.

Confirmed directly against the config loader on v13.12.4 (`132b4634`): with the setting present in `settings.json`, `loadContextConfig().observationTypes` returns all 8 `code`-mode types rather than the 3 configured.

## Expected Behavior

A non-empty `CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES` / `CLAUDE_MEM_CONTEXT_OBSERVATION_CONCEPTS` should scope session-start injection to the listed values. Empty (the default) should keep the current behavior — every type/concept in the active mode.

This is the same defect class as #3358 (`CLAUDE_MEM_EXCLUDED_PROJECTS` ignored on the injection path), and the empty-means-mode-wide default follows the `CLAUDE_MEM_EXCLUDED_PROJECTS: ''` precedent already in `DEFAULTS`.

## Environment

- **Claude-mem**: 13.12.4 (`132b4634`)
- **OS**: macOS 26.1 (arm64)
- **Bun**: 1.3.14
- **Runtime**: worker

I have a fix and regression tests ready and will open a PR against this issue.

ChingEnLin:
## Symptom

Setting `CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES` in `~/.claude-mem/settings.json` does nothing. A user who narrows their injected context to `"bugfix,decision,discovery"` still gets every observation type in the active mode injected at session start — no error, no warning. Same for `CLAUDE_MEM_CONTEXT_OBSERVATION_CONCEPTS`.

Both keys are documented in `docs/public/configuration.mdx:289` and accepted by `POST /api/settings` (`SettingsRoutes.ts:99-100`), so they are written to `settings.json` and then silently discarded.

Closes #3409

## Root cause

The value is lost twice, which is why neither the settings-file nor the env-var form works.

**1. Dropped at load time.** Neither key exists in `SettingsDefaultsManager`. `loadFromFile` copies only keys present in `DEFAULTS`:

```ts
const result: SettingsDefaults = { ...this.DEFAULTS };
for (const key of Object.keys(this.DEFAULTS) as Array<keyof SettingsDefaults>) {
  if (flatSettings[key] !== undefined) result[key] = flatSettings[key];
}
```

`applyEnvOverrides` iterates the same key list, so `CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES=...` in the environment is dropped as well.

**2. Never read.** `loadContextConfig` built both sets from the mode file alone, ignoring the settings it had just loaded:

```ts
const observationTypes = new Set(mode.observation_types.map(t => t.id));
const observationConcepts = new Set(mode.observation_concepts.map(c => c.id));
```

Those two sets are exactly what `ObservationCompiler.queryObservationsMulti` filters the injection query on (`AND type IN (...)`, `ObservationCompiler.ts:66`).

## Changes

- `src/shared/SettingsDefaultsManager.ts` — add both keys to the interface and to `DEFAULTS` with `''`. Empty means the active mode's full list, matching the existing `CLAUDE_MEM_EXCLUDED_PROJECTS: ''` precedent.
- `src/services/context/ContextConfigLoader.ts` — parse the comma-separated lists and use them when non-empty; fall back to the mode lists otherwise.
- `tests/context/context-config-observation-filters.test.ts` — regression tests covering both halves of the chain.

Semantics are an override, not an intersection: the configured list is used verbatim. Intersecting with the mode list would silently drop a value the user explicitly asked for, which is the same class of failure this PR fixes.

Default behavior is unchanged — with both settings empty (the default), `loadContextConfig` returns exactly the sets it returned before.

## Verification

```
bun test tests/context/context-config-observation-filters.test.ts   # 5 pass, 0 fail
bun test tests/context/ tests/shared/ tests/context-injection.test.ts
                                                                    # 227 pass, 0 fail, 19 files
npm run typecheck                                                   # exit 0
npm run build                                                       # exit 0
```

On `main` the new suite fails 4 of 5 — `loadFromFile` drops the keys, and `loadContextConfig` returns all 8 `code`-mode types instead of the 3 configured. The one passing test is the empty-settings fallback, which asserts the unchanged behavior.

The tests set the values via `process.env` and clean up in `afterEach` rather than relying on module-load ordering of `paths.DATA_DIR`, so they pass identically alone and inside a full-suite run (no order dependence — refs #3392).

Branched from `main` at `132b4634` (v13.12.4). Build artifacts under `plugin/` are intentionally not included.

### DIFF
--- a/src/services/context/ContextConfigLoader.ts
+++ b/src/services/context/ContextConfigLoader.ts
@@ -4,13 +4,26 @@ import { paths } from '../../shared/paths.js';
 import { ModeManager } from '../domain/ModeManager.js';
 import type { ContextConfig } from './types.js';
 
+function parseCsvSetting(raw: string | undefined): string[] | null {
+  const values = (raw ?? '').split(',').map(v => v.trim()).filter(v => v !== '');
+  return values.length > 0 ? values : null;
+}
+
 export function loadContextConfig(): ContextConfig {
   const settingsPath = paths.settings();
   const settings = SettingsDefaultsManager.loadFromFile(settingsPath);
 
   const mode = ModeManager.getInstance().getActiveMode();
-  const observationTypes = new Set(mode.observation_types.map(t => t.id));
-  const observationConcepts = new Set(mode.observation_concepts.map(c => c.id));
+  // CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES / _CONCEPTS are documented in
+  // configuration.mdx and persisted by SettingsRoutes, but these two sets were
+  // built from the mode file alone — so a user narrowing their injected context
+  // got the mode's full list regardless. Empty keeps the mode-wide default.
+  const observationTypes = new Set(
+    parseCsvSetting(settings.CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES) ?? mode.observation_types.map(t => t.id)
+  );
+  const observationConcepts = new Set(
+    parseCsvSetting(settings.CLAUDE_MEM_CONTEXT_OBSERVATION_CONCEPTS) ?? mode.observation_concepts.map(c => c.id)
+  );
 
   return {
     totalObservationCount: parseInt(settings.CLAUDE_MEM_CONTEXT_OBSERVATIONS, 10),
--- a/src/shared/SettingsDefaultsManager.ts
+++ b/src/shared/SettingsDefaultsManager.ts
@@ -31,6 +31,8 @@ export interface SettingsDefaults {
   CLAUDE_MEM_CONTEXT_SHOW_WORK_TOKENS: string;
   CLAUDE_MEM_CONTEXT_SHOW_SAVINGS_AMOUNT: string;
   CLAUDE_MEM_CONTEXT_SHOW_SAVINGS_PERCENT: string;
+  CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES: string;
+  CLAUDE_MEM_CONTEXT_OBSERVATION_CONCEPTS: string;
   CLAUDE_MEM_CONTEXT_FULL_COUNT: string;
   CLAUDE_MEM_CONTEXT_FULL_FIELD: string;
   CLAUDE_MEM_CONTEXT_SESSION_COUNT: string;
@@ -125,6 +127,8 @@ export class SettingsDefaultsManager {
     CLAUDE_MEM_CONTEXT_SHOW_WORK_TOKENS: 'false',
     CLAUDE_MEM_CONTEXT_SHOW_SAVINGS_AMOUNT: 'false',
     CLAUDE_MEM_CONTEXT_SHOW_SAVINGS_PERCENT: 'true',
+    CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES: '',  // Comma-separated observation types to inject. Empty = every type in the active mode
+    CLAUDE_MEM_CONTEXT_OBSERVATION_CONCEPTS: '',  // Comma-separated observation concepts to inject. Empty = every concept in the active mode
     CLAUDE_MEM_CONTEXT_FULL_COUNT: '0',
     CLAUDE_MEM_CONTEXT_FULL_FIELD: 'narrative',
     CLAUDE_MEM_CONTEXT_SESSION_COUNT: '10',
--- a/tests/context/context-config-observation-filters.test.ts
+++ b/tests/context/context-config-observation-filters.test.ts
@@ -1,0 +1,82 @@
+import { describe, it, expect, beforeAll, afterEach } from 'bun:test';
+import { mkdtempSync, rmSync, writeFileSync } from 'fs';
+import { join } from 'path';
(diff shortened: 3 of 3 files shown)


thedotmack (approved):
Verified locally on macOS (bun 1.3.9):
- `bun test tests/context/context-config-observation-filters.test.ts` — 5/5 pass
- `bun test tests/context/` — 80/80 pass (no regressions in the context suite)
- `tsc --noEmit` clean

Fix matches the root cause in #3409 exactly: both keys now survive `loadFromFile`/`applyEnvOverrides` (added to `SettingsDefaults` + `DEFAULTS` with the empty-string = mode-wide default, same pattern as `CLAUDE_MEM_EXCLUDED_PROJECTS`), and `loadContextConfig` honors them with sensible CSV parsing. Repo CI hasn't run yet (needs first-contributor approval) — once it's green this looks ready to merge.

One non-blocking note: a typo'd type name (e.g. `decsion`) silently narrows injection to fewer rows. Fine for now given the documented contract, but a startup warning for unknown type/concept ids would be a nice follow-up.

thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3608 ([plan-20] Project Identity & Injection Scope)** as **standalone / merge-now**: it is small and self-contained enough to land on its own ahead of the plan-20 bundle once CI is green.

> step 3 (`SettingsDefaults` gains the missing context keys) — maintainer-approved, waiting on CI approval

thedotmack:
Routed to **#3608 (plan-20, Project Identity & Injection Scope)** — a documented context knob that `SettingsDefaultsManager` drops at load and `loadContextConfig` never reads is squarely the "context config that never reaches the query" defect in that master.

I re-verified against main today: `CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES` still appears only in `SettingsRoutes.ts`, so the settings write succeeds and the value is still discarded — nothing has superseded this. It is also one of the few PRs in the entire open queue still MERGEABLE / CLEAN, three files, with regression coverage. Best merge-now candidate in plan-20 and the one #3310 should sequence behind.

PR #3414, opened 2026-07-26
feat(memory): ACT-R reinforcement — rank context by re-confirmation, not just recency

kejwojew:
## What

Ranks injected context by **how often the world re-confirms an observation**, not by recency alone. Strength follows the ACT-R activation curve — each reinforcement contributes `age^-0.5`, summed and log-compressed, so a gotcha you keep hitting over three months outranks something you touched yesterday.

Five phases, all behind tunables that collapse to today's behavior at `ALPHA=0, BETA=0`:

| Phase | What it does |
|---|---|
| 1 — reinforcement | `reinforcement_dates` + `last_reinforced` on observations; every write path seeds them, exact-duplicate writes reinforce instead of silently dropping |
| 2 — ranking | `queryObservationsMulti` pulls a 5× candidate pool and re-ranks by `ln(1 + age_created^-d + α·Σ age_reinf^-d) + β·log1p(relevance_count)` |
| 3 — dedup judge | **Opt-in** (`CLAUDE_MEM_DEDUP_JUDGE_ENABLED`). LLM pass folds *near*-duplicates into the existing row. Off by default — it costs a model call per candidate |
| 4 — surfacing | `relevance_count` counts genuine injections (not human previews or full dumps) and feeds back into ranking |

No backfill. Existing observations start with empty history, keep their current ranking, and accrue strength going forward.

## Migrations

Versions **50** (reinforcement columns) and **51** (`relevance_count`), both `ALTER TABLE ... ADD COLUMN` guarded by `PRAGMA table_info` checks — the version row is bookkeeping, the PRAGMA is the real guard, so they are idempotent on databases that already have the columns.

## One thing I'd like a second opinion on

`main` made the context read path **read-only** and defends it with `tests/context/context-builder-readonly.test.ts`, including a case where another process holds an `EXCLUSIVE` lock. Phase 4's surfacing counter writes on exactly that path.

The resolution here: the read connection stays read-only, and the write-back gets its own short-lived connection with a 250 ms busy timeout and every error swallowed. A missed surfacing count costs a little ranking accuracy; it can never cost an injection. That readonly test still passes.

That is still a write on a path you deliberately made read-only, so if you'd rather route it through the worker (the `observation_record_event` MCP tool from Phase 4 already exists) or drop the write-back entirely, say the word — it is one function.

## Note on the merge

This branch was cut at v13.6.1 and `main` has since restructured the SQLite layer, so a good chunk of the diff history is reconciling with that:

- `migrations/runner.ts` is gone — `SessionStore` is the single migration chain now, and its duplicate copy of the reinforcement migration went with it
- the migrations originally claimed versions 33/34, which collide with unrelated migrations already shipped; renumbered to 50/51
- the standalone `observations/store.ts` `storeObservation()` is gone — Phase 1c seeding lives only in `storeObservations`, singular delegates
- `ObservationCompiler` moved from `SessionStore` to a raw `Database` and dropped the single-project query variants; Phase 2 ranking moved onto `queryObservationsMulti`

Plugin bundles are deliberately **not** included — they get regenerated from merged sources at release.

## Testing

- 57 new tests across 6 files (strength math, ranking, dedup, migration, surfacing, write path)
- Full suite: 2642 pass / 18 skip. The 6 failures are `workers/sync-hub/test/*` failing on `Cannot find package 'cloudflare:test'` — environmental, untouched by this branch
- `tsc --noEmit` clean; `npm run build` produces bundles carrying the feature


thedotmack:

> ACT-R ranking + facts store + Kimi adapter + eval harness in one branch; ~15 P1s

thedotmack:
Routed to **#2785 (plan-12, Provider & Extensibility Roadmap)** — ACT-R reinforcement ranking is a net-new capability rather than one of the plan-14…25 defect clusters, so it is roadmap work, not bundle material.

One practical note so this does not stall forever: the diff is ~29k added lines and carries `plans/`, a vendored `plugin/vendor/chroma-mcp` tree and rebuilt `plugin/` bundles, which makes it unreviewable as a single unit and guarantees permanent conflicts. Phase 1 alone (the `reinforcement_dates` / `last_reinforced` columns plus write-path seeding, no ranking, no judge) would be a reviewable PR against current main and would let the rest land incrementally.

Issue #2659, opened 2026-05-25
Windows: worker daemon spawn fails silently; stale supervisor.json + hook noise blocks Claude Code tools


zyrill:
## Summary

On Windows (Git Bash + Claude Code), the claude-mem worker daemon fails to stay alive after spawn, leaving `supervisor.json` claiming a dead PID and emitting hook noise that gets surfaced by Claude Code as `printf: write error: Permission denied` blocking other tool calls.

This affects every Claude Code session on Windows: PreToolUse/PostToolUse/Stop hooks all attempt to talk to a non-running worker, fail, emit noise, and tool calls downstream get blocked.

## Environment

- **OS:** Windows 11 Enterprise 10.0.26200
- **Shell:** Git Bash + PowerShell
- **Bun:** 1.3.14 at `C:\Users\pk\.bun\bin\bun.exe` (confirmed runnable: `bun --version` → `1.3.14`, exit 0)
- **Node:** v24.13.0
- **Claude Code:** running in VSCode native extension
- **claude-mem:** 13.3.0 (cached at `~/.claude/plugins/cache/thedotmack/claude-mem/13.3.0/`)

### 1. Stale `supervisor.json` claims dead PID alive

```json
{ "processes": { "worker": { "pid": 20068, "type": "worker", "startedAt": "2026-05-25T10:35:15.231Z" } } }
```

`Get-Process bun` returns empty. Port 37777 is not listening (`Test-NetConnection 127.0.0.1 -Port 37777` → False). PID 20068 is dead, but the file isn't cleaned. New sessions trust this file and skip spawning a fresh worker until the cooldown period expires.

### 2. PowerShell `Start-Process` daemon spawn fails silently

From `~/.claude-mem/logs/claude-mem-2026-05-25.log`:

```
[ERROR] [SYSTEM] Failed to spawn worker daemon on Windows {runtimePath=C:\Users\pk\.bun\bin\bun.exe} 
        Command failed: powershell -NoProfile -EncodedCommand <base64>
[ERROR] [SYSTEM] Failed to spawn worker daemon
[WARN ] [SYSTEM] Worker failed to start before hook, handler will proceed gracefully
[WARN ] [SYSTEM] Cannot lazy-spawn worker: Bun runtime not found on PATH
```

Decoded base64 of the failed command:
```
[...]

Issue #3128, opened 2026-07-04
Windows: version-mismatch recycle orphans children (chroma-mcp) that inherit the listening socket — port stays bound under a dead PID, blocking all respawns for hours


Rashere:
## Summary

Sibling of #3031, different mechanism, different fix required. There, the recycled worker survives and holds the port itself. Here, the recycled worker **dies**, but its spawned children — confirmed: the **chroma-mcp sidecar** — **inherit the listening-socket handle**, so the port stays bound while `netstat` attributes it to a **PID that no longer exists**. Every lazy-spawn then fails with `Failed to start server. Is port 37778 in use?` for hours (13+ hours overnight in one incident), and claude-mem is fully down (no observations captured) until the orphans are manually killed or the machine reboots.

Because the port owner is a ghost, #3031's suggested fix (escalate the recycle to a hard kill of the worker) would **not** fix this variant — the worker is already dead. The socket lives on in the children.

## Environment

- Windows 11 Home 10.0.26200
- claude-mem 13.9.3 → 13.10.0 → 13.10.1 (each version bump triggered an incident)
- Bun runtime at `~/.bun/bin/bun.exe`, worker port 37778
- Claude Code 2.1.197 (desktop app), multiple concurrent sessions

## Incident timeline (2026-07-03, from `~/.claude-mem/logs/`)

```
19:24:09  [INFO ] Worker PID file points to a live process, skipping duplicate spawn
19:24:09  [INFO ] Worker became healthy while waiting on live PID
19:24:09  [INFO ] Worker version mismatch — recycling stale worker {pluginVersion=13.10.0, workerVersion=13.9.3}
19:24:24  [WARN ] No successor worker appeared after recycle; falling through to lazy-spawn
19:24:24  [INFO ] Worker not running — lazy-spawning {runtimePath=...bun.exe, scriptPath=...worker-service.cjs}
19:25:07  [WARN ] Worker port did not open after lazy-spawn within the cold-boot wait (~15s)
19:34:30  [ERROR] ✗ Worker failed to start Failed to start server. Is port 37778 in use?
...       (repeats every ~30s: 2,555 occurrences until 08:19 the next morning)
```

Note the third line: the recycle path itself concluded the worker was healthy moments before recycling — after the recycle, the worker process is gone but the port never frees.

## Key diagnostic evidence (captured live during a recurrence on 2026-07-04)

[...]

PR #3416, opened 2026-07-26
fix(windows): recover from an orphaned listen socket after the upgrade recycle

MythGuru:
## What happens

On Windows, the version-mismatch recycle SIGKILLs the stale worker. `SIGKILL` is `TerminateProcess`, which runs no cleanup — and the dead worker's listen socket can survive it in the kernel. The port stays `LISTENING` against the dead PID, connections to it are **refused**, and `bind()` still returns **EADDRINUSE**.

Observed on a 13.12.1 → 13.12.4 upgrade, and still true on that machine hours later:

```
127.0.0.1:37777  LISTENING  PID 28296
PID 28296 IS DEAD (orphaned listener)
bind() -> NOT BINDABLE: AddressAlreadyInUse
```

## Why it never recovers

Two port oracles disagree in exactly that state:

| | method | verdict |
|---|---|---|
| `waitForWorkerPortClosed()` (`worker-utils.ts`) | HTTP connect; any failure ⇒ free | **free** |
| `HealthMonitor.isPortInUse()` | falls through to a real `bind()` | **in use** |

The old comment states the premise directly: *"A rejected connection is the port-free signal."* That holds on POSIX and is false here.

So the hook believes the port was released, lazy-spawns, and the successor can never bind:

```
Worker version mismatch — killing stale worker
Port in use but worker not responding to health checks
Port already in use, refusing to start duplicate
```

Nothing mutates state, so every subsequent hook repeats it identically. The reporting machine reached `{"consecutiveFailures":70}`.

## Why it took the session down

Past the fail-loud threshold, `recordWorkerUnreachable` calls `emitBlockingError`, which `process.exit(2)`s. On `UserPromptSubmit`, exit 2 **blocks the prompt** — so an unreachable memory worker stopped the user's Claude session entirely, and kept stopping it.

This re-opens the failure class of #729 through the path added in #2292. #2659 is the same Windows family.

Ran both oracles against the genuinely wedged port, unmocked:

```
isWorkerPortReleased(37777) : false (correct — wedged)
old HTTP-connect oracle says free : true (this is the bug)
```

## Changes

- **One oracle.** `isPortBindable()` / exported `isWorkerPortReleased()` decide release by binding — the only question that matters before spawning. A connect probe cannot distinguish "free" from "orphaned listener".
- **Escape the wedge.** When the port never becomes bindable, `relocateWorkerPort()` claims a free loopback port and persists it atomically to the settings file `getWorkerSettings()` actually reads, so worker, hooks, MCP server and CLI all agree on the move.
- **Clean up.** The dead worker's `worker.pid` and `supervisor.json` still name the wedged port; both are removed so the next boot cannot re-adopt it.
- **Fail open.** `recordWorkerUnreachable` keeps the #2292 visibility — same message, same bypass channel — but no longer exits 2. Memory injection degrades while the worker is down; the prompt is never blocked.

I deliberately did **not** build the fix on *why* Windows keeps the socket alive (most likely a surviving child inheriting the listening handle — the worker spawns a long-lived `chroma-mcp` child with piped stdio, which forces `bInheritHandles=TRUE`). That is unproven, and whatever holds the socket, claude-mem should detect it correctly and route around it.

## Tests

New `tests/shared/worker-utils-windows-port-wedge.test.ts` (4 tests): unreachable-but-unbindable port, wedge escape with persistence, non-blocking behavior while unavailable, counter reset.

**Before** — reproduces the incident:

```
error: process.exit(2) called — the hook blocked the prompt
      at emitBlockingError (src/shared/hook-io.ts:145:13)
      at recordWorkerUnreachable (src/shared/worker-utils.ts:750:5)
2 fail
```

**After: 4 pass.** No-regression run across the 9 test files touching this code:

| | pass | fail |
|---|---|---|
| baseline (patch stashed) | 95 | 12 |
| with patch | **99** | **12** |

Identical 12 failures both runs (11 Codex-adapter + 1 stderr-discipline), pre-existing in my environment from an incomplete `node_modules` (`Cannot find module '@anthropic-ai/claude-agent-sdk'`). Zero regressions.

`worker-utils-version-recycle.test.ts` gains a `net` seam — port release is now decided by a real bind, so without it the suite binds the developer's actual worker port and blocks for the full release wait.

## Known limits

The orphaned socket is not reclaimed before a reboot; this routes around it rather than freeing it. Each wedged boot costs a one-time ~5s release wait before relocating. Relocation and persistence are verified under test doubles plus a real-OS check of the exact wedged port; I did not spawn a live successor on the reporting machine.

### DIFF
--- a/src/shared/worker-utils.ts
+++ b/src/shared/worker-utils.ts
@@ -1,5 +1,6 @@
 import path from "path";
-import { readFileSync, existsSync, writeFileSync, renameSync, mkdirSync, readdirSync, statSync } from "fs";
+import net from "net";
+import { readFileSync, existsSync, writeFileSync, renameSync, mkdirSync, readdirSync, statSync, rmSync } from "fs";
 import { spawnHidden } from "./spawn.js";
 import { logger } from "../utils/logger.js";
 import { HOOK_TIMEOUTS, getTimeout } from "./hook-constants.js";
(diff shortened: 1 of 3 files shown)


thedotmack:

> step 3 (single bind-based oracle, relocate + persist); P1s: relocation deletes supervisor.json (orphans children); racy port allocation

Konaki987:
**Reproduction + new data point for this bug family on v13.18.0** (cross-refs: #3692, #3128, #3309, #3834, #3603). Filed as a comment here since #3416's trigger — the upgrade recycle — matches exactly; happy to open a separate issue if preferred.

## Environment
- claude-mem-plugin **13.18.0** (bun worker runtime, `CLAUDE_MEM_RUNTIME=worker`, provider claude/subscription)
- Windows 11 Pro 26200, Claude Desktop (MSIX/Store package) + Claude Code CLI 2.1.255
- Relocated `CLAUDE_CONFIG_DIR`; chroma via **uv-managed python**

## Trigger
Claude Desktop **MSIX self-update** terminated the app uncleanly mid-session (servicing kill). Worker + children survived the parent's death.

## Observed deadlock state
- `netstat` showed `127.0.0.1:37777 LISTENING` **owned by a dead PID** (process confirmed gone via `Get-Process`), persisting for 30+ minutes.
- `Get-NetTCPConnection -LocalPort 37777 -State Listen` → `OwningProcess` = the dead PID.
- `worker.pid` never rewritten; `spawn.lock` held.
- Every new launcher looped:
  ```
  [SYSTEM] Another launcher holds the spawn lock — skipping lazy-spawn and waiting for its worker
  [SYSTEM] Worker port did not open after lazy-spawn within the cold-boot wait (~15s)
  ```
- Worker API on 37777: connection accepted then timeout/ECONNRESET (zombie socket, nothing serving). `claude-mem.db-wal` frozen.
- Escalation (the #3834 symptom): UserPromptSubmit began **blocking all user prompts** — "claude-mem worker unreachable for 12/13 consecutive hooks" — turning a memory-plugin outage into an unusable editor.

## Key new finding: the inherited handle lived in chroma's PYTHON process
#3128 identified orphaned **chroma-mcp (node)** children as the socket holders. On 13.18.0 that was **not sufficient**: killing **every bun worker and every node process** matching `claude-mem|chroma`, plus clearing `spawn.lock`/`worker.pid`/`supervisor.json`, did NOT release the port (verified free-port check after each sweep + 30s waits). The port stayed bound to the dead PID.

What finally released it: **manually killing the uv-managed `python.exe` chroma server processes** (plus one bun zombie). The python child holds an inherited duplicate of the worker's listening-socket handle, so the kernel keeps the socket bound (attributed to the dead creator PID) as long as python lives. After the python kill, the next session-start spawned one clean worker that wrote a proper `worker.pid` and served `/health` normally.

So whatever inheritance fix closed #3692/#3128 doesn't cover the **python (chroma server)** branch of the child tree on 13.18.0 — netstat pointing at a dead PID sends operators hunting the wrong processes.

## Suggestions (mostly restating open work, with this data point)
1. **#3309**: mark the listen socket non-inheritable (`SetHandleInformation`/`bInheritHandles=false`) for ALL child spawns — explicitly including the uv/python chroma launch path, not just node children.
2. Spawn-lock reclaim: if the lock/port holder PID is dead, break the lock instead of waiting (~15s loop forever).
3. **#3489/#2892/#3225**: fail OPEN — a memory plugin should never block user prompts on worker unavailability.
4. Health/liveness (#3603): "port is bound" must be validated with "owner PID alive + API responds", not treated as "worker exists".

## Workaround that recovers reliably (Windows)
Kill in order: all `bun.exe` → `node.exe` matching `claude-mem|chroma` → **`python.exe`/`pythonw.exe` matching `chroma|claude-mem`** → clear `spawn.lock`, `worker.pid`, `supervisor.json` → wait for 37777 to free → start any Claude session (SessionStart hook respawns one clean worker).

src/shared/worker-utils.ts
```diff
@@ -1,5 +1,6 @@
 import path from "path";
-import { readFileSync, existsSync, writeFileSync, renameSync, mkdirSync, readdirSync, statSync } from "fs";
```
monikabielak12-lgtm, line 2: https://apple.com/icloud

monikabielak12-lgtm (approved):

thedotmack:
Routed to **#3603 (plan-15, Worker Port & Liveness Authority)** — two port oracles disagreeing about an orphaned Windows listener is the exact "one verified answer to is a worker serving this port" defect that master exists to end, and the table contrasting `waitForWorkerPortClosed()` with `HealthMonitor.isPortInUse()` is the clearest statement of it anywhere in the queue.

Not closing it against #3408 or #3309: those reclaim the port by different routes (taskkill reaper; clearing `HANDLE_FLAG_INHERIT` at bind), so plan-15 has to reconcile the three into one oracle rather than merge them in sequence. I am treating this PR as the reference for what that oracle should assert. Needs a rebase.

PR #3418, opened 2026-07-26
fix(logging): roll log file path onto the current date (closes #3415)

### LINKED ISSUE #3415: Long-running worker daemon logs to its startup-date file forever; GET /api/logs reads today's file — daemon errors invisible for weeks
**Version:** 12.3.9, Linux (Unraid), worker as a long-running shared daemon (`worker-service.cjs --daemon`), many concurrent Claude Code sessions.

### Summary
The worker daemon resolves its log file path once and keeps appending to the file named for its **start date**. A daemon started July 7 was still writing every line — including all `[ERROR]`/`[WARN]` — to `claude-mem-2026-07-07.log` (54 MB) on July 25. Meanwhile the per-day files for later dates only receive lines from short-lived processes (hooks, MCP servers, `worker-service.cjs start` checks).

`GET /api/logs` computes the filename from the **current** date (`new Date().toISOString().split("T")[0]`), so it serves a file the daemon never writes. Net effect: 18 days of generator failures (`Timed out waiting for agent pool slot after 60000ms`, `Generator failed`, `Marked messages as failed`) were invisible to anyone tailing "today's" log or using the API/viewer — the pipeline looked error-free while silently failing a few percent of captures per day.

### Repro
1. Start the worker daemon; note the date.
2. Let it run past midnight (UTC) and produce activity.
3. `ls -la ~/.claude-mem/logs/` → the startup-date file keeps growing; today's file contains only hook/short-process lines.
4. `GET /api/logs` → returns today's file: no daemon `[SESSION]`/`[QUEUE]`/`[ERROR]` lines at all.

### Observed
```
-rw-rw-rw- 1 root root 54803217 Jul 26 01:49 claude-mem-2026-07-07.log   <- daemon (started Jul 7) still writing
-rw-rw-rw- 1 root root  2629605 Jul 25 23:55 claude-mem-2026-07-25.log   <- hooks/MCP only, zero ERROR/WARN
```
The 07-07 file contained hundreds of ERROR lines from the same day the API-served file showed none.

### Ask
- Recompute the log path per write (roll at the date boundary), or use a stable `claude-mem.log` with size-based rotation.
- Either way, `/api/logs` should read the file the daemon actually writes.

### Related
This is how the failure modes in #3344 stayed invisible for us for weeks — the failure-path logging existed but landed in a file nothing reads.

Alan5168:
## Problem

`Logger.ensureLogFileInitialized()` resolved the dated log path only once, so long-running workers kept writing to the startup-day file after UTC midnight. The read path already selects today's file, which made logs appear empty after rollover.

## Fix

Cache the UTC date only after the dated path is initialized successfully. A date change selects the new file, while a transient path/filesystem failure remains retryable on the next same-day log call.

The regression test now cleans generated files in `afterEach` and covers both midnight rollover and same-day recovery after a transient initialization failure.

## Test

- `bun test tests/utils/logger-log-path-rollover.test.ts`: 2 pass / 0 fail
- `bun test tests/utils`: 223 pass / 0 fail
- `bunx tsc --noEmit --pretty false`: clean
- Earlier full-suite run: 2539 pass / 18 skip / 1 pre-existing timeout in `tests/worker/sync/mutation-sites.test.ts`, reproduced on clean `main`

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -79,16 +79,22 @@ class Logger {
   private level: LogLevel | null = null;
   private useColor: boolean;
   private logFilePath: string | null = null;
-  private logFileInitialized: boolean = false;
+  private logFileDate: string | null = null;
 
   constructor() {
     this.useColor = process.stdout.isTTY ?? false;
     // Don't initialize log file in constructor - do it lazily to avoid circular dependency
   }
 
+  // Re-resolves the log path whenever the date changes so long-running
+  // daemons roll onto the new day's file instead of appending to the file
+  // named for their start date forever (#3415). The date string doubles as
+  // the "already initialized for today" cache, so same-day calls stay a
+  // cheap comparison.
   private ensureLogFileInitialized(): void {
-    if (this.logFileInitialized) return;
-    this.logFileInitialized = true;
+    const date = new Date().toISOString().split('T')[0];
+    if (this.logFileDate === date) return;
+    this.logFileDate = date;
 
     try {
       const logsDir = paths.logsDir();
@@ -97,7 +103,6 @@ class Logger {
         mkdirSync(logsDir, { recursive: true });
       }
 
-      const date = new Date().toISOString().split('T')[0];
       this.logFilePath = join(logsDir, `claude-mem-${date}.log`);
     } catch (error: unknown) {
       console.error('[LOGGER] Failed to initialize log file:', error instanceof Error ? error.message : String(error));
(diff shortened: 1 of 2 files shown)


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3603 ([plan-15] Worker Port & Liveness Authority)** as **standalone / merge-now**: it is small and self-contained enough to land on its own ahead of the plan-15 bundle once CI is green.

> step 9 (log path recomputed per write / date roll)

Alan5168:
Closing as hygiene: no activity since 2026-08-17. Can reopen against current main if the dated log-path fix is still wanted.

PR #3422, opened 2026-07-27
feat(chroma): expose embedding function as a setting

chang-in:
# feat(chroma): expose embedding function as a setting

## Problem

`ChromaSync.ensureCollectionExists()` creates the collection without passing `embedding_function_name`:

```ts
await chromaMcp.callTool('chroma_create_collection', {
  collection_name: this.collectionName
});
```

chroma-mcp then falls back to its `default` — `all-MiniLM-L6-v2`, which is tuned for English. Users whose observations are written in another language get noticeably weaker semantic search, and there is currently no way to change this short of patching the source.

## Change

Adds `CLAUDE_MEM_CHROMA_EMBEDDING_FUNCTION` (default `'default'`) and forwards it at collection creation. Behaviour is identical unless the setting is changed.

## Impact for non-English users

Measured on a real Korean-language observation corpus from this database (short summary as query, full narrative as document, 200 candidates, two independent samples):

| Embedding function | Dim | R@1 (A / B) | MRR (A / B) |
|---|---|---|---|
| `default` — all-MiniLM-L6-v2 | 384 | 12.0% / 10.0% | 0.166 / 0.148 |
| `multilingual` — paraphrase-multilingual-MiniLM-L12-v2 | 384 | **34.5% / 32.5%** | **0.462 / 0.457** |

R@1 improves by **+22.5pp in both samples**. Both models emit 384-dimensional vectors, so storage size and query latency are unchanged.

> Note: the `multilingual` value requires the corresponding option in chroma-mcp (PR filed separately). This change is still useful without it — it also unlocks the API-backed functions chroma-mcp already supports (`openai`, `cohere`, `jina`, `voyageai`) for users who want them.

## Notes

- Default is `'default'` — **no behaviour change** for existing installs.
- Embedding functions are fixed at collection creation time, so changing this setting only affects a **newly created** collection. Existing users must delete `~/.claude-mem/chroma/` and the sync watermark to re-index. This is documented in the setting's comment.
- Follows the existing settings pattern (`SettingsDefaultsManager.loadFromFile(USER_SETTINGS_PATH)`) already used in `ChromaMcpManager`.

## Files changed

- `src/shared/SettingsDefaultsManager.ts` — interface field + default value
- `src/services/sync/ChromaSync.ts` — read setting, pass to `chroma_create_collection`

### DIFF
--- a/src/services/sync/ChromaSync.ts
+++ b/src/services/sync/ChromaSync.ts
@@ -11,6 +11,8 @@ import { ParsedObservation, ParsedSummary } from '../../sdk/parser.js';
 import type { SessionStore as SessionStoreType } from '../sqlite/SessionStore.js';
 import { logger } from '../../utils/logger.js';
 import { ChromaUnavailableError } from '../worker/search/errors.js';
+import { SettingsDefaultsManager } from '../../shared/SettingsDefaultsManager.js';
+import { USER_SETTINGS_PATH } from '../../shared/paths.js';
 import { normalizePlatformSource } from '../../shared/platform-source.js';
 import type * as SqliteFilesModule from '../sqlite/observations/files.js';
 
@@ -121,9 +123,13 @@ export class ChromaSync {
     }
 
     const chromaMcp = ChromaMcpManager.getInstance();
+    const settings = SettingsDefaultsManager.loadFromFile(USER_SETTINGS_PATH);
+    const embeddingFunction =
+      settings.CLAUDE_MEM_CHROMA_EMBEDDING_FUNCTION || 'default';
     try {
       await chromaMcp.callTool('chroma_create_collection', {
-        collection_name: this.collectionName
+        collection_name: this.collectionName,
+        embedding_function_name: embeddingFunction
       });
     } catch (error) {
       const message = error instanceof Error ? error.message : String(error);
--- a/src/shared/SettingsDefaultsManager.ts
+++ b/src/shared/SettingsDefaultsManager.ts
@@ -64,6 +64,7 @@ export interface SettingsDefaults {
   CLAUDE_MEM_CHROMA_TENANT: string;
   CLAUDE_MEM_CHROMA_DATABASE: string;
   CLAUDE_MEM_CHROMA_PREWARM_TIMEOUT_MS: string;
+  CLAUDE_MEM_CHROMA_EMBEDDING_FUNCTION: string;  // chroma-mcp embedding function for new collections
   // Worker-native cloud sync. Active ⇔ TOKEN, USER_ID, and HUB_URL are all
   // non-empty — there is no separate enabled flag. HUB_URL points at the
   // two-lane sync hub (workers/sync-hub); while it is empty, sync is OFF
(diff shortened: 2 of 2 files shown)


chang-in:
I looked into the settings-API finding and I think it's a false positive: this setting follows the existing Chroma convention, so I don't believe `settingKeys` needs to change here.

`settingKeys` in `SettingsRoutes.ts:77` holds 29 of the 84 keys declared in `SettingsDefaults` — it's the set the viewer settings page can write, and the UI sends 21 distinct keys. All ten `CLAUDE_MEM_CHROMA_*` keys sit outside it: the nine that predate this PR (`CHROMA_ENABLED`, `CHROMA_MODE`, `CHROMA_HOST`, `CHROMA_PORT`, `CHROMA_SSL`, `CHROMA_API_KEY`, `CHROMA_TENANT`, `CHROMA_DATABASE`, `CHROMA_PREWARM_TIMEOUT_MS`) plus the one added here. `src/ui/` has no Chroma settings form at all — `CHROMA` appears there only as a log-component filter in `LogsModal.tsx`.

The Chroma group is configured by file or environment, and every consumer reads it that way: `worker-service.ts:479`, `DatabaseManager.ts:31`, `dependency-preflight.ts:154`, `ChromaRoutes.ts:16`. This PR does the same at `ChromaSync.ts:126` via `SettingsDefaultsManager.loadFromFile(USER_SETTINGS_PATH)`, which applies environment overrides by default (`applyEnvOverrides = true`, `SettingsDefaultsManager.ts:221`). So the setting is reachable through `settings.json` or `CLAUDE_MEM_CHROMA_EMBEDDING_FUNCTION` in the environment — the same two paths as the rest of the group.

Adding only this key to the allowlist would make it the single Chroma setting writable over HTTP, with no UI to write it from. That seems worse than keeping the group uniform. If you'd rather `POST /api/settings` accept these, I'm glad to add the whole Chroma group — but that's a separate decision about which settings are UI-editable, and I didn't want to make it inside this PR.

Flagging it here since the bot marked it as needing a fix before merge; the PR is otherwise ready as-is.

seokjushin:
+1 — confirming the problem from an independent real-world Korean-language setup (claude-mem 13.13.1, macOS, `chroma-mcp 0.2.6`).

**Confirmed the default is `all-MiniLM-L6-v2`:** our local onnx cache (`~/.cache/chroma/onnx_models/all-MiniLM-L6-v2`) shows the English-tuned default is what every user silently gets today, exactly as this PR describes.

One observation that may explain why this doesn't get reported more often: **the hybrid search path masks the embedding weakness at small scale.** On our DB (~490 observations), CJK queries still return results because several retrieval legs overlap — FTS5 exact matches on Korean tokens embedded in English observations, prompt-side matches (prompts are stored verbatim in the user's language), and loanword tokens (e.g. 인스턴스/instance) that happen to land near their English counterparts. So the degradation is invisible until the corpus grows or the query uses native Korean vocabulary with no English-adjacent tokens — which is consistent with #3173 reporting 0-observation results at 45k scale while small DBs look fine.

The benchmark in this PR (+22.5pp R@1, MRR 0.166 → 0.462 on a Korean corpus with `paraphrase-multilingual-MiniLM-L12-v2`) matches what we'd expect from that analysis, and since the default stays `'default'`, the change is zero-risk for existing English users.

Would love to see this land — it's the root fix for the CJK search family (#3173) rather than a fallback workaround.

thedotmack:

> step 1 spec surface: `CLAUDE_MEM_CHROMA_EMBEDDING_FUNCTION`; rebase (SettingsDefaultsManager moved)

thedotmack:
Routed to **#3610 (plan-22, Chroma Sidecar Contract)** — `chroma_create_collection` silently accepting the sidecar's `all-MiniLM-L6-v2` default is a sidecar-contract gap, and naming what we embed with belongs with the pinning work in that master.

Worth cross-reading with **#3982 (plan-25, Search Read Path / CJK)**: your Korean measurements (R@1 12% to 34.5%) are the strongest evidence in this repo that the non-English retrieval gap is an embedding-model problem on the semantic side, not only an FTS5 tokenizer problem. Two files, no behavior change at the default, no duplicate open — it just needs a rebase.

chang-in:
Rebased onto current `main` (d8bc9755) — the branch is conflict-free again.

The only conflict was in `SettingsDefaultsManager.ts`: `CLAUDE_MEM_CHROMA_MAX_PENDING_MUTATIONS` landed at the same spot in both the interface and `DEFAULTS`, so I kept both entries. `ChromaSync.ts` auto-merged cleanly and `loadFromFile` / `USER_SETTINGS_PATH` are unchanged upstream, so no code changes beyond the rebase. Still 2 files, +13/−1.

Verified locally: `npm run typecheck` passes, and `bun test tests/services/sync tests/shared tests/infrastructure/version-consistency.test.ts` → 409 pass / 0 fail.

The CI and Windows workflows are sitting at `action_required` since this is my first PR here — would you mind approving the run when you get a chance? Thanks for routing this.

PR #3428, opened 2026-07-27
fix(hooks): suppress resolver stderr

### LINKED ISSUE #3426: Windows/Git Bash: hooks intermittently fail with 'printf: write error: Permission denied' (pipe race in plugin-path resolution)
## Environment

- Windows 11, Claude Code running hooks via Git Bash (`/usr/bin/bash`)
- claude-mem 12.7.5 and 13.12.4 (both affected — verified the pattern is unchanged in 13.12.4's `hooks/hooks.json`)

## Symptom

Hooks intermittently fail with:

```
```

Claude Code surfaces this as blocking hook feedback ("Stop hook feedback" at end of turn, and PostToolUse blocking errors after tool calls), so it interrupts the session repeatedly even though the memory worker itself is healthy.

## Root cause

Every hook command resolves the plugin install path with this pattern:

```bash
_P=$({ [ -n "$_E" ] && printf '%s\n' "$_E"; ls -dt "$_C/plugins/cache/thedotmack/claude-mem"/[0-9]*/ 2>/dev/null; printf '%s\n' "$_C/plugins/marketplaces/thedotmack/plugin"; } | while IFS= read -r _R; do ...; break; done)
```

The `while` loop `break`s at the first valid candidate, closing the read end of the pipe. The left-hand group's remaining `printf`/`ls` writes then race against the teardown; on Windows/Git Bash the closed pipe manifests as `printf: write error: Permission denied` on stderr (rather than a silently-ignored SIGPIPE as on Linux/macOS). It's timing-dependent, so it appears intermittently — in my session it fired on most Stop and PostToolUse events.

Running the same command manually confirms the worker resolves the path correctly and exits 0 — the only problem is the stray stderr, which Claude Code treats as hook failure output.

## Fix

Suppress stderr on the candidate-list group (one-line change, applied to all 7 hook commands):

```bash
_P=$({ ...candidates... ; } 2>/dev/null | while IFS= read -r _R; do ...; done)
```

I've verified this locally on both 12.7.5 and 13.12.4 — the noise disappears and the genuine failure path (`claude-mem: plugin scripts not found`) still reports correctly, since that message is emitted outside the suppressed group.

An alternative that avoids the race entirely is collecting the candidates into a variable (or `for` loop) instead of piping into `while ... break`, which also drops the subshell.

ousamabenyounes:
Fix #3426

## Summary

- Suppress stderr on the whole POSIX hook candidate producer group before the resolver pipe.
- Regenerate Claude and Codex hook templates from the canonical generator.
- Add a regression test that fails if the group-level stderr redirect is removed.

## Validation

- RED: `bun test tests/worker-script-resolution.test.ts --test-name-pattern 'POSIX hook candidate enumeration suppresses stderr for the whole producer group'` failed when the source fix was temporarily reverted (`0 pass / 1 fail`).
- GREEN: `bun test tests/worker-script-resolution.test.ts tests/infrastructure/plugin-distribution.test.ts` (`70 pass / 0 fail`).
- GREEN: `npm run typecheck`.
- GREEN: `npm run build`.
- GREEN: `npm run lint:hook-io`.
- GREEN: `npm run lint:spawn-env`.
- GREEN: `git diff --check`.

## Baseline

`bun test tests` is not currently clean on `origin/main` in this checkout. Baseline on `origin/main` reported `2437 pass / 18 skip / 27 fail / 5 errors`; this branch reported `2438 pass / 18 skip / 27 fail / 5 errors`, with the same pre-existing failure families around `ProcessManager`, `HealthMonitor`, and server mode boot.


thedotmack:

> step 1 (every remaining producer gets `2>/dev/null`; group-level redirect)

thedotmack:
Routed to **#3605 (plan-17, Hook Wrapper Contract)** — suppressing stderr across the whole POSIX candidate-producer group before the resolver pipe is the "plugin-root discovery pipe" line of that master, and the regression test that fails when the group-level redirect is removed is the kind of guard plan-17 wants on every wrapper invariant.

Distinct fix, no duplicate in the queue. It does share `src/build/hook-shell-template.ts` with #3431 and #3403, so all three want sequencing behind one template rewrite instead of merging separately. Conflicting against main now — regenerate and rebase.

PR #3430, opened 2026-07-27
fix(chroma): tolerate CJK list backfill

### LINKED ISSUE #3423: Title: [BUG] ChromaDB backfill fails with "JSON Parse error: Unrecognized token" for CJK (Chinese/Japanese/Korean) observations
Body:

## Describe the bug

ChromaDB backfill consistently fails when observations contain CJK text
(Chinese, Japanese, Korean characters). The backfill process passes raw CJK
text to `JSON.parse()`, causing an unrecoverable parse error that leaves
ChromaDB permanently empty for CJK-heavy projects.

## Steps to reproduce

1. Use claude-mem in a project where observations are generated with Chinese
   titles and content (e.g. `观察: 用户身份定位——轻量数字化改造枢纽`)
2. Let the worker attempt ChromaDB backfill on startup
3. Observe the error in `~/.claude-mem/logs/claude-mem-YYYY-MM-DD.log`

## Expected behavior

Observations with CJK text should be synced to ChromaDB successfully, just as
English-only observations are.

## Actual behavior

Backfill fails with `JSON Parse error: Unrecognized token` on the first CJK
character encountered. Context injection (`/api/context/inject`) also fails
with `malformed JSON` because ChromaDB has no observation embeddings.

## Log output

[CHROMA_MCP] Connected to chroma-mcp successfully
[CHROMA_SYNC] Backfilling observations {project=Lenovo, missing=15, pending=0, watermark=0, total=15}
[CHROMA_SYNC] Backfill failed {project=Lenovo} JSON Parse error: Unrecognized token '用'
[CHROMA_SYNC] Backfill failed for project: Lenovo Backfill failed: JSON Parse error: Unrecognized token '用'
[HTTP  ] Context injection failed {projects=Lenovo} malformed JSON

`chroma-sync-state.json` remains stuck at 0 observations / 0 summaries
despite 15 observations existing in SQLite.

## Environment

- **claude-mem version:** 13.12.4 (also reproduced on 13.11.0)
- **chroma-mcp version:** 0.2.6
- **OS:** Windows 10 Pro (x86_64)
- **Bun:** 1.3.14
- **uv:** 0.11.15
- **Provider:** gemini (also reproduced with claude provider)
- **ChromaDB:** persistent, local, clean (issue persists after wiping
  `~/.claude-mem/chroma/`)

## Impact

- ChromaDB semantic search is non-functional for CJK projects
- Session-start memory injection (`context/inject`) fails
- SQLite FTS search still works, but without vector similarity
- 100% reproducible — happens on every worker start

## Possible root cause

In the backfill pipeline, observation content (which may contain multi-byte
UTF-8 CJK characters) is likely being passed directly to `JSON.parse()`
without proper encoding/escaping, or the MCP stdio transport between the
worker and chroma-mcp is mishandling non-ASCII bytes in the JSON-RPC framing.

ousamabenyounes:
Fix #3423

## Summary

- Make Chroma observation backfill tolerate legacy/plain-string `facts` and `concepts` fields instead of assuming every stored value is a JSON array.
- Decode valid JSON string scalars before indexing so their JSON quote delimiters are not retained.
- Preserve malformed and non-string scalar values as indexable Chroma content without logging stored memory text.
- Add regression coverage for CJK text, JSON string scalars, non-string scalars, and malformed JSON privacy.

- RED: `bun test tests/services/sync/chroma-sync-watermarks.test.ts` failed because the Chroma metadata retained `"数字化改造"` instead of decoded `数字化改造`.
- GREEN: the focused Chroma watermark suite passed (`6 pass / 0 fail`).
- Local CI baseline (`origin/main`): `2554 pass / 18 skip / 22 fail / 3 errors`.
- Local CI final: `2556 pass / 18 skip / 22 fail / 3 errors`, with identical pre-existing failure identities.
- GREEN: typecheck, build, 100% changed-production-line coverage, OpenClaw (`44 pass / 0 fail`), hook-I/O lint, spawn-env lint, and `git diff --check`.

### DIFF
--- a/src/services/sync/ChromaSync.ts
+++ b/src/services/sync/ChromaSync.ts
@@ -95,6 +95,39 @@ interface StoredUserPrompt {
   platform_source: string;
 }
 
+const MALFORMED_LIST_FIELD_PREVIEW_CHARS = 100;
+
+function parseStringListField(
+  rawValue: string | null | undefined,
+  fieldName: 'facts' | 'concepts',
+  rowId: number,
+): string[] {
+  if (!rawValue) {
+    return [];
+  }
+
+  try {
+    const parsed = JSON.parse(rawValue);
+    if (!Array.isArray(parsed)) {
+      logger.warn('CHROMA_SYNC', 'Expected JSON array in observation list field, using plain string fallback', {
+        fieldName,
+        rowId,
+        parsedType: typeof parsed,
+      });
+      return typeof parsed === 'string' && parsed.trim() ? [parsed] : [];
+    }
+    return parsed.filter((value): value is string => typeof value === 'string' && value.trim().length > 0);
+  } catch (error) {
+    logger.warn('CHROMA_SYNC', 'Malformed observation list field, using plain string fallback', {
+      fieldName,
+      rowId,
+      preview: rawValue.slice(0, MALFORMED_LIST_FIELD_PREVIEW_CHARS),
+      error: error instanceof Error ? error.message : String(error),
+    });
+    return rawValue.trim() ? [rawValue] : [];
+  }
+}
+
 export class ChromaSync {
   private project: string;
   private collectionName: string;
@@ -153,8 +186,8 @@ export class ChromaSync {
   private formatObservationDocs(obs: StoredObservation): ChromaDocument[] {
     const documents: ChromaDocument[] = [];
 
-    const facts = obs.facts ? JSON.parse(obs.facts) : [];
-    const concepts = obs.concepts ? JSON.parse(obs.concepts) : [];
+    const facts = parseStringListField(obs.facts, 'facts', obs.id);
+    const concepts = parseStringListField(obs.concepts, 'concepts', obs.id);
     // parseFileList is SQLite-shaped (`bun:sqlite` in the import chain) —
     // resolve it through the deferred loader so this method stays out of
     // the SDK bundle's import graph. Plan §3.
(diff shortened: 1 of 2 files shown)


thedotmack:

> step 3 (UTF-8/CJK-safe framing; tolerant facts/concepts parse); P1: JSON scalar handling + raw preview leaks content into logs

ousamabenyounes:
Thanks for the routing review @thedotmack. Addressed the remaining JSON-scalar blocker in `90d6c491`: valid JSON string scalars are decoded before indexing, while malformed and non-string values retain the raw fallback without logging stored content.

Validation: reproduced the quoted-scalar mismatch RED, then the focused Chroma test passed (`6 pass / 0 fail`). Local CI remained iso-baseline (`22 fail / 3 errors` on both `origin/main` and this branch, with two additional passing tests); typecheck, build, changed-production-line coverage, OpenClaw (`44 pass`), and lint checks passed. Remote checks are now running on the pushed SHA.

thedotmack:
**All-night ship gate: PASS**

1. RCA logged — `JSON.parse` assumed array-shaped `facts`/`concepts`; CJK/plain-string legacy rows aborted backfill (#3423 / #3610 step 3). JSON-scalar decode + no raw-content logs addressed after the rework note.
2. No second-system — tolerant parse lives in `formatObservationDocs`, the existing backfill path.
3. Narrow / not opinionated — helper replaces two `JSON.parse` calls; no new store/mode.

Rebased onto `2714e869` locally (fork branch update blocked: GitHub App `workflows` permission). Focused tests: `bun test tests/services/sync/chroma-sync-watermarks.test.ts` — 11 pass. CI on this head was green. Merging independently; no version bump.

PR #3431, opened 2026-07-28
fix(hooks): fail open Claude runtime hooks

### LINKED ISSUE #3412: bash hooks fail silently on Windows when the default WSL distro is docker-desktop
## Summary

Every prompt in Claude Code is blocked with "A hook blocked your prompt" and the message "No stderr output". The plugin is unusable until it is disabled.

The cause is the `"shell": "bash"` declaration on every hook in `plugin/hooks/hooks.json`. On Windows, `bash` resolves to `C:\Windows\System32\bash.exe`, which is the WSL launcher rather than Git Bash. My only registered WSL distribution is `docker-desktop`, the internal utility VM that Docker Desktop installs. It is therefore the default, so the hook body executes inside that VM.

Inside the docker-desktop VM there is no Node, no `cygpath`, and `$HOME` is `/root`. The plugin root cannot resolve and `node` is not found. The hook exits non-zero with nothing useful on stderr, and Claude Code treats that as a block.

Git Bash 2.52.0 is installed on the same machine. The identical hook body runs correctly under it.

## Environment

| Item | Value |
|---|---|
| OS | Windows 11 Home 10.0.26200 (64-bit) |
| claude-mem | 12.6.2 |
| Claude Code | 2.1.177 |
| Node (Windows) | v20.20.0 at `C:\Program Files\nodejs\node.exe` |
| Bun | present at `~/.bun/bin/bun.exe` |
| `bash` on PATH | `C:\Windows\System32\bash.exe` (WSL launcher) |
| Git Bash | installed at `C:\Program Files\Git\bin\bash.exe`, not first on PATH |
| WSL distributions | `docker-desktop` only, running, version 2 |
| `SHELL` env var | not set |

## Evidence

The plugin files are all present and intact. Running the hook body directly under Windows Node succeeds:

```powershell
node "$r\scripts\bun-runner.js" "$r\scripts\worker-service.cjs" hook claude-code session-init
# exit code 0, no output
```

Running the same resolution logic under Git Bash resolves everything correctly:

```text
SHELL var: [/usr/bin/bash]
HOME: /c/Users/Robert Cumming
RESOLVED ROOT: C:\Users\Robert Cumming\.claude\plugins\cache\thedotmack\claude-mem\12.6.2
node: FOUND (v20.20.0)
bun-runner.js: FOUND
exit=0
```

Running a login shell in the environment that `bash` actually reaches gives this:

```text
HOME=/root
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
node: not found
```

`/root/.claude` does not exist in that VM, so the `ls -dt "$HOME/.claude/plugins/cache/..."` lookup returns nothing and `_R` falls through to a marketplace path that also does not exist.

## Why the existing Windows handling does not catch this

The hook body guards Windows with `command -v cygpath`. That check is sound for Git Bash and MSYS2, both of which ship `cygpath`. The docker-desktop VM does not ship it, so the branch is skipped and `_R` keeps a Linux path that will never resolve.

The `$SHELL -lc 'echo $PATH'` step is also a no-op here. `SHELL` is unset, and the shell it reaches has a bare Linux PATH with no Node on it.

## Suggested fixes

Three options, in the order I would rank them.

1. Make the SessionStart and UserPromptSubmit hooks exit 0 on any failure and write the reason to stderr. The installation docs state that the Setup hook never blocks a session for exactly this reason. The other hooks carry no such guarantee, and a memory plugin failing should degrade to no memory rather than to no Claude Code.

2. Resolve `bash` explicitly on Windows rather than relying on PATH order. Prefer `C:/Program Files/Git/bin/bash.exe` when it exists.

3. Detect the docker-desktop VM and bail out early with a clear message. Checking for `/root` as `$HOME` alongside a missing `~/.claude` would be enough.

Fix 1 alone would have turned this into a minor annoyance rather than a full block.

## Notes

Docker Desktop plus no other WSL distribution is a common Windows setup, so I doubt I am the only person hitting this. `npx claude-mem repair` will not help, because the install is not broken. The plugin cannot be reached from the shell the hooks run in.

Happy to test a patch.

ousamabenyounes:
Refs #3605 (plan-17, step 2) — originally filed as #3412.

## Summary

- Claude Code **runtime** hooks (`SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `Stop`) now fail open: when the plugin root cannot be resolved, or when the resolved runner exits non-zero, the hook reports the reason on stderr and exits `0` instead of blocking the prompt.
- The **Setup** hook stays fail-loud, so install/setup problems are still surfaced outside the per-prompt hot path.
- `plugin/hooks/hooks.json` is regenerated from `src/build/hook-shell-template.ts`; the exit statuses of the two contracts are named constants (`FAIL_OPEN_EXIT_CODE` / `FAIL_LOUD_EXIT_CODE`).

The wrapper leaves the child's stdout untouched, so a hook that already produced its JSON envelope still delivers it even when the runner then exits non-zero:

```text
$ bash -c "$UserPromptSubmit_command"   # runner writes {"continue":true}, then exit 2
exit=0
stdout=[{"continue":true}]
stderr=[claude-mem: hook command failed (exit 2)]
```

RED — upstream `main` (`e2d1df5`) with only this PR's tests applied:

```text
$ bun test tests/infrastructure/plugin-distribution.test.ts
(fail) Plugin Distribution - Startup Root Resolution > Claude runtime hooks fail open when plugin scripts cannot be resolved (#3412)
   expect(received).toBe(expected)   Expected: 0   Received: 1
(fail) Spawn-Contract Templating - Rule A shell resolution matrix > keeps Setup fail-loud while runtime hooks fail open when no candidate exists
   expect(received).toBe(expected)   Expected: 0   Received: 1
(fail) Spawn-Contract Templating - Rule A shell resolution matrix > keeps runtime hooks fail-open when the resolved command exits non-zero (#3412)
   expect(received).toBe(expected)   Expected: 0   Received: 7

 56 pass
 3 fail
```

GREEN — this branch:

```text
$ bun test tests/infrastructure/plugin-distribution.test.ts
 59 pass
 0 fail
Ran 59 tests across 1 file.
```

## Validation

Rebased onto `main` (`e2d1df5`). Full suite compared against the same `main` checkout:

```text
                     main (baseline)      this branch
bun test tests       2554 pass / 18 skip / 22 fail    2556 pass / 18 skip / 22 fail
```

Identical failure set on both sides (all `ProcessManager` / `HealthMonitor` cases, environment-dependent locally — they pass on CI); the +2 are this PR's new tests.

Also green: `npm run typecheck`, `npm run build` (bundle-size guardrails), `bun test openclaw`, the Durable Object anti-pattern grep, the plugin frozen-lockfile drift check, `npm run smoke:clean-room`, `npm run lint:hook-io`, `npm run lint:spawn-env`, `git diff --check`.


ousamabenyounes:
Follow-up pushed for the Greptile coverage gap: added a regression test for the resolved runtime-hook command path where bun-runner.js exits non-zero.

Local validation after the follow-up: ./run-ci.sh passed with plugin-distribution now at 57 pass / 0 fail, plus typecheck, build, lint:hook-io, lint:spawn-env, and git diff --check.

thedotmack:

> step 2 (runtime hooks exit 0 on unresolved root; only Setup is fail-loud)

thedotmack:
Routed to **#3605 (plan-17, Hook Wrapper Contract)** — fail-open runtime hooks with a fail-loud Setup hook is precisely the contract that master is meant to settle, and naming the two exit codes as constants is the right way to make it testable.

It overlaps in *intent* with the worker-unreachable fail-open PRs (#3269, #3225, #3243, #2892) but at a different layer — this one is the generated shell wrapper, those are the hook command — so I am not treating any of them as duplicates. It shares `src/build/hook-shell-template.ts` with #3428 and #3403, so those three have to be sequenced behind one template rewrite rather than merged independently. Needs a rebase.

PR #3432, opened 2026-07-28
fix(supervisor): jail SDK subprocess cwd

### LINKED ISSUE #3357: 0-byte files named after transcript tokens written into the user's working directory (unquoted `>` reaching a shell)
# 0-byte files named after transcript tokens appear in the project working directory

**Version:** claude-mem 13.11.0 · Windows 11 · Claude Code

## Symptom

0-byte files keep appearing in the active working directory (repo root, or whichever
subdirectory the last shell command `cd`'d into). Examples seen in one repo:

    500)   Claims   at   Asset   Finding   Entity}   None`   services/twin/None

They are picked up by `git add -A` and silently land in commits.

## What the names are

Every stray name is **the token immediately following a `>` in recent transcript text**:

| stray file | source text |
|---|---|
| `Finding`, `None`, `Asset` | Python return annotations: `def detect_findings(...) -> Finding`, `-> None` |
| `Entity}` | `-> dict[str, Entity}` (truncated in transcript) |
| `500)` | prose: `TypeError -> 500)` |
| `Claims`, `at` | `-> Claims`, `-> at` in reviewed code/prose |

`->` is `-` followed by `>`. So this is **shell output redirection**, not tokenization:
observed transcript text is being interpolated **unquoted** into a shell command line
that runs with the user's project cwd, and the shell creates one empty file per `>`.

## Live reproduction

During ordinary Python editing, seven strays appeared within four minutes. Each name is
the token after a `->` in code written **or merely read** in that same minute:

    0 bytes  22:26:00  list[dict[str          <- `-> list[dict[str, object]]`   (written)
    0 bytes  22:27:28  Finding                <- `deviation->Finding`            (written)
    0 bytes  22:27:53  per-tenant             <- `same partition -> per-tenant ordering` (written)
    0 bytes  22:28:07  Entity}                <- `org -> {entity id -> Entity}`  (READ ONLY)
    0 bytes  22:28:15  list[Deviation]        <- `-> list[Deviation] | None`     (written)
    0 bytes  22:29:06  Iterator[tuple[str     <- `-> Iterator[tuple[str, ...]]`  (written)
    0 bytes  22:30:15  None                   <- `-> None`                       (written)

mtimes track individual tool calls. The `Entity}` case is the decisive one: that text came
from a file that was only **opened for reading**, never edited and never part of any command.
So the string reaching the shell is the **observation payload**, not the executed command.

## Evidence / what was ruled out

* Files appear **after** the tool call returns; mtime tracks the tool call — consistent with
  the `async: true` PostToolUse observation hook (`worker-service.cjs hook claude-code observation`).
* They are **cwd-relative**, and follow the cwd of the most recent shell command.
* Names are drawn from **transcript history**, not only the current tool call (a stale token
  reappeared with no current reference) — so the source is the observation payload, not the live command.
* Invoking the hook one-shot in a sandbox does **not** reproduce it: the hook returns `{}` and
  queues to the persistent background worker started at SessionStart with the real project cwd.
* Not the host: Bash-tool commands containing literal `>` tokens (quoted) do not create files.
* No `shell: true` / `bash -c` found in `worker-service.cjs`, but it spawns a headless Claude
  process whose cwd defaults to `resolve(t ?? ".")` — i.e. inherits the project cwd. The unquoted
  interpolation is likely in that subprocess's command construction.

## Impact

Repo contamination. Without a strict staging discipline these empty files get committed.
Caught twice here before push; a `git add -A` user would ship them.

## Suggested fix

Quote/escape observed text before it reaches any shell command line (or pass it via stdin/argv
array rather than a shell string), and give the worker a data-dir cwd rather than the project cwd
so any residual writes cannot land in the user's repo.

## Workaround

Never `git add -A`; stage explicit paths.

ousamabenyounes:
Fix 3357

## Summary
- Jail each supervised SDK subprocess in its own `~/.claude-mem/observer-sessions/<sessionDbId>` directory, never the user's project cwd.
- Create the session directory before both direct and Windows `.cmd` wrapper spawns; `spawnSdkProcess` returns `null` cleanly if it cannot be created.
- Keep the SDK callback `cwd` field for compatibility while deliberately ignoring it at the trust boundary.

Rebased onto the latest `main`: the observer-session path helpers (`OBSERVER_SESSIONS_DIR`, `ensureDir`) have since landed there, so this branch now carries only the supervisor cwd-jailing change in `src/supervisor/process-registry.ts` plus its tests.

RED — unmodified `main` lacks the jailing (the focused probe cannot resolve `normalizeSpawnSdkCwd` / the child inherits the supplied project directory):

```text
Export named 'normalizeSpawnSdkCwd' not found in module '.../src/supervisor/process-registry.ts'
```

GREEN — with this change:

```text
bun test tests/supervisor/process-registry.test.ts  →  27 pass  0 fail
```

Full local suite `bun test tests` → 3053 pass / 0 fail / 27 skip, no regression. The `typecheck · build · test · bundle-size` job's `install.ts` errors are pre-existing on `main` (unrelated `@clack/prompts` symbol types) — identical set with and without this branch.

### DIFF
--- a/src/supervisor/process-registry.ts
+++ b/src/supervisor/process-registry.ts
@@ -4,7 +4,7 @@ import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
 import path from 'path';
 import { logger } from '../utils/logger.js';
 import { sanitizeEnv } from './env-sanitizer.js';
-import { OBSERVER_SESSIONS_DIR, paths } from '../shared/paths.js';
+import { ensureDir, OBSERVER_SESSIONS_DIR, paths } from '../shared/paths.js';
 // Moved to shared/ so kill-process-tree.ts can use it without closing an
 // import cycle (process-registry already imports kill-process-tree). Re-exported
 // here so every existing caller keeps its import path.
@@ -586,6 +586,8 @@ export interface SpawnSdkOptions {
   command: string;
   args: string[];
   extraArgs?: string[];
+  // Part of the Claude SDK callback contract. Accepted at this trust boundary
+  // so callers remain compatible, but deliberately ignored for isolation.
   cwd?: string;
   env?: NodeJS.ProcessEnv;
   signal?: AbortSignal;
@@ -615,8 +617,8 @@ export function normalizeSpawnSdkArgs(args: string[], extraArgs: string[] = []):
   return filteredArgs;
 }
 
-export function normalizeSpawnSdkCwd(_cwd: string | undefined): string {
-  return OBSERVER_SESSIONS_DIR;
+export function normalizeSpawnSdkCwd(sessionDbId: number): string {
+  return path.join(OBSERVER_SESSIONS_DIR, String(sessionDbId));
 }
 
 export function spawnSdkProcess(
@@ -628,7 +630,17 @@ export function spawnSdkProcess(
   const useCmdWrapper = process.platform === 'win32' && options.command.endsWith('.cmd');
   const env = sanitizeEnv(options.env ?? process.env);
   const filteredArgs = normalizeSpawnSdkArgs(options.args, options.extraArgs);
-  const cwd = normalizeSpawnSdkCwd(options.cwd);
+  const cwd = normalizeSpawnSdkCwd(sessionDbId);
+  try {
+    ensureDir(cwd);
+  } catch (error: unknown) {
+    const cause = error instanceof Error ? error : new Error(String(error));
+    logger.error('SDK_SPAWN', `[session-${sessionDbId}] failed to create observer session directory`, {
+      sessionDbId,
+      cwd,
+    }, cause);
+    return null;
+  }
 
   const isWin = process.platform === 'win32';
   const child = useCmdWrapper
--- a/src/supervisor/process-registry.ts
+++ b/src/supervisor/process-registry.ts
@@ -4,7 +4,7 @@ import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
 import path from 'path';
 import { logger } from '../utils/logger.js';
 import { sanitizeEnv } from './env-sanitizer.js';
-import { paths } from '../shared/paths.js';
+import { OBSERVER_SESSIONS_DIR, paths } from '../shared/paths.js';
 // Moved to shared/ so kill-process-tree.ts can use it without closing an
 // import cycle (process-registry already imports kill-process-tree). Re-exported
 // here so every existing caller keeps its import path.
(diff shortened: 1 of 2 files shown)


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3607 ([plan-19] Observer Subprocess Isolation)** as a **bundle candidate**: the fix is sound and is slated to be folded into the plan-19 bundle PR with author credit (`Co-authored-by` + `Closes` reference), sequenced by the step noted below. It stays open until that bundle lands.

> step 2 (observer cwd is `~/.claude-mem/observer-sessions/<id>`, never the user's project)

ousamabenyounes:
Rebased onto the latest `main` to clear the merge conflict in `src/supervisor/process-registry.ts` (the observer-session path helpers `OBSERVER_SESSIONS_DIR`/`ensureDir` had since landed on `main`, so this branch now carries only the supervisor cwd-jailing change plus its tests). Pushed as `e9db152`.

The fix is unchanged: each supervised SDK subprocess is jailed to `~/.claude-mem/observer-sessions/<sessionDbId>`, the directory is created before spawn, and `spawnSdkProcess` returns `null` on a creation failure.

Validation (RED → GREEN):
- RED: on unmodified `main` the focused probe fails — `normalizeSpawnSdkCwd` / the jailed cwd are absent.
- GREEN: `bun test tests/supervisor/process-registry.test.ts` → 27 pass / 0 fail.
- Full suite `bun test tests` → 3053 pass / 0 fail / 27 skip (no regression). The `typecheck` job's `install.ts` errors are pre-existing on `main` (unrelated `@clack/prompts` symbol types), identical before and after this branch.

Still happy for this to be folded into the plan-19 bundle (#3607) — this just keeps the branch mergeable in the meantime.

thedotmack:
**issue-patch-ship Gate: PASS** — rebasing/shipping via #4054 (this head is DIRTY/CONFLICTING; fork push is not available from this agent).

1. **RCA-later logged** — #3357: Windows `cmd.exe` treats `>` in transcript text as redirection; spawn cwd was the user project. This is the cwd-jail slice only. Deeper no-shell/argv stays under plan-19 / #3607.
2. **No second-system** — existing `spawnSdkProcess` + `OBSERVER_SESSIONS_DIR`; SDK callback `cwd` ignored at the trust boundary.
3. **Not opinionated** — 2 files, per-session observer dir, `null` on create failure.

Leaving this PR open until #4054 lands.

thedotmack:
Landed via squash-merge of the rebase PR #4054 (`91ad9dd`). Closing this DIRTY fork head as superseded.

thedotmack:
Superseded by same-repo rehost #4054 (merged 91ad9dd). Thanks!

PR #3433, opened 2026-07-28
feat(settings): surface OpenRouter base URL

### LINKED ISSUE #3188: Surface BYOK / custom OpenAI-compatible endpoint (CLAUDE_MEM_OPENROUTER_BASE_URL) in the CLI installer and Settings UI
## Before submitting

---

`CLAUDE_MEM_OPENROUTER_BASE_URL` already lets the OpenRouter provider talk to any OpenAI-compatible endpoint — self-hosted models, LM Studio, Ollama, DeepSeek, a custom gateway — per #2382/#2590/#2622/#2393. But the setting is invisible to anyone who doesn't read source: it's missing from `ContextSettingsModal.tsx`, missing from the settings-save whitelist in `SettingsRoutes.ts` (so it can't persist even if added to the UI), and the `npx claude-mem install` wizard's OpenRouter path only ever prompts for an API key. Today the only way to bring your own hosted model is to hand-edit `~/.claude-mem/settings.json` after install.

- Add a "Base URL" field to the OpenRouter section of the Settings UI, and whitelist `CLAUDE_MEM_OPENROUTER_BASE_URL` in the settings-save route.
- Extend the `npx claude-mem install` wizard: when picking the OpenRouter provider, ask openrouter.ai vs. a custom OpenAI-compatible endpoint, then prompt for base URL + model id + optional API key — mirroring the existing Claude "gateway" flow.
- Document it on the OpenRouter provider page with concrete examples (LM Studio, Ollama, DeepSeek, a self-hosted gateway).
- Relabel the provider option so BYOK is discoverable, e.g. "OpenRouter / Custom endpoint (BYOK)".

Leaving it settings.json-only — rejected, it's effectively undiscoverable and defeats the point of #2382 landing in the first place.

**Additional context**
I have a working patch (CLI wizard + Settings UI + settings-route whitelist/validation + docs), typechecked clean against both `tsc --noEmit` and the viewer's `tsconfig.json`. Will open a PR against `community-edge` per the branch strategy in `docs/public/branches.mdx`.

ousamabenyounes:
Fix #3188

## Summary
- Add `CLAUDE_MEM_OPENROUTER_BASE_URL` to the viewer settings defaults/types and OpenRouter settings panel.
- Allow the worker `/api/settings` save route to persist and validate the OpenRouter base URL (HTTP(S) only).
- Make the installer expose OpenRouter as a custom OpenAI-compatible BYOK path and prompt for custom endpoint base URL + model id.
- Resolve the endpoint by extending the URL's `pathname`, so a base URL carrying a query string or fragment keeps `/chat/completions` in the request path.

## Endpoint resolution (addresses the P1)

`resolveOpenRouterChatCompletionsUrl` previously concatenated onto the raw base string, so a base
such as `https://gw.example.com/v1?key=abc` produced
`https://gw.example.com/v1?key=abc/chat/completions` — the request pathname stayed `/v1` and the
required suffix ended up inside the query. Fragments failed the same way. The resolver now parses
the validated input, normalizes trailing slashes on `pathname`, and extends that pathname while
`search` and `hash` are preserved.

## RED
`bun test tests/shared/openrouter-base-url.test.ts` on the previous branch state:

```
(fail) keeps the chat-completions suffix in the path when the base URL has a query string
  Expected: "https://api.example.com/v1/chat/completions?key=abc"
  Received: "https://api.example.com/v1?key=abc/chat/completions"

(fail) keeps the chat-completions suffix in the path when the base URL has a fragment
  Expected: "https://api.example.com/v1/chat/completions#frag"
  Received: "https://api.example.com/v1#frag/chat/completions"

(fail) normalizes a trailing slash before the query string
  Expected: "https://api.example.com/v1/chat/completions?key=abc"
  Received: "https://api.example.com/v1/?key=abc/chat/completions"

(fail) preserves the query string on a base URL already ending in /chat/completions
  Expected: "https://api.example.com/v1/chat/completions?key=abc"
  Received: "https://api.example.com/v1/chat/completions?key=abc/chat/completions"

 12 pass
 4 fail
```

## GREEN
```
bun test tests/shared/openrouter-base-url.test.ts \
         tests/shared/openrouter-settings-surface.test.ts \
         tests/worker/settings-routes-validation.test.ts

 22 pass
 0 fail
```

Full suite on this branch: `2566 pass / 18 skip / 22 fail`, against `2554 pass / 18 skip / 22 fail`
on `main` at `e2d1df5` — the same 22 pre-existing `HealthMonitor` / `ProcessManager` failures, no
new ones. `npm run typecheck`, `npm run build`, `npm run smoke:clean-room`, the plugin
frozen-lockfile drift check, `npm run lint:hook-io`, `npm run lint:spawn-env` and
`npm run e2e:server:docker` all pass locally.

## Docs
`docs/public/cursor/openrouter-setup.mdx` — the base-URL resolution note now states the HTTP(S)
restriction and the query/fragment behavior.


ousamabenyounes:
Addressed the Greptile protocol validation gap.

Changes:
- shared the HTTP(S)-only URL guard between the installer prompt and SettingsRoutes
- reject unsupported schemes such as ftp:// before persistence/fetch
- added runtime coverage for SettingsRoutes plus resolver/helper coverage

Local validation:
- RED: ftp:// was accepted by SettingsRoutes before the fix
- GREEN: bun test tests/shared/openrouter-base-url.test.ts tests/worker/settings-routes-validation.test.ts tests/shared/openrouter-settings-surface.test.ts
- GREEN: npm run typecheck
- GREEN: npm run build

Note: a full local `bun test tests` run still has unrelated existing local-environment failures in process/health/mode boot tests; the targeted tests, typecheck, build, and GitHub CI are the relevant gates for this fix.

thedotmack:

> step 4 (installer/Settings UI expose the base URL); P1: query/fragment base URLs; installer file rewritten by #3524

ousamabenyounes:
Addressed in `e96723c`, against both items in the routing note.

**P1 (query/fragment base URLs):** `resolveOpenRouterChatCompletionsUrl` now parses the base URL and extends its `pathname` instead of concatenating onto the raw string, so `search` and `hash` are preserved and the suffix stays in the request path — `https://gw.example.com/v1?key=abc` resolves to `https://gw.example.com/v1/chat/completions?key=abc` rather than leaving the request on `/v1`. Covered for query, fragment, trailing-slash-before-query, and an already-suffixed base carrying a query string. One behavior note worth flagging since it is not visible in the diff: parsing returns the origin canonicalized (host lower-cased, default port dropped), which is semantics-preserving for the request but no longer byte-identical to the configured string; the path keeps its original case.

**Installer rewritten by #3524:** the branch is rebased onto current `main`, so the custom-endpoint prompt now sits on top of that rewrite — it runs inside `promptProvider` after the Pro-trial path returns, and applies cleanly there. I also updated the base-URL note in `docs/public/cursor/openrouter-setup.mdx` to match the new resolution.

Happy to fold this into the plan-18 bundle instead if you would rather take it that way.

thedotmack:
Routed to **#3606 (plan-18, Observer Response Pipeline)** as the OpenRouter / OpenAI-compatible endpoint-config slice.

Important heads-up: most of this has already landed on main in `d13fc437f` — `src/shared/openrouter-base-url.ts` exists there and `CLAUDE_MEM_OPENROUTER_BASE_URL` is wired through `SettingsDefaultsManager`, the installer and `OpenRouterProvider`. What is **not** on main is your actual bug fix: `resolveOpenRouterChatCompletionsUrl` still appends to the trimmed raw string, so a base URL carrying a query string or fragment still ends up with `/chat/completions` inside the query. The viewer panel is also still missing.

Leaving this open, but please rebase it down to that resolver delta plus the viewer surface — the rest is now duplicate of merged code.

PR #3440, opened 2026-07-29
fix(hooks): bound session init prompt timeout

### LINKED ISSUE #3434: UserPromptSubmit session-init hook blocks up to 60s under worker saturation — add fail-fast + configurable timeout
## Summary

The `UserPromptSubmit` → `session-init` hook is the only synchronous hook (correctly so — it injects memory context into the prompt), but it ships with `timeout: 60` in `hooks/hooks.json`. When the worker is alive but momentarily saturated, every prompt blocks the full 60 s before proceeding.

## Environment

- claude-mem 13.12.4, Windows 11, `CLAUDE_MEM_RUNTIME=worker`, managed worker on 127.0.0.1:37777
- `CLAUDE_MEM_MAX_CONCURRENT_AGENTS=2`, provider `claude`

## Evidence

- One session logged **16/16 `UserPromptSubmit` hook timeouts at ~60310 ms** (all hit the ceiling).
- Worker was **alive** the whole time (single PID, ~4 h uptime), `state/hook-failures.json` → `consecutiveFailures:0`. So not a dead/cold worker.
- Warm-path measurement of the exact `worker-service.cjs hook claude-code session-init` call: **~1.6 s**.
- Trigger correlated with a `/compact` shortly before, i.e. heavy concurrent summarization saturating the worker event loop; interactive `session-init` requests queued behind it until the 60 s cap.

## Impact

Normal path is ~1.6 s, so the 60 s budget is ~37× oversized. Any transient saturation (post-`/compact`, batch summarization) freezes the user's prompt for a full minute per turn.

## Proposed fix

1. Lower the default `session-init` hook timeout to ~10–15 s (still ~9× the warm path). On timeout, skip that turn's injection gracefully instead of blocking.
2. Make it configurable, e.g. `CLAUDE_MEM_SESSION_INIT_TIMEOUT_MS`, so users can tune per machine.
3. Optionally: shed/deprioritise summarization load while an interactive `session-init` is in flight so it isn't starved.

## Workaround

Set `"timeout": 15` on the `UserPromptSubmit` hook in `hooks/hooks.json` (lost on plugin update).

ousamabenyounes:
## Summary

plan-17 step 3 — keep the UserPromptSubmit session-init budget strictly inside the wire timeout.

- lower the UserPromptSubmit hook timeout from 60s to 15s in the Claude and Codex manifests (generated from `scripts/build-hooks.js`)
- add `CLAUDE_MEM_SESSION_INIT_TIMEOUT_MS`, a settings/env-backed request budget that is never multiplied to the wire timeout: it defaults to 12s unscaled, and an override above 14s is rejected back to the default so it cannot outlive the 15s host cap
- spend that budget once per hook invocation: the server-runtime request gets half of it, and the worker fallback, `/api/context/semantic`, the readiness wait and the port wait each get only what is left rather than a fresh full budget
- bound the version probe: `checkVersionMatch` / `getRunningWorkerVersion` take a probe timeout and pass an `AbortSignal` down to the `/api/health` read, and the hook path hands them its remaining budget
- follow-up: if the version probe consumes the whole remaining budget, the matching-version path now returns the graceful worker fallback instead of starting a fresh readiness probe with the default health timeout
- rebase fix: preserve `main`'s simple Codex `SessionStart` context hook and Codex-specific `{ workerStartupTimeoutMs: 15000, timeoutMs: 2000 }` session-init worker options while keeping this PR's deadline accounting

## Issues

Fix #3434 · tracked under #3605 (plan-17)

## Review follow-up

The P1s raised on this PR are addressed:

- **unbounded version check** — after the bounded liveness probe, `ensureWorkerRunningWithinBudget` awaited `checkVersionMatch`, whose `/api/health` read went through a plain `fetch()` with no `AbortSignal`. A worker that answered the first probe and then stalled on the version request held session-init open past the host deadline. The probe is now bounded by the remaining budget (and by the liveness timeout for callers outside a budget); an expired probe reads as an unknown version, which is the existing no-recycle fallback. The same bound is applied to the post-readiness amplifier check.
- **second full budget on fallback** — the budget clock starts before the server request, so the worker fallback receives `budget - elapsed`; below a 500 ms floor it skips straight to the graceful result.
- **oversize env value** — `CLAUDE_MEM_SESSION_INIT_TIMEOUT_MS` is clamped to `[500, 14000]`; anything outside that logs a warning and falls back to the 12s default, so no override can push the request past the 15s manifest timeout.
- **readiness after budget exhaustion** — the matching-version path no longer turns a zero remaining budget into the default readiness health timeout.

RED — the stalled-version scenario on the branch before this fix, `bun test tests/infrastructure/health-monitor-version-timeout.test.ts`:

```
(fail) worker version probe deadline (3434) > gives up on a stalled /api/health instead of outliving the caller budget [5001.67ms]
  ^ this test timed out after 5000ms.
(fail) worker version probe deadline (3434) > returns null from the version read once the probe budget expires [5001.52ms]
  ^ this test timed out after 5000ms.
 1 pass
 2 fail
```

RED — the budget wiring, with the remaining-budget argument removed from the `checkVersionMatch` call in `ensureWorkerRunningWithinBudget`, `bun test tests/shared/worker-utils-version-recycle.test.ts`:

```
206 |     expect(versionMatchCalls.length).toBe(1);
207 |     expect(versionMatchCalls[0].timeoutMs).toBeLessThanOrEqual(SESSION_INIT_BUDGET_MS);
(fail) ensureWorkerRunning — stale-worker recycle on version mismatch > bounds the version probe by the hook budget, not the standalone health timeout (3434)
 4 pass
 1 fail
```

RED — the review follow-up for readiness after a consumed version-probe budget, `bun test tests/shared/worker-utils-version-recycle.test.ts` before the latest fix:

```
228 |     expect(readinessCalls.length).toBe(0);
(fail) ensureWorkerRunning — stale-worker recycle on version mismatch > does not start readiness when the matching-version probe exhausts the hook budget (3434)
Expected: 0
Received: 1
 5 pass
 1 fail
```

RED — rebase regression caught before the follow-up commit, `bun test tests/cli/handlers/session-init-semantic-platform-source.test.ts tests/cli/handlers/session-init-server-beta-context.test.ts tests/cli/handlers/context-session-start.test.ts`:

```
error: Codex hook options mismatch: ... "options":{"timeoutMs":12000}
(fail) sessionInitHandler semantic injection platform source > includes normalized platformSource in semantic context request payload
 3 pass
 1 fail
```

GREEN — focused follow-up suite after the fixes:

```
bun test tests/shared/worker-utils-version-recycle.test.ts \
         tests/infrastructure/plugin-distribution.test.ts \
         tests/shared/worker-utils-timeout.test.ts \
         tests/cli/handlers/session-init-timeout.test.ts \
         tests/cli/handlers/session-init-semantic-platform-source.test.ts \
         tests/cli/handlers/context-session-start.test.ts \
         tests/cli/handlers/session-init-server-beta-context.test.ts

 78 pass
 0 fail
```

## Local CI replay

Rebased onto `origin/main` `1627ded4`.

- `npm install --no-audit --no-fund`: pass
- `npm run typecheck`: pass
- `npm run build`: pass
- `bun test tests`: iso-baseline with `origin/main`; both report `FAILED_STEPS=bun-test-tests:1` from the same known local `HealthMonitor` / `ProcessManager` failures
- `bun test openclaw`: pass, 44 tests
- `npm run smoke:clean-room`: pass
- sync-hub vitest + WS suite: pass, 113 + 9 tests
- `npm run e2e:server:docker`: pass, including restart/persistence and local-dev auth rejection guards

Coverage note: `bun test --coverage` covers the changed production guard (`src/shared/worker-utils.ts:399` in LCOV). The generic `diff-coverage-gate.py` currently treats changed `.test.ts` lines as production and therefore reports a false failure for the test file, which Bun intentionally omits from LCOV.


thedotmack:

> step 3 (`CLAUDE_MEM_SESSION_INIT_TIMEOUT_MS`, budget never multiplied to the wire timeout); P1s: unbounded version check, second full budget on fallback, oversize env value

ousamabenyounes:
Reworked in `9995bf8` against the plan-17 step-3 design, rebased onto current `main`.

The three P1s are handled. The unbounded one was real: after the bounded liveness probe, the version check went through a plain `fetch()` with no abort signal, so a worker that answered the first probe and then stalled on the version request held session-init past the 15s cap. It now inherits the hook's remaining budget, and an expired probe reads as an unknown version — the existing no-recycle fallback. The worker fallback gets `budget - elapsed` instead of a fresh budget, and `CLAUDE_MEM_SESSION_INIT_TIMEOUT_MS` is clamped so no override can outlive the manifest timeout.

The red check was a test this branch had broken itself: the session-init tests stub `loadFromFileOnce` but not the new timeout reader, so the real one seeded `settings.json` and its creation notice tripped their empty-stderr assertion. Fixed here; checks are re-running on the new head.

Happy to fold this into the plan-17 bundle instead if that's cleaner on your side.

ousamabenyounes:
Thanks for the review. Addressed the remaining session-init budget finding in `0828fb32`: when the version probe consumes the hook budget, the matching-version path now returns the graceful fallback instead of starting readiness with the default timeout.

Also fixed a rebase regression against current `main` so Codex keeps the simple `SessionStart` context hook and its Codex-specific session-init worker options.

Validation:
- RED: new `worker-utils-version-recycle` exhausted-budget case failed with `readinessCalls.length === 1`; GREEN after the fix, 6/6 in that file.
- Focused follow-up suite: 78 pass / 0 fail.
- Local CI replay is iso-baseline with `origin/main` (`FAILED_STEPS=bun-test-tests:1` on both, same known local HealthMonitor/ProcessManager failures); typecheck, build, OpenClaw, clean-room smoke, and sync-hub suites pass.
- Docker server-runtime e2e passed, including restart/persistence and local-dev auth rejection guards.

thedotmack:
Still routed to **#3605 (plan-17, Hook Wrapper Contract)** — a session-init request budget that can never outlive the 15s wire timeout is the "hook timeouts and budget" line of that master, and the `AbortSignal` you thread down into the version probe is the part plan-17 most needs.

Nothing else open bounds the UserPromptSubmit budget, so this is not a duplicate. It is now CONFLICTING against main: both hook manifests and `worker-utils` have moved since this was written, so it needs a regenerate-and-rebase before it can be sequenced into the plan-17 work.

PR #3441, opened 2026-07-29
fix(install): ship Claude marketplace manifest

### LINKED ISSUE #3424: Fresh install broken: marketplace cache-miss on load, then subscription-auth fails for worker compression spawn
# Two install/runtime bugs: missing marketplace manifest, and subscription-auth failure in worker compression spawn

**Version:** claude-mem 13.12.4
**Claude CLI:** v2.1.207 → auto-updated to v2.1.220 during testing
**OS:** macOS (Darwin), Apple Silicon
**Install method:** `npx claude-mem install`, IDE = Claude Code only, runtime = Worker, provider = Claude Agent SDK, auth = Subscription plan, model = Opus 4.8

## Bug 1: Plugin fails to load after a clean install — "cache-miss"

After a normal `npx claude-mem install` (no errors, "Installation Complete" banner shown), `claude plugin list` reports the plugin as failed:

```
❯ claude-mem@thedotmack
  Status: ✘ failed to load
  Error: Marketplace thedotmack failed to load: cache-miss
```

Root cause (confirmed via `claude plugin validate ~/.claude/plugins/marketplaces/thedotmack`):

```
✘ directory: No manifest found in directory. Expected .claude-plugin/marketplace.json or .claude-plugin/plugin.json
```

The installed marketplace repo (`~/.claude/plugins/marketplaces/thedotmack/`) has no `.claude-plugin/marketplace.json` at its root. The actual plugin manifest is nested one level down at `plugin/.claude-plugin/plugin.json`, which validates fine on its own — but Claude Code's marketplace loader looks at the repo root and finds nothing, so the plugin never loads despite `known_marketplaces.json` and `settings.json` both being written correctly by the installer.

**Workaround used:** manually created a `.claude-plugin/marketplace.json` in a new local marketplace directory that sources `./plugin`, and registered that instead via `claude plugin marketplace add`. This gets the plugin to `Status: ✔ enabled`.

**Note:** the worker's own logs also reference a hardcoded path check independent of whatever marketplace name Claude Code actually has it registered under:

```
[ERROR] [SYSTEM] claude-mem MCP started but no marketplace directory was found at
~/.claude/plugins/marketplaces/thedotmack or the XDG equivalent. ... Without it,
no new memories will be captured. To self-heal, run:
node ~/.claude/plugins/cache/thedotmack/claude-mem/*/scripts/smart-install.js
```

That suggested self-heal script (`scripts/smart-install.js`) does not exist anywhere under `~/.claude/plugins/cache/thedotmack/claude-mem/*/` or `~/.claude/plugins/marketplaces/thedotmack/` in this install — the remediation message points at a file that isn't shipped.

## Bug 2: Compression never succeeds — "OAuth session expired and could not be refreshed", using Subscription plan auth

Once the plugin loads (via the workaround above), `SessionStart`/`UserPromptSubmit` hooks fire correctly and sessions/prompts are written to `~/.claude-mem/claude-mem.db` (`sdk_sessions`, `user_prompts` tables populate as expected). However, `observations` and `session_summaries` remain empty in every attempt — the compression step never succeeds.

Every attempt (3 separate sessions, across two different shells, one right after running `/login` to refresh credentials) fails identically:

```
[SDK] Starting SDK query {authMethod=Claude Code OAuth token (read from system keychain at spawn)}
[SDK] ← Response received (72 chars) {promptNumber=N} Failed to authenticate: OAuth session expired and could not be refreshed
[ERROR] [PARSER] SDK authentication failed; run /login to preserve queued batch
[WARN] [SDK_SPAWN] Claude process exited {code=1, signal=null}
```

Key detail: the *interactive* `claude` session (the one the user is actually typing into) authenticates and responds normally the whole time. The failure is specific to the **separate background `claude` subprocess the worker spawns** for compression, which reads a token from the system keychain "at spawn." This failed:

- Immediately after a fresh install
- After a full quit/relaunch of Claude Code
- After running `/login` inside an interactive terminal `claude` session and confirming success
- On both Claude CLI v2.1.207 and v2.1.220 (auto-updated mid-testing)

Since re-authenticating the interactive session has no effect on this background spawn's ability to authenticate, this looks like a bug in how the worker acquires/passes credentials to its child process for the "Subscription plan" auth method — not an expired-credentials issue on the user's end.

## Impact

With both issues combined, a clean `npx claude-mem install` on Claude Code (marketplace-installed IDE) using the default recommended options (Worker runtime, Claude Agent SDK provider, Subscription plan auth) does not produce any working memory capture out of the box: the plugin doesn't load without a manual marketplace-manifest fix, and even after that fix, no memory content is ever generated due to the background auth failure.

## Suggested next steps for maintainers

1. Add `.claude-plugin/marketplace.json` at the repo root (or otherwise fix the installer to register the marketplace pointing at `plugin/`).
2. Either ship `scripts/smart-install.js` or remove/fix the self-heal remediation message that references it.
3. Investigate why the worker's spawned child `claude` process fails to read a valid OAuth token from the system keychain when `CLAUDE_MEM_CLAUDE_AUTH_METHOD=subscription`, independent of the interactive session's own auth state.

Happy to provide full logs / `~/.claude-mem/claude-mem.db` schema dumps if useful.

ousamabenyounes:
Fix #3424

## Summary
- include the root `.claude-plugin` marketplace manifest in the npm package
- copy `.claude-plugin` into the durable marketplace install root
- cover the install/package allowlists so clean Claude Code installs do not cache-miss on marketplace load

## Tests
- RED: `bun test tests/install-non-tty.test.ts --timeout 30000` failed when the two implementation lines were temporarily reverted
- GREEN: `./run-ci.sh`

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "files": [
     "dist",
     ".agents/plugins/marketplace.json",
+    ".claude-plugin",
     ".codex-plugin",
     "plugin/.claude-plugin",
     "plugin/.codex-plugin",
--- a/src/npx-cli/commands/install.ts
+++ b/src/npx-cli/commands/install.ts
@@ -663,6 +663,7 @@ function copyPluginToMarketplace(): void {
 
   const allowedTopLevelEntries = [
     '.agents',
+    '.claude-plugin',
     '.codex-plugin',
     'plugin',
     'package.json',
--- a/tests/infrastructure/plugin-distribution.test.ts
+++ b/tests/infrastructure/plugin-distribution.test.ts
@@ -245,7 +245,7 @@ describe('Plugin Distribution - package.json Files Field', () => {
     expect(packageJson.files).toContain('plugin/sqlite');
   });
 
-  it('npm tarball includes sqlite runtime modules required by the worker', () => {
+  it('npm tarball includes runtime and Claude marketplace root manifests required by clean installs (#3424)', () => {
     const result = spawnSync('npm', ['pack', '--dry-run', '--json'], {
       cwd: projectRoot,
       encoding: 'utf-8',
@@ -257,6 +257,8 @@ describe('Plugin Distribution - package.json Files Field', () => {
 
     expect(filePaths.has('plugin/sqlite/SessionStore.js')).toBe(true);
     expect(filePaths.has('plugin/sqlite/observations/files.js')).toBe(true);
+    expect(filePaths.has('.claude-plugin/marketplace.json')).toBe(true);
+    expect(filePaths.has('.claude-plugin/plugin.json')).toBe(true);
   });
 });
 
--- a/tests/install-non-tty.test.ts
+++ b/tests/install-non-tty.test.ts
@@ -119,10 +119,17 @@ describe('Install Non-TTY Support', () => {
       );
       expect(copyRegion).toContain("'.agents'");
       expect(copyRegion).toContain("'.codex-plugin'");
+      expect(copyRegion).toContain("'.claude-plugin'");
       // Root .mcp.json was dropped in #2411; the MCP manifest now ships
       // exclusively as plugin/.mcp.json (bundled inside the 'plugin' entry).
       expect(copyRegion).toContain("'plugin'");
       expect(copyRegion).not.toContain("'.mcp.json'");
+    });
+
+    it('publishes the Claude marketplace root manifest in the npm package (#3424)', () => {
+      const packageJson = JSON.parse(readFileSync(join(__dirname, '..', 'package.json'), 'utf-8'));
+      expect(packageJson.files).toContain('.claude-plugin');
+      expect(packageJson.files).toContain('plugin/.claude-plugin');
     });
 
     it('validates the bundled plugin as the Codex marketplace source', () => {


ousamabenyounes:
Addressed Greptile's packaging-test gap in 3297bf5e by extending the existing npm pack dry-run test to assert .claude-plugin/marketplace.json and .claude-plugin/plugin.json are actually present in the generated tarball. Local gate: ./run-ci.sh. CI is green.

thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3604 ([plan-16] Canonical Install Identity & Bundle Integrity)** as a **bundle candidate**: the fix is sound and is slated to be folded into the plan-16 bundle PR with author credit (`Co-authored-by` + `Closes` reference), sequenced by the step noted below. It stays open until that bundle lands.

> step 5 (marketplace registration writes `.claude-plugin/marketplace.json` at the discovered root; ship it in the tarball)

thedotmack:
Routing for this pass: **#3604 (plan-16, Canonical Install Identity & Bundle Integrity)** — shipping the root `.claude-plugin/marketplace.json` in the tarball and copying it into the durable marketplace root is exactly the install-identity defect that master owns.

Re-reading the open queue, #3658 is the same fix on the same two seams (the `package.json` `files` list and the `copyPluginToMarketplace()` allowlist) and goes further by teaching `isPluginInstalled()` to assert the root manifest, so install/repair/doctor stop reporting a broken install as healthy. Closing this as a duplicate of #3658; your packaging assertion in the npm-pack dry run is the better test of the two and plan-16 should carry it with author credit. Thanks for getting there first.

PR #3452, opened 2026-07-30
fix(sdk): isolate observer session transcripts

### LINKED ISSUE #3417: Observer transcripts accumulate in ~/.claude/projects/ and flood the Claude Desktop session list
## Summary

Observer-agent transcripts accumulate indefinitely in `~/.claude/projects/`, where Claude Code Desktop indexes them as if they were real conversations. On my machine 452 of them (and counting) buried ~66 genuine sessions in the Desktop "Recents" sidebar.

They are normally invisible, so this only bites once something imports them — for me, Claude Code Desktop's **"import hidden Claude CLI sessions"** option, which appears when a session's content pane fails to populate. After clicking it, **443 of my first 500 listed sessions were observer transcripts**.

I've since cleaned it up manually (details below), but the cleanup is unpleasant enough that I think it's worth reporting — and the root cause is on the write side, which is claude-mem's.

## Where they land

```
~/.claude/projects/C--Users-<user>--claude-mem-observer-sessions/*.jsonl
```

One JSONL per observed session. Titles come from the first line of the prompt, so the sidebar fills with repeats of:

- `Hello memory agent, you are continuing to observe the primary Claude session. <…`
- `You are a Claude-Mem, a specialized observer tool for creating searchable memor…`

`~/.claude-mem/observer-sessions/` (the cwd the observer runs in) is itself empty — the transcripts are Claude Code's own session logs for those subagent runs, landing in `projects/` because that's the cwd the CLI was invoked with.

## Diagnosis — two findings that make this hard to fix from the user side

**1. The Desktop session list reads a cached index, not the filesystem.**

I tested this directly:

| Test | Result |
|---|---|
| Move one transcript into a subfolder | Still listed, counts unchanged |
| Move it **entirely out** of `~/.claude/projects/` | **Still listed**, counts unchanged |

With the file physically absent from the tree, the sidebar still showed the session. So deleting or relocating transcripts does not remove them from the list — the obvious user-side remedy simply doesn't work. (I restored the file afterwards; nothing was lost.)

**2. Bulk archiving is hard-gated.**

The Desktop archive tool states: *"This tool requires explicit approval regardless of permission mode."* Every session needs an individual manual confirmation — no permission setting or allowlist suppresses it. At 443 sessions that's 443 separate clicks.

I also could not find any documented way to exclude a directory from the session list, relocate observer transcripts, or undo the import.

## Manual workaround (what actually worked)

Posting this in case someone else hits the same wall — but it is a slog, not a fix:

1. **`Group by → Project`** in the sidebar filter. This buckets all observer sessions under a single `claude-mem-observer-sessions` group, out of the way. Zero risk, instant, and honestly may be enough for most people.
2. **Archive them.** Filter strictly on `cwd == ~/.claude-mem/observer-sessions` so real sessions are untouched, then archive each one. Archiving is reversible (they move to the Archived list, not deleted). I cleared 443 this way.
3. **Do not bother deleting the transcript files** — per finding 1, it has no effect on the list.

A note for anyone automating step 2: batching many archive calls into a single request is dramatically faster than issuing them one at a time, because each separate request costs an inference round-trip while the approval dialogs themselves queue instantly.

## Suggested directions

1. **Write observer transcripts outside `~/.claude/projects/`** — e.g. under `~/.claude-mem/` — so no client indexes them as conversations. This seems like the cleanest fix and is entirely within claude-mem's control.
2. **Or mark them** with metadata clients can filter on.
3. **Or document a supported cleanup path**, given deletion doesn't work.

I recognise this is partly a Claude Code Desktop concern — the import feature is what surfaces them, and the cached index is what makes them un-deletable. Happy to file there too if you'd rather. But the write location is claude-mem's, so it looks like the more tractable end of the problem.

## Environment

- claude-mem **13.11.0** (cache also holds 13.10.2 and 13.6.0)
- Latest release at time of writing is **v13.12.4** — I have **not** retested on latest, so this may already be addressed
- Claude Code Desktop on Windows 11
- 452 observer transcripts accumulated, newest written minutes before filing

Alan5168:
## Summary

Store Observer and KnowledgeAgent SDK session transcripts under claude-mem's own data directory instead of the user's Claude config directory.

The Agent SDK persists resume-capable sessions under `CLAUDE_CONFIG_DIR/projects`. The observer already uses an isolated `cwd` (`OBSERVER_SESSIONS_DIR`), but without an isolated config directory those transcripts still land in `~/.claude/projects` and get indexed by Claude Desktop as user sessions.

This sets a dedicated internal `CLAUDE_CONFIG_DIR` (`<DATA_DIR>/observer-claude-config`) in the shared hardened SDK options. `resume` is still passed through unchanged, so resume behavior is preserved — this only isolates where the transcript lives.

## Tests

- Added a regression test in `tests/security/observer-tool-enforcement.test.ts` that is red on current `main` and green with the fix (verified locally by reverting just the production fix and re-running).
- `bun test tests/security`: 14 pass / 0 fail.
- `bun run typecheck:root`: clean.
- Full suite: 2539 pass / 18 skip / 1 pre-existing failure (unrelated timeout in `tests/worker/sync/mutation-sites.test.ts`, reproduces identically on clean `main`).

Closes #3417

### DIFF
--- a/src/sdk/hardened-options.ts
+++ b/src/sdk/hardened-options.ts
@@ -38,7 +38,10 @@
  */
 
 import type { Options } from '@anthropic-ai/claude-agent-sdk';
-import { OBSERVER_SESSIONS_DIR } from '../shared/paths.js';
+import {
+  OBSERVER_CLAUDE_CONFIG_DIR,
+  OBSERVER_SESSIONS_DIR,
+} from '../shared/paths.js';
 import { recordObserverToolAttempt } from '../utils/observer-audit.js';
 import { logger } from '../utils/logger.js';
 
@@ -115,7 +118,9 @@ export function buildHardenedSdkOptions(input: HardenedSdkOptionsInput): Options
   return {
     model: input.model,
     cwd: input.cwd ?? OBSERVER_SESSIONS_DIR,
-    env: input.env,
+    // Keep resume-capable SDK transcripts, but persist them under claude-mem's
+    // own data directory so Claude Desktop does not index them as user sessions.
+    env: { ...input.env, CLAUDE_CONFIG_DIR: OBSERVER_CLAUDE_CONFIG_DIR },
     pathToClaudeCodeExecutable: input.pathToClaudeCodeExecutable,
     ...(input.abortController ? { abortController: input.abortController } : {}),
     ...(input.resume ? { resume: input.resume } : {}),
--- a/src/shared/paths.ts
+++ b/src/shared/paths.ts
@@ -67,6 +67,7 @@ export const USER_SETTINGS_PATH = join(DATA_DIR, 'settings.json');
 export const DB_PATH = join(DATA_DIR, 'claude-mem.db');
 
 export const OBSERVER_SESSIONS_DIR = join(DATA_DIR, 'observer-sessions');
+export const OBSERVER_CLAUDE_CONFIG_DIR = join(DATA_DIR, 'observer-claude-config');
 
 export const OBSERVER_SESSIONS_PROJECT = basename(OBSERVER_SESSIONS_DIR);
 
--- a/tests/security/observer-tool-enforcement.test.ts
+++ b/tests/security/observer-tool-enforcement.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, beforeEach, afterEach } from 'bun:test';
 import { readFileSync, existsSync, rmSync } from 'fs';
+import { join } from 'path';
 import {
   buildHardenedSdkOptions,
   OBSERVER_DISALLOWED_TOOLS,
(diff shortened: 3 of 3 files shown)


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3607 ([plan-19] Observer Subprocess Isolation)** as **standalone / merge-now**: it is small and self-contained enough to land on its own ahead of the plan-19 bundle once CI is green.

> step 1 (isolated `CLAUDE_CONFIG_DIR` so observer transcripts never land in `~/.claude/projects`)

Alan5168:
Closing as hygiene: no activity since 2026-08-17. Happy to reopen/rebase if you still want this isolation fix.

PR #3458, opened 2026-07-31
Session ID Display, Filter, and Delete

Krs7N:
image
image


thedotmack:

> viewer session navigation + backup/restore grab-bag; P1s: cascade delete without tombstones, platform_source dropped

thedotmack:
Routing to **#2785 (plan-12 Provider & Extensibility Roadmap)** — session-ID display, filtering and deletion in the viewer is a net-new capability, and plan-12 is where viewer/UX features live rather than any of the defect clusters. Fair warning on size: +6.3k / −877 across the viewer, `SessionStore`, pagination and new backup/data routes, with a session-delete path that a user would notice, so this wants review attention independent of the cluster pass. Not a duplicate; leaving open.

Issue #3454, opened 2026-07-30
Observer discards 63% of turns — 170/day empty (idle) responses, plus dead consecutiveInvalidOutputs counter


danemil:
> **Correction (author):** my original report attributed the dominant failure to conversational prose being misclassified as `idle`. That was wrong — after reading `src/sdk/output-classifier.ts` the two classes are distinct and logged separately, and the counts are very different. Corrected body below; the substance of defects 2 and 3 is unchanged.

## Summary

The observer discards the large majority of its turns. On one day of ordinary use, **192 of 304 turns (63%) were dropped**, split across two distinct classes:

| `outputClass` | Count | Meaning | `preview=` |
|---|---|---|---|
| `idle` | **170** | model returned an empty / whitespace-only string | empty |
| `prose` | 22 | model returned conversational text explaining it had nothing to record | populated |
| — (`STORED`) | 112 | parsed and persisted successfully | — |

The **`idle` case dominates and is the real problem**: the observer is returning nothing at all, 170 times in a day. `<observation>` and `<summary>` parse correctly and account for all 112 stored batches, so the tag grammar itself is sound.

## Environment

| | |
|---|---|
| claude-mem | 13.12.4 |
| Host | Claude Code 2.1.220, npm global |
| OS | macOS (darwin 25.5.0), Apple Silicon |
| Provider | Claude via SDK |

## Log signature

```
[WARN] [PARSER] [session-97] observer returned non-XML idle response — ignoring queued batch
       {outputClass=idle, preview=, consecutiveInvalidOutputs=0}
```

Per `classifyObserverOutput`, `idle` is returned when `typeof raw !== 'string' || raw.trim() === ''`. So these 170 turns produced genuinely empty output — not prose, not a malformed tag.

## Defect 1 — 170 empty responses per day

The dominant failure. Because `previewOutput('')` returns `''`, the log line carries no diagnostic payload at all, so it is not possible from logs alone to tell whether the SDK returned an empty completion, the response was dropped in transport, or the generator yielded nothing.

Suggestion: distinguish "SDK returned empty string" from "no SDK response received" at the call site, and log which. They are different faults with different fixes.

## Defect 2 — no no-op sentinel for observations

For the 22 `prose` turns, the model is behaving sensibly — it has been handed a turn with nothing observable and says so:

```
I don't have enough information to create an observation. The primary session shows only a user request…
I cannot generate an observation from this input as there are no tool executions…
I'm observing the primary session, but there's no work to record…
[...]

PR #3460, opened 2026-07-31
fix(observer): preserve queued batch when a transport failure is returned as text

danemil:
Fixes one of the defects reported in #3454.

## Problem

A dropped or truncated connection can surface as an assistant **message** rather than a structured SDK error:

```
API Error: Connection closed mid-response. The response above may be incomplete.
```

That text reaches `classifyObserverOutput()`, is classified as ordinary `prose`, and the claimed batch is confirmed away by `confirmClaimedMessages()`. Confirming is correct for genuine no-op prose — the observer looked and found nothing worth recording — but a truncated response means the batch was **never actually considered**. The observations are lost silently.

Observed on claude-mem 13.12.4 in normal use.

## Change

Adds `isTransportErrorObserverOutput()` to `src/sdk/output-classifier.ts`, deliberately mirroring the existing `isQuotaLimitedObserverOutput` and `isAuthFailureObserverOutput` detectors — same guard clauses, same XML bail-out, same whitespace/case normalisation, same regex-alternation shape.

`ResponseProcessor` checks it after the auth case and routes matches to `resetProcessingToPending()` so the batch is retried next round rather than dropped.

### Design note

Quota and auth both preserve the batch **and** abort the generator, because those conditions persist until the user intervenes. A truncated connection is transient, so this path preserves the batch but **leaves the generator running**.

That is a deliberate divergence from the two neighbouring handlers. Happy to align it with them instead if you'd prefer consistency — it's a one-line change.

The XML guard is what stops an observation *about* a dropped connection (`<observation><title>Connection closed mid-response in the retry path</title></observation>`) from being mistaken for an actual transport failure.

## Tests

12 new cases in `tests/sdk/output-classifier.test.ts`: the exact string seen in the wild, four other transport phrasings, case/whitespace insensitivity, empty and non-string input, XML containing transport wording, and negatives against ordinary observer prose plus quota/auth text so the detectors don't overlap.

| | |
|---|---|
| `tsc --noEmit` | clean |
| `bun test tests/sdk` | 60 pass / 0 fail |
| `bun test tests/worker` | 426 pass / 0 fail (3 consecutive runs) |
| `lint:hook-io` | OK |
| `lint:spawn-env` | OK |

Purely additive — 98 insertions, 0 deletions, no existing behaviour modified.

## Scope

This addresses only the transport-error defect from #3454. The primary finding in that issue — ~170 empty (`idle`) observer responses per day, a 63% discard rate — is **not** addressed here; its cause is upstream of this file.

### DIFF
--- a/src/sdk/output-classifier.ts
+++ b/src/sdk/output-classifier.ts
@@ -72,6 +72,34 @@ export function isQuotaLimitedObserverOutput(raw: unknown): boolean {
 }
 
 /**
+ * Detect transport-level failure text surfaced as an assistant message rather
+ * than a structured SDK error — e.g. a connection dropped part-way through a
+ * response. Unlike ordinary observer prose, this represents *unfinished* work:
+ * the batch was never actually considered, so confirming it silently loses
+ * observations. Callers should preserve the claimed batch for retry instead.
+ */
+export function isTransportErrorObserverOutput(raw: unknown): boolean {
+  if (typeof raw !== 'string' || raw.trim() === '') {
+    return false;
+  }
+
+  if (/<(observation|summary)\b/i.test(raw) || /<skip_summary\b/i.test(raw)) {
+    return false;
+  }
+
+  const text = raw.toLowerCase().replace(/\s+/g, ' ').trim();
+
+  return (
+    /\bconnection closed\b.{0,40}\bmid-?response\b/.test(text) ||
+    /\bresponse\b.{0,40}\bmay be incomplete\b/.test(text) ||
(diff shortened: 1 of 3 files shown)


thedotmack:

> step 1 (`transport` preserves the batch); already bundled in #3614; P2: anchor keywords

thedotmack:
Routing to **#3606 (plan-18 Observer Response Pipeline)** — "Non-XML replies silently confirming batches" is the first line of that master and this is the cleanest single instance of it: `API Error: Connection closed mid-response` arrives as an assistant message, classifies as ordinary prose, and the claimed batch is confirmed away. Keeping it open: it is already bundled into #3614, which deliberately leaves its source PRs open until the bundle lands. Related but not duplicated by #3467 (empty turn) and #3475 (malformed closed XML).

PR #3462, opened 2026-07-31
fix(chroma): prevent concurrent write storms

ChenglinWei97:
## Summary

- coalesce concurrent collection creation into one Chroma call
- serialize and bound local Chroma mutations so timed-out imports cannot continue as thousands of concurrent writes
- keep read-only queries and remote Chroma mutation concurrency unchanged
- cancel queued or retrying mutations when worker shutdown begins

## Root cause

Historical imports could start many fire-and-forget Chroma writes at once. Callers timed out while the underlying writes remained active, so retries added more work and amplified local Chroma disk growth.

This change applies backpressure at the existing Chroma MCP boundary. The 5,000-call default preserves the observed import burst while preventing an unbounded queue; rejected writes remain unsynced for the existing backfill path.

## Scope

This intentionally does not add a filesystem watchdog, directory scanner, hard disk cap, or new process-killing circuit. Normal search remains concurrent, remote Chroma behavior is unchanged, and only local mutations are serialized.

## Validation

- `npm run typecheck`
- `npm run lint:hook-io`
- `npm run lint:spawn-env`
- `npm run build`
- `npm test` — 2,543 passed, 18 environment-gated skips, 0 failed
- focused Chroma regression tests — 23 passed, 0 failed

### DIFF
--- a/src/services/sync/ChromaMcpManager.ts
+++ b/src/services/sync/ChromaMcpManager.ts
@@ -29,6 +29,8 @@ const RECONNECT_BACKOFF_MS = 10_000;
 const CHROMA_WRITER_LOCK_FILENAME = '.claude-mem-chroma-writer.lock';
 const CHROMA_SUPERVISOR_ID = 'chroma-mcp';
 const CHROMA_OUTPUT_TAIL_MAX_CHARS = 2048;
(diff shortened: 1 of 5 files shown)


ChenglinWei97:
@greptileai please re-review the latest commit 7040b94c.

PR #3463, opened 2026-07-31
fix: accept camelCase dateStart/dateEnd in build_corpus and warn on empty corpus

dmeneghel82:
## Summary

The `build_corpus` MCP tool declares `dateStart`/`dateEnd` (camelCase) in its input schema (`src/servers/mcp-server.ts`), but the worker corpus route only reads `date_start`/`date_end` (snake_case). Since `buildCorpusSchema` uses `.passthrough()`, a camelCase date filter passes validation and is then silently dropped by the destructuring — the corpus is built with no date filter, returns `observation_count: 0`, and the response looks like a clean success.

Measured reproduction (2026-07-31, plugin 13.12.4): `build_corpus` with `dateStart: "2026-07-31", limit: 500` returned `"filter": {"limit": 500}` and `observation_count: 0`, with no error, warning, or log. The same call with `query` instead of a date worked, because every other parameter has the same name on both sides of the boundary — only `dateStart`/`dateEnd` diverge.

## Changes

- **Name bridge at the HTTP boundary**: `buildCorpusSchema` now normalizes `dateStart`/`dateEnd` to `date_start`/`date_end` in a `z.preprocess` step, mirroring the `dateStart ?? date_start ?? date_from` convention the search endpoints already use. snake_case wins when both conventions are sent, so existing clients are unaffected. Fixing the route (rather than the MCP server) means any HTTP client gets the fix. `handleRebuildCorpus` needs no bridge — it reuses the persisted snake_case filter.
- **Empty corpus can no longer read as clean success**: build and rebuild responses gain a non-breaking `warning` field when `observation_count` is 0, pointing at the filters to check. This closes the defect class: any future filter-name mismatch will be visible immediately instead of producing a silently empty corpus.

## Test plan

- [x] Added coverage in `tests/worker/http/routes/corpus-routes-coercion.test.ts`:
  - camelCase `dateStart`/`dateEnd` map to the snake_case filter
  - snake_case wins when both conventions are sent
  - `warning` present when the corpus matched 0 observations
  - `warning` absent when observations matched
- [x] Verified end-to-end against a live worker (13.12.4 with this patch applied): `build_corpus` with `dateStart: "2026-07-31"` now echoes the date in `filter` and returned 265 observations; an impossible date (`2027-01-01`) returned 0 observations **with** the warning.

https://claude.ai/code/session_013dYTvpBchC6w4vF72cm5fc

### DIFF
--- a/src/services/worker/http/routes/CorpusRoutes.ts
+++ b/src/services/worker/http/routes/CorpusRoutes.ts
@@ -36,7 +36,18 @@ const positiveIntegerLike = z.preprocess((value) => {
   return value;
 }, z.number().int().positive().optional());
 
-const buildCorpusSchema = z.object({
+// The MCP tool schema and the search endpoints use camelCase (dateStart/dateEnd)
+// while the corpus filter uses snake_case. Normalize before validation so a
+// camelCase date filter is never silently dropped at the HTTP boundary.
+const buildCorpusSchema = z.preprocess((value) => {
+  if (typeof value !== 'object' || value === null || Array.isArray(value)) return value;
+  const body = value as Record<string, unknown>;
+  return {
+    ...body,
+    date_start: body.date_start ?? body.dateStart,
+    date_end: body.date_end ?? body.dateEnd,
+  };
+}, z.object({
   // Validate the raw name — do NOT .trim() first, or a padded name like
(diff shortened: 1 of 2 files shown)


thedotmack:

> step 4 (MCP↔HTTP param naming: `dateStart` → `date_start`)

PR #3464, opened 2026-07-31
fix: accept camelCase dateStart/dateEnd in build_corpus and warn on e…

dmeneghel82:
…mpty corpus

The MCP build_corpus tool declares dateStart/dateEnd (camelCase) in its input schema, but the corpus route only destructured date_start/date_end (snake_case), silently dropping the date filter and returning an empty corpus as a clean success.

- Bridge both conventions in handleBuildCorpus (snake_case wins when both are sent, so existing clients are unaffected). handleRebuildCorpus needs no bridge: it reuses the persisted snake_case filter.
- Add a non-breaking warning field to build and rebuild responses when observation_count is 0, so a corpus that matched nothing can never again read as a clean success.
- Cover both behaviors in corpus-routes-dates.test.ts.

Upstream PR (ported to current main): thedotmack/claude-mem#3463

Claude-Session: https://claude.ai/code/session_013dYTvpBchC6w4vF72cm5fc

### DIFF
--- a/src/services/worker/http/routes/CorpusRoutes.ts
+++ b/src/services/worker/http/routes/CorpusRoutes.ts
@@ -47,7 +47,10 @@ export class CorpusRoutes extends BaseRouteHandler {
       return;
     }
 
-    const { name, description, project, types, concepts, files, query, date_start, date_end, limit } = req.body;
+    const { name, description, project, types, concepts, files, query, limit } = req.body;
+    // MCP clients send camelCase (dateStart/dateEnd); accept both conventions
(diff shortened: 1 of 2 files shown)


thedotmack:
Closing in the PR consolidation pass — superseded: duplicate of #3463 (author confirms #3463 is the port to current main). The analysis here fed the plan-23 design (#3611); if a residual piece is not covered on main, please open it as a fresh PR against current `main` referencing that plan. Thank you for the fix and the write-up.

Issue #3012, opened 2026-06-20
Runaway ChromaDB growth to 157 GB filled disk to 99% and caused kernel panics; dual chroma-mcp writers on same data dir; auto-reinstalls after uninstall


anzy-renlab-ai:
## Summary

On a personal MacBook Pro (Intel, 16 GB, macOS Ventura 13.7), claude-mem's ChromaDB store grew to **157 GB**, filling a 500 GB SSD to **99% (~3.5 GB free)**. The resulting disk + memory exhaustion caused **two macOS kernel panics / forced reboots in a single day**. While cleaning up I also found **two `chroma-mcp` processes writing the same data dir**, and after a full uninstall the plugin **re-installed and relaunched itself**, recreating the data dir and resuming growth.

## Environment

- **claude-mem**: 13.7.0 (worker/MCP observed running both 13.6.2 and 13.7.0)
- **chroma-mcp**: 0.2.6 (`uv tool uvx ... chroma-mcp==0.2.6`, `onnxruntime>=1.20`, `protobuf<7`)
- **Host**: Claude Code on macOS 13.7 (22H730), MacBookPro14,2, 2 cores / 16 GB, 500 GB SSD

## 1. Runaway disk growth (157 GB)

```
$ du -sh ~/.claude-mem/*
157G   ~/.claude-mem/chroma          <-- vector store
 84M   ~/.claude-mem/claude-mem.db   <-- actual SQLite store
217M   ~/.claude-mem/logs
 56M   ~/.claude-mem/backups
```

The SQLite observation DB is only **84 MB**, but the Chroma/HNSW index ballooned to **157 GB**. Disk before cleanup:

```
$ df -h /System/Volumes/Data
... 447Gi used   6.0Gi avail   99%
```

After deleting `~/.claude-mem` the volume reported **164 Gi free (64%)** — i.e. claude-mem alone accounted for ~158 GB.

## 2. Disk/memory exhaustion → kernel panics

With only ~3.5 GB free, macOS could not grow swap. Two watchdog panics occurred the same day:

- `watchdog timeout: no successful checkins from opendirectoryd in 200 seconds`, alongside `Compressor Info: ... 11 swapfiles and LOW swap space`.
- `watchdog timeout: no checkins from watchdogd in 94 seconds` under heavy load.

Both are resource-starvation watchdog panics; the full disk was a direct contributor to the first.

## 3. Two chroma-mcp instances on the same data dir

```
# (a) chroma launched as the MCP server:
[...]

Issue #3266, opened 2026-07-15
Chroma addDocuments() delete+add on duplicate-ID conflict causes unbounded HNSW index growth (disk exhaustion, OOM/system freeze)


MilesP46:
## Summary

`ChromaSync.addDocuments()`'s duplicate-ID conflict handler uses a delete-then-readd recovery path instead of an in-place update, which causes the on-disk Chroma HNSW index to grow **unboundedly** every time the same batch of documents gets resynced (e.g. after a failed/retried sync). A single collection's `link_lists.bin` can grow to tens or hundreds of GB this way, filling the disk and causing `chroma-mcp` to attempt loading the full index into memory on every worker restart — which can exhaust available system RAM and freeze the machine entirely (no clean shutdown, no panic log, just total unresponsiveness requiring a hard power cycle).

## Root cause (verified in `scripts/worker-service.cjs`, v13.11.0)

In `addDocuments()`, documents are written in batches of 100 via `chroma_add_documents` (a true `add`, not `upsert`). Document IDs are deterministic (`obs_<sqlite_id>_narrative`, `obs_<sqlite_id>_fact_<n>`, `summary_<sqlite_id>_notes`, etc.), so resyncing the same observation always collides on the same IDs. When that collision happens, the current recovery path is:

```js
try {
  await client.callTool("chroma_add_documents", {collection_name, ids, documents, metadatas})
} catch (e) {
  if (e.message.includes("already exist")) {
    await client.callTool("chroma_delete_documents", {collection_name, ids})
    await client.callTool("chroma_add_documents", {collection_name, ids, documents, metadatas})
    // logged as "Batch reconciled via delete+add after duplicate conflict"
  }
}
```

Chroma's local persistent client backs collections with an on-disk HNSW graph. HNSW deletions are **soft** — the graph structure is not compacted or reclaimed on delete. Every delete+add cycle leaves the old node's graph entries behind in `link_lists.bin` *and* appends a brand-new node on top. The file only ever grows, no matter how many times identical content is resynced.

**This looks avoidable, not an inherent Chroma limitation**: `worker-service.cjs` already calls `chroma_update_documents` elsewhere (in the project-merge metadata-patch path) to update existing documents in place with no delete involved. The `addDocuments()` conflict handler could use the same tool instead of delete+add and would not grow the index at all.

## How this gets triggered

Anything that causes an already-synced batch of documents to be resubmitted will hit the "already exist" collision path above, and therefore the growth bug:

[...]

PR #3466, opened 2026-08-01
fix(chroma): refuse to open a local store stamped by a newer writer epoch (refs #3012)

rodboev:
## Summary

A local persistent Chroma store carries no record of which claude-mem writer generation last wrote it, so a runtime opens it blind. `ChromaMcpManager` now stamps `.claude-mem-chroma-store.json` in the data dir while holding the existing writer lock, checks that record before prewarm and again after acquiring the lock, and refuses to open the store when it names a writer epoch strictly newer than its own. Provenance is published only after the MCP connection succeeds and the cancellation check passes.

This does **not** close the corruption reported in #3012. The writers in that report, 13.11.0 and 13.12.2, are already published and will never read this record. What it adds is the enforceable half: a downgrade-safe boundary going forward, and durable machine-readable writer provenance in the data dir so the next report of this class carries evidence instead of `ps` output and bundle diffs.

## Why

`acquireChromaWriterLock()` (`src/services/sync/ChromaMcpManager.ts:385-457`) already refuses a live foreign writer and reaps a dead one, but it only models *concurrent* exclusion. The lock payload records `pid`, `ownerId`, `dataDir`, `acquiredAt`, `startToken` and nothing about writer identity, and `releaseChromaWriterLock()` deletes the lock file on clean release (`:460-493`). The durable store record remains after release so the next startup can inspect provenance.

The first compatibility read runs before prewarm, so a known-incompatible store is rejected without paying the prewarm cost. After prewarm, the manager acquires the writer lock and repeats the same compatibility check before constructing `StdioClientTransport`. The locked read is authoritative for the interval through provenance publication, so an older writer cannot overwrite a newer record that appeared during prewarm.

The record is written only after `Client.connect()` completes and `assertConnectionNotCancelled()` succeeds. Rejected, timed-out, and cancelled startups release the lock without creating or advancing provenance. The record remains atomic via temp-file-plus-rename, following the pattern already used at `src/services/sync/ChromaSyncState.ts:74-83`.

The refusal is deliberately asymmetric. Only `recorded > current` refuses. A missing record is an existing installation and must connect; an *older* record is the safe direction and must connect; an unparseable record must connect rather than let one corrupt byte remove semantic search. Refusal degrades through the existing `ChromaUnavailableError` + `recordChromaVectorSearchUnavailable` contract, so search falls back to FTS5 exactly as the current lock refusal does.

The predicate keys only on `writerEpoch`. `claudeMemVersion`, `chromaMcpVersion`, `depOverrides`, and `clientType` ride in the record as diagnostics and never participate, so ordinary upgrades never trip the guard and a future chromadb pin shows up in the record without changing behavior.

## Scope

Explicitly not addressed:

- **The corruption in #3012 itself.** Reporter forensics in https://github.com/thedotmack/claude-mem/issues/3012#issuecomment-5144394853 show the divergence was claude-mem's own mutation strategy, `chroma_delete_documents` present in 13.11.0, replaced by `chroma_update_documents` in 13.12.3/13.12.4 per #3266. Both writers launched identical `chroma-mcp==0.2.6` with identical dep overrides, so this is not a store-format divergence and no guard keyed on format would have fired. Old binaries also cannot honor a record introduced after they shipped.
- **Why stale versions linger at all**, open issues #3205 and #3216.
- **Index growth bounding, compaction, size caps, and the `link_lists.bin` ratio heuristic**, requests 1 and 4 of #3012. Those depend on chromadb internals not verifiable here.
- **Uninstall and marketplace `autoUpdate`**, request 3 of #3012.
- Shutdown/teardown deadlines (#3465), corrupt-collection drop-and-derive (#3203), and the chromadb pin (#3384) are untouched.

`CHROMA_WRITER_EPOCH` ships at 1 and is not bumped here. The guard is inert until a future write-path change bumps it; that bump is the maintainer's call.

## Risk

Writer-lock semantics are unchanged in every direction: live-owner refusal, dead-PID reap, start-token liveness, same-owner re-entry, and deletion on `stop()` keep their current behavior, and the existing tests cover them with a record present. Remote mode reads and writes nothing. The record survives release, `stop()`, and `reset()` by design, and a failed record write logs and continues rather than aborting an otherwise successful connection. Failed, timed-out, and cancelled MCP startups leave the prior record absent or unchanged. The one new failure mode is refusal to open a store stamped by a newer epoch, which degrades to FTS5 rather than failing the worker.

Everything here is proven in-process against the existing fake-transport harness. No real `uvx`, `chroma-mcp`, `chromadb`, or persistent store is exercised, and no claim is made about chromadb's on-disk behavior.

## Verification

- [x] `bun test tests/services/sync/chroma-mcp-manager-singleton.test.ts`
- [x] `bun test tests/services/sync/chroma-mcp-manager-ssl.test.ts tests/services/sync/chroma-sync-unavailable.test.ts`
- [x] `npm run build`
- [x] `npm run lint:hook-io`

Refs #3012

### DIFF
--- a/src/services/sync/ChromaMcpManager.ts
+++ b/src/services/sync/ChromaMcpManager.ts
@@ -13,7 +13,7 @@ import { getUvxBinDirs } from '../../shared/uvx-bin-dirs.js';
 import { sanitizeEnv } from '../../supervisor/env-sanitizer.js';
 import { getSupervisor } from '../../supervisor/index.js';
 import { captureProcessStartToken, isPidAlive } from '../../supervisor/process-registry.js';
-import { clearDependencyStatus, recordChromaVectorSearchUnavailable, recordUvxVectorSearchUnavailable, CHROMA_VECTOR_SEARCH_REMEDIATION } from '../../shared/dependency-health.js';
+import { clearDependencyStatus, recordChromaVectorSearchUnavailable, recordUvxVectorSearchUnavailable } from '../../shared/dependency-health.js';
 import { ChromaUnavailableError } from '../worker/search/errors.js';
 
 const execFileAsync = promisify(execFile);
@@ -32,6 +32,9 @@ const CHROMA_STORE_RECORD_FILENAME = '.claude-mem-chroma-store.json';
 // longer safely follow a newer one's store layout. Deliberately not tied to
 // the package version so routine releases do not trigger spurious refusals.
 const CHROMA_WRITER_EPOCH = 1;
+// Size cap for the store record. A legitimate record is a small JSON envelope;
+// anything larger is damaged or adversarially written and must fail open.
+const CHROMA_STORE_RECORD_MAX_BYTES = 64 * 1024;
 const CHROMA_SUPERVISOR_ID = 'chroma-mcp';
 const CHROMA_OUTPUT_TAIL_MAX_CHARS = 2048;
 const DEFAULT_MAX_PENDING_MUTATIONS = 5_000;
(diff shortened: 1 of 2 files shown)


rodboev:
Both startup windows are fixed in commit `a89b9003` within the local persistent writer transaction in `ChromaMcpManager`:

1. `connectInternal()` keeps the early check before prewarm, then calls the shared `assertChromaStoreCompatible()` helper immediately after acquiring the writer lock and before constructing the transport. If a newer writer published while this process was waiting, the older process releases the lock and refuses without replacing the newer record.

2. The store-record write now runs after `Client.connect()` completes and the cancellation check passes. Failed, timed-out, and cancelled handshakes leave the prior record absent or unchanged; a successful local startup publishes provenance while it still owns the writer lock.

The singleton regression coverage exercises the prewarm interleaving, rejected handshake, and stop-triggered cancellation. The record schema, epoch comparison, remote mode, and existing lock lifecycle remain unchanged.

thedotmack:

> step 3 (writer lock stamped with writer version/epoch; older worker refuses to write)

thedotmack:
Routing to **#3610 (plan-22 Chroma Sidecar Contract)** — writer arbitration on the local store is squarely that master's "single-writer upsert sync, writer lock/arbitration" item. Your own framing is right and worth keeping in the cluster notes: this does not close #3012, because 13.11.0 and 13.12.2 will never read the stamp; it is the forward-going downgrade boundary plus durable writer provenance. Distinct from the chroma work in #3614 (orphan reap / static `parseFileList`), so not a duplicate. Leaving open.

PR #3468, opened 2026-08-01
feat(observations): add "sensitive" type with Telegram notification

thedotmack:
## What

Adds a ninth observation type to the code mode: **`sensitive`** — information that isn't quite private, but that you wouldn't want leaking into further content development in the wrong context. Internal URLs, unreleased plans, personal details, business metrics, client or partner names.

These fire a **Telegram notification** by default, the same way `security_alert` does.

## How

The notifier was already generic — it matches `obs.type` against `CLAUDE_MEM_TELEGRAM_TRIGGER_TYPES` — so wiring the notification needed only the new type in the default trigger set plus its own emoji.

| File | Change |
|---|---|
| `plugin/modes/code.json` | New `sensitive` observation type + `type_guidance` rewrite |
| `src/shared/SettingsDefaultsManager.ts` | Default trigger types → `security_alert,sensitive` |
| `src/services/integrations/TelegramNotifier.ts` | Distinct emoji for `sensitive` |
| `src/utils/bmp-safe.ts` | Astral fallback for the new emoji |
| `src/services/worker/http/routes/CorpusRoutes.ts` | Allow the type in corpus filters |
| `openclaw/src/index.ts` | Render `sensitive` in the detailed feed |
| `plugin/skills/weekly-digests/SKILL.md` | Emoji legend |

Parser validation, the observer prompt's type enum, context legends, and the viewer all derive dynamically from the mode file, so they pick this up with no change.

## Drive-by fix

`type_guidance` was stale: it still said *"MUST be EXACTLY one of these 6 options"* and never listed `security_alert` / `security_note` from #2084. That comment is **the only type prose the observer agent actually sees** — the per-type `description` fields in `observation_types` are never injected into any prompt. A type missing from it is effectively invisible to the model, so the security types have been under-emitted since they were added. Now enumerates all 9.

## Note on built bundles

`plugin/scripts/*.cjs` are **intentionally not rebuilt here.** They are already stale at `main` — they predate the chroma fix in a90066f9c, verified by the absence of `CLAUDE_MEM_CHROMA_MAX_PENDING_MUTATIONS` in the committed bundle. Rebuilding produces ~900 lines of churn and would pull an unrelated contributor's unbuilt work into this diff. Left to the release build, matching what recent commits did.

The mode file itself ships as-is (it's a runtime data file, not bundled), so the schema half of this change is live the moment it merges.

## Testing

- `tsc --noEmit` clean
- 87 tests pass across `bmp-safe`, `sdk/parser`, `context/formatters/agent-formatter`, `worker/agents/response-processor`
- `code.json` validated: parses, 9 unique type ids


thedotmack:
Confirmed the P1 and fixed it in 7a926d2 — thanks, this was a real dead-on-arrival bug and I'd missed it.

Verifying the mechanism against the code: `loadFromFile` seeds a brand-new `settings.json` with **every** default (`getAllDefaults()` → `writeJsonFileAtomic`), and persisted values then win over `DEFAULTS` in the merge loop. So any install created after the notifier shipped in #2084 has `"security_alert"` frozen on disk, and changing the default literal could never reach it. The users affected are precisely the ones who bothered to configure Telegram.

Fix follows your suggestion: rewrite only the **exact** legacy value to the current default, mirroring the nested→flat migration already in that function (in-memory fix still applies if the write fails). Anything else is treated as user-customized and left untouched.

On the ambiguity you implied: a user who deliberately set exactly `security_alert` is indistinguishable from the seeded default, so they do get migrated. I took that trade deliberately — it's opt-out via the same key, whereas the alternative ships the feature dead for every pre-existing install. It's called out in the code comment rather than left implicit.

Regression coverage added (6 cases): legacy-default migration, write-back with unrelated keys preserved, customized list preserved, superset list `security_alert,security_note` preserved, empty opt-out preserved, and idempotency across repeated loads.

Local: full typecheck clean, `bun test tests` 2550 pass / 18 skip / 0 fail, `bun test openclaw` 44 pass.

Issue #3378, opened 2026-07-23
Version-mismatch recycle respawns the STALE version from its own path — self-perpetuating loop (2,424 recycles/day, hooks fail every prompt) on 13.12.2


jp63670:
## Summary

On 13.12.x the version-mismatch recycle is **self-perpetuating**: the restart successor is spawned from the **stale worker's own script path**, so the replacement worker is the same old version that triggered the recycle. The next hook re-detects the identical mismatch and recycles again, forever.

Measured on one machine, one day: **2,424 recycles**, **1,464 successor spawns**, and of those `1464 / 1465` spawned the *stale* `13.11.0` binary. Every `UserPromptSubmit` ends in a hook error after ~40s of timeouts. I had to disable the plugin to use Claude Code at all.

This is adjacent to #3298 / #3303 (which resolver picks *which* version) but is a distinct mechanism: those are about **selection**; this is about the **successor inheriting the old script path** after selection has already correctly identified 13.12.2.

## The loop, from the plugin's own log

```
[2026-07-23 20:12:37.557] [WARN ] Live PID detected but worker did not become ready before timeout
[2026-07-23 20:12:37.609] [INFO ] Worker version mismatch — recycling stale worker {pluginVersion=13.12.2, workerVersion=13.11.0}
[2026-07-23 20:12:37.891] [INFO ] Shutdown initiated
[2026-07-23 20:12:37.892] [ERROR] Graceful shutdown failed — proceeding {reason=restart} Server is not running.
[2026-07-23 20:12:37.907] [INFO ] Restart successor spawned {pid=50864, script=/Users/…/claude-mem/13.11.0/scripts/worker-service.cjs, port=37777}
                                                                                    ^^^^^^^ stale version respawned
[2026-07-23 20:12:38.110] [INFO ] HTTP server started {host=127.0.0.1, port=37777, pid=50864}
[2026-07-23 20:12:47.612] [WARN ] No successor worker appeared after recycle; falling through to lazy-spawn {pluginVersion=13.12.2, workerVersion=13.11.0}
[2026-07-23 20:12:47.613] [INFO ] Worker not running — lazy-spawning {scriptPath=/Users/…/claude-mem/13.12.2/scripts/worker-service.cjs}
[2026-07-23 20:12:47.795] [INFO ] Port already in use, refusing to start duplicate {port=37777}
[2026-07-23 20:12:57.634] [WARN ] Worker lazy-spawned but did not become ready before hook readiness timeout
```

[...]

PR #3476, opened 2026-08-01
fix(worker): recycle a worker that is healthy but never becomes ready

jbelke:
## The bug

A worker whose background init throws stays `initialized:false` **forever**, while still answering `200` on `GET /api/health`. Because it also reports the **correct version**, the version-mismatch recycle in `ensureWorkerRunning()` never fires:

```ts
const { matches } = await checkVersionMatch(...);
if (matches) {
  const ready = await waitForWorkerReadiness();
  if (!ready) {
    logger.warn('SYSTEM', 'Worker is healthy but not ready; skipping hook API call');
    return false;          // <-- forever. nothing can ever replace this process
  }
```

So the process is effectively immortal. Nothing in the system can replace it.

I hit this on a real machine: **a worker wedged for 7.15 days of continuous uptime.**

```
$ curl 127.0.0.1:37701/api/health
{"status":"ok","version":"13.12.4","uptime":618146,"initialized":false,"mcpReady":false, ...}

$ curl 127.0.0.1:37701/api/readiness
{"status":"initializing","message":"Worker is still initializing, please retry"}   # 503
```

Every skipped hook increments the hook-failure counter, so once it crossed `CLAUDE_MEM_HOOK_FAIL_LOUD_THRESHOLD` the hooks started blocking outright:

> Hook blocked with message: claude-mem worker unreachable for 11 consecutive hooks.

That message is the symptom users actually report, which makes the true cause hard to find — `/api/health` looks perfectly fine.

## Root cause of that particular wedge

Schema migration v46 in `SessionStore.ts` calls `Statement.iterate()`. `bun:sqlite` only gained `iterate()` in **Bun v1.1.31**. On Bun 1.1.26 it threw from inside background init:

```
[ERROR] [SYSTEM] Background initialization failed
  this.db.query(`SELECT CAST(id AS TEXT) AS row_id, typeof(...) ...`).iterate is not a function
```

`engines.bun` claimed `">=1.0.0"`, so nothing warned at install time.

## The fix

**1. Recycle a wedged worker** (the general fix — any future init failure is now self-healing, not just this one). Reuses the existing SIGKILL + lazy-spawn path.

Readiness alone is deliberately **not** the trigger. A cold boot is legitimately un-ready for a while — Chroma prewarm alone defaults to 120s — and killing during boot is exactly the restart storm #3378 documents. The worker's **self-reported uptime** separates a wedge from a boot and needs no new state file: a fresh replacement starts near 0s and is immune until it has had the full window, so this cannot feed back on itself. Defaults to 300s, override with `CLAUDE_MEM_WEDGED_WORKER_UPTIME_S`.

**2. Correct `engines.bun` to `>=1.1.31`** (both `package.json` and `plugin/package.json`) — the verified floor for `Statement.iterate()`.

**3. Fall back to `.all()` when `iterate()` is absent**, since `engines` is advisory and is not enforced when the plugin is installed through the Claude Code marketplace. Logs the real cause once instead of surfacing a cryptic `TypeError` that wedges the worker.

Fix 1 is the important one; 2 and 3 close the specific hole that triggered it.

## Verification

On the affected machine, after upgrading Bun 1.1.26 → 1.3.14, init completes and the wedge clears:

```
[INFO] [DB    ] Database initialized (shared connection)
[INFO] [SYSTEM] v12.4.3 cleanup complete
[INFO] [SYSTEM] Core initialization complete (DB + search ready)

$ curl 127.0.0.1:37701/api/readiness
{"status":"ready","mcpReady":true}          # 200
```

Both edited files transpile clean. I could not run `tsc`/`bun test` locally — this was a marketplace install with no `node_modules` — so CI is the real check on types and tests.


thedotmack:

> step 6 (wedge watchdog keyed to evidence, not fixed uptime); P1s: kills slow init; concurrent recycles kill healthy replacement

thedotmack:
Routing to **#3603 (plan-15 Worker Port & Liveness Authority)** — "wedged-but-bound worker stalling hooks" and "healthy but never ready" are both listed in that master by name, and this is the exact case: `initialized:false` forever while `/api/health` answers 200 with a matching version, so the version-mismatch recycle never fires. A 7.15-day wedged worker is the best evidence in the cluster. Related to #3219 (unhealthy listener already owning the port) but a different failure — that one is at acquisition, this one is an immortal incumbent. Leaving open.

PR #3477, opened 2026-08-02
docs(cloud-sync): self-update claude-mem before SyncHub

thedotmack:
## What

Adds a **Step 0** to the `cloud-sync` skill that self-updates claude-mem before attempting any SyncHub operation.

## Why

SyncHub requires claude-mem **>= 13.12.0**. Older builds have no `/api/sync/*` routes, so status/write/verify all fail with confusing errors. Hit this live: an installed 13.10.2 had to be `git pull`ed to 13.12.4 before setup would work.

## How

Step 0 reads the installed plugin version and, if `< 13.12.0`, runs `git pull --ff-only` on the marketplace clone (the step-4 worker restart then loads the new build). If it can't fast-forward or is still too old, it stops and tells the user to update manually rather than probing an unsupported build. Version compare uses `sort -V` — no new dependency.

### DIFF
--- a/plugin/skills/cloud-sync/SKILL.md
+++ b/plugin/skills/cloud-sync/SKILL.md
@@ -17,6 +17,25 @@ the three connection values issued by **cmem.ai → Connect**.
 Confirm only its length. Preserve every unrelated setting and keep
 `~/.claude-mem/settings.json` mode `0600`.
 
+## 0. Ensure a SyncHub-capable build
+
+SyncHub requires claude-mem **>= 13.12.0**. Self-update before anything else,
+since older builds have no `/api/sync/*` routes and every later step will fail.
+
+```bash
+DIR="$HOME/.claude/plugins/marketplaces/thedotmack"
+V=$(node -pe "require('$DIR/plugin/.claude-plugin/plugin.json').version" 2>/dev/null || echo 0.0.0)
+if [ "$(printf '%s\n13.12.0\n' "$V" | sort -V | head -1)" != "13.12.0" ]; then
+  echo "claude-mem $V < 13.12.0 — updating"
+  git -C "$DIR" pull --ff-only
+fi
+```
+
+If the pull advances the clone, the worker restart in step 4 loads the new
+build. If it cannot fast-forward (local changes) or `$V` is still `< 13.12.0`
+afterward, stop and tell the user to update the plugin manually — do not try
+SyncHub against an unsupported build.
+
 ## 1. Check status
 
 Resolve the worker port and query the always-registered status route:
--- a/plugin/skills/cloud-sync/SKILL.md
+++ b/plugin/skills/cloud-sync/SKILL.md
@@ -41,7 +41,21 @@ checkout, restart, and re-read `/api/health`:
 DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/plugins/marketplaces/thedotmack"
 git -C "$DIR" pull --ff-only
 curl -s -X POST "http://127.0.0.1:${PORT}/api/admin/restart"
+START=$(date +%s)
+# Poll for the successor; do not read health once. The restart endpoint acks
+# before the outgoing worker releases the listener, so an immediate read is
+# answered by the process being replaced and still reports the pre-update
+# build. Accept a reading only once its uptime is shorter than the time since
+# the restart was requested — that is what proves it is the new process.
+for _ in $(seq 1 15); do
+  sleep 2
+  curl -s -m 2 "http://127.0.0.1:${PORT}/api/health" \
+    | CMEM_SINCE=$(( $(date +%s) - START )) node -pe 'const j=JSON.parse(require("fs").readFileSync(0,"utf8"));if(!(j.uptime<+process.env.CMEM_SINCE))process.exit(1);j.version+" "+j.workerPath' 2>/dev/null && break
+done
 ```
+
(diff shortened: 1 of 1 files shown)


thedotmack:
Addressed in 8869299.

**Issue 1 — CLAUDE_CONFIG_DIR ignored: valid, fixed.** The path now resolves from `${CLAUDE_CONFIG_DIR:-$HOME/.claude}`, matching the root the plugin's own MCP launcher already uses. Without this, a user with a custom config root could have the inactive checkout updated while the active unsupported one kept serving.

**Issue 3 — stale version after pull: valid, fixed.** `V` was captured once before `git pull` and never refreshed, so the documented "still `< 13.12.0` afterward" branch could never observe a successful upgrade. The check is now a `supported()` function re-invoked after the pull.

**Issue 2 — `sort -V` unavailable on macOS: not reproducible here.** Apple's sort does support `-V`:

```
$ sort --version
2.3-Apple (199)
$ printf '13.12.1\n13.12.0\n' | sort -V
13.12.0
13.12.1
```

The comparison moved into node regardless — node is already required by step 1, so this drops a shell dependency instead of arguing about which platforms ship which `sort`.

Verified across the version boundary:

| version | result |
|---|---|
| 12.6.2, 13.11.9 | update |
| 13.12.0, 13.12.4, 13.13.0, 14.0.0 | supported |
| manifest missing | unsupported (fails closed) |

thedotmack:
Both valid. Fixed in 9ca421e.

**Quote-unsafe manifest lookup.** Reproduced exactly as described — a `13.12.0` fixture under a config path containing an apostrophe and a backslash:

```
=== interpolated into JS source (before) ===
  UNSUPPORTED  <-- false negative
=== path via environment (after) ===
  supported (correct)
```

The manifest path now travels through the environment and the node script is single-quoted, so no path content reaches the script text at all.

**Failure branch did not stop.** Confirmed against a `13.11.9` fixture with no upstream to pull from:

| tail | output | exit |
|---|---|---|
| `supported \|\| echo …` | STOP printed | **0** — run continues |
| `supported \|\| { echo …; exit 1; }` | STOP printed | **1** — run stops |
| new tail, compatible build | *(silent)* | 0 |

So the old form printed a warning and then let the run proceed into the SyncHub steps against a build with no `/api/sync/*` routes — which is the exact failure this preflight exists to prevent.

thedotmack:
Valid, and it invalidates the whole approach rather than one line. Rewritten in d75ec98.

Confirmed on a live install — the marketplace clone is current, but there are four cached copies beside it:

```
marketplace manifest: 13.12.4
cached builds:        13.9.1  13.10.1  13.10.2  13.12.4
```

A disk-manifest check cannot tell which of those the worker actually started, which is the only thing the preflight cares about.

**The worker already answers this directly.** `/api/health` reports the running version *and* the file it was loaded from:

```
$ curl -s "http://127.0.0.1:$PORT/api/health"
version=13.12.4  workerPath=/Users/…/plugins/marketplaces/thedotmack/plugin/scripts/worker-service.cjs
```

So step 0 now gates on that, and if the version is still old after the update and restart, `workerPath` names the stale cached copy that is being launched.

Rather than mirroring the launcher's cache-first resolution order into a doc — which would duplicate logic that can drift — this deletes the disk check entirely. That also removes the hand-rolled semver comparison, its quoting hazard, and its exit-status handling, i.e. the three defects from the previous two rounds, since none of that code exists anymore. Net: **+26 / -12**, and the remaining check is one `curl`.

Port resolution is hoisted into step 0 and reused by step 1 instead of being repeated.

thedotmack:
Valid, and measurable. Fixed in c34d574.

Instrumented a real restart on a live worker — all three outcomes the comment predicts occur inside 1.6s:

| elapsed | result |
|---|---|
| 0.02s | **200 from pid 68753** — the outgoing worker, uptime 9s |
| 0.83s | connection failed — handover gap |
| 1.56s | 200 from pid 69139, uptime 0s — the successor |

So a single read after the ack lands on the process being replaced, which still reports the pre-update build. The preflight would then blame a stale cached copy and stop a setup that was seconds from working.

The step now polls, and accepts a reading only once `uptime` is shorter than the time elapsed since the restart was requested — that is what distinguishes the successor from its predecessor, rather than "first 200 wins", which the 0.02s row shows is wrong. Verified both directions: it waits through a real restart (pid 69139 → 72740) before reporting, and a synthetic `uptime: 900` reading is rejected so polling continues.

Connection failures and 503s in that window are now documented as expected rather than as a diagnosis, matching what steps 1 and 4 already say.

Incidental confirmation of the previous round while testing this: the same machine reported

```
workerPath=…/plugins/marketplaces/thedotmack/plugin/scripts/worker-service.cjs   (earlier)
workerPath=…/plugins/cache/thedotmack/claude-mem/13.12.4/scripts/worker-service.cjs  (after this restart)
```

Same host, same session, two different plugin roots — which is exactly why the check reads `workerPath` from the worker instead of inferring a path.

thedotmack:
Routing to **#3618 (plan-24 Server Runtime & Cloud Sync Contract)** — the SyncHub version gate is part of that contract, and a skill that probes an unsupported build is the same class of problem as the runtime paths plan-24 is consolidating. It is a skill doc rather than code, and it is the only mergeable PR in this batch. Confirmed nothing on `main` already adds a Step 0 to `cloud-sync/SKILL.md`, so it is not superseded. Leaving open.

Issue #532, opened 2026-01-03
Memory leak: SessionManager accumulates sessions without cleanup, conversationHistory grows unbounded


mkdelta221:
## Description

After using claude-mem for several days, VS Code (running Claude Code with the plugin enabled) consumed **54GB+ of memory**, triggering macOS "out of application memory" warnings and forcing VS Code/Xcode to be paused.

## Environment

- **Plugin version:** 8.5.1
- - **OS:** macOS (M1 MacBook Pro, 16GB RAM)
- - - **Claude Code version:** 2.0.75
## Root Cause Analysis

After investigating the source code, I identified two related memory leak issues:

### 1. Sessions never auto-cleanup

In `SessionManager.ts`, active sessions are stored in a Map:

```typescript
private sessions: Map<number, ActiveSession> = new Map();
```

Sessions are only deleted via manual API calls (`deleteSession`), but **never automatically after the SDK agent completes**. The "Agent completed" log fires, but the session remains in memory indefinitely.

### 2. conversationHistory grows unbounded

Each `ActiveSession` has a `conversationHistory: ConversationMessage[]` array (defined in `worker-types.ts`). Multiple agents (SDKAgent, GeminiAgent, OpenRouterAgent) continuously `.push()` messages to this array:

```typescript
session.conversationHistory.push({ role: 'user', content: initPrompt });
session.conversationHistory.push({ role: 'assistant', content: text });
```

This array is **never trimmed, cleared, or bounded**. Over time, especially with long coding sessions, these can grow very large.

### Compounding factors

- No session TTL or timeout mechanism
- - No periodic cleanup of stale sessions
- - - No maximum limit on `conversationHistory` length
- - - - Context strings in messages can be very large
## Suggested Fix

```typescript
// Option 1: Auto-cleanup after SDK agent completion
// In SDKAgent.ts after "Agent completed" log:
this.sessionManager.deleteSession(session.sessionDbId);

// Option 2: Add session TTL with periodic cleanup
private SESSION_TTL_MS = 30 * 60 * 1000; // 30 minutes
setInterval(() => this.cleanupStaleSessions(), 5 * 60 * 1000);

// Option 3: Bound conversationHistory
const MAX_HISTORY_LENGTH = 100;
if (session.conversationHistory.length > MAX_HISTORY_LENGTH) {
  session.conversationHistory = session.conversationHistory.slice(-MAX_HISTORY_LENGTH);
}
```

## Steps to Reproduce

[...]

PR #3479, opened 2026-08-02
fix(worker): prevent history amplification after auth failures

KarelRoy:
## Problem

When Claude authentication expires or is revoked, the observer can return 401/authentication-failure prose instead of XML. The worker then:

1. resets the claimed batch to pending;
2. aborts the generator with an `auth:*` reason;
3. preserves the active session and its in-memory buffer; and
4. starts a fresh Claude generator on the next hook event.

The fresh generator replays the preserved buffer, appending the init prompt and every queued observation prompt to the same `conversationHistory`. The failed attempt's additions were never removed. Repeated auth failures therefore retained another copy of the replayed queue on every attempt (`O(retries * queue depth)`; near-quadratic growth while the queue also grows). A few thousand buffered tool events can become millions of retained history entries and drive the worker to multi-GB memory usage.

Quota pauses use the same preserve-and-replay path and had the same amplification risk.

## Fix

- Capture a conversation-history checkpoint before each provider attempt.
- When a Claude generator exits through the preserved `auth` or `quota` path, truncate only the history appended by that failed attempt.
- Keep the buffered work and all history from before the attempt intact.
- Leave Gemini/OpenRouter history behavior unchanged because those providers use the shared history as request context.
- Include the rollback count in the pause log for diagnosis.

The Claude retry already starts a fresh SDK process and uses the in-memory message buffer as the replay source, so retaining the failed attempt's duplicated history is unnecessary.

## Regression coverage

The new test simulates 1,000 consecutive Claude auth retries. Before this change, three replay entries per attempt would leave 3,001 history entries; after the change, history remains at the single pre-attempt entry while the pending batch is still preserved.

## Testing

- `bun run typecheck` — passed
- `bun test tests/worker/poison-respawn.test.ts tests/worker/agents/response-processor.test.ts` — 34 passed
- `bun test tests` — 2,544 passed, 18 skipped, 1 unrelated timeout: `mutation sites > adoptMergedWorktrees emits remap_project through its own connection` exceeded its 5,000 ms limit (5,043 ms; 5,034 ms when rerun alone)

Related historical report: #532 described unbounded `conversationHistory`, but this PR addresses the distinct auth/quota replay path introduced later.

### DIFF
--- a/src/services/worker/http/routes/SessionRoutes.ts
+++ b/src/services/worker/http/routes/SessionRoutes.ts
@@ -165,6 +165,7 @@ export class SessionRoutes extends BaseRouteHandler {
     session.lastGeneratorSource = source;
 
     const myController = session.abortController;
+    const conversationHistoryCheckpoint = session.conversationHistory.length;
 
     let skipGeneratorExitFinalization = false;
     let generatorPromise: Promise<void>;
@@ -254,6 +255,7 @@ export class SessionRoutes extends BaseRouteHandler {
         await handleGeneratorExit(session, reason, {
           sessionManager: this.sessionManager,
           completionHandler: this.completionHandler,
+          conversationHistoryCheckpoint,
         });
       });
     session.generatorPromise = generatorPromise;
--- a/src/services/worker/session/GeneratorExitHandler.ts
+++ b/src/services/worker/session/GeneratorExitHandler.ts
@@ -7,15 +7,18 @@ import { getSdkProcessForSession, ensureSdkProcessExit } from '../../../supervis
 export interface GeneratorExitDependencies {
   sessionManager: SessionManager;
   completionHandler: SessionCompletionHandler;
+  conversationHistoryCheckpoint: number;
 }
 
 /**
  * Post-generator-exit handler.
  *
  * The generator's message iterator only ends on abort (idle / shutdown) or when
- * the SDK stream throws, so most exits mean this session is done. Quota exits
- * are different: claimed work has already been reset to pending, so leave the
- * session and in-RAM buffer alive for a later generator start.
+ * the SDK stream throws, so most exits mean this session is done. Quota/auth
+ * exits are different: claimed work has already been reset to pending, so
+ * leave the session and in-RAM buffer alive for a later generator start. A
+ * fresh Claude generator replays that buffer, so discard history appended by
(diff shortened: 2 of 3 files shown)


SYMBaiEX:
Confirmed this failure mode in a live macOS installation running the latest release, v13.15.0.

Environment:
- claude-mem v13.15.0
- Claude Code CLI 2.1.220
- Bun 1.4.0-canary.1
- `CLAUDE_MEM_PROVIDER=claude`
- Claude CLI auth state: logged out

After about 10h 46m, the worker reached a 40.6 GB physical footprint (43.8 GB peak) and ~188% CPU. It held 6,359 in-memory queue items across 16 active sessions. The worker log contained 5,185,081 `Broadcasting processing status` entries and had grown to ~699 MB. Logs repeatedly showed `SDK authentication failed; run /login to preserve queued batch` followed by `Generator paused for auth; preserving buffered work`.

This matches the PR's history-amplification mechanism: failed auth attempts preserve the buffer, new hook events replay it, and each failed replay appends another copy to `conversationHistory`.

I cherry-picked a7940f5 onto current `main` at d768ba36. Validation passed:
- `bun run typecheck`
- Focused worker suite: 36 passed, 0 failed
- Full suite: 2,557 passed, 18 skipped, 1 unrelated timeout in `adoptMergedWorktrees emits remap_project through its own connection` at 5.01s—the same pre-existing timeout already documented in this PR

The checkpoint rollback is clean against current main and directly addresses the observed multi-GB retention. The remaining logged-out-provider queue replay/status-log churn may merit a follow-up auth circuit breaker or cooldown, but that is separate from this PR's bounded-history fix.

thedotmack:

> step 3 (auth-paused buffer capped; per-attempt history checkpoint) — independently reproduced 40 GB wedge on 13.15.0

thedotmack:
Routing to **#3606 (plan-18 Observer Response Pipeline)** — "conversationHistory unbounded / double-appended" plus the auth classification path, both in that master. The specific defect is good: a failed attempt's appended turns are never removed before the fresh generator replays the preserved buffer, so retries retain O(retries × queue depth). Distinct from #3516 (steady-state compaction) and #3467/#3475 (batch preservation); they should land together but none supersedes another. Leaving open.

PR #3486, opened 2026-08-03
fix(hooks): dedupe file-context injection per session (#3480)

### LINKED ISSUE #3480: PreToolUse `file-context` hook re-injects the same "prior observations" block on every Read of the same unchanged file within a session (no per-session dedupe)
### Bug Description

The `PreToolUse` hook registered for `Read` (`hooks/hooks.json`, `file-context` command) re-fetches and re-injects the "This file has prior observations — supplementary context follows" `additionalContext` block on *every* `Read`/`Grep` call against a given file path, as long as the file hasn't been modified since the last recorded observation. There is no per-(session, file) memoization, so reading the same unchanged file N times in one session injects the same template block N times — pure token cost with zero new information after the first injection.

Where: `scripts/worker-service.cjs`, function `t6e(t, filePath)` (the `file-context` PreToolUse handler). The only existing gate is a content-staleness check:

```js
if (r > 0) {
  let p = Math.max(...u.observations.map(f => f.created_at_epoch));
  if (r >= p) return null; // file modified since last observation, skip
}
```

This only guards against showing a *stale* observation relative to file mtime — it does not guard against showing the *same, still-valid* observation more than once per session. We verified no dedup exists on the injection path: `grep -n "dedup" scripts/worker-service.cjs` → 5 hits, all SQL dedup of `observations`/`pending_messages` (capture side), none related to the `file-context` handler.

### Steps to Reproduce

1. Enable claude-mem with prior observations recorded against some file `foo.md`.
2. In a single session, call `Read` (or `Grep`) on `foo.md` three times in a row without modifying it.
3. Observe the system-reminder "This file has prior observations…" block appended after every single call, not just the first.

Observed in real sessions: dozens of near-identical blocks for the same paths during one long session (2026-08-01); reproduced again 2026-08-02 (2+ injections for the same unchanged file within minutes).

### Expected Behavior

The block is injected once per (session, file) — or re-injected only when NEW observations were recorded for that file since the last injection in the same session.

### Environment

- **Claude-mem version**: 13.13.0
- **Claude Code version**: 2.1.220
- **OS**: Windows 11 Pro (10.0.22631)
- **Platform**: win32 x64

### Logs

Worker logs (`~/.claude-mem/logs/claude-mem-YYYY-MM-DD.log`) only record the capture side (`[HOOK] → PostToolUse: Read`, 213 entries on 2026-08-02); the injection side leaves no log line, so the evidence is the repeated `additionalContext` blocks visible in the session transcript. Example (identical header repeated per tool call on the same path):

```
This file has prior observations — supplementary context follows.
- **Need details on a past observation?** get_observations([IDs]) — ~300 tokens each.
- **Need a structural map first?** smart_outline(...) — line numbers only, cheaper than re-reading.
### Jul 27, 2026 ...
```

### Additional Context

Suggested fix: track `(session_id, resolved_file_path)` pairs already surfaced this session. In-memory won't work since each hook invocation is a fresh spawned process (via bun-runner) — needs a small persistent store, e.g. a row in the existing sqlite DB or a session-scoped JSON file under the plugin's cache dir. Skip re-injection for a pair already shown unless new observations were recorded since.

Impact: for sessions that repeatedly re-read the same handful of files (iterative debugging, long governance sessions), this multiplies context cost with no corresponding value — the second+ injection tells the agent nothing it doesn't already have in its own transcript.

ousamabenyounes:
## Summary

Fixes 3480

The `PreToolUse` file-context handler re-injected the same "This file has prior observations — supplementary context follows" block on **every** `Read`/`Grep` of a given file within a session. The only gate was file-mtime staleness (`fileMtimeMs >= newestObservationMs`); there was no per-`(session, file)` memoization, so reading the same unchanged file N times injected the same template N times — pure token cost with zero new information after the first.

## Fix

Per the routing note on this PR (plan-20 / 3608 **step 4**), the gate is persisted **per `(session, file, observation epoch)` in SQLite — not in a JSON side-store**.

- `src/cli/handlers/file-context-dedupe.ts` keeps a `file_context_injections` table in the main `claude-mem.db`: one row per `(session_id, file_path)` carrying `observation_epoch` + `injected_at_epoch`. In-memory memoization can't work here — each hook runs in a freshly spawned process (bun-runner) — so the state has to be durable, and it now lives next to every other durable fact instead of in a parallel JSON store.
  - The row *carries* the epoch rather than keying on the triple: the gate only ever asks "was this pair already served at an epoch `>=` the current newest one", so an upsert has the exact same semantics as an epoch-keyed row while leaving one row per pair instead of one row per observation.
  - The table is device-local. Sync enumerates the tables it pushes explicitly (`observations`, `session_summaries`, `user_prompts`), so a gate row never leaves the box.
- `wasFileContextInjected` + `recordFileContextInjection` collapse into a single **`claimFileContextInjection`**. Claiming *is* recording, in one `INSERT ... ON CONFLICT ... RETURNING` statement, so two *concurrent* Reads of the same file in one session can no longer both inject — the read-before-write race the previous store shipped with (and documented as a known ceiling) is gone.
- The upsert guard `WHERE excluded.observation_epoch > file_context_injections.observation_epoch` keeps the stored epoch **monotonic**: a hook that finishes late carrying an older epoch neither wins the claim nor rolls the row back to that stale value.
- The claim moved to the very end of `buildFileContextTimeline`, after every other reason to bail out, so a suppressed injection never burns the claim.
- `resolveDbPath()` resolves the database path at **call** time. `DB_PATH` freezes `DATA_DIR` at import — right for the long-lived worker, wrong for a per-invocation hook process that must honour a `CLAUDE_MEM_DATA_DIR` set after the module loaded (and it is what lets the tests point at a temp dir instead of the real `~/.claude-mem`).
- Rows older than 7 days are pruned on a session's first claim, so the table stays bounded.
- Everything fails **open**: an unopenable database or a failed claim injects rather than throwing, which is never worse than the un-deduped behaviour it replaces.

Base for RED is the previous PR head (`148feea7`, the JSON-side-store implementation) with **only** the new test hunk applied — the earlier per-session dedupe bug is already fixed on this branch, so the thing under test here is specifically "the gate is a SQLite row".

RED — `bun test tests/hooks/file-context.test.ts`:

```
385 |     // The gate is a row in the main database keyed by (session, file) and
386 |     // carrying the observation epoch it was served at — see plan-20 3608.
387 |     const db = new Database(resolveDbPath(), { readonly: true });
                     ^
SQLiteError: unable to open database file
      errno: 14,
 byteOffset: -1,
       code: "SQLITE_CANTOPEN"

(fail) fileContextHandler > persists the injection gate as a SQLite row, not a JSON side-store (3608 step 4)
(fail) fileContextHandler > never rolls the stored epoch back to an older observation (3608 step 4)

 18 pass
 2 fail
Ran 20 tests across 1 file.
```

GREEN — same command with the fix applied:

```
bun test v1.3.14 (0d9b296a)

 21 pass
 0 fail
 47 expect() calls
Ran 21 tests across 1 file. [302.00ms]
```

New coverage: the gate is a row in `claude-mem.db` with the right `(session_id, file_path, observation_epoch)`; exactly one of two concurrent claims for the same `(session, file, epoch)` is granted; a late/older epoch never downgrades the stored row; and an unusable gate database degrades to "always inject" rather than to an error or a swallowed injection.

## CI red — addressed

The red run on the previous head (`099d0349`) was based on `v13.15.0` and failed on three `server boot: mode loading (2443)` / `server runtime in-process smoke (2550)` tests, unrelated to this change. This branch is rebased onto current `main` (`e2d1df56`), where those are gone.

## Full suite — no regression

`bun test tests` locally, before and after this commit: **identical** failure set (2559 pass / 18 skip / 22 fail — all pre-existing `HealthMonitor` / `ProcessManager` order-dependent mock-leak flakes off `main`, unrelated to this fix). No new failures.

`npm run typecheck`, `npm run build`, `npm run lint:hook-io` and `npm run lint:spawn-env` all pass. The `worker-service.cjs` advisory bundle warning is pre-existing on `main` (2907.88 KB there vs 2911.24 KB here — the same advisory line appears in `main`'s own CI log).

## Files changed

| File | Change |
|------|--------|
| `src/cli/handlers/file-context-dedupe.ts` | gate persisted as a `file_context_injections` row in `claude-mem.db`; atomic claim-is-record upsert with a monotonic-epoch guard |
| `src/cli/handlers/file-context.ts` | single `claimFileContextInjection` call, made last so a suppressed injection never burns the claim |
| `src/shared/paths.ts` | `resolveDbPath()` — database path resolved at call time so a hook process honours `CLAUDE_MEM_DATA_DIR` |
| `tests/hooks/file-context.test.ts` | 4 new tests: SQLite row shape, single-winner claim, epoch monotonicity, fail-open on an unusable database |

### DIFF
--- a/src/cli/handlers/file-context-dedupe.ts
+++ b/src/cli/handlers/file-context-dedupe.ts
@@ -6,6 +6,7 @@
 // resolved file path to the newest observation epoch that was injected for it.
 // A repeated Read of the same unchanged file is skipped unless a NEWER
 // observation has been recorded since the last injection.
+import { createHash } from 'crypto';
 import { existsSync, mkdirSync, readFileSync, readdirSync, statSync, unlinkSync, writeFileSync } from 'fs';
 import { join } from 'path';
 import { resolveDataDir } from '../../shared/paths.js';
@@ -19,9 +20,12 @@ const SESSION_FILE_TTL_MS = 7 * 24 * 60 * 60 * 1000;
 type SeenMap = Record<string, number>;
 
 function sessionFilePath(sessionId: string): string {
-  // Session ids are UUID-ish, but sanitize defensively against path traversal /
-  // separators so a hostile value can never escape the store directory.
-  const safeId = sessionId.replace(/[^A-Za-z0-9_-]/g, '_');
+  // Hash the COMPLETE session id: a char-replace scheme (e.g. `[^A-Za-z0-9_-] -> _`)
+  // is not injective, so distinct ids like `a.b` and `a:b` would collide on the
+  // same store file and cross-suppress each other's injection. A sha256 hex digest
+  // is collision-resistant, fixed-length, path-safe, and can never escape the store
+  // directory (no separators / traversal), regardless of how hostile the input is.
+  const safeId = createHash('sha256').update(sessionId).digest('hex');
   return join(resolveDataDir(), DEDUPE_SUBDIR, `${safeId}.json`);
 }
 
--- a/src/cli/handlers/file-context-dedupe.ts
+++ b/src/cli/handlers/file-context-dedupe.ts
@@ -88,8 +88,8 @@ export function recordFileContextInjection(
   const filePath = sessionFilePath(sessionId);
   const storeDir = join(resolveDataDir(), DEDUPE_SUBDIR);
   try {
-    const firstWriteForSession = !existsSync(filePath);
     mkdirSync(storeDir, { recursive: true });
+    const firstWriteForSession = !existsSync(filePath);
     if (firstWriteForSession) pruneStaleSessions(storeDir);
     const seen = readSeen(filePath);
     seen[resolvedPath] = newestObservationEpoch;
--- a/src/cli/handlers/file-context-dedupe.ts
+++ b/src/cli/handlers/file-context-dedupe.ts
@@ -1,105 +1,159 @@
-// #3480 — per-(session, file) dedupe for the file-context PreToolUse handler.
+// #3480 — per-(session, file, observation-epoch) injection gate for the
+// file-context PreToolUse handler, persisted in the main SQLite database
+// (plan-20 #3608 step 4: the gate is a DB row, not a JSON side-store).
 //
 // Each hook invocation is a freshly spawned process (via bun-runner), so an
-// in-memory Set cannot remember what was already surfaced this session. We keep
-// a tiny persistent store — one JSON file per session under DATA_DIR — mapping a
-// resolved file path to the newest observation epoch that was injected for it.
-// A repeated Read of the same unchanged file is skipped unless a NEWER
-// observation has been recorded since the last injection.
-import { createHash } from 'crypto';
-import { existsSync, mkdirSync, readFileSync, readdirSync, statSync, unlinkSync, writeFileSync } from 'fs';
-import { join } from 'path';
-import { resolveDataDir } from '../../shared/paths.js';
+// in-memory Set cannot remember what was already surfaced this session. The
+// gate therefore lives in `claude-mem.db` next to every other durable fact:
+// one row per (session, file) carrying the newest observation epoch already
+// injected for it. A repeated Read of the same unchanged file is skipped
+// unless a NEWER observation has landed since that injection.
+//
+// The row is keyed on (session_id, file_path) and *carries* the epoch rather
+// than keying on the triple: the gate only ever asks "was this pair already
+// served at an epoch >= the current newest one", so an upsert keeps the exact
+// same semantics as an epoch-keyed row while leaving one row per pair instead
+// of one row per observation.
+//
+// Rows are device-local scratch state. Sync enumerates the tables it pushes
+// explicitly (observations, session_summaries, user_prompts — see
+// SessionStore.initializeSyncHubLaunchBaseline), so a table it does not name is
+// never drained: a gate row never leaves the box.
+import { Database } from 'bun:sqlite';
+import { mkdirSync } from 'fs';
+import { dirname } from 'path';
+import { resolveDbPath } from '../../shared/paths.js';
+import { applySqliteConnectionPragmas } from '../../services/sqlite/connection.js';
 import { logger } from '../../utils/logger.js';
 
-const DEDUPE_SUBDIR = 'file-context-seen';
-// Sessions rarely outlive a few days; prune stale session files so the store
-// never grows unbounded. Cheap because we only sweep on a session's first write.
-const SESSION_FILE_TTL_MS = 7 * 24 * 60 * 60 * 1000;
+const GATE_TABLE_DDL = `
+  CREATE TABLE IF NOT EXISTS file_context_injections (
+    session_id TEXT NOT NULL,
+    file_path TEXT NOT NULL,
+    observation_epoch INTEGER NOT NULL,
+    injected_at_epoch INTEGER NOT NULL,
+    PRIMARY KEY (session_id, file_path)
+  )
+`;
 
-type SeenMap = Record<string, number>;
+// Claiming and recording are ONE statement, so two concurrent Reads of the same
+// file in the same session cannot both pass the gate: SQLite serializes the
+// writers, the first inserts and gets its row back, the second conflicts and is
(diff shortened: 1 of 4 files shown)


ousamabenyounes:
Thanks @greptile-apps — both P1 findings were real. Fixed in `edd446be`:

**1. Sanitized session ids collide** (`file-context-dedupe.ts`)
The `[^A-Za-z0-9_-] -> _` replace was not injective, so `a.b` and `a:b` both mapped to `a_b.json` and the second distinct session read the first's marker. Now the store filename is a `sha256` hex digest of the **complete** session id — collision-resistant, fixed-length, still path-safe and traversal-proof.

**2. Dot-segment aliases bypass dedupe** (`file-context.ts`)
Already-absolute inputs were kept verbatim, so `/p/src/../src/f.ts` and `/p/src/f.ts` produced two keys for one file. The key is now `path.resolve(cwd, filePath)`, which collapses `.`/`..` for absolute inputs too (it still ignores `cwd` when the path is absolute). Symlink aliases are intentionally left out of scope (would need `realpath` + extra fs failure modes).

**Validation (RED → GREEN):** two new regression tests in `tests/hooks/file-context.test.ts` — one asserting `a.b`/`a:b` stay isolated, one asserting a dot-segment alias dedupes to the canonical path. Both fail on the pre-fix code and pass after. Full `file-context` suite: 17 pass. `npm run typecheck` clean. Full `bun test tests` baseline unchanged (same 27 pre-existing cross-file mock-leak failures before and after; +2 new passing tests).

ousamabenyounes:
Pushed a small follow-up in `64576689` for the remaining review nit: `recordFileContextInjection` now creates the dedupe store directory before checking whether the session marker file exists, so the prune ordering is explicit.

Validation: `bun test tests/hooks/file-context.test.ts` passed with 17/17, `npm run typecheck` passed, and all GitHub checks on `64576689` are green.

thedotmack:

> step 4 (file-context gate persisted per (session,file,epoch) in SQLite, not a JSON side-store); CI red

ousamabenyounes:
Reworked against plan-20 (3608) **step 4** as requested — the file-context gate is now a row in `claude-mem.db` keyed per `(session, file)` and carrying the observation epoch, instead of the JSON side-store under `DATA_DIR/file-context-seen/`. Pushed as `d4b5fce9`.

Three things came out of the move beyond the storage swap:

1. `wasFileContextInjected` + `recordFileContextInjection` collapsed into one `claimFileContextInjection` — a single `INSERT ... ON CONFLICT ... RETURNING`, so claiming *is* recording. The read-before-write race the JSON store shipped with (documented there as a known ceiling: two concurrent Reads of the same file in one session could both inject) is gone.
2. The upsert guard `excluded.observation_epoch > stored` keeps the epoch monotonic, so a hook that finishes late carrying an older epoch neither wins the claim nor rolls the row back.
3. The claim moved to the end of `buildFileContextTimeline`, after every other bail-out, so a suppressed injection never burns the claim.

The table is device-local — sync enumerates the tables it pushes explicitly, and this is not one. Everything still fails open: an unopenable database injects rather than throwing.

On the CI red: that run was based on `v13.15.0` and failed on three `server boot: mode loading (2443)` / `server runtime in-process smoke (2550)` tests unrelated to this change. The branch is rebased onto current `main` (`e2d1df56`) and those are gone.

RED → GREEN and the full-suite no-regression comparison are in the PR description. Happy to fold this into the plan-20 bundle instead if that is the cleaner path.

thedotmack:
**All-night ship gate: PASS**

1. RCA logged — PreToolUse re-injected the same file-context block on every Read/Grep in a session (#3480 / #3608 step 4). JSON side-store rework completed: gate is a `file_context_injections` row in `claude-mem.db`.
2. No second-system — persisted in the main database (not a JSON side-store); sync does not enumerate this table.
3. Narrow / not opinionated — claim-is-record upsert scoped to `(session, file)` + observation epoch; fail-open matches prior un-deduped behavior.

Rebased onto `2714e869` locally (fork branch update blocked: GitHub App `workflows` permission). Focused tests: `bun test tests/hooks/file-context.test.ts` — 21 pass. CI on this head was green. Merging independently; no version bump.

PR #3493, opened 2026-08-04
fix(openrouter): never send an empty messages array to the API (Fixes #3491)

### LINKED ISSUE #3491: OpenRouter truncateHistory returns an empty array when one message exceeds MAX_TOKENS, causing a 400 and a permanently lost observation
# OpenRouter `truncateHistory` returns an empty array when one message exceeds `MAX_TOKENS`, causing a 400 and a permanently lost observation

**Version:** 12.4.7
**Provider:** OpenRouter (`CLAUDE_MEM_PROVIDER=openrouter`)
**Platform:** macOS 26.5.2, arm64, node v26.4.0, bun 1.3.13

## Summary

The OpenRouter copy of `truncateHistory` is missing a `s.length > 0` guard that the Gemini copy has. When the newest message on its own estimates above `CLAUDE_MEM_OPENROUTER_MAX_TOKENS`, the function returns `[]`. That empty array is sent as `messages`, OpenRouter rejects it with `400 Input required`, and the observation is marked failed and never written.

On one affected install this silently discarded **1,231 observations, 4.9% of all attempts**. Nothing surfaces to the user: the session continues normally and the memory simply is not there later.

The date correlation is the clearest evidence. Logs spanning 27 April 2026 to the present contain **zero** occurrences before **18 July**, which is two days after `CLAUDE_MEM_OPENROUTER_MAX_TOKENS` was changed from the default `100000` to `12000`. From that point they run continuously.

## The two implementations

Both live in `scripts/worker-service.cjs` in the published bundle. The sibling methods are `estimateTokens` / `conversationToOpenAIMessages` for the OpenRouter one, and `conversationToGeminiContents` for the Gemini one.

```js
// Gemini
if (s.length > 0 && (s.length >= n || o + u > i)) { warn(...); break }

// OpenRouter
if (                s.length >= n || o + u > i ) { warn(...); break }
```

The loop walks newest-first. On the first iteration `s` is empty and `o` is 0, so a single message with `estimateTokens(content) > i` satisfies `o + u > i` immediately and breaks having kept nothing.

`estimateTokens` is `Math.ceil(content.length / 4)`, so the practical threshold is `MAX_TOKENS * 4` characters in any one message.

## Log fingerprint

```
[WARN ] [SDK] Context window truncated to prevent runaway costs
        {originalMessages=4, keptMessages=0, droppedMessages=4, estimatedTokens=0, tokenLimit=12000}
[ERROR] [SDK] OpenRouter API error: 400 - {"error":{"message":"Input required: specify \"prompt\" or \"messages\"","code":400}}
[ERROR] [SESSION] Marked messages as failed after generator error {failedCount=1}
```

`keptMessages=0` with `estimatedTokens=0` is diagnostic. Note that `estimatedTokens=0` is the accumulator that was never incremented, not a failed estimator, which is an easy thing to misread when triaging.

To count occurrences in an affected install:

```bash
# total events
grep -rho "keptMessages=0" ~/.claude-mem/logs/*.log | wc -l

# which days, useful for correlating against a MAX_TOKENS change
grep -rl "keptMessages=0" ~/.claude-mem/logs/claude-mem-*.log | sed 's|.*claude-mem-||;s|\.log||' | sort
```

## Reproduction

```js
const est = c => Math.ceil(c.length / 4);

// verbatim from the OpenRouter branch
function truncateHistory(e, n, i) {
  if (e.length <= n && e.reduce((c, u) => c + est(u.content), 0) <= i) return e;
  let s = [], o = 0;
  for (let a = e.length - 1; a >= 0; a--) {
    const c = e[a], u = est(c.content);
    if (s.length >= n || o + u > i) return s;
    s.unshift(c); o += u;
  }
  return s;
}

const msgs = [
  { content: 'a'.repeat(500) },
  { content: 'b'.repeat(500) },
  { content: 'c'.repeat(500) },
  { content: 'd'.repeat(60000) },   // newest, ~15,000 tokens
];

console.log(truncateHistory(msgs, 6, 12000));  // []  <-- sent as `messages`, 400
```

The Gemini version with the same input returns one message, as intended.

## Why the default hides it

`CLAUDE_MEM_OPENROUTER_MAX_TOKENS` defaults to `100000`, so it takes a single 400,000 character message to trigger. Lowering the setting, which is exposed as a user-configurable option in `SettingsDefaultsManager.DEFAULTS`, moves it into the range where ordinary large tool outputs trip it. At `12000` the threshold is 48,000 characters, which a single file read or grep result clears easily.

That makes this look like a rare edge case in testing while being routine for anyone who tunes the setting down.

## Suggested fix

Add the same guard the Gemini branch already has, so at least one message always survives:

```js
if (s.length > 0 && (s.length >= n || o + u > i)) { warn(...); break }
```

Two things worth considering alongside it:

1. **Never send an empty `messages` array.** Even with the guard, an explicit check before dispatch would turn a silent data loss into a clear error.
2. **Truncate the oversized message rather than dropping it.** A single message above the ceiling currently cannot be summarised at all. Slicing its content to fit would preserve the observation instead of discarding it.

## Workaround for affected users

Raise the ceiling. This does not increase normal cost, since the limit is a ceiling rather than a target and `CLAUDE_MEM_OPENROUTER_MAX_CONTEXT_MESSAGES` is what binds in practice:

```jsonc
// ~/.claude-mem/settings.json
"CLAUDE_MEM_OPENROUTER_MAX_TOKENS": "50000"
```

No worker restart is needed, as `truncateHistory` calls `loadFromFile` (a plain `readFileSync`) on every invocation.

After that change: 0 failures, and truncation events now correctly report `keptMessages=6, tokenLimit=50000`, which is the message-count cap binding rather than the token cap.

This is only a mitigation. It moves the threshold from 48,000 to 200,000 characters, it does not remove the failure mode.

stantheman0128:
## Summary

- Normalize OpenRouter conversation history the same way Gemini already does: skip blank turns, merge adjacent same-role messages, and fall back to the newest non-empty prompt.
- Prevents OpenRouter requests with messages: [], which returned HTTP 400 and dropped observations.

## Context

v13.9.2 removed centralized 	runcateHistory(), which was the original source of the empty-array bug on 12.4.7. This patch closes the remaining provider gap: OpenRouter still forwarded blank history verbatim while Gemini filtered it.

## Verification

\\\
cd claude-mem-wt/3491-openrouter-truncate-guard
bun test ./tests/worker/openrouter-messages.test.ts
# 4 pass
\\\

## AI assistance

Fixes #3491

### DIFF
--- a/src/services/worker/OpenRouterProvider.ts
+++ b/src/services/worker/OpenRouterProvider.ts
@@ -95,6 +95,7 @@ export function classifyOpenRouterError(input: {
 }
 
 const CHARS_PER_TOKEN_ESTIMATE = 4;
+const OPENROUTER_EMPTY_HISTORY_FALLBACK = '(context unavailable)';
 
 interface OpenAIMessage {
   role: 'user' | 'assistant' | 'system';
@@ -180,10 +181,42 @@ export class OpenRouterProvider extends OpenAICompatibleProvider<OpenRouterConfi
   }
 
   private conversationToOpenAIMessages(history: ConversationMessage[]): OpenAIMessage[] {
-    return history.map(msg => ({
-      role: msg.role === 'assistant' ? 'assistant' : 'user',
-      content: msg.content
-    }));
+    let newestNonEmptyContent: string | null = null;
+    for (const msg of history) {
+      const trimmed = msg.content.trim();
+      if (trimmed.length > 0) {
+        newestNonEmptyContent = trimmed;
+      }
+    }
+
+    const messages: OpenAIMessage[] = [];
+    for (const msg of history) {
+      const trimmed = msg.content.trim();
+      if (!trimmed) {
+        continue;
+      }
+
+      const role = msg.role === 'assistant' ? 'assistant' : 'user';
+      if (messages.length === 0 && role === 'assistant') {
+        continue;
+      }
+
+      const previous = messages[messages.length - 1];
+      if (previous?.role === role) {
+        previous.content = `${previous.content}\n\n${msg.content}`;
+      } else {
+        messages.push({ role, content: msg.content });
+      }
+    }
+
+    if (messages.length === 0) {
+      return [{
+        role: 'user',
+        content: newestNonEmptyContent ?? OPENROUTER_EMPTY_HISTORY_FALLBACK,
+      }];
+    }
+
+    return messages;
   }
 
   protected async query(history: ConversationMessage[], config: OpenRouterConfig): Promise<ProviderQueryResult> {
@@ -234,7 +267,7 @@ export class OpenRouterProvider extends OpenAICompatibleProvider<OpenRouterConfi
   ): Promise<ProviderQueryResult> {
     const messages = this.conversationToOpenAIMessages(history);
     const totalChars = history.reduce((sum, m) => sum + m.content.length, 0);
-    const estimatedTokens = this.estimateTokens(history.map(m => m.content).join(''));
+    const estimatedTokens = this.estimateTokens(messages.map(m => m.content).join(''));
 
     logger.debug('SDK', `Querying OpenRouter multi-turn (${model})`, {
       turns: history.length,
--- a/tests/worker/openrouter-messages.test.ts
+++ b/tests/worker/openrouter-messages.test.ts
@@ -1,0 +1,89 @@
+import { describe, it, expect, beforeEach, spyOn, mock } from 'bun:test';
+import { OpenRouterProvider } from '../../src/services/worker/OpenRouterProvider.js';
+import { DatabaseManager } from '../../src/services/worker/DatabaseManager.js';
+import { SessionManager } from '../../src/services/worker/SessionManager.js';
+import { SettingsDefaultsManager } from '../../src/shared/SettingsDefaultsManager.js';
+import type { ConversationMessage } from '../../src/services/worker-types.js';
+
+class TestOpenRouterProvider extends OpenRouterProvider {
+  buildMessages(history: ConversationMessage[]) {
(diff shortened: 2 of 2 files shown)


stantheman0128:
Thanks for the re-review.

**Mock isolation (P1):** Addressed in 3185ff7d — `global.fetch` and `SettingsDefaultsManager.loadFromFile` spies are both restored in `afterEach`.

**Leading assistant skip (P1):** Intentional parity with `GeminiProvider.conversationToGeminiContents` (see `GeminiProvider.ts` ~270: `contents.length === 0 && role === 'model'` also skips). OpenAI-compatible APIs expect a user-leading payload; this PR targets #3491 (HTTP 400 on `messages=[]`), not a broader history-shape change. Happy to preserve leading assistant in a follow-up if maintainers want that behavior unified across providers.

thedotmack:

> step 4 (provider adapter parity: history normalization for OpenRouter); one open thread on leading-assistant turn

thedotmack:
**All-night ship gate: PASS**

1. RCA logged — v13.9.2 removed centralized `truncateHistory()`, so OpenRouter still forwarded blank history as `messages: []` and the API returned HTTP 400 (Fixes #3491 / #3606 step 4).
2. No second-system — same blank-turn skip / same-role merge / newest-non-empty fallback Gemini already uses.
3. Narrow / not opinionated — one converter + token estimate from the payload actually sent + tests.

Rebased onto `7335467f` locally (fork branch update blocked). Focused tests: `bun test ./tests/worker/openrouter-messages.test.ts` — 4 pass. CI on this head was green. Merging independently; no version bump.

PR #3498, opened 2026-08-06
Add secure CMEM Managed Worker connection

ChenglinWei97:
## What changed

- add `npx claude-mem cloud connect` with masked local prompts for the sync token and CMEM-only worker key
- validate the fixed HTTPS gateway path and configure the server-selected `cmem-managed` model alias
- write credential-bearing settings atomically with mode 0600 from the first byte
- update the cloud-sync skill so agents never request or print customer secrets
- rebuild the tracked plugin runtime artifacts

## Why

CMEM Pro needs a local setup path that does not put credentials in command arguments, shell history, AI prompts, or logs. The customer receives only a CMEM gateway key; the provider credential and actual model remain server-side.

The paired CMEM Pro application change is tracked privately.

## Validation

- `bun test tests`: 2,563 passed, 18 environment-dependent skips, 0 failed
- focused connection/settings tests: 66 passed
- `npm run typecheck`
- `npm run lint:hook-io`
- `npm run lint:spawn-env`
- `npm run build`
- `git diff --check`

## Rollout note

Publish this CLI before deploying the Pro Connect page that emits the new command.


ChenglinWei97:
Addressed the Greptile P1 in c2873a32. cloud connect now rejects non-CMEM hosts, look-alike domains, custom ports, and wrong worker paths before prompting for or persisting a secret. It still accepts both existing 24-character and current 32-character CMEM sync tokens. Focused tests, root typecheck, and the full distribution build pass.

thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3618 ([plan-24] Server Runtime & Cloud Sync Contract)** as a **capability / product decision** — tracked on the plan master but not part of the defect fix sequence.

> `cloud connect` CLI for CMEM Pro; depends on private app change

thedotmack:
Routing to **#2785 (plan-12 Provider & Extensibility Roadmap)** — plan-12 collects Pro/billing surfaces and new provider paths, and `cloud connect` plus the server-selected `cmem-managed` model alias is exactly that. The 0600 atomic credential write overlaps #3472, which is rebuilding the settings-document boundary under plan-21 (#3609); worth rebasing onto that rather than adding a second atomic-write path. Leaving open (draft, paired with a private app-side change).

PR #3501, opened 2026-08-06
fix(observer): skip textless assistant frames before batch confirm (Fixes #3492)

### LINKED ISSUE #3492: Observer never leaves the INIT prompt: idle eviction discards memory_session_id on a worker that never restarted — every observation batch dropped (observations=0, Windows)
## Summary

On Windows, the observer never advances past its **INIT prompt**. Every SDK query for the
session is created with `isInitPrompt=true, promptType=INIT` and `historyLength=0`, so the
agent is asked to observe while holding no tool data. It correctly answers *"No observations
to record yet"* (or returns an empty string), the parser classifies that as `prose`/`idle`,
and **the queued batch is dropped**. The result is a permanent `observations = 0`.

The `idle`/`prose` discard itself is already reported (#3454, #3351, #3193). What this issue
adds is the **upstream reason the agent has nothing to say**: the SDK conversation is thrown
away between batches, so the observer restarts from prompt #1 every time.

```
observations = 0   sessions = 1   summaries = 2
```

Summaries work. `<summary>` XML parses and persists correctly on the same SDK, same auth —
so this is not an auth or grammar failure.

## Environment

| | |
|---|---|
| claude-mem | plugin cache `13.13.0`, running worker reports `13.13.1` (see *Version split* below) |
| Host | Claude Code 2.1.220 (npm global) |
| OS | Windows 11 Pro 10.0.26200 |
| Runtime | node v24.11.0, `CLAUDE_MEM_RUNTIME=worker` |
| Provider | Claude Code OAuth token (read from system keychain at spawn) |
| uv | uvx 0.12.1 |

## Root cause

Two code sites combine into a loop.

**1. `src/services/worker/SessionManager.ts:88-94`** — when the session is not in the
in-memory map, the stored `memory_session_id` is discarded unconditionally:

```ts
if (dbSession.memory_session_id) {
  logger.warn('SESSION', `Discarding stale memory_session_id from previous worker instance (Issue #817)`, {
    sessionDbId,
    staleMemorySessionId: dbSession.memory_session_id,
    reason: 'SDK context lost on worker restart - will capture new ID'
  });
}
```

The stated reason — *"SDK context lost on worker restart"* — did not hold in my capture.
**The worker had 26.2 minutes of continuous uptime and never restarted.** Absence from the
in-memory map is being read as evidence of a restart, and it is not.

**2. `src/services/worker/SessionManager.ts:316-336` + the 180 s idle abort** — what actually
empties the map:

```
[SESSION] Idle timeout reached, triggering abort to kill subprocess {idleDurationMs=180016, thresholdMs=180000}
[SESSION] Generator exited — finalizing session {reason=idle}
[SESSION] Session removed from active sessions {project=importacao}
```

Three minutes of thinking, reading, or waiting on a long command is ordinary in a coding
session. Every time it happens, the next hook rebuilds the session from the DB, hits site 1,
discards the SDK conversation, and the observer is back on prompt #1.

### A. Unassisted session — continuous, no idle gap, still 0 stored

```
09:20:22.667 [HOOK  ] [session-1] INIT_COMPLETE | sessionDbId=1 | promptNumber=1
09:20:23.391 [SDK   ] Starting SDK query {hasRealMemorySessionId=false, shouldResume=false,
                                          resume_parameter=(none - fresh start), lastPromptNumber=1}
09:20:23.477 [SDK   ] Creating message generator {isInitPrompt=true, promptType=INIT}
09:20:31.064 [WARN ] [PARSER] SDK returned non-XML idle response — ignoring queued batch
                              {outputClass=idle, preview=, consecutiveInvalidOutputs=0}
09:20:31.290 [SDK   ] ← Response received (99 chars)
                      (No observations to record yet - waiting for tool executions and results
                       from the primary session.)
09:20:31.291 [WARN ] [PARSER] SDK returned non-XML prose response — ignoring queued batch
09:20:32.692 [QUEUE ] ENQUEUED | messageId=1 | type=observation | tool=Read(…/package.json)
09:20:41.630 [WARN ] [PARSER] … idle … ignoring queued batch
09:20:53.415 [QUEUE ] ENQUEUED | messageId=2 | type=observation | tool=PowerShell
09:21:03.070 [WARN ] [PARSER] … idle … ignoring queued batch
09:21:10.570 [WARN ] [PARSER] … idle … ignoring queued batch
09:21:21.048 [HOOK  ] → Stop: Requesting summary {hasLastAssistantMessage=true}
```

Note the ordering at the top: the agent is answering *"waiting for tool executions"* **before**
the first observation is even enqueued. It is replying to the INIT prompt, and the batch queued
behind it is consumed and discarded by that reply.

### B. The Stop path succeeds on the same SDK

```
09:22:05.080 [SDK] ← Response received (1318 chars) <summary>…
09:22:05.080 [DB ] STORING | sessionDbId=1 | obsCount=0 | hasSummary=true
09:22:05.081 [DB ] STORED  | sessionDbId=1 | obsCount=0 | obsIds=[] | summaryId=2
```

`obsCount=0 | obsIds=[]` on a successful store — the summarizer works, the observer contributed
nothing.

### C. Idle eviction, then the discard, then INIT again

```
09:24:57.495 [SESSION] Idle timeout reached {idleDurationMs=180016, thresholdMs=180000}
09:24:57.998 [SESSION] Session removed from active sessions
09:26:50.460 [HOOK   ] → PostToolUse: Read(…/src/lib/status.ts)
09:26:50.673 [WARN  ] [SESSION] Discarding stale memory_session_id from previous worker instance (Issue #817)
                                {staleMemorySessionId=68a7f6f2-7d20-4ffa-9e60-9a9f3a58ea90,
                                 reason=SDK context lost on worker restart - will capture new ID}
09:26:50.673 [SESSION] Session initialized {queueDepth=0, hasGenerator=false}
09:26:51.290 [SDK   ] Creating message generator {isInitPrompt=true, promptType=INIT}
09:26:59.691 [WARN ] [PARSER] … idle … ignoring queued batch
09:26:59.894 [SDK   ] ← Response received (258 chars)
                      I have no substantive tool execution data or outcomes to record from the
                      observed session. The provided context only shows a generic request to run
                      the project with no resulting technical signal, logs…
```

Worker uptime at 09:26:50 was **1572 s**, unbroken since 09:03 — no restart occurred.

## Reproduction

Capture **A** above is from an ordinary, unassisted session (`"run the project"`). I then
reproduced capture **C** deliberately by piping a hand-built payload into the hook command
extracted from `hooks/hooks.json`, to confirm the path end to end:

```bash
# PostToolUse payload with tool_name + tool_input + tool_response fully populated
cat obs-payload.json | bash obs-hook.sh   # → {}  exit 0
```

The hook returns cleanly, the batch enqueues (`messageId=7 | type=observation`), and the batch
is discarded. Both captures show the same `promptType=INIT`.

To be explicit about method: capture C followed a >180 s gap, so its eviction was self-inflicted.
Capture A had no such gap and dropped every batch anyway — the conclusion does not rest on the
manual run.

## Suggested direction

1. Stop treating *"absent from the in-memory map"* as *"the worker restarted."* The discard at
   `SessionManager.ts:88` could be gated on an actual worker-instance epoch/boot id compared
   against one persisted with the session, instead of firing whenever the map lookup misses.
2. Re-attach to the stored `memory_session_id` via `resume` when the worker is provably the
   same instance, so an idle eviction costs a subprocess but not the conversation.
3. Failing both, do not let a batch be consumed by the INIT prompt: hold the queued batch until
   the generator is past INIT, rather than dropping it against the INIT reply.

Related symptom reports, same pipeline, different layer: #3454 (idle volume + dead
`consecutiveInvalidOutputs`), #3351 (no salvage/retry on discard), #3193, #3083 (Windows).

## Secondary: Chroma failure is cold-start budget, not connectivity

Every vector sync failed:

```
[CHROMA_MCP] Connecting to chroma-mcp via MCP stdio {command=…\uvx.exe,
             args=--python 3.13 --with onnxruntime>=1.20 --with protobuf<7 --from chroma-mcp==0.2.6 …}
[WARN ] [CHROMA_MCP] Connection failed, killing subprocess tree {error=MCP connection to chroma-mcp
             timed out after 30000ms, stderrTail=Installed 104 packages in 25.27s}
[ERROR] [CHROMA] SDK chroma sync failed, continuing without vector search {summaryId=2}
```

`stderrTail` gives it away: `uvx` spent 25.27 s of the 30 s budget installing 104 packages. I
timed the same command warm:

```
uvx --python 3.13 --with 'onnxruntime>=1.20' --with 'protobuf<7' --from chroma-mcp==0.2.6 chroma-mcp --help
→ 8.1 s
```

So the 30 s ceiling is only ever breached on a cold uv cache — i.e. on exactly the first run
after install, when a new user is forming their first impression. The prewarm at
`INIT_COMPLETE` already allows 120 s (`timeoutMs=120000`); the connect path allowing 30 s
means the prewarm and the connect can race on a fresh machine. Worth aligning the two budgets,
or making the first connect inherit the prewarm's.

## Also noted: version split between hooks and worker

`installed_plugins.json` pins the plugin at `13.13.0`:

```
C:\Users\<user>\.claude\plugins\cache\thedotmack\claude-mem\13.13.0
```

and the hooks/skills resolve there, but the running worker reports `13.13.1` and was started
from the marketplace checkout:

```
[SYSTEM] Starting worker daemon {workerScriptPath=…\plugins\marketplaces\thedotmack\plugin\scripts\worker-service.cjs}
GET /api/stats → {"worker":{"version":"13.13.1", …}}
```

The path list in `hooks.json` ends with `…/plugins/marketplaces/thedotmack/plugin` as a
fallback, so on a machine where the marketplace checkout is newer than the installed version,
hooks and worker come from two different builds. Probably benign here, but it makes version
numbers in bug reports unreliable.

stantheman0128:
## Summary

- Skip `processAgentResponse` when a Claude assistant frame has no `text` content block (thinking-only or tool_use-only frames).
- Prevents `ResponseProcessor` from classifying the flattened `''` as idle and calling `confirmClaimedMessages()` before the real observation XML frame arrives.

## Context

Fixes #3492. Reporter logs showed `idle` with an empty preview immediately before a 258-char XML response; observations stayed at 0 while summaries still stored.

**Level of fix:** `ClaudeProvider` only (Gemini already skips empty responses; OpenRouter uses `forwardEmptyMessageResponse`). Frames that include a `text` block still flow through, including empty or idle prose, so the existing confirm-and-drop path for true idle answers is unchanged.

**Bot / edge-case checks**

| Path | Before | After |
| --- | --- | --- |
| tool_use-only frame then XML frame | `''` classified idle, batch dropped | batch kept, XML stores observation |
| empty `text` block frame | confirm-and-drop | unchanged |
| idle prose text frame | confirm-and-drop | unchanged |

## Verification

```
cd claude-mem-wt/3492-observer-init-loop
bun test tests/worker/claude-provider-assistant-frames.test.ts
# 6 pass
bun test tests/worker/agents/response-processor.test.ts
# 23 pass
bun run typecheck
```

## AI assistance

Fixes #3492

### DIFF
--- a/src/services/worker/ClaudeProvider.ts
+++ b/src/services/worker/ClaudeProvider.ts
@@ -273,6 +273,10 @@ export class ClaudeProvider {
       // Textless frames are not dispatched (see below), so their usage rolls
       // into the next dispatch instead of going unattributed.
       let discoveryTokenBaseline = session.cumulativeInputTokens + session.cumulativeOutputTokens;
+      // Whether the current turn has dispatched any text yet. A turn that ends
+      // without one still needs the idle hand-off, otherwise the claimed batch
+      // is left dangling for session teardown to discard.
+      let turnDispatchedText = false;
 
       for await (const message of queryResult) {
         // Quota-aware wall-clock guard (#2234): the SDK pushes `system` events
@@ -339,7 +343,9 @@ export class ClaudeProvider {
           // confirm-and-drop the claimed batch before the real XML frame
           // arrives (#3492). A frame that does contain a text block still goes
           // through even when that text is empty: that is the provider saying
-          // it had nothing to record, which stays a confirmed no-op batch.
+          // it had nothing to record, which stays a confirmed no-op batch. A
+          // turn that never produces a text frame gets the same idle hand-off
+          // once, from the `result` branch below.
           const hasTextBlock = Array.isArray(content)
             ? content.some((c: any) => c?.type === 'text')
             : typeof content === 'string';
@@ -422,6 +428,7 @@ export class ClaudeProvider {
           );
 
           discoveryTokenBaseline = session.cumulativeInputTokens + session.cumulativeOutputTokens;
+          turnDispatchedText = true;
         }
 
         if (message.type === 'result') {
@@ -467,6 +474,30 @@ export class ClaudeProvider {
                   : undefined,
             });
           }
+
+          if (!turnDispatchedText) {
+            // The whole turn carried no text, so no frame handed the claimed
+            // batch to the parser. Do it once here with the empty response the
+            // turn actually produced: the batch is classified idle and
+            // confirmed, exactly as before #3492's frame-level skip. Skipping
+            // this would leave the batch claimed until session teardown
+            // disposes the in-RAM buffer, which drops it with no hand-off.
+            await processAgentResponse(
+              '',
+              session,
+              this.dbManager,
+              this.sessionManager,
+              worker,
+              (session.cumulativeInputTokens + session.cumulativeOutputTokens) - discoveryTokenBaseline,
+              session.earliestPendingTimestamp,
+              'SDK',
+              cwdTracker.lastCwd,
+              modelId,
+              activeResponseContext.current
+            );
+            discoveryTokenBaseline = session.cumulativeInputTokens + session.cumulativeOutputTokens;
+          }
+          turnDispatchedText = false;
         }
       }
     } finally {
--- a/src/services/worker/ClaudeProvider.ts
+++ b/src/services/worker/ClaudeProvider.ts
@@ -269,6 +269,11 @@ export class ClaudeProvider {
         }),
       });
 
+      // Baseline for the next dispatched response's discovery-token delta.
+      // Textless frames are not dispatched (see below), so their usage rolls
+      // into the next dispatch instead of going unattributed.
+      let discoveryTokenBaseline = session.cumulativeInputTokens + session.cumulativeOutputTokens;
+
       for await (const message of queryResult) {
         // Quota-aware wall-clock guard (#2234): the SDK pushes `system` events
         // with subtype `rate_limit` carrying live subscription quota state.
@@ -327,13 +332,22 @@ export class ClaudeProvider {
 
         if (message.type === 'assistant') {
           const content = message.message.content;
+          // A turn can arrive as several assistant messages, and a frame that
+          // holds only thinking or tool_use blocks carries no text at all.
+          // Flattening such a frame to '' and handing it to
+          // processAgentResponse makes the parser read the turn as idle and
+          // confirm-and-drop the claimed batch before the real XML frame
+          // arrives (#3492). A frame that does contain a text block still goes
+          // through even when that text is empty: that is the provider saying
+          // it had nothing to record, which stays a confirmed no-op batch.
+          const hasTextBlock = Array.isArray(content)
+            ? content.some((c: any) => c?.type === 'text')
+            : typeof content === 'string';
           const textContent = Array.isArray(content)
             ? content.filter((c: any) => c.type === 'text').map((c: any) => c.text).join('\n')
             : typeof content === 'string' ? content : '';
 
           const responseSize = textContent.length;
-
-          const tokensBeforeResponse = session.cumulativeInputTokens + session.cumulativeOutputTokens;
 
           const usage = message.message.usage;
           if (usage) {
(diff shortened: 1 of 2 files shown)


thedotmack:

> step 1 (thinking-only/tool_use-only frames are not `idle`; text-block gating)

thedotmack:
Night-ship gate (plan-18 batch A, after #3363):
1. RCA: YES — logged in PR body + #3492; tracked under #3606 (thinking-only/tool_use-only frames are not `idle`).
2. No second-system: YES — ClaudeProvider frame dispatch only; Gemini already skips empty responses.
3. Narrow: YES — skip textless assistant frames; empty/idle text blocks still confirm-and-drop.

Rebase onto latest main is clean locally. Focused tests next, then squash-merge if CI-equivalent local suite is green. Cannot push rebase to the fork (maintainer update-branch 403).

Issue #3497, opened 2026-08-05
Observer conversationHistory is never trimmed — ~10x token amplification billed silently to the user's subscription


sideby-alex:
## Before submitting

- [x] I searched [existing issues](https://github.com/thedotmack/claude-mem/issues) and confirmed this is not a duplicate

---

## 📝 Manual Bug Report

> Filed manually rather than via `npm run bug-report`: the plugin has already been uninstalled from this machine (that was the mitigation), so the generator can no longer collect live diagnostics. All numbers below were measured from the on-disk state before removal.

### Bug Description

**The observer's `conversationHistory` is append-only and never trimmed, so every tool call re-bills the entire accumulated observer conversation. On a busy day this amplified observer input tokens by ~10x over the actual volume of new material.**

Because `CLAUDE_MEM_CLAUDE_AUTH_METHOD` defaults to `subscription`, that amplified cost lands on the user's own Claude subscription, and there is nothing in the UI or in INFO-level logs that surfaces it.

The mechanism:

1. `hooks/hooks.json` registers `PostToolUse` with matcher `*`, so every tool call in every session is forwarded to the worker.
2. `ClaudeProvider.ts:528` does `session.conversationHistory.push({ role: 'user', content: obsPrompt })` for each observation. The same push-only pattern appears at `ClaudeProvider.ts:490` / `:552`, `OpenAICompatibleProvider.ts:105` / `:215` / `:265`, and `ResponseProcessor.ts:294`.
3. Nothing ever removes entries. `grep -rn 'conversationHistory.\(splice\|shift\|slice\|length =\|pop\)\|trimHistory\|pruneHistory' src/` returns **zero matches** across the whole tree. `SessionManager.ts:127` initializes it to `[]` and it only grows from there.
4. For the Claude CLI provider the array isn't resent explicitly — but the persistent `claude --input-format stream-json` child holds the equivalent context itself, so the billing shape is identical. For the OpenAI-compatible provider it *is* resent literally (`this.query(session.conversationHistory, config)` at `OpenAICompatibleProvider.ts:218`), which makes the growth explicit and easy to confirm.

[...]

PR #3508, opened 2026-08-06
fix(observer): surface cumulative token usage on session completion (Related to #3497)

stantheman0128:
## Summary

- Add shared `observer-usage` helpers so OpenAI-compatible providers accumulate **real** `inputTokens`/`outputTokens` when reported (Gemini, OpenRouter) instead of always applying a 70/30 split on `tokensUsed`.
- Track Claude `cache_read_input_tokens` in `cumulativeCacheReadTokens` (separate from `cumulativeInputTokens` so `discovery_tokens` semantics stay intact).
- Attach cumulative observer token fields to the once-per-session **completion** and **error** log lines at INFO for Claude and OpenAI-compatible providers.

## Context

Related to #3497. The reporter's ~10x amplification on the **default Claude subscription path** comes from the persistent `claude` CLI child context, not from trimming `conversationHistory` (that array is write-only on the Claude path). Maintainer removed client-side history caps in `29af0284` ("second system" anti-pattern), so this PR does **not** reintroduce truncation.

What it does address from the issue: cost was invisible because cumulative counters were DEBUG-only (Claude) or never logged (OpenAI-compatible), and HTTP providers used a 70/30 estimate that understates input on history-resend sessions.

**Level of fix:** shared `observer-usage.ts` + session completion boundaries. Rejected: sliding window / `CLAUDE_MEM_MAX_HISTORY_TOKENS` (reverted upstream; competing PRs #3151/#2957 on community-edge).

## Verification

```
cd claude-mem-wt/3497-trim-conversation-history
bun test tests/worker/observer-usage-visibility.test.ts
# 11 pass
bun run typecheck
```

## AI assistance

Related to #3497

### DIFF
--- a/src/services/worker-types.ts
+++ b/src/services/worker-types.ts
@@ -58,6 +58,14 @@ export interface ActiveSession {
   pendingCompressionEvent?: Record<string, unknown> | null;
   /** Cumulative total_cost_usd from the SDK's latest result message — per-compression cost is the delta between results. */
   lastResultTotalCostUsd?: number | null;
+  /**
+   * Cumulative cache_read_input_tokens across the session. Kept apart from
+   * cumulativeInputTokens because discovery_tokens is the delta of that
+   * counter; on a long observer session this is where most of the context the
+   * model re-reads shows up, so it is the number that makes resend growth
+   * visible.
+   */
+  cumulativeCacheReadTokens?: number;
 }
 
 export interface PendingMessage {
--- a/src/services/worker/ClaudeProvider.ts
+++ b/src/services/worker/ClaudeProvider.ts
@@ -28,6 +28,7 @@ import { query } from '@anthropic-ai/claude-agent-sdk';
 import { buildHardenedSdkOptions } from '../../sdk/hardened-options.js';
 import { ClassifiedProviderError } from './provider-errors.js';
 import { resolveTierAlias } from './model-aliases.js';
+import { accumulateClaudeUsage, observerUsageLogFields } from './observer-usage.js';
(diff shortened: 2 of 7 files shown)


stantheman0128:
Greptile P2 addressed in 4bb6a592: \Generator failed\ now spreads \observerUsageLogFields(session)\ so accrued input/output/cache-read totals survive Claude stream failures. Contract covered in observer-usage-visibility.test.ts.

stantheman0128:
Greptile P1 addressed in 007bab0b: \ccumulateObserverUsage\ / \lastUsage\ now run for init, observation, and summary turns even when the provider returns an empty body, as long as the gateway reports usage. History insertion and \processAgentResponse\ stay behind the content guard.

Regression: \ccumulates billed usage from an empty-bodied observation response\ in \	ests/worker/observer-usage-visibility.test.ts\ (init 90/10 + empty obs 990/10 → cumulative 1080/20).

@greptile retest

thedotmack:

> step 3 (cumulative tokens logged INFO, exposed on health; real usage counts, cache-read tokens); conflicting

thedotmack:
Routing to **#3606 (plan-18 Observer Response Pipeline)** — accurate per-provider token accounting is what makes plan-18's "budget/recycle policy" enforceable, so the shared `observer-usage` helpers belong in that cluster. Deliberately not a duplicate of #3516: that one bounds `conversationHistory`, this one stops guessing a 70/30 split and reports what the provider actually returned, and your own note that the Claude-path amplification comes from the persistent CLI child rather than the history array is the useful finding. Leaving open.

PR #3510, opened 2026-08-07
fix: release SDK agent subprocess when a session ends (missing SessionEnd hook)

jkaizzol:
## Problem

`SDKAgent.startSession()` keeps **one `claude` subprocess alive per session**, held open by `SessionManager.getMessageIterator()` — an infinite async generator. That's by design: the memory agent should live for the whole session.

The only thing that aborts it is `SessionManager.deleteSession()` (which calls `abortController.abort()`), and that is reachable only from `shutdownAll()` or the `sessionDbId`-keyed HTTP endpoints.

**No `SessionEnd` hook is registered.** `plugin/hooks/hooks.json` declares only `SessionStart`, `UserPromptSubmit`, `PostToolUse` and `Stop`, even though `CLAUDE.md` documents *"5 Lifecycle Hooks: SessionStart → UserPromptSubmit → PostToolUse → Summary → SessionEnd"* and `SessionRoutes.ts` already carries this comment:

```
cleanup-hook expects POST /sessions/:sessionDbId/complete instead of DELETE
```

The endpoint exists, waiting for a hook that was never wired up. So nothing ever tells the worker a session finished, and each ~400MB subprocess survives until reboot.

### Measured impact

On a 24GB M4 Pro running `CLAUDE_MEM_PROVIDER=claude`:

- 4 subprocesses accumulated in 11 minutes of uptime, ~400MB each
- 11 alive simultaneously during a working day
- swap reached 20GB sustained, peaking at 36.6GB
- the machine froze with only a couple of apps open

They aren't visible across reboots — the accumulation happens within a single boot.

## Fix

- **`src/hooks/cleanup-hook.ts`** — new hook on `SessionEnd`, posts `session_id` to the worker. It deliberately does *not* call `ensureWorkerRunning()`: booting a worker just to tell it about a session it never had would spawn the very process this hook exists to reclaim.
- **`POST /api/sessions/complete`** — mirrors the existing `contentSessionId` endpoints, because hooks only know Claude Code's `session_id`, not `sessionDbId`.
- **`SessionStore.getSessionIdByContentSessionId()`** — a pure lookup. Reusing `createSDKSession()` here would `INSERT` a row for every unknown `session_id` the hook sees — creating sessions while tearing one down.
- **`scripts/build-hooks.js`** — the `HOOKS` list is hardcoded, so the new hook is added there.

## Verification

- With a live session holding a **399MB** subprocess, running the compiled hook **released it**.
- Unknown session → `{"status":"skipped","reason":"unknown session"}`; missing `contentSessionId` → `400`.
- `tsc --noEmit` reports the **same 308 pre-existing errors as `main`, none new** (compared branch vs. base, not eyeballed).

## Note on a related dead field

While tracing this I noticed `sdk_sessions.status` is never set to `'completed'` anywhere in the codebase — it's written once with the schema `DEFAULT 'active'` and never updated. My local database had 184 rows, all `'active'`, the oldest from January. That's cosmetic and independent of this leak, so I left it alone, but you may want to either populate it or drop it. Happy to fix in a follow-up if useful.


jkaizzol:
Closing this — my mistake, and apologies for the noise.

I diagnosed this against the version installed on my machine (**8.5.9**) without first checking `main`. On re-checking, the leak is **already fixed upstream**, just not the way I proposed:

- `SessionManager` now aborts on an idle timeout — the log line is literally `'Triggering abort due to idle timeout to kill subprocess'`
- `src/supervisor/process-registry.ts` tracks SDK processes and reaps them with SIGTERM/SIGKILL; it doesn't exist in 8.5.9 at all
- `GeneratorExitHandler` calls `ensureSdkProcessExit()` and finalizes the session on generator exit

So the `SessionEnd` hook I proposed isn't needed — the teardown happens on idle abort instead, which is a better design since it doesn't depend on a hook firing.

For anyone landing here from a search: the symptom on 8.5.9 was `claude` subprocesses (~400MB each) accumulating for the lifetime of a boot — I measured 4 in 11 minutes and 11 alive simultaneously, which pushed a 24GB machine to 20GB of swap and hard UI freezes. **The fix is to update the plugin**, not to patch it.

One unrelated observation that may still hold on `main`, offered in case it's useful: `sdk_sessions.status` appears to never be set to `'completed'` — written once via the schema `DEFAULT 'active'` and never updated. My local DB had 184 rows, all `'active'`, oldest from January. Harmless, but the column may be dead weight. Happy to open a separate issue if you'd like.

Thanks for the project.

Issue #2393, opened 2026-05-09
Feature: Model inheritance for local LM Studio users


0xLeathery:
# Feature Request: Model Inheritance for Local LM Studio Users

## Problem Description

When using claude-mem with locally-hosted models via LM Studio (or similar local inference servers), the plugin's hardcoded `CLAUDE_MEM_MODEL` setting causes issues:

### The Issue

1. **Auto-loading unwanted models**: Every time an observer session is spawned, claude-mem explicitly requests a specific model (e.g., `qwen/qwen3.6-35b-a3b`). LM Studio sees this request and auto-loads that model into memory.

2. **Memory pressure**: Large models (30GB+) are loaded unnecessarily, even when the user's parent Claude Code session is using a different/smaller model.

3. **Inflexible workflow**: Users who want to switch between models in their parent session find that claude-mem observer sessions keep triggering loads of the hardcoded model.

### Root Cause Analysis

The issue stems from two places:

**1. `src/shared/SettingsDefaultsManager.ts` (line 79):**
```typescript
CLAUDE_MEM_MODEL: 'claude-haiku-4-5-20251001',
```
This hardcoded default means even new installations get a specific model.

**2. User settings (`~/.claude-mem/settings.json`):**
Once set, this value persists and is used for every observer session.

### Why This Happens

When claude-mem spawns an observer session via the Anthropic SDK:
```typescript
const modelId = session.modelOverride || this.getModelId();
```

The `modelId` is explicitly passed to LM Studio's API. LM Studio then auto-loads that specific model if not already in memory — regardless of what model the parent Claude Code session is using.

---

### Change Default Behavior

Change `CLAUDE_MEM_MODEL` default from a hardcoded model to an **empty string**, signaling "inherit from parent session":

```diff
- CLAUDE_MEM_MODEL: 'claude-haiku-4-5-20251001',
+ CLAUDE_MEM_MODEL: '', // Empty by default — inherits model from parent session
```

### How It Works

When `CLAUDE_MEM_MODEL` is empty:
- Observer sessions don't explicitly request a specific model
- LM Studio uses whatever model is currently active/loaded (the parent session's model)
- No unnecessary model switching or memory pressure

Users who want explicit control can still set `CLAUDE_MEM_MODEL` to a specific value.

### Benefits

1. **Respects user workflow**: Local hosting users often switch models based on task complexity
2. **Reduces memory pressure**: No auto-loading of large models unnecessarily  
[...]

PR #3514, opened 2026-08-08
feat: external worker clients (CORS allowlist + API-key auth) and free local-model memory

saltoriousSIG:
## Summary

Two features, both opt-in with **zero behavior change for existing installs**:

1. **External worker clients** — lets browser apps on real `https` origins (and remote agents) talk to the local worker API. Defaults change nothing: new behavior activates only when the user configures an allowed origin.
2. **Free local-model memory** — `npx claude-mem local-model check|setup` tests whether the machine can run memory compression on a local model (Ollama / LM Studio) and switches to it in one command: free, private, no API costs.

Most of the machinery already existed in the repo (the sqlite API-key service, the `server api-key` CLI which already targets the worker's own `claude-mem.db`, the `version` field on `/api/health`) — this PR mostly wires it up where the worker's HTTP surface needed it.

## What changed

**CORS** (`src/services/worker/http/middleware.ts`)
- `createCorsMiddleware` accepts an exact-match origin allowlist from the new `CLAUDE_MEM_WORKER_ALLOWED_ORIGINS` setting (comma-separated full origins; empty = today's localhost-only behavior). The `http://localhost:` / `http://127.0.0.1:` prefix family stays implicitly allowed.
- Disallowed origins now get a clean `403 {"error": "origin_not_allowed"}` instead of `next(new Error(...))` surfacing as an HTML 500.
- Preflights for allowed origins answer `Access-Control-Allow-Private-Network: true` when the request carries `Access-Control-Request-Private-Network: true` — without this, Chrome's private-network preflight blocks any public site from reaching a local server even with permissive CORS.
- `X-Api-Key` added to `Access-Control-Allow-Headers`.

**Auth** (new `src/services/worker/http/worker-auth.ts`, wired via opt-in `ServerOptions.workerAuth`)
- New `CLAUDE_MEM_WORKER_AUTH` setting: `origin` (default) | `all` | `off`.
- `origin` mode requires an API key only where the surface actually opens: requests carrying a non-localhost `Origin`, and originless requests whose `Host` header is a non-localhost DNS name (the DNS-rebinding shape — a same-origin fetch after rebinding always carries the attacker's domain in `Host`, never an IP literal). Hooks, curl, the bundled viewer, LAN-IP setups, and `0.0.0.0` binds all stay tokenless.
- `all` mode covers every `/api` + `/v1` request (hosted deployments); health probes (`/health`, `/readiness`, `/version`, `/chroma/status`, `/settings/dependency-health` — the same set the init gate exempts) stay open.
- Key verification delegates to the existing `sqlite-api-key-service` (scrypt-salted, timing-safe) against the worker's own DB. No new token scheme, no new storage, no schema migration.
- Only the worker runtime sets `workerAuth` — the server runtime authenticates in its own route layer (`requireServerAuth`) and is untouched.

**CLI** — `claude-mem worker api-key create|list|revoke`: thin aliases routing to the existing key CLI (which already operates on the worker DB).

**Docs** — new "Connecting External Clients" page (`docs/public/external-clients.mdx`).

## Behavior matrix

| Client | Origin header | Config | Result |
|---|---|---|---|
| Claude Code hooks / curl | none | any default | unchanged, works |
| Bundled viewer UI | localhost | default | unchanged, works |
| External web app | its `https` origin | allowlisted + token | works (the new capability) |
| Random website drive-by | its origin | not allowlisted | 403; preflight failure blocks reads |
| DNS-rebinding page | none (Host: attacker domain) | default | 401 (new guard; previously reachable) |
| Hosted worker | any | `CLAUDE_MEM_WORKER_AUTH=all` | token on everything |

## Security notes

Adding CORS for a real web origin is what makes token auth necessary rather than optional — today "localhost-only CORS" does double duty as the auth layer. Once any `https` origin is allowlisted, the token becomes the actual boundary, and `origin` mode ensures the boundary exists exactly where the surface opens. The `Host`-header check for originless requests closes the DNS-rebinding hole using the same pattern as the existing `requireServerAuth` local-dev bypass. The three `/api/admin/*` routes keep their `requireLocalhost` behavior.

## Tests

27 new tests in `tests/worker/middleware/worker-auth.test.ts`: CORS allow/deny/normalization matrix, PNA header presence/absence, the no-Origin-no-token invariant (the most important test in the PR), Bearer + X-Api-Key, invalid/revoked keys, all three modes, preflight-without-credentials, exempt health paths, and helper units. Existing `cors-restriction.test.ts` passes unchanged against the new signature.

Full suite: **2,600 tests, 0 failures**; both typecheck configs clean.

## Local-model memory (second feature, commit 80acf14b)

- `npx claude-mem local-model check` (read-only): RAM-based capability tiers (<8GB not recommended, 8–15GB ~3B, 16–31GB ~7–8B, 32GB+ ~14B), Apple Silicon detection, and probes for a running Ollama (`:11434`) / LM Studio (`:1234`) with their installed models. Exits 0 when capable — script-friendly like `doctor`.
- `npx claude-mem local-model setup [--model id] [--endpoint url]`: writes `CLAUDE_MEM_PROVIDER=openrouter` + `CLAUDE_MEM_OPENROUTER_BASE_URL/MODEL` for the detected runtime, preferring the tier-recommended model among those actually installed. Rides entirely on the existing OpenAI-compatible base-URL support (#2393).
- Two fixes the keyless-local path needed: `isOpenRouterAvailable()` now treats a configured base URL as available (previously the provider silently fell back to Claude unless the user set a dummy OpenRouter key), and the `Authorization` header is omitted entirely when no key is configured instead of sending an empty `Bearer `.
- New "Local Models" docs page; 10 new tests (tiers, fake Ollama/LM Studio probe servers, dead-port/non-OK cases, settings delta).

### DIFF
--- a/docs/public/docs.json
+++ b/docs/public/docs.json
@@ -87,7 +87,8 @@
           "troubleshooting",
           "platform-integration",
           "openclaw-integration",
-          "external-clients"
+          "external-clients",
+          "local-models"
         ]
       },
       {
--- a/docs/public/docs.json
+++ b/docs/public/docs.json
@@ -86,7 +86,8 @@
           "branches",
           "troubleshooting",
           "platform-integration",
-          "openclaw-integration"
+          "openclaw-integration",
+          "external-clients"
         ]
       },
       {
--- a/docs/public/external-clients.mdx
+++ b/docs/public/external-clients.mdx
@@ -1,0 +1,76 @@
+---
+title: "Connecting External Clients"
+description: "Let browser apps and remote agents call the local worker API — CORS allowlist, private-network preflight, and API keys."
+---
+
+# Connecting External Clients
+
+By default the worker at `http://127.0.0.1:<port>` only accepts browser requests
+from `localhost` origins, and nothing else changes until you configure it: Claude
+Code hooks, `curl`, and the bundled viewer UI keep working with no token and no
+setup.
+
+To let an external client — a web app on a real `https` origin, or a remote
+agent — talk to your worker, you allowlist its origin and hand it an API key.
+
+## 1. Allowlist the origin
+
+Add the client's full origin (comma-separated for several) to your settings —
+either in `~/.claude-mem/settings.json` or as an environment variable:
+
+```json
+{
+  "CLAUDE_MEM_WORKER_ALLOWED_ORIGINS": "https://app.example.com"
+}
+```
+
+Allowlisted origins get proper CORS echoes plus the
+`Access-Control-Allow-Private-Network` preflight header Chrome requires before a
+public website may reach a local server. Origins **not** on the list get a clean
+`403 {"error": "origin_not_allowed"}`.
+
+Restart the worker (`npx claude-mem restart`) after changing settings.
+
+## 2. Create an API key
+
+```bash
+npx claude-mem worker api-key create --name my-external-client
+```
+
+This prints the key **once** — paste it into the external client's settings. The
+same command family manages the lifecycle:
+
+```bash
+npx claude-mem worker api-key list
+npx claude-mem worker api-key revoke <id>
+```
+
+## 3. The client sends the key
+
+```bash
+curl -H "Authorization: Bearer cmem_..." \
+  "http://127.0.0.1:37777/api/search?query=auth+refactor"
+```
+
+`X-Api-Key: cmem_...` is accepted as a fallback for clients that can't set an
+`Authorization` header.
+
+External clients writing sessions or observations should identify themselves via
+the `platformSource` field on the session routes, so their activity is
+attributable in the viewer.
+
+## Enforcement modes
+
+`CLAUDE_MEM_WORKER_AUTH` controls when a key is required:
+
+| Mode | Behavior |
+|------|----------|
+| `origin` (default) | A key is required only where the surface actually opens: requests carrying a non-localhost `Origin`, and requests whose `Host` header is a non-localhost DNS name (the DNS-rebinding shape). Hooks, curl, and the viewer are unchanged. |
+| `all` | Every `/api` and `/v1` request needs a key — for hosted or shared-network deployments. Health probes (`/api/health`, `/api/readiness`, `/api/version`) stay open. |
+| `off` | No key checks (trusted-network opt-out). |
+
+## Feature detection
+
+External clients can probe `GET /api/health` (no auth in every mode) — the
+response includes `version`, so a client can feature-detect worker capabilities
+instead of guessing.
--- a/docs/public/external-clients.mdx
+++ b/docs/public/external-clients.mdx
@@ -59,13 +59,31 @@ External clients writing sessions or observations should identify themselves via
 the `platformSource` field on the session routes, so their activity is
 attributable in the viewer.
 
+## Live events over SSE
+
+Native `EventSource` can't attach headers, so `/stream` accepts a short-lived,
+single-use ticket instead. Mint one with an authenticated request, then connect:
+
+```js
+const { ticket } = await fetch("http://127.0.0.1:37777/api/stream-ticket", {
+  method: "POST",
+  headers: { Authorization: `Bearer ${key}` },
+}).then(r => r.json());
+
+const events = new EventSource(`http://127.0.0.1:37777/stream?ticket=${ticket}`);
+```
+
+Tickets expire after 60 seconds, work exactly once, and only open `/stream` —
+mint a fresh one on every (re)connect. Never put the long-lived API key itself
+in a URL.
+
 ## Enforcement modes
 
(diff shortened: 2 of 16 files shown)


saltoriousSIG:
Both P1 findings addressed in 7b7768ee:

**/stream unauthenticated** — the worker-auth middleware now also mounts on the root `/stream` route, so an allowlisted external origin needs a key to open the SSE stream while the originless loopback viewer stays tokenless. Covered by three new tests (allowlisted origin keyless → 401, with key → 200, loopback viewer → 200).

**IPv4-literal Host bypass** — originless trust now requires BOTH a loopback socket peer AND a localhost/IP-literal Host, instead of trusting the caller-controlled Host header alone. A LAN client reaching a 0.0.0.0-bound worker gets 401 without a key in `origin` mode regardless of what Host it sends; the Host-shape check is retained solely to catch DNS rebinding (which arrives from loopback carrying the attacker's domain). Real test sockets always connect via loopback, so the LAN cases are covered by synthetic-peer tests driving the middleware with a non-loopback `socket.remoteAddress`; the suggested real-interface integration test would depend on the CI host's network configuration.

Full suite: 2,608 tests, 0 failures.

saltoriousSIG:
Both round-2 P1 findings addressed in cc5a8160:

**Reverse-proxy bypass** — originless loopback trust now additionally requires the absence of forwarded-client headers (`Forwarded`, `X-Forwarded-For`, `X-Forwarded-Host`, `X-Real-IP`), matching the anti-proxy guard in the existing `requireServerAuth` local-dev bypass. A loopback proxy forwarding for a remote client no longer passes; proxied clients need a key in `origin` mode. Tested with each forwarded header individually voiding trust, plus a proxied-client-with-valid-key pass case.

**EventSource stream auth** — implemented the suggested ticket design rather than declaring `/stream` unavailable to browsers: `POST /api/stream-ticket` (registered under `/api`, so it is gated by the same auth policy — cross-origin minting requires a valid key) returns a single-use ticket with a 60-second TTL, accepted only as `GET /stream?ticket=...` and only for the stream mount. Tickets authenticate nothing else, are dead after one use or expiry, and long-lived keys never appear in URLs. The flow is documented with a browser example in `docs/public/external-clients.mdx`. Lifecycle tests cover: keyless cross-origin mint → 401, mint-then-connect → 200, reuse → 401, bogus/expired → 401, ticket on a non-stream route → 401.

Full suite: 2,616 tests, 0 failures.

saltoriousSIG:
Added in 80acf14b (second feature in this PR): **local-model capability check + free local memory**.

`npx claude-mem local-model check` probes the machine (RAM-based capability tiers, Apple Silicon detection) and looks for a running Ollama or LM Studio with installed models; `npx claude-mem local-model setup` switches memory compression to the detected local endpoint in one command — free, private, no API costs. It rides entirely on the existing OpenAI-compatible base-URL support (#2393), so the diff is mostly detection + CLI + docs.

Two fixes the keyless-local path needed and now includes: `isOpenRouterAvailable()` treats a configured base URL as available (previously a keyless local endpoint silently fell back to the Claude provider unless the user set a dummy OpenRouter key), and the `Authorization` header is omitted when no key is configured instead of sending an empty `Bearer `.

New docs page ("Local Models") and 10 new tests (hardware tiers, fake Ollama/LM Studio probe servers, dead-port and non-OK cases, settings delta). Full suite: 2,626 tests, 0 failures.

saltoriousSIG:
Round-3 P1 addressed in f97b6bba: **localhost-Origin bypass**.

The Origin-bearing branch now applies the same "provably local" rule the originless branch already used: a localhost-shaped `Origin` only keeps a request tokenless when the socket peer is loopback AND no forwarded-client headers are present — header shape (Origin or Host) narrows which provably-local requests skip the key, and never substitutes for the peer check. A LAN client forging `Origin: http://localhost:<port>` gets 401 without a key, as does the same forgery through a loopback proxy; the bundled viewer's clean loopback localhost-Origin requests stay tokenless.

Regression tests cover the exact reproductions: forged localhost Origin from a non-loopback peer (401, then 200 with a valid key), forged localhost Origin with `X-Forwarded-For` over loopback (401), and the clean-loopback viewer invariant (200). Full suite: 2,630 tests, 0 failures.

thedotmack:

> step 4 (CORS same-host + `CLAUDE_MEM_ALLOWED_ORIGINS`); auth-mode half needs the localhost-Origin peer check closed; local-model CLI half → roadmap #2785

thedotmack:
Routing to **#2785 (plan-12 Provider & Extensibility Roadmap)** — both halves are opt-in net-new capability rather than defect repair: a local-model provider path (Ollama / LM Studio) and an external-client surface. The CORS/API-key piece brushes plan-23's (#3611) MCP/HTTP tool surface item, but plan-23 is about validating existing host adapters against real schemas, so the new-capability cluster is the better home. Leaving open.

Issue #3196, opened 2026-07-11
[Windows] findBun() doesn't check for bun.exe via 'where bun' — falls back to bare 'bun', corrupted by shell quoting (related to #2528)


SiNaPsEr0x:
## Environment
- OS: Windows 11 Pro
- claude-mem: 13.10.2
- Bun: 1.3.14, installed via official installer (`npx claude-mem install`) — this creates `bun.exe` only, no `bun.cmd`
- Claude Code: 2.1.207

## Symptom
Every hook fails on startup:
```
SessionStart:startup hook error
Failed with non-blocking status code: ""bun"" non è riconosciuto come comando interno o esterno,
```
(Italian locale; translates to `""bun"" is not recognized as an internal or external command`.)

`where.exe bun` correctly resolves to `C:\Users\<user>\.bun\bin\bun.exe`, and `bun --version` works fine everywhere else — so this is not a PATH problem.

## Root cause

In `scripts/bun-runner.js`, `findBun()`:

```js
if (pathCheck.status === 0 && pathCheck.stdout.trim()) {
  if (IS_WINDOWS) {
    const bunCmdPath = pathCheck.stdout.split('\n').find(line => line.trim().endsWith('bun.cmd'));
    if (bunCmdPath) {
      return bunCmdPath.trim();
    }
  }
  return 'bun';   // <-- reached when only bun.exe is present
}
```

When `where bun` succeeds but only returns a `bun.exe` path (official installer, no npm shim), the function falls through to `return 'bun'` — the bare command name, not a resolved path.

That bare string then hits the Windows spawn block further down:

```js
if (IS_WINDOWS) {
  const quote = (s) => `"${String(s).replace(/"/g, '\\"')}"`;
  spawnOptions.shell = true;
  spawnCmd = [bunPath, ...args].map(quote).join(' ');
  spawnArgs = [];
}
```

`bunPath = 'bun'` gets quoted to `"bun"` and joined into a single shell string executed via `shell:true` → `cmd.exe`. With an unresolved bare command name in this position, cmd.exe receives a malformed token and reports it back doubly-quoted (`""bun""`), failing every hook (`SessionStart`, `UserPromptSubmit`, `PostToolUse`, `Stop`).

Confirmed this is specific to the resolved value, not the spawn mechanism itself:
```js
// Works fine when passed a plain string directly (not through findBun's bare fallback):
spawnSync('cmd', ['/c', 'bun', '--version'], {encoding:'utf8'})
[...]

Issue #748, opened 2026-01-18
[Bug][Windows] Terminal window pops up on every message when realtime hooks are enabled


artwisut:
## 🐞 Bug Description

When using **claude-mem** on **Windows** with **Claude Code**, a new terminal/console window pops up **for every single user message**.

This happens whenever realtime hooks are enabled (e.g. `UserPromptSubmit`, `PostToolUse`, `user-message`, etc).

Even after attempting multiple workarounds (PowerShell `Start-Process -WindowStyle Hidden`, `cmd /B`, reducing hooks, etc.), the terminal window still opens on every message.

---

## 🔁 Steps to Reproduce

1. Install `claude-mem` from the Claude marketplace  
2. Enable the plugin in Claude Code  
3. Use default or realtime hooks (user-message / PostToolUse)  
4. Start a conversation  
5. Send multiple messages  

👉 **Result:**  
Each message spawns a new Windows terminal window.

---

## ✅ Expected Behavior

Realtime memory hooks should run **silently in the background**, without opening any visible terminal or console windows.

---

## ❌ Actual Behavior

- A Windows terminal (console) window pops up **for every message**
- Makes realtime usage practically unusable on Windows

---

## 🖥️ Environment

- OS: **Windows 11**
- Claude Code version: **v2.1.12**
- claude-mem: Installed from marketplace (latest at time of writing)
- Shell: Windows Terminal / cmd / PowerShell (all affected)

---

## 📎 Additional Notes

- Realtime memory **does work** (events appear correctly in `http://localhost:37777`)
- The issue appears to be related to how Claude Code spawns `command` hooks on Windows
- Reducing hooks to `SessionStart` + `Stop` avoids the issue, but disables realtime memory
- This seems to be a Windows-specific runner / console spawn issue

---

## 🙏 Request

Is there a recommended way to:
- Enable **realtime memory**
- While preventing **Windows terminal windows from popping up per message**

Any guidance or fix would be greatly appreciated 🙇

bigph00t:
Hey thanks for the report. 

This is a known issue we are actively working on a fix for, we appreciate your issue report. 

Please join the Claude-mem discord if you havent already!

bigph00t:
https://github.com/thedotmack/claude-mem/pull/751 You can test this PR. It will solve the popup issues. However, chromadb will be disabled for now. In this coming week we are looking to refactor and add Chroma HTTP service which will make windows more compatible. But for now this PR will allow windows to work using SQL

[...]

Issue #676, opened 2026-01-11
Windows Terminal console window repeatedly opens and closes on hook execution


MrViSiOn:
## Description

I recently installed this Claude plugin on my Windows system and integrated it with Claude Code. However, I'm experiencing an issue where a Windows Terminal console window repeatedly opens and closes whenever a hook is triggered during agent interactions. This behavior is quite disruptive and happens consistently.

## Environment
- **OS**: Windows
- - **Plugin**: Claude-mem (freshly installed)
- - - **Node version manager**: fnm installed (fnm_multishell)
- - - - **Server status**: http://localhost:37777/ is working correctly
## Issue Details

Every time I interact with the agent and a hook is triggered, a console window appears and immediately closes, creating a rapid flashing effect. This happens repeatedly and is becoming annoying during normal usage.

## Question

Is this the normal expected behavior of the plugin on Windows, or is there a configuration issue causing this? Should I expect this behavior every time a hook executes?

Any guidance or troubleshooting steps would be appreciated. Thank you!

trade-art:
Same

AinzRimuru:
Same
http://localhost:37777/ will become available the moment the window pops up.

luidpivetta:
same here

imudak:
Same

thedotmack:
Duplicate of #997. Console window flicker is tracked in #997.

PR #3519, opened 2026-08-08
fix(windows): guard the login-shell PATH prelude behind a node check

### LINKED ISSUE #3396: [Windows] 13.11.0 hooks format ("shell": "bash") reintroduces the console-flash-per-hook regression (third shape after #676/#681/#748)

NiceLeader:
Fixes #3396.

## The spawn

Six of the seven `claude-code` hooks begin with:

```sh
export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH";
```

`Setup` is the exception. On Windows `$SHELL` is `/bin/bash.exe`, so this is a
console-attached Git Bash **login** shell, sourcing the whole profile, to read
one variable. `PostToolUse` is registered on `matcher: "*"`, so it runs once per
tool call. That is the console window flashing on every action, and it is the
same symptom class as #676 / #681 / #748.

The same prelude also doubles `$PATH` on every invocation.
`plugin/scripts/bun-runner.js` already documents that as the cause of #3196,
where the doubled value overran cmd.exe's ~8191-char per-variable limit and left
`bun` unresolvable even though `where bun` had just succeeded.

## The change

Guard it on the one thing it exists to guarantee:

```sh
command -v node >/dev/null 2>&1 || export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH";
```

`command -v` is a POSIX builtin, so the common path forks nothing. Where the
prelude was load-bearing, meaning node is genuinely absent from the inherited
PATH (a GUI-launched host on macOS with nvm unsourced, say), the original
prepend still runs unchanged. Nothing is removed, it is only skipped when it is
provably unnecessary.

Two precedents already in-tree: the `mcp` host has no PATH prelude at all and
relies on the inherited PATH, and `command -v cygpath` appears a few tokens
later inside these very command strings.

## Why both files move

`src/build/hook-shell-template.ts` is the generator and
`plugin/hooks/hooks.json` is the hand-maintained output that must match it byte
for byte. Changing only `hooks.json` fails all six Rule A generator-parity
assertions in `tests/infrastructure/plugin-distribution.test.ts`; changing only
the generator fails the same six. They have to land together.

`plugin/hooks/codex-hooks.json` needs nothing: the `codex-cli` prelude already
guards its login-shell fallback behind an empty-PATH check.

## Verification

`bun test tests/infrastructure/plugin-distribution.test.ts` on Windows 11,
before and after the change: **55 pass, 2 fail**, identical either way. The two
failures are pre-existing on `main` and are path-format assertions
(`RESOLVED=/tmp/...` vs `C:\...`) unrelated to this change.

Root cause isolated by @HolaRoy in #3396. This takes the same finding from the
generator side so the parity tests stay green.


stantheman0128:
This still leaves bash in the hook, so the node.exe grandchild still flashes on every tool call. #3564 drops that bash layer (exec-form node), which is the product-path fix for #3521 / #3559.

I would take #3564 over this for those issues. The node-on-PATH guard is a real narrowing of the login-shell prelude, just not enough to close the flash.

thedotmack:

> step 1 (`command -v node` guard on the prelude); already bundled in #3614; note #3564 argues the flash persists via the bash layer

rodrigogs:
Two findings from running this exact guard on macOS 26.6 / Apple M1 Pro with zsh + mise, claude-mem 13.15.3. **The change is safe and I would land it** — but one of its stated premises does not hold, and the guard is one binary short.

## 1. The fallback this PR preserves cannot do its job on macOS + zsh

`src/build/hook-shell-template.ts` justifies keeping the unconditional prepend for the load-bearing case:

> Where the prelude was load-bearing - node genuinely absent from an inherited PATH, e.g. a GUI-launched host on macOS with nvm unsourced - the original prepend still runs, byte for byte.

`$SHELL -lc` is a **login but non-interactive** shell. zsh sources `.zshrc` only for *interactive* shells — a login non-interactive zsh reads `.zshenv`, `/etc/zprofile`, `.zprofile`, `.zlogin` and stops. nvm, mise, asdf, fnm and rbenv all conventionally install their activation into `.zshrc`. So on the exact host that sentence describes, the probe returns a PATH with no version-manager entries at all.

Marker test — appended a write to a temp file at the end of `~/.zshrc`, then removed it:

```
zsh -lc  'true'   →  .zshrc NOT sourced
zsh -ilc 'true'   →  .zshrc sourced
```

End to end under `env -i`, so nothing leaks in from the parent environment:

```console
$ env -i HOME="$HOME" USER="$USER" SHELL=/bin/zsh PATH=/usr/bin:/bin \
    /bin/zsh -lc 'echo $PATH' | tr : '\n' | grep -c 'mise/installs'
0
$ env -i HOME="$HOME" USER="$USER" SHELL=/bin/zsh PATH=/usr/bin:/bin \
    /bin/zsh -lc 'command -v node; command -v bun'
        (both empty)
```

So the prelude never was load-bearing on this class of host. It only ever appeared to work because the inherited PATH already carried node — which is precisely the case this PR short-circuits. That makes the change **strictly safe** (it cannot make the broken case worse), but the comment overstates what is being preserved, and the fallback is effectively dead code on macOS + zsh until it either uses `-ilc` or reads a cached PATH — which #3605 fix-sequence item 1 already asks for ("resolves nothing it can cache").

## 2. The guard gates on `node`, but the chain also needs `bun`

`bun-runner.js`'s `findBun()` resolves bun independently: `which bun`, then four hardcoded paths (`~/.bun/bin/bun`, `/usr/local/bin/bun`, `/opt/homebrew/bin/bun`, linuxbrew), then returns `null`. When `command -v node` is satisfied by a system or Homebrew node while bun is only version-manager-managed, the guard short-circuits, the probe is skipped, and none of the four fallbacks match:

```console
# PATH carrying Homebrew node, no bun anywhere on PATH
$ command -v node   →  /opt/homebrew/bin/node  (v26.7.0)   # guard passes, probe skipped
$ command -v bun    →  not found
```

On my host this is **latent rather than fatal**, only because `~/.bun/bin/bun` happens to exist alongside mise's bun. For a mise/asdf/proto user whose bun is managed exclusively by the version manager, `findBun()` returns `null`.

Gating on both is one token:

```sh
command -v node >/dev/null 2>&1 && command -v bun >/dev/null 2>&1 || export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH";
```

## What I run locally instead

This skips the login shell entirely by resolving the version manager's stable major symlink, and pays for the probe only if that is genuinely gone:

```sh
_mn="$HOME/.local/share/mise/installs/node/24/bin"
_mb="$HOME/.local/share/mise/installs/bun/latest/bin"
[ -x "$_mb/bun" ] || _mb="$HOME/.bun/bin"
if [ -x "$_mn/node" ] && [ -x "$_mb/bun" ]; then export PATH="$_mn:$_mb:$PATH"
else export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH"; fi
```

mise repoints `installs/node/<major>` and `installs/bun/latest` on upgrade, so this survives patch bumps with no edit. Wrapper prelude cost, median of 7 runs, same machine and load:

| prelude | median | min | max |
|---|---|---|---|
| unconditional probe (current `main`) | 1.512 s | 0.910 | 3.484 |
| this PR's guard, node present | 0.867 s | 0.508 | 0.944 |
| direct symlink resolution | 0.691 s | 0.635 | 0.830 |

Absolute values include roughly 0.5 s of harness overhead and the machine was under heavy memory pressure, so read the ordering rather than the magnitudes — it held on every run. Measured with `/usr/bin/time -p` on a quieter sample, the same comparison was 0.21 s versus 0.06 s.

Happy to open a PR for either variant against whichever branch the plan-17 bundle uses, if that would help #3614 along.

NiceLeader:
Both taken, thanks — this is exactly the review that makes a one-line change
safe to bundle.

1. Guard now gates on `node` **and** `bun` (ecb3e22). A node-only guard leaves
   `findBun()` to its four hardcoded paths on a version-manager-only host, and
   the `~/.bun/bin` coincidence on your machine is not something to rely on.

2. Comment corrected. The "load-bearing on macOS + nvm" sentence overstated
   what the preserved prepend does; your `env -i` marker test settles it —
   `-lc` is login-non-interactive, `.zshrc` never runs, so the probe was dead
   on that host class all along. The comment now says the fallback is
   preserved unchanged but known-dead on macOS + zsh, and points at #3605
   item 1 (cached PATH or `-ilc`) for the real fix rather than growing this PR
   into it.

Parity still holds (55/2, the 2 being the pre-existing Windows path-format
asserts).

Your symlink-resolution variant is the better end state and the timing table
shows why (0.69 s vs 0.87 s median, ordering held on every run). I'd keep it
out of this PR only because it hardcodes a mise layout, which is a per-host
decision — it reads like the shape item 1 should take once "resolve nothing it
can cache" gets designed. If you open that against the bundle branch, happy to
review.

thedotmack:
Routing to **#3605 (plan-17 Hook Wrapper Contract)** — "login-shell PATH prelude" and "Windows console windows / bash.exe" are the first two items in that master, and guarding the prelude on `command -v node` is the minimal version of the fix. Keeping this open: it is already bundled into #3614, which explicitly leaves its sources open until the bundle lands. Note #3470 and #2598 attack the same prelude from slightly different angles — plan-17 should converge on one template, and this is the smallest of the three.

Issue #2967, opened 2026-06-16
tracking: project name detection resolves wrong directory in monorepos, worktrees, and submodules


thedotmack:
## Root Cause

Project name resolution has three related failure modes: (1) in monorepos/subdirectories, the git root is used instead of the cwd basename, silently losing context after upgrading; (2) worktree observations accumulate under a `parent/leaf` compound key, and the one-shot `.cwd-remap-applied-v1` migration misses worktrees created afterward, orphaning them when the worktree is deleted; (3) sessions inside a git submodule resolve to the empty submodule leaf project rather than the parent repo, showing the empty-state banner despite rich history.

### Issues
- Closes #2882 — https://github.com/thedotmack/claude-mem/issues/2882
- Closes #2864 — https://github.com/thedotmack/claude-mem/issues/2864
- Closes #2842 — https://github.com/thedotmack/claude-mem/issues/2842

### PRs
- Related PR #2883 — https://github.com/thedotmack/claude-mem/pull/2883
- Related PR #2856 — https://github.com/thedotmack/claude-mem/pull/2856
- Related PR #2665 — https://github.com/thedotmack/claude-mem/pull/2665

## Resolution

This is the canonical tracking issue for this root cause. Individual issues and PRs listed above will be closed in favor of this one.

thedotmack:
Consolidating into #3608 (plan-20). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

Issue #2882, opened 2026-06-10
Project detection uses git root instead of working directory, causing issues in monorepos/subdirectories


anthonypdawson:
When working in a subdirectory of a git repository (e.g., `/monorepo/automation`), claude-mem detects the project name as the repository root (`monorepo`) instead of the working directory (`automation`). This causes observations captured in one context to not be injected in future sessions because the project names don't match.

**Note**: This is distinct from #2842 (submodules). This issue is about plain subdirectories in a standard git repo, not git submodules or worktrees.

## Expected Behavior
- Sessions started in `/monorepo/automation` should be tagged with project name `automation`
- Observations captured in that directory should use project name `automation`
- Context injection should recognize `automation` as an established project

## Actual Behavior
- Sessions are tagged with project name `monorepo` (git root)
- Historical observations were tagged with `automation` (working dir)
- Context injection shows "This project has no memory yet" despite 58 observations existing

## This is a Regression
Previous versions of claude-mem (May 2026) correctly used the working directory name (`automation`), creating observations tagged with that project name. Current versions (June 2026, 13.4.1+) switched to using git root (`monorepo`), breaking context injection since the project names no longer match.

This means existing users with subdirectory-based memories are silently losing context injection after upgrading, requiring manual database surgery to fix.

## Steps to Reproduce
1. Work in a git repository with subdirectories (not submodules) that have distinct contexts
2. Run `cd /path/to/repo/subdirectory`
3. Start Claude Code session
4. Check `~/.claude-mem/claude-mem.db`: `SELECT project FROM sdk_sessions ORDER BY started_at DESC LIMIT 1;`
5. Observe it uses the git root name, not the subdirectory name

## Environment
- claude-mem version: 13.4.1 → 13.5.0
- Runtime: worker mode
- Working directory: `/path/to/monorepo/automation` (regular subdirectory, not a submodule)
- Git root: `/path/to/monorepo`

## Workaround
Manually update project names in the database to match future session detection:
```bash
sqlite3 ~/.claude-mem/claude-mem.db "UPDATE observations SET project = 'monorepo' WHERE project = 'automation'; "
sqlite3 ~/.claude-mem/claude-mem.db "UPDATE sdk_sessions SET project = 'monorepo' WHERE project = 'automation'; "
```

## Suggested Fix
Add a configuration setting for project name resolution:

```json
{
  "CLAUDE_MEM_PROJECT_NAME_SOURCE": "git-root" | "working-dir" | "explicit"
[...]

PR #3525, opened 2026-08-09
fix(worktree): reconcile orphaned worktree and submodule observations

DanielCarmingham:
## Summary

Two failures under tracking issue #2967, fixed together because they interact (see Assumptions).

**#2864 — worktree observations orphaned.** `adoptMergedWorktreesForAllKnownRepos` discovered repos by reading distinct `cwd` values from `pending_messages`. That table stopped receiving writes in 61fe70a2 ("remove retry/observation queue, replace with SessionMessageBuffer", v13.10.0) — the only remaining `INSERT INTO pending_messages` statements are in tests. The startup sweep therefore found zero repos on every boot, and adoption only ever ran when a user invoked `npx claude-mem adopt` by hand. Worktree rows kept their `parent/leaf` key with `merged_into_project = NULL`; once the worktree directory was deleted, the parent↔worktree association could no longer be recomputed from disk and no read path could reach them.

`cwd` was not lost, only unrecorded: it still arrives on every ingest (`SessionRoutes.ts` → `ingestObservation`), and `cwd TEXT` existed in exactly one schema — `pending_messages`.

**#2842 — submodule sessions see no memory.** `detectWorktree` only matched `.git/worktrees/<name>` gitdir pointers. A submodule's `.git` file points at `.git/modules/<name>`, so it fell through to the standalone-project branch and resolved to the submodule's own leaf name — a new, empty project — producing the "no memory yet" banner despite a full superproject history.

## Changes

| Area | Change |
|---|---|
| `SessionStore.ts` | `ensureSessionCwdColumn()` (PRAGMA-guarded) + `setSessionCwd()`, first-write-wins |
| `worker/http/shared.ts` | Stamp cwd at ingest |
| `WorktreeAdoption.ts` | Sweep reads `sdk_sessions.cwd`; orphan adoption; live-submodule exclusion; `resolveMainRepoPath` walks `.git/modules` |
| `worktree.ts` / `project-name.ts` | Submodule detection → superproject composite key; `isSubmodule` added |
| `worker-service.ts` | CLI reports orphans |

Notable details for review:

- **`sdk_sessions.cwd` is local-only.** Sync entities are `observation | summary | prompt` (`CanonicalContent.ts:9`); `sdk_sessions` is joined for reads but never pushed, so no `CanonicalContent`/`SyncApply`/`CloudSync` changes are needed. `SessionStore.ts` already documents this ("sdk_sessions rows do not sync").
- **First-write-wins on cwd** — a session's cwd drifts when the agent `cd`s into a subdirectory; the launch directory is the one that identifies the repo.
- **Discovery UNIONs `pending_messages`** so installs upgrading with pre-13.10 rows still on disk keep their known-repo history until sessions repopulate `sdk_sessions.cwd`.
- **Orphan matching uses a half-open range, not `LIKE`.** `_` is a single-character LIKE wildcard and common in repo names, so `my_app/%` would also match `myXapp/...`. Range comparison removes the hazard at the source. It is *not* a perf change — `EXPLAIN QUERY PLAN` shows both forms drive off `idx_observations_merged_into` for the `IS NULL` term.
- **Two early returns removed** (`childWorktrees.length === 0`, `targets.length === 0`) — orphans exist precisely when there are no live worktrees left.
- **`--branch` skips the orphan sweep**, staying a precise escape hatch for squash-merges.

## Assumptions / decisions

1. **Deleted worktrees are adopted regardless of merge status.** Existing adoption required the branch to be merged into HEAD. A worktree can be deleted with unmerged or abandoned work, and for those rows the alternative to adopting is not "kept separate" — it is unreachable forever. Widening the rule is deliberate; it is surfaced through `AdoptionResult.orphanedWorktrees`, the applied-adoption log line, and the CLI so a run never widens silently. Live worktrees, merged or in-flight, are untouched.

2. **#2882 is excluded.** The third item under #2967 asks for project names derived from the subdirectory again, reversing the git-root behavior of #2663 (closed COMPLETED) / #2665. That is a product decision about what a "project" means, not a technical one, and it touches `getProjectName` — far hotter than anything here. **#2967 will not auto-close on merge**; it needs the #2882 decision.

3. **The two fixes had to land together.** Submodules now share the `parent/leaf` composite key but are *not* reported by `git worktree list`, so the orphan sweep initially classified a live, actively-used submodule as a removed worktree and folded it into the superproject. Caught by test, fixed by enumerating `git submodule status --recursive` into the live set (existence-filtered, so a *deleted* submodule checkout is still adopted). Shipping either issue alone would not have produced this bug — which is the argument for one PR.

4. **Pre-existing submodule rows are not repaired.** Observations captured in a submodule *before* this change are keyed with the bare leaf (e.g. `peerless`); after it, sessions key `react/peerless`, so the old flat key goes dark. Historical rows have no cwd to resolve against (the column is new), so precise repair is impossible. Adding the bare leaf to `allProjects` would recover them but risks silently merging an unrelated top-level project of the same name. New submodule data is correctly keyed from the first session and needs no reconciliation. Structurally identical to the #2882 legacy-key gap; one explicit user-invoked remap could serve both.

## Verification

- `bun test tests/` — **2574 pass, 18 skip, 0 fail** (18 new tests across 5 files)
- `npm run typecheck` — clean
- `npm run lint:hook-io`, `npm run lint:spawn-env` — clean
- `npm run build` — clean (rebuilt `plugin/` artifacts reverted to keep the diff source-only)

Every fix was written test-first and the failing state observed before implementing; the orphan-reporting field and the `isWorktree`-for-submodules correction were each reverted to confirm RED before being restored.

Known-failing, **not** caused by this branch (both verified by stashing and re-running on clean `main`):

- `workers/sync-hub` — 6 failures, `Cannot find package 'cloudflare:test'` (needs vitest-pool-workers, not bun). Identical on `main`.
- `npm run strip-comments:check` — exits 1 with 384 changed files; **381 on clean `main`**. The 3 added are source files where this branch added issue-referencing comments, matching the surrounding house style.

## Test coverage

| File | Covers |
|---|---|
| `worktree-adoption-sweep.test.ts` | Discovery from `sdk_sessions.cwd` with `pending_messages` empty |
| `ingest-session-cwd.test.ts` | cwd stamped at ingest; launch cwd survives a subdirectory observation |
| `worktree-adoption-orphans.test.ts` | Removed worktree adopted; unmerged orphan adopted; live unmerged worktree untouched; `_`-wildcard sibling not swept; orphans reported |
| `worktree-adoption-submodule.test.ts` | Live submodule not adopted (repo + sweep); deleted submodule adopted; superproject resolved from a submodule-only cwd |
| `project-name-submodule.test.ts` | Composite key, `allProjects`, subdirectory parity, superproject unaffected, `isWorktree` not claimed |

https://claude.ai/code/session_01LshL8n6PcqmStKHBpamv1j

### DIFF
--- a/src/services/infrastructure/WorktreeAdoption.ts
+++ b/src/services/infrastructure/WorktreeAdoption.ts
@@ -16,6 +16,12 @@ export interface AdoptionResult {
   parentProject: string;
   scannedWorktrees: number;
   mergedBranches: string[];
+  /**
+   * #2864 — composite project keys adopted because their worktree directory is
+   * gone, regardless of merge status. Reported separately from mergedBranches
+   * so a run never silently widens what it folded into the parent.
+   */
+  orphanedWorktrees: string[];
   adoptedObservations: number;
   adoptedSummaries: number;
   chromaUpdates: number;
@@ -86,6 +92,21 @@ function resolveMainRepoPath(cwd: string): string | null {
   ]);
   if (!commonDir) return null;
 
+  // #2842 — inside a submodule, --git-common-dir is `<super>/.git/modules/<name>`
+  // (nested submodules append further `/modules/<name>` segments). Neither
+  // branch below matches that shape, so discovery used to hand back the
+  // gitdir itself: a path that exists, is not a working tree, and resolves to
+  // the wrong project. Walk back past `/.git/modules/` to the superproject
+  // root — the same anchor detectWorktree uses for the composite key, so
+  // discovery and key derivation agree.
+  const modulesMarker = `${path.sep}.git${path.sep}modules${path.sep}`;
+  const normalized = commonDir.split('/').join(path.sep);
+  const modulesIndex = normalized.indexOf(modulesMarker);
+  if (modulesIndex !== -1) {
+    const superprojectRoot = normalized.slice(0, modulesIndex);
+    return existsSync(superprojectRoot) ? superprojectRoot : null;
+  }
+
   const mainRoot = commonDir.endsWith('/.git')
     ? path.dirname(commonDir)
     : commonDir.replace(/\.git$/, '');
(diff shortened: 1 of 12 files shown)


thedotmack:

> step 1/2 (submodule → superproject key; reconciler sweeps from stored cwd, not a dead table)

thedotmack:
Routing to **#3608 (plan-20 Project Identity & Injection Scope)** — worktrees, submodules and `merged_into_project` are all named in that master, and the root cause you found (the startup sweep discovering repos from `pending_messages`, a table that stopped receiving writes in 61fe70a2) is the identity resolver drifting away from the capture path, which is exactly what plan-20 wants collapsed into one resolver. Overlaps with #3536 on `src/utils/project-name.ts`; not a duplicate of it, they fix different halves of the same key derivation. Leaving open.

PR #3547, opened 2026-08-10
feat: Kimi Code CLI support (follow-up to #2908, rebased + payload fixes)

usbrandon:
## Summary

This builds on @T3mmyKay's Kimi Code CLI integration in #2908 (credit preserved via co-author trailer on the port commit) and:

1. **Rebases it onto current main** (v13.14.0). #2908 currently has 59 merge conflicts, mostly in checked-in build artifacts; this PR ports the source files cleanly onto `main` and regenerates the bundles.
2. **Fixes hook payload handling** based on testing against a live Kimi Code CLI install — #2908 as written would silently drop tool results and inject a JSON envelope into the model's context.

## Corrections on top of #2908

- **Tool result normalization**: Kimi's `PostToolUse` payload sends `tool_output` / `tool_call_id`, not `tool_response`. The adapter now accepts both, so tool results actually reach the observation pipeline.
- **Transcript synthesis for Stop/summarize**: Kimi's `Stop` payload has no `transcript_path`. Added `src/cli/adapters/kimi-transcript.ts`, which converts `$KIMI_CODE_HOME/sessions/*/session_<id>/agents/main/wire.jsonl` (`context.append_message` / `context.append_loop_event` records) into a Claude-style transcript in tmpdir so `summarize` has something to read.
- **Plain-text `formatOutput`**: Kimi appends raw stdout to the model's context on `UserPromptSubmit`, so returning the Claude-style JSON envelope would inject literal JSON into the conversation. The adapter now returns a bare `additionalContext` string, or `undefined` (Kimi ignores `SessionStart` stdout entirely). `hook-io.ts` prints string output raw.
- **Expanded hook coverage (4 → 7)**: added the `SessionStart[startup|resume]` bare `start` command (previously the worker was never started by hooks), `PreToolUse[Read]` → `file-context`, and `PostToolUseFailure` → `observation`. SessionStart matcher scoped to `startup|resume`.
- **Uninstall/status** now match any `worker-service.cjs` hook command instead of a hardcoded subset.

## Test plan

- `bun run typecheck` — clean
- `bun test tests/cli/adapters/kimi.test.ts` — 21/21 pass (incl. `tool_output` remap, plain-text formatOutput, wire.jsonl transcript synthesis with a fake `KIMI_CODE_HOME`)
- `bun test tests/integration/kimi-code-installer.test.ts` — 7-hook install/uninstall/status expectations pass
- Full suite: 2577 pass / 18 skip / 1 fail (`mutation-sites.test.ts` adoptMergedWorktrees — 5s timeout flake, passes in isolation, unrelated)
- Smoke-tested the built `plugin/scripts/worker-service.cjs` against a live worker: `hook kimi-code observation` and `hook kimi-code summarize` both enqueue and the SDK generator runs

Happy to split the rebase from the behavior fixes, or fold changes back into #2908, whichever you prefer.


thedotmack:

> new host adapter (Kimi Code CLI) — plan-23 out-of-scope routes new hosts to the roadmap; P1s: transcript truncation, basename hook match deletes foreign hooks

thedotmack:
Routing to **#3611 (plan-23 Host Integration Contracts)** — a new non-Claude-Code harness adapter is precisely the cluster, and the payload-shape bugs it fixes are the reason plan-23 asks for contract tests against the host's real schema.

Heads up on overlap: #3676 is a second open Kimi Code CLI integration (newer, adds docs + wire fixtures + a transcript parser) and the two collide on `src/cli/adapters/kimi.ts`, the installer, `ide-detection` and `platform-source`. I checked #3676's diff and it does **not** handle Kimi's `tool_output`/`tool_call_id` PostToolUse shape or the missing `transcript_path` on `Stop`, which is the part this PR verified against a live install. Leaving both open — whichever branch wins needs those two corrections carried over.

PR #3555, opened 2026-08-11
Bound scheduled projection repair work

ChenglinWei97:
## What changed

- Bound the secret-gated scheduled projection repair endpoint to one projection page per user and invocation.
- Release the fencing lease after that page is durably checkpointed so the next cron can continue immediately.
- Return HTTP `202 Accepted` with the existing result schema while the captured repair target remains ahead of the durable checkpoint, and HTTP 200 once the target is reached.
- Keep client-triggered projection unchanged: calls that omit a page budget still drain through their requested target.
- Reject invalid page budgets before acquiring a lease.

## Why

The repair endpoint previously tried to drain an arbitrarily large backlog in one HTTP request. In production this produced timeouts and prolonged lease contention even though individual projection pages were succeeding. One page is already bounded to 100 operations or 4 MB, so scheduled repair can make deterministic progress without monopolizing the request or the per-user lease.

## Impact

Lagging users recover incrementally through successive bounded requests during the selected cron attempt. The paired Pro caller in [claude-mem-pro #99](https://github.com/thedotmack/claude-mem-pro/pull/99) ([exact scheduler commit](https://github.com/thedotmack/claude-mem-pro/blob/ce66fc0b3043eb496871ab3ea4026b48b1114a1b/src/lib/cmem/content-v2-repair.ts)) pins the first returned head, requeues incomplete users fairly, and stops starting pages early enough to preserve Vercel's route-response margin. The unchanged JSON schema keeps either deployment order compatible. Ordinary sync and client-triggered projection semantics do not change.

## Deployment

Merging this repository does not deploy the Cloudflare Worker. After merge, an authorized operator must run the documented `workers/sync-hub` Wrangler deployment and then verify the next Pro repair cron reports bounded progress.

## Validation

- `npm run lint` in `workers/sync-hub`
- `npm test` in `workers/sync-hub` — 116 tests passed
- `npm run typecheck:root`
- Added regression coverage for the 202-to-200 continuation contract, partial checkpoint progress, immediate lease release, continuation on the next repair call, and invalid budgets

### DIFF
--- a/workers/sync-hub/src/index.ts
+++ b/workers/sync-hub/src/index.ts
@@ -866,7 +866,13 @@ async function handleRepairDrain(request: Request, env: Env): Promise<Response> 
 			projected_through_seq: finalState.projected_seq,
 		});
 	}
-	return json(200, {
+	// A bounded repair that checkpointed one page is successful progress, but it
+	// is not a completed request until the original target has been reached.
+	// HTTP 202 preserves the existing JSON schema while giving the Pro caller an
+	// explicit continuation signal. Explicit through_seq callers are complete
+	// once their requested target is reached even if newer ops arrived meanwhile.
+	const complete = decimalAtLeast(finalState.projected_seq, target);
+	return json(complete ? 200 : 202, {
 		protocol_version: 1,
 		user_id: record.user_id,
 		epoch: finalState.epoch,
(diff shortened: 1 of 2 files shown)


ChenglinWei97:
@greptileai

ChenglinWei97:
@greptileai

ChenglinWei97:
@greptileai

thedotmack:

> step 5 (paged sync-hub repair with lease release); P1: Pro cron caller treats any 2xx as done

thedotmack:
Superseded by same-repo rehost #4046 (merged to `main`). This fork tip was behind and had messy indent in `drainProjection`; we rehosted the bound (one page per scheduled repair, 202 until target) on current `main` rather than merging it.

Issue #3540, opened 2026-08-10
[Windows] chroma-mcp respawn teardown correlates with orphaned uv builds-v0 temp dirs — 144GB confirmed


MartinHoblisch:
# [Windows] chroma-mcp respawn teardown correlates with orphaned `uv` `builds-v0` temp dirs — 144GB confirmed

## Before submitting

- [x] I searched existing issues and confirmed this is not an exact duplicate (closest related: #2897, #3175, #3138 — see "Relation to existing issues" below)

## Edit history

*2026-08-10: corrected the size figure from an unverified extrapolation ("~140GB, estimated") to a fully measured total (144.21GB, confirmed by summing every file in every one of the 696 directories — see Evidence). Also reworded the causal claims in Summary and Suggested Fix #1 that were originally stated as confirmed fact but were actually inference from code reading, not directly observed. Apologies for the initial imprecision.*

## Summary

On Windows, `ChromaMcpManager.killProcessTree()` tears down a `chroma-mcp` process tree via:

```ts
await execFileAsync('taskkill', ['/PID', String(pid), '/T', '/F'], { timeout: 5_000, windowsHide: true });
```

(`src/services/sync/ChromaMcpManager.ts`, Windows branch — called from both the singleton pre-existing-process cleanup and the normal `stop()` teardown path)

Unlike the POSIX branch (`SIGTERM`, ~500ms wait, then `SIGKILL` stragglers), the Windows branch goes straight to `/F` (force-terminate) with no wait step. **Hypothesis, not directly observed:** if `uv` is mid-build when this lands (materializing the ephemeral `uvx` venv into `builds-v0`), it never gets to run its own temp-dir cleanup, leaving a full partially/fully-built venv behind as `%LOCALAPPDATA%\uv\cache\builds-v0\.tmpXXXXXX` (chromadb, onnxruntime, grpcio, huggingface_hub, cohere, etc.). I have **not** correlated an actual kill-event timestamp with a specific orphaned directory's creation timestamp — I don't have per-event logs granular enough for that. What I can show is the aggregate state below, which is consistent with this mechanism but doesn't, on its own, rule out other causes of build interruption.

## Evidence (real install, Windows 10)

- `%LOCALAPPDATA%\uv\cache\builds-v0` contained **696** orphaned `.tmpXXXXXX` directories (direct count, `Get-ChildItem` / `Measure-Object`)
[...]

Issue #3482, opened 2026-08-02
Windows: orphaned chroma child inherits the worker's listening socket, permanently blocking port 37777


NeilWong4424:
﻿## Relationship to #2578 / #2599

#2578 (and its fix #2599) covers **PID reuse** — a stale `worker.pid` matching an unrelated new process. This is a **different failure**: the PID is *not* reused, it simply no longer exists, yet the socket it opened is still in `Listen` state because a surviving descendant inherited the handle. #2599's `captureProcessStartToken` fix does not help here — there is no process to token-check.

## Summary

When the worker dies, the `chroma-mcp` tree it spawned survives. On Windows those children inherit the worker's **listening socket handle** for port 37777. The port stays `Listen`-owned by a dead PID, so every subsequent worker start aborts with `Port already in use, refusing to start duplicate`. The worker can never recover unattended.

## Evidence

`supervisor.json` after the worker died:

```json
{
  "processes": {
    "worker":     { "pid": 28620, "startedAt": "2026-08-01T19:52:26.451Z" },
    "chroma-mcp": { "pid": 29580, "startedAt": "2026-08-01T19:53:02.991Z", "pgid": 29580 }
  }
}
```

Both PIDs are gone:

```
> tasklist /FI "PID eq 28620"
INFO: No tasks are running which match the specified criteria.
```

Yet the socket is still held by that dead PID:

```
> Get-NetTCPConnection -LocalPort 37777
LocalAddress LocalPort  State OwningProcess
127.0.0.1        37777 Listen         28620
```

Nothing is serving:

```
> Invoke-WebRequest http://127.0.0.1:37777/health -UseBasicParsing
probe failed: Unable to connect to the remote server
```

The surviving orphan tree:

```
worker (28620)            <- dead; socket leaked to descendants
 └─ chroma-mcp (29580)    <- dead
      └─ uv.exe (5996)              ALIVE
           └─ chroma-mcp (19100)    ALIVE
                └─ python (26640)   ALIVE
                     └─ python (20592) ALIVE
```

## Resulting log loop

```
[05:58:43] Worker not running - lazy-spawning {runtimePath=...\bun.exe, scriptPath=...\worker-service.cjs}
[05:58:43] Port already in use, refusing to start duplicate {port=37777}
[05:58:58] Worker port did not open after lazy-spawn within the cold-boot wait (~15s)
[05:59:07] Port in use but worker not responding to health checks
```

[...]

PR #3567, opened 2026-08-12
fix(chroma): isolate uvx children from foreign Python venvs (#3552)

### LINKED ISSUE #3552: [Windows] chroma-mcp inherits the spawning process's Python env, defeating uvx isolation — semantic sync silently stops
## Summary

The worker spawns `chroma-mcp` through `uvx --python 3.13`, which looks hermetic but is not: the child inherits the environment of whatever process started the worker. If the worker happens to be launched from a shell with a Python virtualenv active, the 3.13 interpreter resolves `numpy` out of that venv's `site-packages`. When the venv is a different minor version, the C extension fails to load and **all** Chroma sync stops.

The failure is silent at `WARN` level. SQLite capture keeps working, so the session looks healthy — observations accumulate and are never indexed, and semantic search silently stops returning anything recent.

Version: `13.15.0`, Windows 11, worker runtime.

## Evidence

```
[CHROMA_MCP] Prewarming chroma-mcp uvx environment
  {command=...\uvx.exe, args=--python 3.13 --with onnxruntime>=1.20 --with protobuf<7
   --from chroma-mcp==0.2.6 chroma-mcp --help}
[CHROMA_MCP] chroma-mcp uvx prewarm failed
  File "...\uv\cache\archive-v0\<hash>\Lib\site-packages\chromadb\api\types.py", line 20
    from numpy.typing import NDArray
  File "C:\Users\<user>\AppData\Local\<other-app>\venv\Lib\site-packages\numpy\__init__.py"
    ...
  The NumPy version is: "2.4.3"
  Original error was: No module named 'numpy._core._multiarray_umath'
[CHROMA_SYNC] Chroma unavailable before write; leaving documents unsynced
  {collection=cm__claude-mem, requested=7, written=0}
[CHROMA_SYNC] Observation watermark bump skipped — partial write
  {observationId=325, requested=7, written=0}
```

Note the mixed provenance: `chromadb` is loaded from the uv cache as intended, but `numpy` comes from an unrelated virtualenv outside uv's control. That venv is Python 3.11, so its `_multiarray_umath.cp311-win_amd64.pyd` cannot be loaded by the 3.13 interpreter.

## Observed vs. inferred

Observed:

- The identical `uvx ... chroma-mcp --help` command succeeds from a clean shell on the same machine, same uv cache.
- `PYTHONPATH` and `PYTHONHOME` are empty in user and machine scope.
- 18 previous daily log files contain zero occurrences of this error; it appeared only after the worker was restarted from a shell that had the foreign venv active, then produced 72 occurrences in ~13 minutes.
- Killing the worker and letting it respawn from a clean shell fixed it immediately: prewarm succeeded in 1.6s and the backlog (12 observations, 2 summaries, 4 prompts) backfilled automatically.

Inferred (not directly confirmed):

- The exact variable carrying the contamination. I could not read the worker process's environment block on Windows to distinguish `VIRTUAL_ENV` from a `PATH` entry. The mechanism is env inheritance, but which key is responsible is unverified.

## Impact

The blast radius is larger than "one failed sync" because the watermark is not advanced on partial writes, so every subsequent observation also fails to sync. Semantic search degrades to silence while every other signal — worker health endpoint, hook failure counter, SQLite writes, session-start context injection — reports healthy. There is no user-visible indication that the memory system is now half dead.

## Suggested fix

Sanitize the environment when spawning the uvx child rather than inheriting it. Stripping `VIRTUAL_ENV`, `PYTHONPATH`, `PYTHONHOME`, and any venv `Scripts`/`bin` entries prepended to `PATH` would make the isolation match what `--python 3.13` implies.

Two smaller improvements that would have made this self-diagnosing:

1. Log the resolved interpreter and `numpy.__file__` on prewarm failure. The provenance mismatch is the whole diagnosis and it is currently only visible by reading the traceback carefully.
2. Surface repeated `CHROMA_SYNC` failures somewhere the user actually sees. A `WARN` in a daily log file is not discoverable; the sync had been failing for hours before anyone noticed.

## Workaround

Restart the worker from a shell with no virtualenv active. The stale writer lock and the backlog both clear on their own once the child can start.

---

*Diagnosed and written by Claude Opus 5, running in Claude Code on the affected machine. Every item under "Observed" is the output of a command actually executed there, not inference from reading the source — I never opened claude-mem's code. The one thing I could not verify is separated out under "Inferred" instead of being presented as fact.*

stantheman0128:
## Summary

Fixes #3552.

`uvx --python 3.13` looked hermetic, but the chroma-mcp child still inherited the worker's activated virtualenv (`VIRTUAL_ENV` / `PYTHONPATH` / `PYTHONHOME` / conda keys, plus matching PATH prefixes). On Windows that pulled a foreign `numpy` ABI into the uv cache interpreter and silently stopped Chroma sync while SQLite stayed healthy.

## Fix

- Add `stripForeignPythonEnv` next to `sanitizeEnv` and run it on the chroma/uvx spawn env.
- Strip foreign Python/conda activation keys and venv `Scripts`/`bin` PATH entries.
- On prewarm failure, log parent contamination booleans (`parentHadVirtualEnv` / `PYTHONPATH` / `PYTHONHOME`) so the next diagnosis does not require reading a mixed traceback.

## Level of fix

Fixed at the chroma spawn env boundary because that is the only place `uvx --python` promised isolation. Broader `sanitizeEnv` stays Claude-Code-leak focused so unrelated subprocesses are unchanged.

## Evidence

```
bun test tests/supervisor/env-sanitizer.test.ts
16 pass
0 fail
```

Also re-ran `chroma-mcp-manager-path` (6 pass). Pre-existing `spawn-contract-windows` cmd.exe path expectations failed on this box (`cmd.exe` vs `C:\WINDOWS\system32\cmd.exe`) and are unrelated to this diff.

## What was not tested

- Live Windows repro with an activated foreign venv + real chroma-mcp prewarm (would need a conflicting numpy install). Unit tests cover the env scrub that removes that contamination vector.

## AI assistance

Implemented with Cursor assistance. Human: Stan Shih (@stantheman0128).

### DIFF
--- a/src/services/sync/ChromaMcpManager.ts
+++ b/src/services/sync/ChromaMcpManager.ts
@@ -10,7 +10,7 @@ import { logger } from '../../utils/logger.js';
 import { SettingsDefaultsManager } from '../../shared/SettingsDefaultsManager.js';
 import { USER_SETTINGS_PATH, paths } from '../../shared/paths.js';
 import { getUvxBinDirs } from '../../shared/uvx-bin-dirs.js';
-import { sanitizeEnv } from '../../supervisor/env-sanitizer.js';
+import { sanitizeEnv, stripForeignPythonEnv } from '../../supervisor/env-sanitizer.js';
 import { getSupervisor } from '../../supervisor/index.js';
 import { captureProcessStartToken, isPidAlive } from '../../supervisor/process-registry.js';
 import { clearDependencyStatus, recordChromaVectorSearchUnavailable, recordUvxVectorSearchUnavailable } from '../../shared/dependency-health.js';
@@ -671,6 +671,11 @@ export class ChromaMcpManager {
         timeoutMs,
         ...(pid ? { pid } : {}),
         error: errorMessage,
+        // Parent contamination signals (#3552): child env is scrubbed, but these
+        // tell operators why a prior unclean spawn may have failed.
+        parentHadVirtualEnv: Boolean(process.env.VIRTUAL_ENV),
+        parentHadPythonPath: Boolean(process.env.PYTHONPATH),
+        parentHadPythonHome: Boolean(process.env.PYTHONHOME),
         ...(stdout ? { stdoutTail: stdout } : {}),
         ...(stderr ? { stderrTail: stderr } : {})
       });
@@ -1371,7 +1376,12 @@ export class ChromaMcpManager {
 
   private static getUvxPreflightEnv(): Record<string, string> {
     const baseEnv: Record<string, string> = {};
-    for (const [key, value] of Object.entries(sanitizeEnv(process.env))) {
+    // sanitizeEnv strips host Claude Code vars; stripForeignPythonEnv then
+    // removes activated-venv / conda leakage so uvx --python stays hermetic
+    // even when the worker was started from a foreign virtualenv (#3552).
+    for (const [key, value] of Object.entries(
+      stripForeignPythonEnv(sanitizeEnv(process.env)),
+    )) {
       if (value !== undefined) {
         baseEnv[key] = value;
       }
--- a/src/supervisor/env-sanitizer.ts
+++ b/src/supervisor/env-sanitizer.ts
@@ -59,3 +59,88 @@ export function sanitizeEnv(env: NodeJS.ProcessEnv = process.env): NodeJS.Proces
 
   return sanitized;
 }
+
+/**
+ * Python / venv keys that poison hermetic `uvx --python …` children when the
+ * worker itself was launched from an activated virtualenv (#3552).
+ *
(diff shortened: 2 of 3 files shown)


thedotmack:

> step 1 exactly (sanitized env at the chroma spawn boundary: strip VIRTUAL_ENV/PYTHONPATH/conda)

thedotmack:

stantheman0128:
Thanks. v13.16.1 covering the chroma tree-kill and the foreign-Python isolation is the right height. We'll treat #3567 as absorbed. If orphans still show up on 13.16.1 on this Win11 box I'll file a fresh issue.

PR #3570, opened 2026-08-12
fix(install): register claude-mem MCP server in opencode.json

Bleed00:
## Why

The OpenCode installer only wrote the plugin reference (`plugin: ["./plugins/claude-mem.js"]`), so the `claude-mem` MCP server never appeared in OpenCode. Users had to hand-edit `opencode.json` to add an `mcp.claude-mem` block before the MCP was discoverable.

## Changes

`src/services/integrations/OpenCodeInstaller.ts`:

- `addOpenCodeMcpReference` / `removeOpenCodeMcpReference` write/remove only `mcp["claude-mem"] = { "type": "local", "command": [node, "<absolute mcp-server.cjs>"] }`, preserving all other MCP servers. OpenCode does no `${CLAUDE_PLUGIN_ROOT}` substitution, so the path is baked absolute via the shared resolvers in `install-paths.ts` (Rule B).
- Registration writes plugin + MCP in one pass; deregistration removes both; `checkOpenCodeStatus` now reports MCP registration state.
- Fails safe: if the MCP script cannot be resolved, the plugin is still registered but no `mcp` block is written (with a logged warning), so a broken build never corrupts the user's `opencode.json`.

## Tests

`tests/integration/opencode-installer.test.ts`: MCP add/remove/idempotency, merge with existing MCP servers (e.g. `context7`), removal on deregistration, and the empty-`mcp`-block cleanup. The entry shape is asserted against OpenCode's local-server schema (`type: "local"`, `command: string[]`).

## Verification

- `npx tsc --noEmit` — clean
- `npm run build` — clean
- `tests/integration/` + `tests/integrations/`: 73/73 pass (incl. opencode-plugin-contract)

### DIFF
--- a/src/services/integrations/OpenCodeInstaller.ts
+++ b/src/services/integrations/OpenCodeInstaller.ts
@@ -6,8 +6,10 @@ import { existsSync, readFileSync, writeFileSync, mkdirSync, copyFileSync, unlin
 import { logger } from '../../utils/logger.js';
 import { CONTEXT_TAG_OPEN, CONTEXT_TAG_CLOSE, injectContextIntoMarkdownFile } from '../../utils/context-injection.js';
 import { getWorkerHost, getWorkerPort } from '../../shared/worker-utils.js';
+import { getMcpServerAbsolutePath, getNodeAbsolutePath } from './install-paths.js';
 
 const OPENCODE_PLUGIN_CONFIG_PATH = './plugins/claude-mem.js';
+const OPENCODE_MCP_SERVER_KEY = 'claude-mem';
 
 type OpenCodeConfig = {
   $schema?: string;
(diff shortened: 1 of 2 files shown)


Bleed00:
Thanks for the review — P1 addressed in `64382007`.

`resolveDefaultOpenRouterModel` now has a final fallback for a reachable, non-empty catalogue with no curated-model overlap: it returns a deterministic advertised free model (first id in sorted order) instead of persisting the static default. The static default is now only used when the catalogue is empty or unreachable.

Coverage added in `tests/npx-cli/openrouter-models.test.ts`: `uses an advertised free model when the catalogue has no curated overlap` (mocked catalogue returning only `acme/catalogue-only-model:free`).

thedotmack:

> OpenCode installer must also register the MCP entry (step 2/3); OpenRouter-model half partly landed via #3524

Bleed00:
Realigned per the plan-23 routing note (#3611). Two changes:

1. **Dropped the OpenRouter free-model picker half.** It's partly landed via #3524 (install now writes `CLAUDE_MEM_OPENROUTER_MODEL` on the CMEM Pro path), so this PR no longer touches `src/npx-cli/cmem-pro-costs.ts`, `src/npx-cli/commands/install.ts`, or the `openrouter-models` test.

2. **Kept and rebased only the OpenCode MCP registration** onto current `main` (fae697a4). It now relies on the shared resolvers already on main (`getMcpServerAbsolutePath` / `getNodeAbsolutePath` in `install-paths.ts`, Rule B) rather than adding any new path plumbing, and the entry shape is asserted against OpenCode's local-server MCP schema (`type: "local"`, `command: string[]`) in `tests/integration/opencode-installer.test.ts`.

This is the "OpenCode installer must also register the MCP entry" piece only. Typecheck + build clean; integration suites 73/73 pass. Ready to be folded into the plan-23 bundle (step 2/3) or reviewed standalone.

Bleed00:
Superseded by #3621 — reopened under branch `fix/opencode-mcp-registration` (name now matches the narrowed MCP-only scope) and extended with the plan-23 step-1 host fixture (`fixtures/hosts/opencode-mcp.json`).

PR #3571, opened 2026-08-13
fix(chroma): make persistent backfills durable

HDK1218:
## Summary

Make local persistent Chroma rebuilds durable, bounded, and source-complete without leaving a sustained foreground load.

## Root cause

Several independent costs compounded during a large rebuild:

- MCP mutations inherited the SDK's 60-second request deadline. A timeout triggered reconnect teardown while Chroma could still be committing.
- `chroma-mcp` 0.2.6 scans all existing IDs before every add, making repeated backfill calls quadratic.
- Chroma's default ONNX embedding function was reconstructed for each collection facade/request.
- Observations and summaries emitted one vector per field, and exact prompt replays were indexed repeatedly.
- NUL-containing documents can make Chroma 1.5.9's trigram FTS index fail `PRAGMA quick_check` with a malformed inverted index.
- Project discovery omitted prompt-only namespaces, and pending write failures could still end with an all-project completion message.

## Changes

- Add a bounded, configurable 10-minute timeout for mutation tools only; queries retain the normal SDK deadline.
- Ship a pinned `chroma-mcp` bridge that:
  - bypasses the full-collection duplicate scan for adds,
  - reuses one locked ONNX embedding runtime,
  - sanitizes NUL text defensively,
  - lowers local Chroma priority to nice 15.
- Pack source rows into 100-document mutations while preserving pending-gap/watermark correctness.
- Emit one prioritized semantic projection per observation or summary.
- Deduplicate exact prompt replays within each session, keeping the latest row.
- Discover projects from observations, summaries, and prompts.
- Retry incomplete projects once, then fail the rebuild if any failed or pending rows remain.
- Package the Python bridge and add regression coverage for timeout scoping, batching, repair, prompt deduplication, prompt-only projects, and NUL sanitization.

## Validation

- `bun test tests/services/sync`: 53 passed, 0 failed.
- `npm run typecheck`: root and viewer typechecks passed.
- Package dry-run: 116 files; Python bridge and worker bundle both included.
- Full persistent rebuild validation:
  - 57 project namespaces,
  - 31,888 Chroma documents exactly matching 10,781 observations + 3,358 summaries + 17,749 deduplicated prompts,
  - zero pending rows and zero watermark mismatches,
  - stopped-state `PRAGMA quick_check` and `integrity_check` both returned `ok`,
  - semantic API search returned results for a query with zero SQLite FTS matches,
  - one worker and one persistent Chroma Python process; both measured 0% CPU while idle after rebuild.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "plugin/modes",
     "plugin/scripts/*.js",
     "plugin/scripts/*.cjs",
+    "plugin/scripts/*.py",
     "plugin/sqlite",
     "plugin/skills",
     "plugin/ui",
--- a/plugin/scripts/chroma-mcp-bridge.py
+++ b/plugin/scripts/chroma-mcp-bridge.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python3
+"""claude-mem's pinned chroma-mcp bridge.
+
+chroma-mcp 0.2.6's ``chroma_add_documents`` fetches every existing ID before
+each add. That makes a rebuild quadratic and pushes ordinary batches past the
+MCP client's 60-second deadline once a collection grows. Chroma's native
+``Collection.add`` already rejects duplicate IDs, so this bridge replaces only
+that tool with a direct, validation-equivalent call and leaves every other
+upstream tool untouched.
+"""
+
+import os
+from threading import Lock
+from typing import Dict, List
+
+
+def _lower_process_priority() -> None:
+    """Keep local embedding/index work from competing with interactive apps."""
+    if not hasattr(os, "setpriority") or not hasattr(os, "PRIO_PROCESS"):
+        return
+
+    try:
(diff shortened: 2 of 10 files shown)


thedotmack:

> forces step 0 (contract or retire): pinned Python bridge, batched mutations, mutation-only timeout; P1: live prompt sync still duplicates

thedotmack:
Routing to **#3610 plan-22 (Chroma Sidecar Contract)** — pinned `chroma-mcp`, mutation-vs-query deadlines, watermarks, and backfill durability are all named in the master, and this PR is essentially plan-22's fix list in one branch.

Leaving open. Note the master's open question is whether the sidecar is repaired or retired; this is the strongest existing "repair it" implementation, so it should be weighed against that decision rather than merged in isolation. It also overlaps the chroma slice of the #3614 bundle and the pinning work in #3384 — those three need reconciling before any of them land.

PR #3572, opened 2026-08-13
fix(cleanup): tolerate negative statfs availability

### LINKED ISSUE #3551: v12.4.3 cleanup permanently fails with "Insufficient disk" on linux-x64 (free<0, not the #2683 bsize=0 case) — Bun statfsSync int32 truncation
## Before submitting

- [x] Searched `statfs`, `bavail`, `Insufficient disk`, `int32 truncation`, `8TB`, `negative free`. The closest matches, #2683 / #2681, cover the **darwin-x64 `bsize=0`** case. This is the **linux-x64 negative-`bavail`** variant, which the #2681 guard does not catch. No existing issue covers it.

## Bug Description

On linux-x64 with more than ~8 TiB available on the `$HOME` filesystem, the v12.4.3 cleanup permanently fails with `Insufficient disk for v12.4.3 backup; skipping cleanup` despite petabytes free. Unlike #2683, `bsize` here is perfectly valid — only `bavail` is corrupted, into a *negative* number, which slips past the guard added in #2681.

Root cause chain:

1. Bun truncates the non-bigint `fs.statfsSync()` block fields to signed int32 on linux-x64 — oven-sh/bun#31510, fixed by oven-sh/bun#36503 (merged 2026-07-31). **No released Bun contains the fix**: the newest release, 1.3.14, predates the merge by two months, so upgrading is not a remedy.
2. Where available blocks exceed `2^31 - 1`, the value wraps negative.
3. `worker-service.cjs` computes `free = bavail * bsize` — negative.
4. The #2681 guard tests `!Number.isFinite(bsize) || !Number.isFinite(bavail) || bsize <= 0`. A negative `bavail` is finite and `bsize` is correct, so the reading is accepted as credible.
5. `free < required` is therefore always true → early `return`. The `.cleanup-v12.4.3-applied` marker is never written, so the cleanup retries on every worker cold-boot and re-emits the same `[ERROR]` forever (28 times in one day here).

### Evidence

Four sources, same path, one sampling window (`bsize = 1048576`; the kernel value was re-read afterwards and had not changed, so all four rows describe the same filesystem state):

| source | `bavail` | |
| --- | --- | --- |
| kernel `statfs(2)` | `68702437328` | reference |
| node v20.19.6 | `68702437328` | correct |
| **Bun 1.3.14** (newest release) | **`-17039408`** | **wrong** |
| Bun 1.4.0-canary (`23d233b20`) | `68702437328` | correct — has oven-sh/bun#36503 |

`68702437328 mod 2^32` as a signed int32 is exactly `-17039408`, confirming plain int32 truncation. The canary row is the practical point: the upstream fix works, but exists only in canary, so every user on a released Bun is affected with no version to upgrade to.

Resulting log line:

```
[ERROR] [SYSTEM] Insufficient disk for v12.4.3 backup; skipping cleanup
(will retry on next startup) {dbSize=93118464, free=-178622772}
```

## Steps to Reproduce

1. On linux-x64, put `$HOME` on a filesystem whose available blocks exceed `2^31` (>8 TiB at 4 KiB blocks; here a 64 PiB NFS mount with 1 MiB blocks, ~1% used).
2. Compare Bun against the kernel:

   ```sh
   bun -e "const s=require('fs').statfsSync(process.env.HOME);console.log(s.bavail, s.bavail*s.bsize)"
   stat -f -c 'bavail=%a bsize=%s' "$HOME"
   ```

   Bun reports a negative product; the kernel (and node, same call) reports the correct positive value.
3. Cold-boot the worker: the `Insufficient disk` ERROR appears on every startup and the cleanup never completes.

## Expected vs Actual

**Expected:** a negative free-space reading is rejected as non-credible, exactly as `bsize <= 0` is in #2681, and the gate is skipped with a `[WARN]`. A genuine out-of-space condition still surfaces from the subsequent `VACUUM INTO` / `copyFileSync`.

**Actual:** the negative reading passes the guard and is read as a real out-of-space condition. The cleanup never runs on any affected host and the database grows unbounded (89 MB here).

## Suggested Fix

One-line extension of the #2681 guard:

```js
if (!Number.isFinite(bsize) || !Number.isFinite(bavail) || bsize <= 0 || bavail < 0)
```

Worth doing independently of Bun's release cycle: claude-mem resolves Bun from `PATH` rather than bundling it (`scripts/bun-runner.js`) and `engines.bun` is `>=1.0.0`, so users on an older Bun stay broken indefinitely even after a fixed Bun ships.

Alternatively `fs.statfsSync(path, { bigint: true })` is unaffected by the truncation and fixes the reading today — on this host, under the same Bun 1.3.14 and the same sampling window, it returns `68702437328`, matching the kernel exactly.

## Environment

claude-mem 13.15.0 · Bun 1.3.14 (worker runtime, newest release) · Bun 1.4.0-canary `23d233b20` (comparison only) · node v20.19.6 (comparison only) · linux-x64, Ubuntu, kernel 5.15 · `$HOME` on NFS, 64 PiB, 1 MiB blocks, ~1% used

dajiaohuang:
## Summary
- treat negative `statfsSync().bavail` values as non-credible disk telemetry
- bypass only the pre-flight disk-space gate for that case
- add a regression test that verifies cleanup and backup still complete

Closes #3551

## Test plan
- `npm run typecheck` — passed
- `npm run build` — passed
- `npx -y bun@1.3.14 test tests/infrastructure/cleanup-v12_4_3.test.ts` — assertions complete, but Windows cleanup hits the file's existing SQLite `EBUSY` in `afterEach` (also affects the existing cases); Linux CI should provide the authoritative result

### DIFF
--- a/src/services/infrastructure/CleanupV12_4_3.ts
+++ b/src/services/infrastructure/CleanupV12_4_3.ts
@@ -142,10 +142,10 @@ function executeCleanup(dbPath: string, effectiveDataDir: string, markerPath: st
     // and will ship in the next Bun release after 1.3.14. Until then, any
     // `bavail * bsize` math returns 0 and this gate would permanently skip
     // the cleanup with a misleading `free=0` error. Treat non-credible
-    // readings (bsize <= 0, NaN, or non-finite) as "skip the gate" rather
+    // readings (bsize <= 0, bavail < 0, NaN, or non-finite) as "skip the gate" rather
     // than "disk is full" -- a real out-of-space condition will still
     // surface from the subsequent VACUUM INTO / copyFileSync.
-    if (!Number.isFinite(bsize) || !Number.isFinite(bavail) || bsize <= 0) {
+    if (!Number.isFinite(bsize) || !Number.isFinite(bavail) || bsize <= 0 || bavail < 0) {
       logger.warn(
         'SYSTEM',
         'statfsSync returned non-credible values; proceeding without disk-space pre-flight',
--- a/tests/infrastructure/cleanup-v12_4_3.test.ts
+++ b/tests/infrastructure/cleanup-v12_4_3.test.ts
@@ -211,6 +211,41 @@ describe('runOneTimeV12_4_3Cleanup', () => {
     );
   });
 
+  it('proceeds with cleanup when statfsSync returns a negative available block count', () => {
+    const dbPath = path.join(tmpDataDir, 'claude-mem.db');
+    seedDatabase(dbPath, { observerSessions: 2, stuckCount: 10 });
+
+    const statfsSpy = spyOn(fs, 'statfsSync').mockImplementation(() => ({
+      type: 0,
+      bsize: 4096,
+      blocks: 4096,
+      bfree: 1048576,
+      bavail: -1,
+      files: 0,
+      ffree: 0,
+    }) as unknown as ReturnType<typeof fs.statfsSync>);
+
+    try {
+      runOneTimeV12_4_3Cleanup(tmpDataDir);
+    } finally {
+      statfsSpy.mockRestore();
+    }
+
+    const markerPath = path.join(tmpDataDir, '.cleanup-v12.4.3-applied');
+    expect(existsSync(markerPath)).toBe(true);
+    const payload = JSON.parse(readFileSync(markerPath, 'utf8'));
+    expect(payload.counts.observerSessions).toBe(2);
+    expect(payload.counts.stuckPendingMessages).toBe(10);
+    expect(payload.backupPath).toBeTruthy();
+    expect(existsSync(payload.backupPath)).toBe(true);
+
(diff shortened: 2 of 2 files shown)


dajiaohuang:
﻿Thanks for flagging the distribution boundary. This PR targets `core-dev` per `docs/public/branches.mdx`; that branch is run from source with `npm run build-and-sync`, while stable artifacts are produced when changes are promoted to `main`.

I ran `npm run build` on this commit and verified that the generated worker contains the negative-`bavail` guard. I have kept the PR source/test-only, consistent with other `core-dev` changes, rather than committing broad bundle churn from the repository's unlocked root dependency install.

thedotmack:

> step 5 exactly (free-space check negative-guarded)

thedotmack:
Rebased onto current `main` and shipped as #4014.

PR #3574, opened 2026-08-13
fix(transcripts): scan configured dot-directories

### LINKED ISSUE #3512: transcript-watcher: Bun.Glob needs dot: true — transcripts behind dot-directories are never discovered (blocks Antigravity IDE capture)
### Summary

`TranscriptWatcher.scanGlob()` calls `Bun.Glob(...).scanSync()` without `dot: true`, so **any transcript whose path contains a dot-directory is invisible to the watcher** — the glob silently returns zero matches and no session is ever ingested.

This is what blocks passive capture for the **Antigravity desktop IDE**, whose transcripts live at:

```
~/.gemini/antigravity/brain/<conversation-id>/.system_generated/logs/transcript_full.jsonl
                                              ^^^^^^^^^^^^^^^^^^
```

Related: #3239 (Antigravity GUI IDE gets no automatic capture). That issue proposes a native GUI integration, but the GUI [reportedly does not execute plugin hooks at all](https://discuss.ai.google.dev/t/do-antigravity-ide-2-0-actually-execute-plugin-hooks-pretooluse-posttooluse-or-is-that-cli-only-right-now/176814) — users report "zero hook invocations" on IDE and Desktop 2.0 while the `agy` CLI works. So the transcript watcher is arguably the *only* viable capture route there, and this one flag is what stands in the way.

### Offending code

`scripts/transcript-watcher.cjs` (and the same helper bundled into `scripts/worker-service.cjs`):

```js
scanGlob(e) {
  return Array.from(new Bun.Glob(e).scanSync({ absolute: !0, onlyFiles: !0 }))
}
```

### Reproduction

```js
const base = 'C:/Users/stefa/.gemini/antigravity/brain';
const pat  = base + '/**/.system_generated/logs/transcript_full.jsonl';
const scan = (p, o) => Array.from(new Bun.Glob(p).scanSync({ absolute: true, onlyFiles: true, ...o }));

console.log('without dot:', scan(pat).length);        // 0
console.log('with dot   :', scan(pat, {dot:true}).length); // 2
```

```
without dot: 0
with dot   : 2
```

Note the pattern names `.system_generated` **explicitly** and still does not match — Bun.Glob requires the `dot` option regardless of whether the dot component is literal or a wildcard.

### Suggested fix

```diff
-return Array.from(new Bun.Glob(e).scanSync({absolute:!0,onlyFiles:!0}))
+return Array.from(new Bun.Glob(e).scanSync({absolute:!0,onlyFiles:!0,dot:!0}))
```

Low risk: watch patterns are explicit, user-configured paths, so enabling `dot` widens matching only where the configured pattern already points.

### Verified end-to-end after patching

With `dot: true` applied and a schema added for the Antigravity transcript format, capture works completely:

```
[TRANSCRIPT] Watching transcript file {file=...\.system_generated\logs\transcript_full.jsonl, watch=antigravity}
[SESSION]    CREATED | contentSessionId=<uuid> → sessionDbId=1456 | project=antigravity
[DB]         STORED  | obsCount=1 | obsIds=[12109]
```

```
sdk_sessions by platform_source:  claude: 1446 | codex: 9 | antigravity: 1
observation #12109 [discovery] proj=antigravity
```

`extractSessionIdFromPath()` already works unmodified here, since the Antigravity path embeds the conversation UUID.

### Environment

- claude-mem 13.12.4
- Windows 11, Bun 1.3.10
- Antigravity desktop IDE (VS Code fork), transcripts under `~/.gemini/antigravity/brain/`

### Note on the schema side

Two things worth documenting for anyone writing a schema for a transcript that puts the tool call and its result on the **same** line (as Antigravity does):

1. `handleToolUse` emits the observation immediately when both `toolName` and `toolResponse` resolve — no `tool_result` event and no id pairing needed.
2. `toolId` is only honored when it resolves to a **string**; a numeric field (e.g. `step_index`) silently yields `undefined` and the pending-tool path never fires.

Happy to open a PR with the one-line fix if useful.

dajiaohuang:
## Summary
- enable Bun glob matching through dot-directories for transcript watches
- keep matching constrained to the configured watch pattern
- add a behavioral regression test using a real `.system_generated/logs` path

Closes #3512

## Test plan
- focused dot-directory transcript watcher test: passed
- npm run typecheck: passed

### DIFF
--- a/src/services/transcripts/watcher.ts
+++ b/src/services/transcripts/watcher.ts
@@ -223,7 +223,7 @@ export class TranscriptWatcher {
   }
 
   private scanGlob(pattern: string): string[] {
-    return Array.from(new Bun.Glob(pattern).scanSync({ absolute: true, onlyFiles: true }));
+    return Array.from(new Bun.Glob(pattern).scanSync({ absolute: true, onlyFiles: true, dot: true }));
   }
 
   private normalizeGlobPattern(inputPath: string): string {
--- a/tests/transcripts/watcher-start-at-end.test.ts
+++ b/tests/transcripts/watcher-start-at-end.test.ts
@@ -1,7 +1,7 @@
 import { afterEach, beforeEach, describe, expect, it, mock, spyOn } from 'bun:test';
 import { appendFileSync, mkdirSync, rmSync, writeFileSync } from 'fs';
 import { tmpdir } from 'os';
-import { join } from 'path';
+import { join, resolve } from 'path';
 import type { NormalizedHookInput } from '../../src/cli/types.js';
 import type { TranscriptSchema, WatchTarget } from '../../src/services/transcripts/types.js';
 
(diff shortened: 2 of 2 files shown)


thedotmack:

> step 3 exactly (`Bun.Glob(..., {dot: true})` for hidden transcript dirs)

thedotmack:
Rebased onto current `main` and shipped as #4016.

PR #3576, opened 2026-08-13
fix(openrouter): accept empty reasoning responses

### LINKED ISSUE #3569: OpenRouter init fails on empty content from reasoning models (DeepSeek V4 Flash)
## Version

claude-mem 13.15.0 (worker-service.cjs)

## Symptom

Intermittent `[SDK] Empty response from OpenRouter` followed by `Empty OpenRouter init response - session may lack context` on session init, when using the OpenRouter provider slot with a **custom OpenAI-compatible base URL** (OpenCode Zen/Go gateway, `https://opencode.ai/zen/go/v1`) and model `deepseek-v4-flash`.

## Root cause

`queryOpenRouterMultiTurn` rejects any response whose `choices[0].message.content` is empty:

```js
if(!d.choices?.[0]?.message?.content)return h.error("SDK","Empty response from OpenRouter"),{content:""};
```

DeepSeek V4 Flash is a reasoning model. For prompts of the form "extract an observation in XML, or output nothing if nothing notable" (which is what the init query sends against conversation history), it legitimately returns `content: ""` with `finish_reason: "stop"`, while `reasoning_content` is non-empty (100–300 chars). The SDK treats this legitimate no-op as a hard failure.

## Reproduction

6 sequential calls to the endpoint with an "output nothing if nothing notable" style prompt:

- 5/6 returned `content: ""`, `finish_reason: "stop"`, non-empty `reasoning_content`
- 1/6 returned short filler content

10/10 control calls with a concrete instruction always returned content.

## Impact

- Session-init observation dropped (context observation at session start lost)
- Error-log noise on every restart with thin history
- Reasoning tokens billed for empty output
- Observation generation via the message loop is **unaffected** — init failure does not abort the session

## Suggested fix

When `content` is empty but `reasoning_content` is present (or `finish_reason === "stop"`), treat the response as an idle/no-op instead of an error — or retry once with an explicit "you must output at least one character" instruction.

Likely affects any reasoning model served through OpenAI-compatible endpoints (OpenRouter, Zen, etc.), not just DeepSeek.

dajiaohuang:
## Summary
- distinguish a structurally missing OpenRouter message from a present empty string response
- accept empty messages produced by reasoning models without logging an SDK error
- preserve usage, served-model metadata, assistant history, and initialization bookkeeping
- retain the malformed-response guard when the message or content field is absent

Closes #3569

## Test plan
- OpenRouter empty-content and initialization tests: 3 passed
- npm run typecheck: passed
- npm run build: passed

### DIFF
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -45,7 +45,7 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
   /** Prefix for the synthetic memorySessionId (e.g. 'gemini', 'openrouter'). */
   protected abstract readonly syntheticIdPrefix: string;
   /**
-   * When a query returns empty content for an observation/summary message:
+   * When a query returns empty content:
    * OpenRouter still calls processAgentResponse('') (forwards the empty batch
    * to the parser/recovery path); Gemini skips it and logs a warning. This flag
    * preserves that per-provider divergence.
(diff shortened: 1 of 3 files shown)


dajiaohuang:
Addressed the empty-init bookkeeping finding in ae732aaf. handleInitResponse now follows the provider's existing forwardEmptyMessageResponse contract, so an explicit empty OpenRouter message retains assistant history, token usage, and served-model processing while structurally missing content still takes the error path. The expanded provider tests pass (3/3).

thedotmack:

> step 4 exactly (empty content vs missing content; empty init response forwarded)

thedotmack:
Rebased onto current `main` and shipped as #4017.

Diff stayed on the OpenRouter / OpenAICompatible empty-content path (query accepts `content: ''` and keeps usage; init records the empty turn for `forwardEmptyMessageResponse` providers without sending it to storage).

Issue #2459, opened 2026-05-13
Add operator re-queue command for stranded summary jobs


jxt1009:
**Summary**

When the worker or queue path is unhealthy, session summary jobs can be persisted in `observation_generation_jobs` but never make it back onto the summary queue after the system recovers. Today the generic retry flow is aimed at failed/cancelled jobs and treats already-queued jobs as a no-op, while startup reconciliation only republishes on startup. We need an operator command/flow to republish stranded queued `session_summary` jobs without requiring a restart.

**Why this matters**

- Temporary config changes or worker issues can leave summaries unprocessed.
- Once the worker is healthy again, those queued summaries may stay stranded.
- Operators need a safe, auditable recovery path.

**Proposed direction**

- Add a targeted re-queue operator flow for `session_summary` jobs in `queued` state.
- Reuse the persisted payload and deterministic BullMQ job ID.
- Keep Postgres canonical and BullMQ transport-only.
- Reuse the existing audit/logging patterns from retry/cancel.

**Acceptance criteria**

- Operators can republish stranded queued summary jobs without a restart.
- The first implementation supports **bulk** requeue for scoped queued summary jobs.
- The flow is idempotent and scoped.
- Existing retry behavior for failed/cancelled jobs remains unchanged.
- Tests cover the new behavior and operator docs explain when to use it.

## Notes

- The per-ID retry path already exists for failed/cancelled jobs; the real gap is **bulk queued summary republish**.
- The main product choice still open is the initial operator surface: CLI, HTTP, or both.

PR #3583, opened 2026-08-14
fix(server-beta): re-pin payload.generation_job_id on idempotent re-create

alessandropcostabr:
## Problem

A generation job whose `payload.generation_job_id` points at a row that does not exist is dropped **silently**.

`ProviderObservationGenerator` resolves the canonical row through the payload:

```ts
const fresh = await this.lockOutbox(payload.generation_job_id, payload.team_id, payload.project_id);
if (!fresh) {
  logger.info('SYSTEM', 'job no longer exists or is in terminal status; nothing to do', ...);
  return { jobId: payload.generation_job_id, status: 'completed', observationCount: 0 };
}
```

When the id is stale, `lockOutbox` returns `null` and the worker reports **`completed` to BullMQ without touching Postgres**. The queue drains clean, the row stays `queued` forever, and nothing is ever recorded as `failed`. The only trace is a `logger.info` line, which is below the default log level.

## Why the pointer goes stale

Every re-create mints a fresh id for `$1` and carries that same id in the payload it hands over. On `ON CONFLICT (idempotency_key)` the existing row wins, so the minted id is discarded for the `id` column — but the payload merge still lets it overwrite `generation_job_id`:

```sql
ON CONFLICT (idempotency_key) DO UPDATE SET
  payload = observation_generation_jobs.payload || excluded.payload,
```

The row now points at an id that was never inserted.

## Why only `session_summary` is affected

`buildObservationGenerationJobIdempotencyKey` derives the key from `sourceId`:

| source type | `sourceId` is | collides? |
|---|---|---|
| `agent_event` | the individual event id | never |
| `session_summary` | the **session** id | on every re-run of that session's summary |

Summaries are re-enqueued for the same session, so they hit `ON CONFLICT` routinely. Event jobs mint a new key each time and never conflict.

## Fix

Re-pin `generation_job_id` to the surviving row inside the same merge, so the payload can never outlive the id it names.

## Evidence from a production deployment

Running server-beta since May:

- **190** summary jobs stranded in `queued` across four months
- **190/190** had `payload->>'generation_job_id'` absent from the table
- event jobs on the same instance: **0** divergences in ~221k rows
- after re-pointing the payload of those rows, they started draining immediately — the same jobs that had been inert for months

## Test

Added a regression test to `tests/storage/postgres/postgres-storage.test.ts`. Verified it fails without the fix and passes with it:

```
# without the patch
Expected: "9f2e1624-5970-4b83-b82f-14aa5c88afba"
Received: "id-discarded-by-on-conflict"
(fail) re-pins payload.generation_job_id to the surviving row on idempotent re-create

# with the patch
23 pass, 0 fail   (full file)
```

## Relation to #2459 / discussion #2460

#2459 asked for an operator re-queue command for stranded summary jobs, with the guidance to *"reuse the persisted payload and deterministic BullMQ job ID"*.

Worth flagging: replaying the persisted payload replays the stale pointer, so those rows are dropped again, just as silently. I re-enqueued the 190 rows that way first and they evaporated exactly as before. An operator tool built on that guidance would appear not to work. This patch fixes the data going forward; existing stranded rows also need their payload pointer repaired once.

### DIFF
--- a/src/storage/postgres/generation-jobs.ts
+++ b/src/storage/postgres/generation-jobs.ts
@@ -125,7 +125,16 @@ export class PostgresObservationGenerationJobRepository {
         )
         VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13::jsonb)
         ON CONFLICT (idempotency_key) DO UPDATE SET
-          payload = observation_generation_jobs.payload || excluded.payload,
+          -- The conflicting INSERT carries a freshly generated job id in both $1
+          -- and payload.generation_job_id. ON CONFLICT keeps the existing row, so
+          -- $1 is discarded -- but the merge below would otherwise let the new
+          -- payload overwrite generation_job_id with that discarded id, leaving
+          -- the row pointing at a job that never existed. The worker resolves the
+          -- job via payload.generation_job_id, so such a row can never be locked
+          -- and is silently dropped. Re-pin the pointer to the surviving row.
+          payload = observation_generation_jobs.payload
+                    || excluded.payload
+                    || jsonb_build_object('generation_job_id', observation_generation_jobs.id),
           updated_at = now()
         RETURNING *
       `,
--- a/tests/storage/postgres/postgres-storage.test.ts
+++ b/tests/storage/postgres/postgres-storage.test.ts
@@ -504,6 +504,39 @@ describe('server beta postgres observation storage', () => {
     })).resolves.toBeNull();
   });
 
+  it('re-pins payload.generation_job_id to the surviving row on idempotent re-create', async () => {
+    const { project, event } = await createFixtureScopeWithEventJob(storage);
+
+    const first = await storage.observationGenerationJobs.create({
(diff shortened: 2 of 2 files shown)


thedotmack:

> step 2 (job upsert must not point `payload.generation_job_id` at a phantom row)

alessandropcostabr:

```
```

thedotmack:
**All-night ship gate: PASS**

1. RCA logged — `ON CONFLICT (idempotency_key)` keeps the surviving row but the payload merge overwrites `generation_job_id` with a discarded id; worker `lockOutbox` then reports completed and the row stays `queued` forever.
2. No second-system — re-pins `generation_job_id` to the surviving row in the same SQL merge.
3. Narrow / not opinionated — one `jsonb_build_object` pin + one regression test.

Rebased onto `7335467f` locally (fork branch update blocked). Focused Postgres suite is environment-gated here; CI on this head was green (including server-runtime e2e). Merging independently; no version bump.

PR #3584, opened 2026-08-14
fix(server-beta): bound session-summary input by payload size, not event count

alessandropcostabr:
## Problem

`listUnprocessedEvents` caps the event **count** (default 500), not the payload **volume**. Event payloads vary by orders of magnitude, so that cap does not bound what actually reaches the provider. Long sessions overflow the context window, the job is classified `unrecoverable`, and the session ends up with **no summary at all**.

## Measurements

From a production deployment running server-beta since May, over the sessions that failed with `Anthropic context overflow`:

| session | unprocessed events | payload | ≈ tokens |
|---|---:|---:|---:|
| A | 3,099 | **34 MB** | ~9,000k |
| B | 2,412 | 19 MB | ~5,011k |
| C | 2,411 | 15 MB | ~4,060k |

And the part that matters: **even truncated to the existing 500-event default**, those sessions still averaged **~634k tokens**, worst case **~1.55M** — against a ~200k window.

So the existing limit is not merely too generous; it is the wrong dimension. No context window currently on offer absorbs 9M tokens, which means raising the model tier does not fix this either.

## Fix

Bound the summary input by size:

- `CLAUDE_MEM_SUMMARY_INPUT_BUDGET_BYTES`, default `600_000` (≈150k tokens, leaving room for prompt + output)
- Keep the **head and the tail** of the session rather than an arbitrary prefix: the opening carries the goal, the close carries the outcome and what was left pending
- Sessions that already fit are untouched — same objects, same order

## Trade-off, stated plainly

For very long sessions this produces a **partial** summary: the middle of the session is dropped. That is a real loss of fidelity.

It is still strictly better than the current behaviour *for those same sessions*, which is an unrecoverable failure and no summary whatsoever. Sessions within budget see no change at all.

A fuller answer would be hierarchical summarisation (chunk → summarise → summarise the summaries). That is a much larger change and a different PR; this one stops the data loss.

## Tests

`tests/server/generation/summary-input-budget.test.ts` — 4 cases: passthrough under budget, bounding over budget, both ends preserved, chronological order preserved.

```
4 pass, 0 fail
```

`tsc --noEmit -p tsconfig.json` clean for the touched files.

### DIFF
--- a/src/server/generation/ProviderObservationGenerator.ts
+++ b/src/server/generation/ProviderObservationGenerator.ts
@@ -58,6 +58,57 @@ export interface ProviderObservationGeneratorOptions {
   workerId?: string;
 }
 
+
+// The `limit` on listUnprocessedEvents caps the event COUNT, not the payload
+// volume, and event size varies by orders of magnitude. Long sessions therefore
+// still blow the provider context window: measured on a production deployment,
+// sessions that failed with "context overflow" carried up to 34 MB of event
+// payload (~9M tokens), and even truncated to the 500-event default they still
+// averaged ~634k tokens with a 1.55M worst case — against a ~200k window. No
+// model currently on offer absorbs that, so the input has to be bounded by size.
+//
+// Keep the head and the tail of the session: the opening carries the goal, the
+// close carries the outcome and what was left pending. Dropping the middle
+// yields a partial summary, which is strictly better than the current outcome
+// for these sessions — an unrecoverable failure and no summary at all.
+const SUMMARY_INPUT_BUDGET_BYTES = Number.parseInt(
+  process.env.CLAUDE_MEM_SUMMARY_INPUT_BUDGET_BYTES ?? '',
+  10,
+) || 600_000;
+
+export function capSummaryInput<T>(events: T[]): T[] {
+  const size = (e: T): number => {
+    try {
+      return JSON.stringify(e)?.length ?? 0;
+    } catch {
+      return 0;
+    }
+  };
+  let total = 0;
+  for (const e of events) total += size(e);
+  if (total <= SUMMARY_INPUT_BUDGET_BYTES) return events;
+
+  const half = SUMMARY_INPUT_BUDGET_BYTES / 2;
+  const head: T[] = [];
+  let headBytes = 0;
+  for (const e of events) {
+    const s = size(e);
+    if (headBytes + s > half) break;
+    head.push(e);
+    headBytes += s;
+  }
+  const tail: T[] = [];
+  let tailBytes = 0;
+  for (let i = events.length - 1; i >= head.length; i -= 1) {
+    const e = events[i] as T;
+    const s = size(e);
+    if (tailBytes + s > SUMMARY_INPUT_BUDGET_BYTES - headBytes) break;
+    tail.unshift(e);
+    tailBytes += s;
+  }
+  return [...head, ...tail];
+}
+
 export class ProviderObservationGenerator {
   constructor(private readonly options: ProviderObservationGeneratorOptions) {}
 
@@ -518,7 +569,7 @@ export class ProviderObservationGenerator {
         projectId: job.projectId,
         teamId: job.teamId,
       });
-      return events;
+      return capSummaryInput(events);
     }
 
     if (job.sourceType !== 'agent_event') {
--- a/src/server/generation/ProviderObservationGenerator.ts
+++ b/src/server/generation/ProviderObservationGenerator.ts
@@ -73,6 +73,13 @@ export interface ProviderObservationGeneratorOptions {
 // close carries the outcome and what was left pending. Dropping the middle
 // yields a partial summary, which is strictly better than the current outcome
 // for these sessions — an unrecoverable failure and no summary at all.
+// Bounds the SESSION INPUT, not the finished prompt: buildServerGenerationPrompt
+// wraps the selected event blocks in request/project/job tags, the instructions and
+// the observation schema. That envelope is a fixed cost — 1,244 bytes with the mode
+// active here, independent of event count and payload size — so it is left out of
+// the budget rather than reserved from it. At the default that is 0.2%; only a
+// budget small enough to be unusable anyway (~1 KB buys one truncated tool call)
+// would be decided by it.
 const SUMMARY_INPUT_BUDGET_BYTES = Number.parseInt(
   process.env.CLAUDE_MEM_SUMMARY_INPUT_BUDGET_BYTES ?? '',
   10,
--- a/src/server/generation/ProviderObservationGenerator.ts
+++ b/src/server/generation/ProviderObservationGenerator.ts
@@ -76,14 +76,35 @@ const SUMMARY_INPUT_BUDGET_BYTES = Number.parseInt(
   10,
 ) || 600_000;
 
+// Mirrors MAX_PAYLOAD_CHARS in providers/shared/prompt-builder.ts: each event's
+// payload is truncated there before it reaches the model.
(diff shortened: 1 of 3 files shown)


alessandropcostabr:
Both measurement findings were right — fixed in e90dd7f4.

The budget now measures the footprint the event will actually contribute: payload truncated to the same 16 KiB cap `buildEventBlock` applies, sized with `Buffer.byteLength` in UTF-8. Two tests pin it: an oversized single event is kept instead of returning nothing, and a UTF-8-heavy batch stays within budget.

On regenerating the shipped bundle — I've left that out deliberately, and I think it's the right call for this repo:

- `plugin/scripts/*.cjs` appears to be regenerated by the maintainer at release time; the recent commits touching it are all `chore: bump version to 13.x`
- my previous merged PR in this area (#2634) shipped source only, no bundle, and was merged that way
- a 2 MB regenerated bundle in a feature PR would conflict with every other open PR touching the same runtime

The finding itself is correct — a Docker deployment built from the committed `plugin/` does keep the old behaviour until the bundle is rebuilt. Happy to add it if you'd rather have PRs self-contained; just say the word and I'll push it.

alessandropcostabr:
Confirmed, and the root cause is narrower than "the budget is wrong": `promptFootprint` re-derives what `buildEventBlock` does instead of calling it, so the two drift apart in four places.

Against `prompt-builder.ts`:

| # | `promptFootprint` (`ProviderObservationGenerator.ts:93-102`) | `buildEventBlock` |
|---|---|---|
| 1 | `JSON.stringify(payload ?? {})` — compact | `JSON.stringify(event.payload ?? {}, null, 2)` — pretty-printed (`:106`) |
| 2 | no escaping | `escapeXml(truncatedPayload)` (`:126`) |
| 3 | payload bytes only | wrapped in `<agent_event>` + 4 metadata tags + indentation (`:120-131`) |
| 4 | no `stripTags` | `stripTags(rawPayload)` before truncation (`:108`), and the block is dropped entirely when the stripped payload is blank (`:113-115`) |

Note that 1-3 make the estimate too low and 4 can make it too high: an event whose payload strips to nothing still consumes budget here while contributing zero bytes to the prompt.

**Measured on the deployment this came from.** I replayed both paths over the 2,000 most recent `agent_events` rows (`stripTags` not applied, so the rendered side is a floor, never inflated):

| | bytes |
|---|---:|
| sum of `promptFootprint` | 5,005,161 |
| sum of rendered blocks | 6,390,262 |

- aggregate: **1.28x**
- per event: p50 **1.37x**, p95 **1.91x**, worst **4.76x**

One nuance against your reproduction: on real traffic the gap is a factor, not an order of magnitude. At the 600,000-byte default an accepted set renders to roughly 766,000 bytes, which a 200k-token window still absorbs. That does not make it safe — the relative error is budget-independent, so it bites exactly when someone lowers `CLAUDE_MEM_SUMMARY_INPUT_BUDGET_BYTES` to fit a smaller model, which is the one case the knob exists for.

**Direction for the fix:** stop estimating. `PROMPT_PAYLOAD_CAP_CHARS` is already a hand-copy of `MAX_PAYLOAD_CHARS` — the code comment says "Mirrors" out loud — and that duplication is what produced this bug. Exporting the block construction and measuring `Buffer.byteLength(block.body, 'utf8')` closes all four rows above at once and cannot drift again, at the cost of building each candidate block twice.

Thanks for running it rather than asserting it; the artifacts made this fast to check.

alessandropcostabr:
Fixed in `804a7183`.

The estimate is gone rather than corrected. `prompt-builder.ts` now exports `eventBlockBytes`, which builds the block and weighs it, and `promptFootprint` is one line delegating to it. `PROMPT_PAYLOAD_CAP_CHARS` — the hand-copy of `MAX_PAYLOAD_CHARS` that produced the drift — is deleted. All four divergences from the table above close at once, and they cannot reopen: the number is now the text, not a model of it.

Two decisions worth flagging, since neither is free:

**The `try/catch` returning 0 is gone, not ported.** An event that cannot be turned into a block here cannot be turned into one during prompt construction a few milliseconds later, so swallowing the error never saved the job — it only replaced a loud failure with "size 0, everything fits", which is the context overflow this budget exists to prevent.

**Measuring now depends on the whole event, not just its payload.** The tests had to move from `{id, payload}` stubs to real `PostgresAgentEvent` rows. That is a genuine cost of the reuse, and the honest trade for it is that the test can no longer pass against a prompt shape the builder does not produce.

Evidence, rather than assertion — I reverted only the production file to `e90dd7f4` and re-ran the same suite:

| | result |
|---|---|
| against `e90dd7f4` | **2 fail**, 6 pass |
| against `804a7183` | **8 pass**, 0 fail |

The two that fail on the old code are `bounds an oversized batch to the byte budget` and the new `counts XML escaping and block overhead`, which builds 100 events of 3 KB of `<`: ~300 KB raw, so the old filter passed them untouched, rendering past 1 MB once every `<` becomes `&lt;`.

Wider suites on the new code: `tests/server/generation/` 36 pass / 3 skip / 0 fail, `tests/server/` 189 pass / 13 skip / 0 fail, `tsc --noEmit` clean on both touched files.

alessandropcostabr:
Correct, and reproduced: I built the prompt with zero events and the envelope is **1,244 UTF-8 bytes** here (request wrapper, project/session/job tags, instructions, observation schema).

I'd argue it should stay out of this PR anyway, and the reason is the shape of the error rather than its existence.

The defect you found earlier was **proportional**: the estimate under-counted by a factor that grew with the content, 1.28x aggregate and 4.76x worst case across 2,000 real events, so no budget value was safe from it. This one is a **constant offset**. The envelope does not scale with event count or payload size, so at the 600,000-byte default it is 0.207% of the budget, and no realistic session gets pushed past a provider window by 1.2 KB of instructions.

Where your reproduction is right and mine agrees with it: at a 1,000-byte budget the envelope alone exceeds the cap before a single event is selected. But that configuration is already meaningless for a different reason — 1,000 bytes buys roughly one truncated tool call, so the summary would be worthless even if the accounting were exact. The knob exists to fit smaller context windows, and the smallest models on offer still take tens of thousands of tokens.

So the honest statement is that `CLAUDE_MEM_SUMMARY_INPUT_BUDGET_BYTES` bounds **the session input**, not the finished prompt, and callers should leave the envelope out of their reckoning. If you'd like that made explicit rather than left implicit, the cheap version is one line on the constant saying the budget covers event blocks only and the envelope adds a fixed ~1.2 KB on top. Reserving it in the selection loop is also a two-line change, but it buys accuracy that only matters at budgets where the feature is already unusable, and every extra thing this function knows about the prompt is another thing that can drift out of sync with it — which is how we got here.

Happy to add either if you disagree with the reasoning; I'd rather not expand the PR on my own judgement.

alessandropcostabr:
Documented in `c621b7a7` — the constant now states that it bounds the session input, that the envelope is a fixed 1,244 bytes on top, and why that is left out rather than reserved.

Not enforced in the selection loop, for the reason in the comment: every extra thing this function knows about the prompt shape is another thing that can drift out of sync with the builder, and that drift is the bug this PR set out to fix. Say the word if you'd rather have the two-line reservation instead and I'll swap it.

thedotmack:

> step 1 (summary input bounded by rendered bytes measured in the prompt builder)

alessandropcostabr:

```
```

thedotmack:

`MERGEABLE` / `CLEAN`, but this is a design change (new size budget + head/tail trim), not a narrow RCA patch. Leaving open.

thedotmack:
Superseded by same-repo rehost #4030, merged as fb0b79c3. Thank you @alessandropcostabr.

**RCA:** Event-count cap did not bound payload volume, so long sessions overflowed and got no summary.

PR #3586, opened 2026-08-14
fix(server-beta): accept contentSessionId on /v1/memories, mirroring /v1/events

alessandropcostabr:
## Problem

#2634 taught the ingest path to resolve `contentSessionId` into a `server_sessions` row, because callers writing from inside a session know the agent's own session id, not the `server_sessions` row UUID.

The memory write path never got the same affordance. `/v1/memories` accepts only `serverSessionId`:

```ts
z.object({
  projectId: z.string().min(1),
  serverSessionId: z.string().min(1).nullable().optional(),   // ← no contentSessionId
  ...
})
```

So any client writing a memory from inside a session lands with `server_session_id NULL`, and those rows can never be grouped, scoped or compared by session afterwards.

## Measurement

On a production deployment running server-beta since May:

| write path | rows | linked to a session |
|---|---:|---:|
| generation path (`/v1/events` → generator) | 187,970 | **85.7%** |
| direct memory writes (`/v1/memories`) | 2,948 | **0.0%** |

Not a sampling artifact — it is 2,948 out of 2,948. The information exists on the caller side (we carry it in `metadata` as a workaround), it simply has nowhere to go in the request.

## Fix

Accept `contentSessionId` and resolve it through the same `findIdByContentSessionId` used by the ingest path.

- an explicit `serverSessionId` still wins, and short-circuits without querying
- the lookup is best-effort: a failure logs and stores the memory unlinked, exactly as today, so no write can begin failing because of this change
- behaviour for callers that do not send `contentSessionId` is unchanged

## Why the #2634 idempotency concern does not apply here

Review of #2634 flagged idempotency-key drift, since the ingest path derives its key partly from session linkage. That does not carry over: `observations` keys off `id`, and `server_session_id` is a plain FK with `ON DELETE SET NULL`. Resolving the link later cannot drift a key.

## Tests

Four cases added to the #2634 linkage suite (`tests/server/runtime/server-session-linkage.test.ts`), same mocked-pool style:

- explicit `serverSessionId` wins **and issues no query**
- `contentSessionId` resolves when `serverSessionId` is absent
- neither id supplied → stores unlinked, no query
- lookup throws → returns null, write proceeds

```
6 pass, 0 fail   (2 pre-existing + 4 new)
```

`tsc --noEmit -p tsconfig.json` clean for the touched file.

### DIFF
--- a/src/server/routes/v1/ServerV1PostgresRoutes.ts
+++ b/src/server/routes/v1/ServerV1PostgresRoutes.ts
@@ -874,6 +874,12 @@ export class ServerV1PostgresRoutes implements RouteHandler {
       z.object({
         projectId: z.string().min(1),
         serverSessionId: z.string().min(1).nullable().optional(),
+        // Callers that write memories from inside a session know the agent's own
+        // session id, not the server_sessions row UUID. /v1/events already accepts
+        // contentSessionId and resolves it (#2634); without the same affordance here
+        // every such memory lands with server_session_id NULL and can never be
+        // grouped, scoped or compared by session.
+        contentSessionId: z.string().min(1).nullable().optional(),
         kind: z.string().min(1).optional(),
         content: z.string().min(1),
         metadata: z.record(z.string(), z.unknown()).optional(),
@@ -882,10 +888,16 @@ export class ServerV1PostgresRoutes implements RouteHandler {
         const teamId = this.requireTeamId(req, res);
         if (!teamId) return;
         if (!this.ensureProjectAllowed(req, res, body.projectId)) return;
+        const linkedSessionId = await this.resolveMemorySessionLink({
+          serverSessionId: body.serverSessionId ?? null,
+          contentSessionId: body.contentSessionId ?? null,
+          projectId: body.projectId,
+          teamId,
+        });
         const createInput = {
           projectId: body.projectId,
           teamId,
-          serverSessionId: body.serverSessionId ?? null,
+          serverSessionId: linkedSessionId,
           kind: body.kind ?? 'manual',
           content: body.content,
           metadata: body.metadata ?? {},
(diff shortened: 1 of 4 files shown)


alessandropcostabr:
Both findings were right — fixed in 7e05fc4c.

**Platform scope**: `resolveMemorySessionLink` now takes the scope through the same `sessionLookupPlatformScope` helper the ingest path uses, so an omitted `platformSource` and an explicit `null` keep meaning different things. Two tests pin it — the platform reaches the query, and the two cases produce different lookups.

Thanks for the reproduction artifacts on that one; the Cursor-memory-linked-to-Codex-session capture made the failure mode unambiguous.

**Client threading**: also correct, and it made the fix inert in practice — the route accepted `contentSessionId` but `ServerAddObservationRequest`, `buildAddObservationPayload` and the MCP `observation_add` args exposed only `serverSessionId`, so no supported caller could reach it. All three now carry `contentSessionId` and `platformSource`.

thedotmack:

> step 3 (`/v1/memories` resolves `contentSessionId` + platform scope like `/v1/events`)

alessandropcostabr:

```
```

thedotmack:
**All-night ship gate: PASS**

1. RCA logged — `/v1/memories` accepted only `serverSessionId`, so in-session writes landed with `server_session_id NULL` (0% of 2,948 production rows linked). Mirrors #2634 on `/v1/events`.
2. No second-system — same `findIdByContentSessionId` + platform-scope helper the ingest path already uses; lookup is best-effort.
3. Narrow / not opinionated — accept the field, resolve, thread through MCP/client so callers can reach it.

Rebased onto `7335467f` locally (fork branch update blocked). Focused tests: `bun test tests/server/runtime/server-session-linkage.test.ts` — 8 pass. CI on this head was green. Merging independently; no version bump.

PR #2058, opened 2026-04-18
fix(summarizer): salvage summary from observation tags, support Gemini transcripts, use user_prompt

thedotmack:
## Summary
- Fix #1908: Add salvage fallback in parseSummary() when LLM emits `<observation>` instead of `<summary>` tags
- Fix #1909: Add Gemini transcript format detection and role mapping in transcript-parser
- Fix #1910: Include user_prompt in summary prompt and add explicit tag guardrails

## Test plan
- [ ] Verify parseSummary handles `<observation>` tags by extracting summary data
- [ ] Verify parseSummary handles raw text output (no tags)
- [ ] Verify Gemini JSONL transcripts parse correctly
- [ ] Verify summary prompt includes user_prompt context
- [ ] Verify summary prompt explicitly instructs model to use only `<summary>` tags

Closes #1908, closes #1909, closes #1910

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1823, opened 2026-04-15
Summary mode receives observation XML — prompt conditioning leaks between modes


SalvoItaly:
## Severity
Low (cosmetic warning). Observations are still saved correctly, but the log gets polluted with a WARN every turn.

## Symptom
The worker emits, on every processed tool-use:

```
WARN [PARSER] "Summary response contained observation tags instead of summary — prompt conditioning may need strengthening"
```

## Likely cause
Looking at `scripts/worker-service.cjs`:

- `xu(t, e)` builds the summary prompt with the explicit instruction:
  > `Do NOT output <observation> tags. This is a summary request, not an observation request.`
  > `Your response MUST use <summary> tags ONLY. Any <observation> output will be discarded.`
- `wu(t, e, r, n)` builds the observer prompt with `<observation>` XML template.

Despite the explicit negative instructions, the model (Sonnet 4.6 in our case) emits `<observation>` in summary mode. Likely causes: strong conditioning carried over from prior observer turns, or positive template > negative.

## Suggested upstream fixes
1. **Parser tolerance**: accept either `<summary>` or `<observation>` in summary responses, optionally converting the latter into a derived summary, without warning.
2. **Prompt strengthening**: switch from negative instruction to a few-shot example of a correct summary + a failure case.
3. **SDK session separation**: use distinct SDK agents for observer vs summary to avoid conditioning crossover.

## Environment
- OS: Windows 11 Pro 10.0.22621
- Plugin: claude-mem 12.1.0
- Provider: claude (Sonnet 4.6 via CLI)

thedotmack:
Closed during the April 2026 backlog cleanup. The underlying bug is now tracked in #1908, which is the single canonical issue being addressed by the maintainer. Thanks for taking the time to report — your symptoms and repro are captured in the consolidated ticket.

PR #3587, opened 2026-08-14
fix(server-beta): stop discarding session-summary responses (recurrence of #1345)

alessandropcostabr:
## This is #1345 reappearing in server-beta

#1345 (merged 13 Mar) described exactly this:

> the LLM responds with `<observation>` XML tags instead of the expected `<summary>` tags. `parseSummary()` finds no `<summary>` block, returns null, and the summary is silently lost. The stop hook reports "Summary completed" (false positive) because the pending queue is empty.

That fix landed in the SDK path (`src/sdk/prompts.ts`, `src/sdk/parser.ts`). server-beta arrived on 8 May with its own prompt builder and its own summary persistence path, and neither inherited the lesson:

- `buildServerGenerationPrompt` is shared by both job types and asks for `<observation>` blocks — including for summary jobs
- `parseAgentXml` sees an `<observation>` root and returns `{ observations: [...], summary: null }`
- `processSessionSummaryResponse` reads only `parsed.summary`, computes an empty body, and completes the job without looking at `parsed.observations`

Same cause, same silent loss, new runtime. The earlier reports of the legacy symptom are #1649, #1823 and #1546; the `WARN [PARSER]` that made it noticeable there was later removed by #2074, so the server-beta recurrence arrived without any signal at all.

## Measured impact

On a deployment running server-beta since May:

| | |
|---|---:|
| `session_summary` jobs | **4,898** |
| completed | 4,836 |
| observations written by any of them | **0** |

No observation in the database has `created_by_job_id` pointing at a summary job, in three months. I confirmed the mechanism by replaying a real call with the current prompt: the model returned a well-formed `<observation>` block with useful content, and it was discarded.

## Changes

**1. Ask summary jobs for what the summary path stores.** The prompt now branches on `job.sourceType` and requests a `<summary>` block with the sub-tags `parseSummaryBlock` already understands. This is the same shape of fix as #1345 — condition the prompt for the task — applied to the runtime that was written after it.

All the summary machinery already exists (`parseSummaryBlock`, `renderSummaryContent`, the `ParsedSummary` fields); it simply was never reachable, because the prompt asked for something else.

**2. Keep `<observation>` answers instead of dropping them.** Defensive: if a provider answers with observation blocks anyway, fold them into the summary body rather than discarding a response that was already billed.

Note that #1604 and #2058 proposed salvage-only fixes for the legacy path and were not merged, while the prompt-conditioning fix (#1345) was. If you'd rather keep this PR to the prompt change alone, say so and I'll drop the second commit — the fallback is there to protect existing deployments during the window before a prompt change reaches them, not to replace the conditioning fix.

## Known limitation, not addressed here

This does **not** make summaries complete. `loadEvents` feeds summary jobs from `listUnprocessedEvents`, which excludes every event that already has a completed per-event generation job. On the same deployment, counting what a summary prompt would actually receive:

| events visible to the summary prompt | sessions |
|---|---:|
| **zero** | **3,399** |
| 1–5 | 1,203 |
| 6–50 | 102 |
| 50+ | 194 |

So for ~69% of sessions the prompt arrives with an empty `<agent_events>` block. With this PR those sessions will emit `<skip_summary />` and correctly persist nothing — which is the right outcome, but it means the fix only recovers sessions that still carry unprocessed events.

Making summaries actually describe a session needs a different input (the session's full event set, or the observations already generated from it). That's a design change and I'd rather not smuggle it in here; happy to open it separately if useful.

## Tests

- 4 cases pinning the prompt per source type (summary asks for `<summary>`, event still asks for `<observation>`, both keep the `<skip_summary />` escape hatch)
- 1 integration test: a summary job answered with `<observation>` now persists — verified failing before the change (0 observations) and passing after

**On the suite:** `tests/server/generation/process-generated-response.test.ts` is flaky when the whole file runs against one Postgres — on a clean `upstream/main` checkout it already reports 5 failures out of 7 in my environment, and every case passes in isolation. Looks related to the shared pool + per-test schema `search_path` workaround noted in the setup comment; untouched here.

### DIFF
--- a/src/server/generation/processGeneratedResponse.ts
+++ b/src/server/generation/processGeneratedResponse.ts
@@ -477,6 +477,15 @@ function renderObservationContent(observation: ParsedObservation): string {
   if (observation.facts && observation.facts.length > 0) {
     parts.push(observation.facts.map(f => `- ${f}`).join('\n'));
   }
+  // The parser treats concepts as content: an observation whose only populated
+  // field is <concepts> survives its empty-observation guard. This renderer did
+  // not, so that observation rendered to '' and was dropped by the empty-content
+  // skip in persistGeneratedObservations - accepted, discarded, job completed.
+  // Rendered last-resort only, so responses that carry prose keep the body they
+  // already have and no existing content changes shape.
+  if (parts.length === 0 && observation.concepts && observation.concepts.length > 0) {
+    parts.push(`Concepts: ${observation.concepts.join(', ')}`);
+  }
   return parts.join('\n\n').trim();
 }
 
--- a/src/server/generation/processGeneratedResponse.ts
+++ b/src/server/generation/processGeneratedResponse.ts
@@ -196,7 +196,17 @@ export async function processSessionSummaryResponse(
 
   const summary = parsed.summary ?? null;
   const skipped = summary?.skipped === true;
-  const summaryContent = summary ? renderSummaryContent(summary) : '';
+
+  // Defence in depth: a provider that answers with <observation> blocks instead of
+  // a <summary> block used to be dropped here silently — parsed.summary was null,
+  // the content was judged empty, the job completed and parsed.observations was
+  // never looked at. Fold those blocks into the summary body rather than
+  // discarding a response the model was billed for.
+  const fallbackObservations = !summary && !skipped ? parsed.observations : [];
+  const summaryContent = summary
+    ? renderSummaryContent(summary)
+    : fallbackObservations.map(o => renderObservationContent(o)).join('\n\n');
+
   const privateContentDetected = skipped || summaryContent.trim().length === 0;
 
   const rendered: RenderedObservation[] = privateContentDetected
--- a/src/server/generation/providers/shared/prompt-builder.ts
+++ b/src/server/generation/providers/shared/prompt-builder.ts
@@ -72,6 +72,42 @@ export function buildServerGenerationPrompt(
 
   const observationOutputSchema = buildObservationOutputSchema(mode);
 
+  // Summary jobs are a different task from event jobs, and the persistence path
+  // for them (processGeneratedResponse -> parsed.summary) only understands a
+  // <summary> block. Asking for <observation> here yields a response the summary
+  // path silently discards, so branch the instruction on the job's source type.
+  const isSessionSummary = context.job.sourceType === 'session_summary';
+
+  const summaryInstruction = [
+    'You are reviewing a complete agent session. Return a single',
+    '<summary>...</summary> XML block describing the arc of the session. If the',
+    'session contains nothing worth recording (e.g., everything was scrubbed by',
+    'privacy filters or the activity was trivial), return a single self-closing',
+    '<skip_summary /> tag and nothing else. Do not include any prose outside the XML.',
+    '',
+    'Schema for the <summary> block (at least one of the first five is required):',
+    '<summary>',
+    '  <request>what the user asked for</request>',
+    '  <investigated>what was explored, and how</investigated>',
+    '  <learned>durable findings worth remembering later</learned>',
+    '  <completed>what was actually done</completed>',
+    '  <next_steps>what is left open</next_steps>',
+    '  <notes>anything else worth keeping</notes>',
+    '</summary>',
+  ];
+
+  const observationInstruction = [
+    'You are observing an agent at work. Return one or more',
+    '<observation>...</observation> XML blocks summarizing durable, useful',
+    'discoveries from the events above. If the events contain nothing worth',
+    'recording (e.g., everything was scrubbed by privacy filters or the',
+    'activity was trivial), return a single self-closing <skip_summary />',
+    'tag and nothing else. Do not include any prose outside the XML.',
+    '',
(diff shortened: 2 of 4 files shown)


alessandropcostabr:
This finding is correct, and it invalidates the premise of the PR. Converting to draft.

I measured it on the deployment this came from — for every session that has a `session_summary` job, how many events `listUnprocessedEvents` would actually return:

| events visible to the summary prompt | sessions |
|---|---:|
| **zero** | **3,399** |
| 1–5 | 1,203 |
| 6–50 | 102 |
| 50+ | 194 |

**69% of sessions would hand the model an empty `<agent_events>` block** while asking it to describe the arc of the session, and another 25% would give it five events or fewer. So this PR would have taken "the response is silently discarded" and turned it into "whatever the model invents gets persisted as the session summary" — which is worse, because it contaminates memory rather than merely wasting a call.

That also reframes the original observation. I opened this because 4,898 summary jobs had produced zero observations and that looked like a persistence bug. Given the above, the far likelier reading is that the discard is the **last line of defence**: the summary path is being fed the unprocessed-event recovery set, not the session, and dropping an ungrounded response is the correct outcome.

What would actually be needed is a different input for summaries — the session's full event set, or the observations already generated from it — rather than the leftovers `listUnprocessedEvents` is designed to return. That's a design change, not a patch, so I'd rather not smuggle it in here.

Leaving this as a draft. The measurement above is the useful part; happy to open a separate issue describing the input problem if that's more useful than a stale draft PR.

alessandropcostabr:
Reopened for review. Updating my own comment above, which says I was leaving this as a draft.

To be clear about what changed: **nothing in the code**. `42c3b640` is still the only commit. What I rewrote is the PR description, because on a second reading the finding kills the *framing* of this PR, not the patch.

The finding stands and I am not arguing with it: `loadEvents` feeds summary jobs from `listUnprocessedEvents`, so for ~69% of sessions on my deployment (3,399 of 4,898) the prompt would arrive with an empty `<agent_events>` block. What I got wrong at 14:18 was the conclusion I drew from it. With this patch those sessions do not fabricate anything — they emit `<skip_summary />` and persist nothing, which is the correct outcome. The fabrication risk you flagged is real for a patch that only stopped the discard; it is not what this one does once the prompt is conditioned per source type.

So the honest scope is now stated in the description as a "Known limitation, not addressed here":

- **What it fixes:** sessions that still carry unprocessed events stop having a well-formed response thrown away.
- **What it does not fix:** summaries do not become complete. Feeding summary jobs the session's full event set, or the observations already generated from it, is a design change and does not belong in this PR.

If you would rather settle the input question first and take this after, say so and I will close it in favour of an issue describing the input problem — the measurement is the part worth keeping either way.

thedotmack:

> step 1 (prompt keyed by job sourceType; summary built from the full session event set, not unprocessed leftovers); P1: concepts-only fallback still dropped

alessandropcostabr:
**Rebased onto v13.15.3, and the P1 that blocked this is fixed.** CI is green — all five checks pass, `MERGEABLE / CLEAN`.

**The red `typecheck · build · test · bundle-size` you'd have seen here was inherited from the base, not introduced by this PR** — same three mode-loading tests that were failing on my other four PRs from that day. The rebase carries it away; nothing in the original commit changed.

**On the blocking finding** — *"P1: concepts-only fallback still dropped"*. It's real, and the root cause sits one layer below the fallback:

`parseObservationBlocks` treats `<concepts>` as content — its empty-observation guard is `!title && !narrative && facts.length === 0 && cleanedConcepts.length === 0`, so a block whose only populated field is concepts survives parsing and arrives at the persist path as a valid observation. `renderObservationContent` never rendered concepts. So that observation rendered to `''`, hit the empty-content `continue` in `persistGeneratedObservations`, and the job completed with nothing written. Accepted, then discarded — the same shape this PR is about.

Because the parser and the renderer disagreed rather than the fallback being incomplete, **this was not only a summary-fallback bug**: the per-event path dropped concepts-only observations too. Fixing it in `renderObservationContent` covers both callers at once, which is why the diff is six lines rather than one branch per pipeline.

Rendered **last resort only** — when no other field produced content:

```ts
if (parts.length === 0 && observation.concepts && observation.concepts.length > 0) {
  parts.push(`Concepts: ${observation.concepts.join(', ')}`);
}
```

The alternative — always appending concepts — would rewrite the body of essentially every observation in the store and change what the embeddings index. Not worth it to fix a drop.

Two regression tests, one per pipeline (`tests/server/generation/process-generated-response.test.ts`). Both verified failing against the renderer as it was and passing after, on a throwaway Postgres 16 — not just added and assumed.

Happy to reshape any of this against the plan-24 design if the sequencing there wants it elsewhere.

thedotmack:

Owner routing is **needs rework** (plan-24). Author also recorded that most summary jobs are fed an empty unprocessed-event set; prompt conditioning alone is not a safe night-ship. Leaving open.

thedotmack:
Superseded by same-repo rehost #4029, merged as 7e4807c6. Thank you @alessandropcostabr.

**RCA:** #1345 recurrence in server-beta — shared prompt asked for `<observation>`, summary path only stored `parsed.summary`.

Issue #3592, opened 2026-08-15
Observation `type` enum is unenforced: parser logs `Invalid observation type` then stores it verbatim; missing `<type>` still defaults to `bugfix`


bxshan:
## Environment

- claude-mem **v13.15.0** (current latest), plugin install, Linux (Ubuntu 26.04)
- Provider `claude` / `claude-haiku-4-5-20251001`, `CLAUDE_MEM_RUNTIME=worker`
- Mode `code--grounded` — a local mode that inherits `code.json` and overrides only `recording_focus` / `summary_instruction` / `summary_footer`, so `observation_types` is the stock 9-entry `code` enum
- Store: ~18.2k observations

## Summary

The active mode's `observation_types` enum is advisory only — it is rendered into the observer's system prompt but never enforced on the way in. Both branches of the type-handling code at the parse site are lossy, in opposite directions.

`scripts/worker-service.cjs` (bundled), parse site:

```js
y = vt.getInstance().getActiveMode().observation_types.map(S => S.id),
v = y[0],
_ = v;
o
  ? (_ = o, y.includes(o) || h.error("PARSER", `Invalid observation type: ${o}, preserving emitted type`, { correlationId: e }))
  : h.error("PARSER", `Observation missing type field, using "${v}"`, { correlationId: e });
```

**Branch A — type present but out of enum:** logged at ERROR, then stored verbatim. `observations.type` is `TEXT NOT NULL` with no CHECK constraint, so the ERROR line describes a value that is nevertheless durably written. The column becomes unbounded free text.

**Branch B — type missing:** falls back to `observation_types[0]`, which in `code` is `bugfix`. This is the mis-tagging reported in #3181; the fix appears to have addressed the out-of-enum path (now "preserving") but left the missing-type default pointing at the first enum entry.

## Impact on this install

- **301** of ~18.2k observations carry a type outside the 9-entry enum, across **183 distinct values**, most appearing exactly once — e.g. `sample-gate`, `candidate-design`, `killer-input-validation`, `p34-selftest-fix-applied`, `phase2-final-integrity-verification`.
- Five days of logs (Aug 10–14) contain **259** `Invalid observation type` lines and **39** `Observation missing type field` lines.
- Those rows fall out of any type-filtered query, and `getTypeIcon()` renders them with the `📝` fallback, which does not appear in the legend that the SessionStart context builds from `observation_types`.
[...]

Issue #3181, opened 2026-07-09
Parser rejects most LLM-emitted observation types → everything mis-tagged as "bugfix"; no retention/prune path for unbounded DB growth


Uncle-Lou-AfterShockFam:
Two issues observed on v13.1.0 (macOS, `code` mode, ~2 months of use):

## 1. Observation-type drift → silent mis-tagging

The summarizer LLM routinely emits `type` values outside the active mode's
`observation_types` enum. The parser falls back to `observation_types[0]` — which in
`code` mode is `bugfix` — so a large share of observations are stored mis-tagged.

One day's log (`claude-mem-2026-07-07.log`) shows **445** `[ERROR] [PARSER] Invalid
observation type` lines. Frequency tally over recent logs:
`code` 63, `milestone` 39, `test` 34, `state` 27, `fix` 21, `what` 20, `architecture` 20,
`bug` 18, `task` 15, `finding` 14, `tool` 10, `skeptic` 10, plus a long tail.

Impact: type-filtered retrieval is degraded — most of the store reads as "bugfix".

Suggestions (either helps):
- Fall back to a neutral type (`change`) instead of `observation_types[0]`, or make the
  fallback configurable.
- Strengthen the summarizer prompt's type constraint, and/or fuzzy-map near-misses
  (`bug`/`fix` → `bugfix`, `architecture` → closest) before falling back.

## 2. No retention/prune path — unbounded growth

`~/.claude-mem/` reached **3.6 GB in ~2 months** (claude-mem.db 1.4 GB + chroma 2.1 GB).
I could find no TTL/retention/VACUUM logic in the bundled scripts, and `worker-cli.js`
exposes only `start|stop|restart|status`. The only shrink on record was the one-time
v12.4.3 migration wipe. The context-injection knobs (`CLAUDE_MEM_CONTEXT_*`) limit
read-back, not storage.

Suggestion: a `worker-cli.js prune --older-than <days>` (SQLite delete + VACUUM + Chroma
collection cleanup), or a documented supported manual procedure.

Happy to provide logs/DB stats. Thanks for the plugin — the multi-session `.in_use`
PID+procStart locking in 13.x has been working well.

PR #3593, opened 2026-08-15
test(parser): pin observation type handling against the mode enum

ntdatt812:
Refs #3592.

## What this does

Pins the current type-handling behaviour at the parse site with tests, and makes the missing-type default legible in the log. No behaviour change — the enum decision in #3592 is yours to make, and this is meant to be the safe step before it.

## Why tests first

`src/sdk/parser.ts` resolves an observation's type against `mode.observation_types`, and `tests/sdk/parser.test.ts` only ever exercises types that are already in the enum. Neither branch of #3592 was covered, so either fix — enforcing the enum, or changing the fallback — would currently land with nothing to catch the blast radius.

Three cases added:

- an out-of-enum type is preserved verbatim
- an observation with no `<type>` falls back to `observation_types[0]`
- that fallback follows enum *order*, not a fixed default — reordering the mode's list changes what an untyped observation becomes

The third is the one worth looking at. The fallback is positional, so in the bundled `code` mode an untyped observation is filed as `bugfix` — the first entry — rather than as anything neutral. That is the mis-tagging in #3181, and it is currently invisible because nothing asserts it.

## The log line

`Observation missing type field, using "bugfix"` reads as though `bugfix` were a deliberate neutral default. It isn't; it is whatever happens to sit first in the active mode's list. Reworded to say so:

```
Observation missing type field, defaulting to the first declared type: "bugfix"
```

Nothing asserts the old string, and the severity and control flow are untouched.

## What I did not change

I left both branches alone. Enforcing the enum is a semantic decision with a data-migration tail — 301 of the reporter's 18.2k observations already carry out-of-enum types across 183 distinct values, so whatever you pick has to say something about the rows already written. Happy to follow up with whichever you want:

- coerce out-of-enum types to a neutral entry, and add one to the mode schema
- keep preserving them but add a `CHECK` constraint or a reconciliation pass
- leave the enum advisory and just fix the missing-type fallback

The tests here should make any of those a small diff.

## Verification

```
bun test tests/sdk/parser.test.ts   24 pass, 0 fail   (21 before)
bun test tests/sdk/                 57 pass, 0 fail
tsc --noEmit                        clean
```


thedotmack:

> step 5 (unknown/missing `type` handling pinned by tests) — test-only

ntdatt812:
Thanks for the routing note — happy to keep this standalone.

One thing that may hold up the "once CI is green" step: **no workflow run has ever been triggered on this PR.** `ci.yml` runs on every `pull_request`, but the API reports 0 check runs and 0 statuses on the head commit here, and the same on #3619 and #3620. Other open PRs from other contributors do show 5–6 check runs, so this looks like the first-time-contributor **"Approve and run workflows"** gate rather than anything failing — the PR sits at `UNSTABLE` simply because nothing has reported. Whenever you get a moment to approve the runs, CI can actually speak.

Meanwhile I rebased this branch onto current `main` (was 7 behind; rebased clean, no conflicts) so the run reflects today's tree, and reproduced the CI job steps locally on that rebased commit:

| CI step | result |
|---|---|
| `npm run typecheck` (`tsc --noEmit` ×2) | exit 0 |
| `bun test ./tests/sdk/parser.test.ts` | **24 pass, 0 fail**, 58 assertions |

Diff is still exactly `src/sdk/parser.ts` + `tests/sdk/parser.test.ts`, no behaviour change — the tests pin today's positional-default behaviour so whichever way you decide to enforce the enum, the change shows up as an intentional test edit rather than a silent drift.

Same status on the other two, both already on current `main`:

- **#3619** — typecheck exit 0, `tests/worker/openai-compatible-history-append.test.ts` 3 pass / 0 fail
- **#3620** — typecheck exit 0, `tests/worker/openrouter-truncation-warning.test.ts` 5 pass / 0 fail

No rush on any of them; just flagging the approval gate since it is invisible from the PR page.

ntdatt812:
Follow-up with the hard evidence for the approval gate I mentioned above — I'd only inferred it before from "0 check runs", which was weak. The runs API states it outright.

The workflow runs **do** get created on these PRs; they just terminate unstarted:

```
$ gh api "repos/thedotmack/claude-mem/actions/runs?head_sha=<head of #3593>"
  name=CI       event=pull_request  status=completed  conclusion=action_required
  name=Windows  event=pull_request  status=completed  conclusion=action_required
```

`conclusion=action_required` is GitHub's code for "waiting on a maintainer to press **Approve and run workflows**". Identical on #3619 and #3620.

It isn't specific to me. Every currently open fork PR from a contributor without a prior merge here sits in the same state:

| PR | author | conclusion |
|---|---|---|
| #3650, #3646 | @anupamme | `action_required` |
| #3633 | @zloglevel | `action_required` |
| #3627 | @FaintFlower | `action_required` |
| #3623 | @akashpandey | `action_required` |
| #3621 | @Bleed00 | `action_required` |
| #3616 | @maksodf | `action_required` |
| #3597 | @SYMBaiEX | `action_required` |
| #3593, #3619, #3620 | @ntdatt812 | `action_required` |

Whereas fork PRs from contributors who already have a merge here run normally — @ChenglinWei97's #3624/#3625/#3626 are `success`, @alessandropcostabr's #3587 is `failure,skipped,success`. That contrast is what makes the first-time-contributor setting the explanation rather than anything wrong with `ci.yml`.

So for this PR, "once CI is green" needs one click on the Actions tab (or the yellow banner on the PR) before it can mean anything. No rush at all — I'm flagging it because the state is completely invisible from the contributor side, and eight PRs are currently parked behind it.

Local results on the current rebased head, unchanged from above: `npm run typecheck` exit 0; `bun test ./tests/sdk/parser.test.ts` 24 pass / 0 fail. #3619 and #3620 likewise green locally.

If you'd rather not approve run-by-run, Settings → Actions → General → "Require approval for first-time contributors" can be relaxed to "Require approval for first-time contributors who are new to GitHub" — but that's a repo-level call, entirely your preference.

thedotmack:

`MERGEABLE` / `UNSTABLE`. Test-only pin plus a log-string tweak; no user-facing bug-fix. CI has never run (`action_required` first-time-contributor gate). Leaving open; not merging in this batch.

thedotmack:

thedotmack:

thedotmack:
Superseded by same-repo rehost #4034, merged as 0d9313ce. Thank you @ntdatt812.

Test-only pin of observation type handling. No behavior change.

PR #3597, opened 2026-08-16
fix(worker): reuse OpenAI-compatible synthetic session IDs

SYMBaiEX:
## Summary

- reuse a persisted synthetic memory-session ID when an OpenRouter or Gemini generator restarts for the same content session
- mint and persist a new synthetic ID only when the stored ID belongs to a different provider/session
- make `updateMemorySessionId` a no-op when the value is unchanged, preventing redundant prompt repair operations from any caller

## Root cause

`SessionManager` intentionally initializes reloaded in-memory sessions without a memory-session ID so Claude SDK sessions are not resumed with stale remote IDs. OpenAI-compatible providers use local synthetic IDs instead, but they treated every reload as a new identity. Each new timestamped ID called `updateMemorySessionId`, which requeued every native prompt in the conversation as a `set_prompt_session` mutation.

That makes the sync outbox grow with `generator restarts × prompts`, so large, active, multi-agent conversations amplify the database and worker overhead.

The provider now reuses a persisted synthetic ID only when its provider/content prefix matches. Provider changes still create a fresh ID once. The store-level equality guard provides a second idempotency boundary.

## Verification

- `bun run typecheck`
- `bun test tests` — 2,560 passed, 18 skipped, 0 failed
- reproduced against an active multi-agent workload: before the patch, the outbox added 163 rows between consecutive reads and reached 3.9 million `set_prompt_session` rows; after the patch, repeated generator restarts logged `MEMORY_ID_REUSED` and the cleaned outbox remained at zero

Related to #2793. Complements #3537 by preventing the redundant requeue work at its source.

### DIFF
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -1883,6 +1883,14 @@ export class SessionStore {
   }
 
   updateMemorySessionId(sessionDbId: number, memorySessionId: string | null): void {
+    const current = this.db.prepare(`
+      SELECT memory_session_id
+      FROM sdk_sessions
+      WHERE id = ?
+    `).get(sessionDbId) as { memory_session_id: string | null } | undefined;
+
+    if (!current || current.memory_session_id === memorySessionId) return;
+
     this.db.prepare(`
       UPDATE sdk_sessions
       SET memory_session_id = ?
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -90,10 +90,18 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
     }
 
     if (!session.memorySessionId) {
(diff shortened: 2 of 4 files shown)


thedotmack:

> step 3/5 (synthetic memory_session_id reused across reloads; `updateMemorySessionId` no-op when unchanged) — halves outbox amplification with #3616

thedotmack:

`MERGEABLE` but owner-routed as a plan-21 **bundle candidate**. Overlaps the same outbox-amplification cluster as #3616/#4024. Not a standalone night-ship.

thedotmack:

thedotmack:

thedotmack:
Superseded by same-repo rehost #4032, merged as 46f527e3. Thank you @SYMBaiEX.

**RCA:** OpenAI-compatible path minted a new synthetic ID on every generator restart and amplified `set_prompt_session` outbox rows. Complementary to #4027.

Issue #3607, opened 2026-08-16
[plan-19] Observer Subprocess Isolation — an explicit env, cwd, config, tool set and auth source for the headless generator, with loud degradation


thedotmack:
# [plan-19] Observer Subprocess Isolation — an explicit env, cwd, config, tool set and auth source for the headless generator, with loud degradation

> **Tracker:** #3607 · **Design doc:** `plans/19-observer-subprocess-isolation.md` · **Series:** plan-14 #3602 · plan-15 #3603 · plan-16 #3604 · plan-17 #3605 · plan-18 #3606 · plan-19 #3607 · plan-20 #3608 · plan-21 #3609 · plan-22 #3610 · plan-23 #3611

## Defect

The headless observation generator is a `claude` CLI child of the worker, and it is spawned as if it were the user's interactive session: it inherits (or is denied) whatever the worker happened to inherit, loads the user's full `~/.claude` config including third-party plugin hooks, runs in a cwd it can write to, keeps peer-messaging tools, resolves the CLI binary once and never again, and picks its credential from a hard-coded chain that ignores what the user actually configured. Each of those is a separate spawn attribute; none of them is declared, and when any of them fails the generator returns prose that plan-18's classifier used to drop silently.

- **Config/hook inheritance:** the SDK child loads all plugin hooks; a third-party SessionStart hook injects "you MUST invoke the skill" and hijacks INIT (empty response, zero observations). `--setting-sources` was emitted with no value in older bundles (`Invalid setting source: --permission-mode`).
- **Tool surface:** `disallowedTools` lists Bash/Read/Write… but not `SendMessage`; observer sessions have sent fabricated instructions to the user's primary session. `CLAUDE_MEM_MAX_CONCURRENT_AGENTS=0` is unrepresentable (`parseInt(...) || 2`).
- **cwd and identity:** on Windows `cross-spawn` routes the SDK spawn through `cmd.exe /d /s /c "<argv joined>"`, so `>` in transcript text creates 0-byte files in the *user's project directory* (and gets committed by `git add -A`); observer transcripts land in `~/.claude/projects/` where Claude Desktop lists them as sessions (452 of them); observer-spawned SDK sessions are re-captured by hooks and filed under a project named after the plugin version dir (`13.12.4`) or the plugin cache basename.
[...]

PR #3615, opened 2026-08-17
refactor: simplify subprocess tool lockdown to tools:[] + canUseTool, add systemPrompt

joeblackwaslike:
## Summary

- Remove three redundant SDK lockdown layers (`allowedTools: []`, `disallowedTools: [list]`, `permissionMode: 'dontAsk'`) that operated on a tool surface already emptied by `tools: []`
- Delete the `OBSERVER_DISALLOWED_TOOLS` constant (no longer needed)
- Add `systemPrompt` pass-through on `buildHardenedSdkOptions()` — defense-in-depth for the system identity text, complementing `settingSources: []`
- Thread `mode.prompts.system_identity` (Observer) and `corpus.system_prompt` (KnowledgeAgent) through all three call sites

## Rationale

Verified against `@anthropic-ai/claude-agent-sdk` v0.3.x:
- `tools: []` is the SDK's TRUE restrictive allowlist — disables ALL built-in tools
- `allowedTools` is an auto-approve list, NOT a restriction — `allowedTools: []` means "nothing auto-approved" but all tools remain callable
- `disallowedTools` and `permissionMode` operate on a surface that `tools: []` has already emptied
- `canUseTool` remains the deny + audit backstop

The `systemPrompt` option (SDK v0.1.0+) sets the system prompt slot directly. `settingSources: []` already suppresses CLAUDE.md in the SDK, but explicit `systemPrompt` documents intent and guards against regressions.

## Test plan

- [x] RED phase: tests rewritten to assert target state (3 expected failures)
- [x] GREEN phase: implementation simplified, all 12 security tests pass
- [x] TypeScript compilation clean (`npx tsc --noEmit` — 0 errors)
- [x] Full test suite: 2283 pass, 0 fail across 198 files
- [x] Verified `process-registry.ts` pair-aware arg filter untouched
- [x] Verified `EnvManager.ts` / `env-sanitizer.ts` untouched
- [x] No dangling references to deleted symbols

### DIFF
--- a/src/sdk/hardened-options.ts
+++ b/src/sdk/hardened-options.ts
@@ -9,58 +9,47 @@
  * guarantee was enforced ONLY by `disallowedTools`. If a future SDK release
  * shipped a new built-in tool that was not in our deny-list, the Observer could
  * autonomously call Edit/Write/Bash on the user's source tree. This helper
- * makes the prompt's guarantee true at the SDK-config layer with
- * defense-in-depth — no single option is load-bearing:
+ * makes the prompt's guarantee true at the SDK-config layer with a simplified
+ * two-layer model:
  *
- *   - belt:        `tools: []`           — the SDK's TRUE restrictive allowlist.
+ *   - restriction: `tools: []`           — the SDK's TRUE restrictive allowlist.
  *                                          Per the SDK type docs, `tools: []`
  *                                          disables ALL built-in tools. (Note:
  *                                          `allowedTools` is an AUTO-APPROVE
- *                                          list, NOT a restriction — see below.)
- *   - empty allow: `allowedTools: []`    — nothing is auto-approved.
- *   - suspenders:  `disallowedTools`     — explicit per-tool deny list.
- *   - braces:      `permissionMode`      — 'dontAsk' = deny unless pre-approved
- *                                          (nothing is pre-approved here).
+ *                                          list, NOT a restriction, so it is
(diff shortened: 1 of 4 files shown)


thedotmack:
**Routing note from the PR consolidation pass.** Every open PR was read in full and clustered by root cause; this one is tracked under **#3607 ([plan-19] Observer Subprocess Isolation)** as **needs rework**: the diagnosis is right, but the item noted below blocks it. If you'd like to keep it moving, address that against the design in the plan doc; otherwise the fix will be absorbed into the plan-19 bundle.

> step 1 (`ObserverSpawnSpec` declares tools/permissionMode once) — plan keeps `disallowedTools`; P1: capability probe still requires `dontAsk` (contract drift between two hand-maintained copies)

thedotmack:
Routing to **#3607 plan-19 (Observer Subprocess Isolation)** — the master explicitly covers `disallowedTools`/SendMessage lockdown and the headless generator's explicit config surface, which is exactly `hardened-options.ts`.

Leaving open. Worth a careful read rather than a quick merge: `main`'s own comment block in `hardened-options.ts` argues the redundant layers are deliberate belt-and-suspenders against a future SDK release changing `tools: []` semantics, so removing `allowedTools: []`, `disallowedTools` and `permissionMode` is a real posture change. The `systemPrompt` pass-through half is uncontroversial and could be split out.

Issue #3609, opened 2026-08-16
[plan-21] SQLite Schema Evolution & Queue State Integrity — one DDL path guarded by introspection, immutable session keys, and queue/session states that can actually transition


thedotmack:
# [plan-21] SQLite Schema Evolution & Queue State Integrity — one DDL path guarded by introspection, immutable session keys, and queue/session states that can actually transition

> **Tracker:** #3609 · **Design doc:** `plans/21-sqlite-schema-evolution-and-queue-state.md` · **Series:** plan-14 #3602 · plan-15 #3603 · plan-16 #3604 · plan-17 #3605 · plan-18 #3606 · plan-19 #3607 · plan-20 #3608 · plan-21 #3609 · plan-22 #3610 · plan-23 #3611

## Defect

The store evolves through two competing DDL paths guarded by version flags instead of the real schema, and its runtime state machines have dead transitions:

- `initializeSchema()` runs before the migration chain and contains `CREATE INDEX … ON sdk_sessions(platform_source)`; on a v23 database `CREATE TABLE IF NOT EXISTS` is a no-op, the index throws `no such column: platform_source`, migrations 24–35 never run, and the worker stays "healthy but not ready" forever. `removeSessionSummariesUniqueConstraint()` (v7) is guarded by a live index probe re-run on every construction; older bundles' lax probe misclassifies the v41 `ux_session_summaries_origin` index and **rebuilds `session_summaries` from a frozen 14-column list**, dropping `discovery_tokens` and the v21 `ON UPDATE CASCADE` FK; `ensureDiscoveryTokensColumn()` is guarded on the v11 ledger stamp so it never self-heals; two rebuild paths emit different DDL at the same schema version. Nothing stamps which binary version last wrote the DB, so an older worker (plan-16 stale-dir spawn) can destroy a newer schema.
- Children (`observations`, `session_summaries`) FK on the **mutable** `sdk_sessions(memory_session_id)`; `startSession` rewrites it (`updateMemorySessionId(id, null)` then a new id) under a live child set → 10k–27k `foreign_key_check` violations, `FOREIGN KEY constraint failed` generator loops every 5–10 s where the cascade is missing.
[...]

PR #3616, opened 2026-08-17
fix: sync_outbox unbounded growth when cloud sync is unconfigured (+28x re-registration amplification)

maksodf:
## Summary

On installs where cloud sync is **not configured** (the default), `sync_outbox` grows forever: the producer writes unconditionally while the only consumer (`CloudSync.drainMutations`) is never even constructed without full credentials. A second defect multiplies the growth: every session re-registration re-enqueues the session's **entire prompt history**.

Observed on a live 13.15.x deployment (macOS, cloud sync never configured):
- `sync_outbox`: **10,457 rows**, zero ever deleted (`sqlite_sequence` == row count), oldest from the moment migration 47 created the table.
- Those rows cover only **369 distinct prompts** across ~20 sessions — a **28× amplification**: sessions get fresh synthetic `memory_session_id`s on (re)start, and each `requeuePromptSync` re-enqueues one `set_prompt_session` op per prompt, every time.
- Growth is unbounded by design today: all three `DELETE FROM sync_outbox` sites are behind either a one-shot migration or the never-constructed CloudSync object. (Cost is currently small — a few MB — but it's also per-write WAL churn on a hot path, forever.)

## The fix (two parts)

1. **Gate the producer on the consumer existing.** `SessionStore` gains a `syncOpsEnabled` constructor option (default `true`, so existing direct constructions keep today's behavior). `DatabaseManager` hoists the exact CloudSync-construction predicate into one place and passes it to both — when cloud sync isn't configured, `enqueueMutationOp` no-ops and `requeuePromptSync` skips its whole transaction (the `sync_rev` bump + `synced_at` re-null serve only sync; skipping also preserves the v47 launch-baseline exclusion, which the bump would otherwise silently promote into sync eligibility).
2. **Supersede, don't append.** For `set_prompt_session`, `enqueueMutationOp` now DELETEs any queued op with the same `(op, target)` (NULL-safe `IS ?` on `origin_device_id`, `json_valid`-guarded per the v49 precedent) before inserting. A newer repair op carries complete fields at a higher rev, so an **unsent** older one is dead weight; replicas apply by the rev guard either way. This bounds the outbox at ≤ one row per (prompt, op-type) regardless of re-registration count. No rev comparison on the DELETE: the mutation-site `sync_rev` bump makes revs monotonic per target (noted in-code), and lexical comparison of canonical decimals would itself be a bug.

## Tests

- New `tests/worker/sync/outbox-growth.test.ts` (7 tests, following `mutation-sites.test.ts` conventions): disabled ⇒ outbox stays empty across registration/re-registration; enabled ⇒ N re-registrations leave exactly one row per prompt (supersede), per-target scoping, fresh-UUID minting, and existing single-registration behavior preserved.
- Red/green: with the src changes stashed, exactly the 7 new tests fail; with them applied, **full suite: 2618 pass, 18 skip (pre-existing), 0 fail** (`bun test tests`). `tsc --noEmit` clean on both tsconfig legs; `lint:hook-io` and `lint:spawn-env` clean.

## Notes / non-goals

- `emitRemapProject` (`src/services/sync/remap-outbox.ts`) also inserts ungated, but those are bounded one-shot worktree-adoption/cwd-remap events, not this amplification class — left alone deliberately; happy to gate it too if you'd prefer symmetry.
- This bounds **future** growth only. No purge of already-accumulated rows is included: on temporarily-unconfigured installs a purge could destroy legitimately queued ops, so existing-row cleanup feels like the operator's/a migration's call, not this patch's.
- Known edge (accepted): if sync was configured, rows pushed, then un-configured, a re-registration during the off window isn't replayed on re-enablement. Never-pushed prompts are unaffected — they resolve join fields at snapshot time (`CloudSync.ts` snapshot path).

### DIFF
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -71,10 +71,25 @@ interface SdkSessionDetailRow {
   status: string;
 }
 
+export interface SessionStoreOptions {
+  /**
+   * Whether this store may enqueue mutation ops into sync_outbox. The only
+   * consumer of that queue is CloudSync's drain, and DatabaseManager
+   * constructs CloudSync iff cloud sync is fully credentialed — so the
+   * worker bootstrap passes the same configuration state here, and an
+   * unconfigured install produces no ops it can never drain. Defaults to
+   * true (the pre-flag behavior) for direct constructions that never wire
+   * the flag.
+   */
+  syncOpsEnabled?: boolean;
+}
+
 export class SessionStore {
   public db: Database;
+  private readonly syncOpsEnabled: boolean;
 
-  constructor(dbPathOrDb: string | Database = DB_PATH) {
+  constructor(dbPathOrDb: string | Database = DB_PATH, options: SessionStoreOptions = {}) {
+    this.syncOpsEnabled = options.syncOpsEnabled ?? true;
     if (dbPathOrDb instanceof Database) {
       this.db = dbPathOrDb;
     } else {
@@ -1899,8 +1914,25 @@ export class SessionStore {
    * RULES, SyncApply.ts). Pure SQL, no notify(): callers on the worker
    * connection nudge CloudSync themselves; the startup drain catches the
    * rest.
+   *
+   * Producer gate: acked ops are DELETEd by CloudSync's drain — the queue's
+   * ONLY retention path — and CloudSync exists iff cloud sync is fully
+   * credentialed. With syncOpsEnabled false (unconfigured install) this
+   * no-ops instead of growing sync_outbox forever.
+   *
+   * Supersede, don't append (set_prompt_session): every session
+   * re-registration re-emits the repair for EVERY prompt in the session
+   * (requeuePromptSync), and the mutation site bumps the prompt's sync_rev
+   * before each enqueue — so per target the newest op always carries the
+   * complete field set at the highest rev, and a still-queued older op is
+   * dead weight. Replicas apply by the op.rev >= row sync_rev guard, so
+   * dropping an unsent superseded op cannot regress them; one already pushed
+   * (ack lost mid-flight) is ordered before the newer op in the hub log and
+   * converges the same way. This bounds the outbox at one
+   * set_prompt_session row per prompt regardless of re-registration count.
    */
   private enqueueMutationOp(rev: string | number, body: CanonicalMutation): void {
+    if (!this.syncOpsEnabled) return;
     // set_prompt_session records NULL as the durable "this device" marker;
     // validate the exact mutation shape/UTF-8 bounds with a temporary valid
     // device id before appending. CloudSync substitutes the resolved device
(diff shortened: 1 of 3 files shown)


thedotmack:

> step 5 (retention: outbox producer gated on the consumer's existence; `set_prompt_session` supersedes); P1: keep a `set_title` backfill path

webandseo:
Second field data point for this, from Windows rather than macOS — and one that
revises the severity note in the summary ("Cost is currently small — a few MB").
On a 16 GB machine it was not small: it ended in OS-level memory exhaustion and
kernel bugchecks.

## Environment

claude-mem 13.15.3 · bun 1.3.13 · `CLAUDE_MEM_RUNTIME=worker` ·
`CLAUDE_MEM_QUEUE_ENGINE=sqlite` · Windows 11 25H2 (build 26200, UBR 9168) ·
16 GB RAM. Cloud sync never configured — `TOKEN`, `USER_ID`, `HUB_URL` all empty,
`GET /api/sync/status` → `{"configured": false}`.

## Outbox

**35,018 rows** and climbing, versus the 10,457 reported here. Measured live over
a clean 3-minute window: **+239 rows (79/min)**, i.e. roughly 114k/day under
normal use. Per-day counts from `created_at_epoch`:

```
2026-08-24 : 1008      2026-08-29 : 8909
2026-08-25 :  385      2026-08-30 : 4139
2026-08-28 : 6723      2026-08-31 : 9164
```

`sync_dead_letter` and `sync_content_outbox` both stay at 0, so nothing fails
loudly — it just accumulates, exactly as the "producer writes unconditionally /
consumer never constructed" analysis predicts.

## Worker memory — the part that made this expensive

The `bun` worker's private working set, sampled every 30 minutes:

```
10:14  386 MB   (3.3 h uptime, machine idle, outbox +6 rows/min)
10:18  417 MB
10:48  570 MB
10:53  641 MB
11:0x  685 MB   (4.0 h uptime, outbox +79 rows/min)
```

Growth tracks **outbox insertion rate, not elapsed time**: flat around 390 MB for
three idle hours, then climbing once sessions became active. Two larger readings
were captured independently by the Windows `Resource-Exhaustion-Detector`
(event 2004), which names top consumers by process:

```
2026-08-29 20:40   bun.exe   1.97 GB
2026-08-31 22:08   bun.exe   7.79 GB
```

`POST /api/admin/restart` reliably returns it to ~400 MB with no data loss, and
after a restart it starts at ~400 MB **even with 31k rows already queued** — so
the backlog is not being loaded into memory; the growth is per-operation. I have
not isolated the allocation site. The unfinalized prepared-statement handles in
`requeuePromptSync` described in #3537 look like a plausible candidate, given
that `requeuePromptSync` is also the amplifier described here, but I can't
confirm that from the outside.

## Downstream impact

The 7.79 GB reading filled the commit charge on 16 GB, produced a Windows
"Out of Virtual Memory" popup at 22:08, and the machine bugchecked minutes later.
Seven bugchecks across August in two families:

- `0x116` VIDEO_TDR_ERROR with parameter 3 = `0xc000009a`
  (`STATUS_INSUFFICIENT_RESOURCES`) — the GPU driver hung and the TDR *recovery*
  then failed for lack of memory
- `0x9F` DRIVER_POWER_STATE_FAILURE, parameter 1 = `0x3`

To be fair to the project: claude-mem does not crash Windows, and the display
driver is the proximate failure. But the memory exhaustion behind it is
attributable to this process in the two events Windows logged by name, and I
spent six days chasing the GPU driver — including a full driver upgrade that
changed nothing — before reading the body of the 2004 event instead of just its
timestamp. On a machine with more headroom this would present only as a slow
creep, which may be why it reads as "a few MB" elsewhere.

## Note on the obvious workaround

Setting `CLAUDE_MEM_CLOUD_SYNC_WS=false` does **not** help: rows keep accumulating
at the same rate. That is consistent with the analysis in this issue — the
producers write unconditionally, so the transport flag is the wrong lever.
Worth knowing before someone suggests it as a stopgap.

Also, none of the `CLOUD_SYNC_*` keys exist in `settings.json` by default (they
are all resolved defaults), so an edit that tries to rewrite an existing value
silently does nothing — the key has to be added.

## Current mitigation

A scheduled probe every 30 minutes that calls `POST /api/admin/restart` above a
1500 MB threshold and verifies `/api/health` afterwards. It works, but it is
babysitting rather than a fix, so I'd be glad to see the producer gate in the
proposed patch land.

Happy to supply the full event-log correlation, the per-day outbox counts, or the
probe's ongoing 30-minute memory series if any of it would help sizing the fix.

thedotmack:

`MERGEABLE` / `UNSTABLE`, but this stays a plan-21 **bundle candidate** (owner routing + Greptile P1: no `set_title` backfill when sync is later enabled). The narrower root cause of the outbox amplification is the per-turn `ensureMemorySessionIdRegistered` overwrite — shipping as #4025. Landing this on top would be a second system. Leaving open.

thedotmack:

thedotmack:

thedotmack:
Superseded by same-repo rehost #4031, merged as 661df742. Thank you @maksodf.

**RCA:** Default installs wrote `sync_outbox` forever because CloudSync (the only drain) is never constructed without credentials. Residual after #4027/#4032.

Issue #3499, opened 2026-08-06
OpenAI-compatible provider appends the same assistant response to conversationHistory twice


bwndlct:
## Bug description

The OpenAI-compatible provider path appears to append the same assistant response to `session.conversationHistory` twice.

### Evidence

In `src/services/worker/OpenAICompatibleProvider.ts`, the provider appends the response before calling the shared response processor:

```ts
session.conversationHistory.push({ role: 'assistant', content: obsResponse.content });
...
await processAgentResponse(obsResponse.content || '', session, ...);
```

Then `src/services/worker/agents/ResponseProcessor.ts` unconditionally appends the same `text` again:

```ts
if (text) {
  session.conversationHistory.push({ role: 'assistant', content: text });
}
```

The same pattern exists for init, observation, and summary responses.

This appears to affect the OpenAI-compatible providers (OpenRouter and Gemini). The Claude SDK provider does not append the assistant response in its provider loop, so the shared processor append is needed there.

### Actual behavior

For one response `R`, the history becomes:

```
user: observation prompt
assistant: R
assistant: R
```

The two assistant messages contain the same response text.

### Expected behavior

Each provider response should occur exactly once in `conversationHistory`, regardless of provider.

### Impact

This duplicates every OpenAI-compatible assistant response in the in-memory conversation history and causes duplicated content to be sent in subsequent requests. It can further amplify context growth and token usage in long-running sessions.

### Suggested regression test

Add an integration-level test for `OpenAICompatibleProvider.startSession()` that processes at least one valid response and asserts that the response text occurs exactly once in `session.conversationHistory`. The existing unit tests for the provider append and the response processor append pass independently, but do not cover their composition.

Relevant source:
- https://github.com/thedotmack/claude-mem/blob/main/src/services/worker/OpenAICompatibleProvider.ts
- https://github.com/thedotmack/claude-mem/blob/main/src/services/worker/agents/ResponseProcessor.ts

Observed against version `13.13.1`.

thedotmack:
Consolidating into #3606 (plan-18). The root cause and fix sequencing are tracked there alongside the rest of the cluster — please follow that issue for progress.

bwndlct:
For attribution: I originally reported this issue and submitted the fix with regression coverage in #3500. Since the same fix later landed in 7a7ada5f6501, could you please preserve credit to #3500 / @bwndlct? Thanks.

PR #3619, opened 2026-08-17
test(worker): pin one assistant turn per observer reply, summary included

ntdatt812:
## Status

The source fix from this PR is on `main`. 7a7ada5f ("bound observer conversation growth…", on `main` since 2026-08-31) removed the provider-side `conversationHistory.push` calls, and its message cites this PR. This branch is now rebased onto current `main` and cut down to the test file. It no longer touches `src/`.

The one assistant push left in `OpenAICompatibleProvider` is in `handleInitResponse`, and it is not a double append. Since a00ce61e the init reply no longer goes through `processAgentResponse`, so the provider records it itself, once. `processAgentResponse` records the observation and summary replies.

## Why the test is still worth having

`main` pins the observation path: `tests/worker/openai-compatible-init-response.test.ts` asserts that roles alternate after init plus one observation.

Two paths have no test:

- **Summary.** Putting the push back in `processSummaryMessage` changes no result in any of the 21 test files that drive the OpenAI-compatible providers or `processAgentResponse`.
- **Empty observation reply, not forwarded.** When a provider does not forward empty replies (`forwardEmptyMessageResponse = false`), an empty observation reply must add no assistant turn. Pushing one there also changes no result in those 21 files.

This file pins both:

- `records each reply exactly once across the init, observation and summary turns`: three queries, roles alternate `user, assistant` three times, and the assistant turns are the three replies in order.
- `adds no assistant turn for an empty observation reply when empty replies are not forwarded`: roles are `user, assistant, user`.

## Verification

Each mutation below was applied to `main`'s `OpenAICompatibleProvider.ts`. The 21 upstream files were compared with an unmutated run by test name, because this Windows machine has pre-existing failures in that set.

| Mutation | 21 upstream files | This file |
|---|---|---|
| Put the push back in `processObservationMessage` | `init-response` fails | test 1 fails |
| Put the push back in `processSummaryMessage` | no change | test 1 fails |
| Push an empty turn for an unforwarded empty observation reply | no change | test 2 fails |

The new file passes (2 pass, 0 fail). It also type-checks clean against the root `tsconfig.json` (strict) when included explicitly; `tests/` is excluded from the regular `typecheck` script.

Refs #3499, #3497. Tracked under #3606 (plan-18).

### DIFF
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -40,6 +40,12 @@ export interface ProviderQueryResult {
  * truncation — is identical between them. Per-provider differences (config
  * resolution, request shape, token estimation, usage/cost reporting) are
  * supplied by abstract members.
+ *
+ * History ownership: this class pushes the *user* half of every turn;
+ * `processAgentResponse` is the single writer of the *assistant* half, for all
+ * providers including Claude. Appending the reply here too left two identical
+ * assistant messages in front of each user turn and doubled the assistant side
+ * of every later request.
  */
 export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string; model: string }> {
   protected dbManager: DatabaseManager;
@@ -177,7 +183,6 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
     responseContext: ReturnType<typeof snapshotResponseContext>
   ): Promise<void> {
     if (initResponse.content) {
-      session.conversationHistory.push({ role: 'assistant', content: initResponse.content });
       const tokensUsed = initResponse.tokensUsed || 0;
       session.cumulativeInputTokens += Math.floor(tokensUsed * 0.7);
       session.cumulativeOutputTokens += Math.floor(tokensUsed * 0.3);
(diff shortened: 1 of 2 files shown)


thedotmack:
Routing to **#3606 plan-18 (Observer Response Pipeline)** — "conversationHistory unbounded / double-appended" is listed in the master, and this is the double-append half.

Verified on `main`: `OpenAICompatibleProvider.ts:234` pushes the assistant reply and `ResponseProcessor.ts:311` pushes it again, so OpenRouter and Gemini really do send `user, assistant, assistant, ...` upstream while `ClaudeProvider` does not. Leaving open — this is the cleanest fix in the cluster at +168/-3 and should land ahead of the larger context work in #3625, which duplicates it as a side effect.

ntdatt812:
Rebased onto current `main` and narrowed to tests: the source change is already on `main` via 7a7ada5f. The assistant push still in `OpenAICompatibleProvider` is `handleInitResponse`'s. Since a00ce61e the init reply no longer reaches `processAgentResponse`, so it is recorded once. What `main` still lacks is a test for the summary path and for an unforwarded empty observation reply. Putting either regression back changes no result in the 21 related test files, and this file catches both. The mutation table is in the description.

Issue #3490, opened 2026-08-04
`max_tokens` hardcoded to 4096 silently truncates 76% of observation responses


ML-o13:
**Version:** 13.11.0
**Provider:** OpenRouter (`CLAUDE_MEM_PROVIDER=openrouter`)
**Model:** `google/gemini-2.5-flash-lite` (supports 65,535 output tokens)

## Summary

The observation synthesis request sends a hardcoded `max_tokens: 4096`. On a normal
working day this truncates the large majority of responses. Because the XML parser only
accepts complete blocks, the truncated tail is discarded silently — no error, no warning,
no log entry. The observations that *do* get stored are well-formed, so nothing looks
broken; the loss is invisible by construction.

## Evidence

From a single day of logs (`~/.claude-mem/logs/claude-mem-2026-08-03.log`):

| Metric | Value |
|---|---|
| Total OpenRouter calls | 1,121 |
| Calls returning exactly 4,096 output tokens | **851 (75.9%)** |
| Observations stored per truncated call | 1.58 |
| Observations stored per non-truncated call | 1.09 |
| Truncated calls storing nothing at all | 2 / 852 (0.2%) |
| Stored observations with missing/aborted fields | **0%** |

Two things stand out:

1. **Truncated responses are the content-rich ones.** They yield ~45% more observations
   than non-truncated calls, which means the cut is removing material from exactly the
   responses that had the most to say.
2. **Nothing detectably fails.** I checked every stored observation from truncated calls
   for empty `narrative`, empty `facts`, empty `concepts`, and narratives ending
   mid-sentence — all came back 0%. The parser cleanly drops the partial trailing block.
   The only symptom is observations that never come into existence.

## Root cause

`scripts/worker-service.cjs`, OpenRouter request path:

```js
body: JSON.stringify({
  model: n,
  messages: s,
  temperature: .3,
  max_tokens: 4096,          // <-- hardcoded
  ...e.includes("openrouter.ai") ? { usage: { include: !0 } } : {}
})
```

The Gemini direct path has the same constant:

```js
generationConfig: { temperature: .3, maxOutputTokens: 4096 }
```

There is no corresponding key in `~/.claude-mem/settings.json` and no environment
variable, so operators cannot raise it without patching the bundle.

## Why raising it is cheap

For input-heavy workloads the output side barely matters. Measured over 86 calls on the
same setup: ~32,264 input tokens vs ~745 output tokens per call, so **input accounts for
[...]

PR #3620, opened 2026-08-17
fix(openrouter): name the max_tokens cap when it cuts a reply off

ntdatt812:
## Defect

`OpenRouterResponse` declares `finish_reason`. The provider reads it in only two places: the empty-message debug line and the Telegram wrap-up error. Every request carries a 4096-token output cap (as `max_tokens`, or as `max_completion_tokens` on the compatibility retry). When a reply stops at a limit (`finish_reason: "length"`), nothing at WARN level says so:

- **A structured block cut off mid-tag never closes.** `parseObservationBlocks` only matches closed `<observation>…</observation>` blocks. If earlier blocks in the reply are complete, they are stored and the cut one disappears without a log line. If none is complete, the reply is classified `xml`, logged as `returned non-XML xml response — ignoring queued batch`, and the batch is confirmed away.
- **A reasoning model that spends the whole cap before answering** comes back as an empty reply, again with nothing naming the limit.

This is the first of the revised suggestions in #3490's correction: *"Log `finish_reason` … `length` should at minimum produce a WARN."* The same correction found that the original 76% figure came from the model repeating itself, not from observations running out of room. So this PR does not raise the cap or retry. It only makes `length` visible.

## Fix

A single `logger.warn` fires when `finish_reason === 'length'`:

```
OpenRouter reply was cut off at the output-token limit
  { model, requestId, maxTokens: 4096, outputTokens, contentChars, messagesInContext }
```

- **Placement.** The warning comes right after the reply text is read, **before** the plain-text (Telegram wrap-up) and empty-reply exits, so every reply shape is covered.
- **Wording.** The message names no response type, because `query()` serves the init, observation, summary and Telegram wrap-up turns. It says "output-token limit" rather than "cap" because `length` can also come from a model limit below the cap; `outputTokens` next to `maxTokens` shows which one it was.
- **Missing usage.** `outputTokens` stays `undefined` when the gateway reports no completion usage.
- **`requestId`.** It lets a user find the generation in the OpenRouter dashboard, as the Telegram error line already does.
- **Named cap.** `4096` becomes `OPENROUTER_MAX_OUTPUT_TOKENS`, used by `buildOpenRouterRequestBody` and by the fallback in `fetchChatCompletion`, so the warning reports the number the request sent.

Returned content, retries and the existing `Empty response from OpenRouter` error are unchanged. `openrouter-empty-content.test.ts` asserts that error's exact two-argument call.

**Overlap with #3868.** #3868 (CONFLICTING, waiting on its rebase) makes the cap configurable through `CLAUDE_MEM_OPENROUTER_MAX_TOKENS` and threads `finish_reason` into the response pipeline. If it lands first, I'll rebase this so the warning reports the configured value instead of the constant.

## Rebased onto current `main`

`main` has restructured this file since the PR was opened:

- the request body is built by `buildOpenRouterRequestBody`
- empty replies are accepted since #4017
- a plain-text path serves the Telegram wrap-up

So the change is reapplied on current code instead of rebased line by line. The earlier version extended the `logger.error` line for empty replies. On current `main` an empty string is no longer an error, so the warning covers that case instead.

## Tests

The new `tests/worker/openrouter-truncation-warning.test.ts` drives the real `query()` path with a mocked `fetch` and asserts the exact warning call. It has 6 tests:

- a cut-off reply warns with `maxTokens: 4096` and the output-token count
- `finish_reason: "stop"` does not warn (negative control)
- an empty reply from a reasoning model that used the whole cap still warns
- a `null` reply warns, and the existing error line is still logged unchanged
- the Telegram wrap-up path warns before it throws on an empty reply
- missing completion usage is reported as `undefined`, not `0`

Each mutation below is caught:

| Mutation | Caught by |
|---|---|
| Remove the warning | 5 of the 6 tests |
| Warn on every `finish_reason` | `stays quiet when the model stopped on its own` |
| Move the warning below the plain-text exit | `names the cap before the Telegram wrap-up gives up on an empty reply` |
| Move the warning below the empty-reply exit | the `null` reply test and the Telegram test |
| Name a response type in the message | 5 of the 6 tests (exact-message assertion) |
| Report missing usage as `0` | `leaves output tokens undefined when the gateway reports no completion usage` |

```
bun test tests/worker/openrouter-*.test.ts tests/worker/telegram-wrapup-provider.test.ts \
  tests/worker/provider-classifiers.test.ts tests/worker/provider-dispatch.test.ts \
  tests/shared/openrouter-attribution.test.ts

143 pass, 0 fail
```

`tsc --noEmit` is clean for the root and viewer configs. The new test file also type-checks clean against the root `tsconfig.json` (strict) when included explicitly; `tests/` is excluded from the regular `typecheck` script.

Refs #3490. Tracked under #3606 (plan-18), step 3.

### DIFF
--- a/src/services/worker/OpenRouterProvider.ts
+++ b/src/services/worker/OpenRouterProvider.ts
@@ -174,6 +174,12 @@ export function classifyOpenRouterError(input: {
 
 const CHARS_PER_TOKEN_ESTIMATE = 4;
 
+/**
+ * Output cap sent on every chat-completions request. Named so the truncation
+ * warning can point at the number that cut the reply off.
+ */
+const OPENROUTER_MAX_OUTPUT_TOKENS = 4096;
+
 interface OpenAIMessage {
   role: 'user' | 'assistant' | 'system';
   content: string;
@@ -292,7 +298,7 @@ export class OpenRouterProvider extends OpenAICompatibleProvider<OpenRouterConfi
         model,
         messages,
         temperature: 0.3,  // Lower temperature for structured extraction
-        max_tokens: 4096,
+        max_tokens: OPENROUTER_MAX_OUTPUT_TOKENS,
         // Ask openrouter.ai for usage accounting (token counts + cost).
         // Only sent to openrouter.ai — strict custom gateways may reject
         // unknown body fields.
@@ -365,8 +371,19 @@ export class OpenRouterProvider extends OpenAICompatibleProvider<OpenRouterConfi
       return responseData;
     }, { label: `OpenRouter ${model}` });
 
+    // `length` means the model was cut off at max_tokens rather than finishing.
+    // The reply is then a partial observation block, the parser rejects it as
+    // prose, and the batch is confirmed and dropped — so without this the only
+    // trace is a "non-XML response" warning that blames the model instead of
+    // the cap that truncated it.
+    const finishReason = data.choices?.[0]?.finish_reason;
+    const truncatedByCap = finishReason === 'length';
+
     if (!data.choices?.[0]?.message?.content) {
-      logger.error('SDK', 'Empty response from OpenRouter');
+      logger.error('SDK', 'Empty response from OpenRouter', {
+        ...(finishReason ? { finishReason } : {}),
+        ...(truncatedByCap ? { maxTokens: OPENROUTER_MAX_OUTPUT_TOKENS } : {}),
+      });
       return { content: '' };
     }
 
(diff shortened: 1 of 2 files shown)


thedotmack:
Routing to **#3606 plan-18 (Observer Response Pipeline)** — "max_tokens + finish_reason=length" is called out verbatim in the master, and the failure mode (a client-set cap producing a truncated `<observation>`, classified as prose, then confirmed away) is the same data-loss path #3624 closes.

Verified still live: `main` reads `choice.finish_reason` only inside the empty-content debug branch (`OpenRouterProvider.ts:580`); a truncated non-empty reply never mentions the cap. Leaving open — small and complementary to #3624, and it should land with whatever raises or configures the cap (see #3635 for the server-side equivalent).

ntdatt812:
Rebased onto current `main`, with the change reapplied on top of #4017 and the Telegram plain-text path. I also reframed it after rereading #3490: its reporter's correction attributes the 76% to repetition rather than a lack of room, and its first revised suggestion is exactly this (`length` should produce a WARN). The warning now fires before both the plain-text and empty-reply exits, names no response type, and carries the request id. The description has a mutation table. If #3868's configurable cap lands first, I'll rebase this to report the configured value.

PR #3624, opened 2026-08-17
fix(worker): preserve rejected observer batches

ChenglinWei97:
## Why

Issue #3606 identified a data-loss path: claimed observer work was confirmed and removed when a model returned malformed XML, prose, or an unrecognized provider failure.

## What changed

- confirm only after successful structured storage or an exact explicit skip sentinel
- preserve rejected and failed-storage batches without committing bad responses to reusable history
- retry recoverable failures once per stable batch, then pause without a third attempt
- apply the same bounded recovery to thrown provider failures, including retry-init failures before reclaim
- keep auth and quota failures paused with typed, content-safe reason telemetry
- regenerate the shipped worker bundle

## Verification

- 2,625 tests passed; 18 existing environment-gated tests skipped; 0 failed
- root and viewer TypeScript checks passed
- full distribution build passed
- hook I/O, spawn environment, postinstall allowlist, worker bundle syntax, and diff checks passed

Refs #3606


ChenglinWei97:
@greptileai

ChenglinWei97:
@greptileai

ChenglinWei97:
@greptileai

ChenglinWei97:
@greptileai please review the current head (9ce17c5c). The prior 4/5 review is on c3495f9f; its concurrent-resume finding was fixed with a per-session single-flight start claim and a concurrent regression test.

thedotmack:
Routing to **#3606 plan-18 (Observer Response Pipeline)** — it's the master's headline defect: confirm only after successful structured storage or an exact skip sentinel, never on prose or malformed XML.

Confirmed not superseded — `src/sdk/output-classifier.ts` on `main` still documents "ordinary observer prose is confirmed and dropped", and both `confirmClaimedMessages` call sites in `ResponseProcessor.ts` are unguarded. Leaving open; this is the strongest existing implementation of the plan-18 confirm contract and the natural spine for the architectural PR.

PR #3625, opened 2026-08-18
fix(worker): bound observer context growth

ChenglinWei97:
## What changed

- Resolve observer context windows from the cached OpenRouter model catalogue, with Gemini mappings and a fail-soft fallback.
- Compact OpenAI-compatible observer history at 70% of the model window and re-inject a recent, token-budgeted project timeline.
- Bound oversized observation and summary payloads before dispatch.
- Build replacement context before mutating history; if timeline construction fails, preserve the existing conversation and continue.
- Remove duplicate assistant history entries and use real input/output token counts when providers report them.
- Add settings and public documentation for the context-window override and compaction kill switch.

## Why

Long observer sessions continually resend their full history. A sampled production request reached roughly 296k input tokens across 313 messages, increasing cost and eventually risking context-window failures. This addresses the context-growth portion of #3606.

This is a standalone current-`main` successor to #3516. It keeps that PR's model-aware budgeting while adding atomic, fail-soft history replacement and removing the stale implementation plan.

## User impact

Gemini and OpenRouter observers retain recent durable project context without allowing the request history to grow indefinitely. Claude SDK sessions are unchanged. A settings kill switch remains available for strict custom gateways.

## Validation

- `bun run typecheck`
- `bun run build`
- Focused observer/context suite: 51 passed
- Full suite: 2,648 passed, 18 environment-dependent tests skipped; two unrelated timing-sensitive tests failed under the loaded full run and both passed immediately in isolated reruns
- `git diff --check`


ChenglinWei97:
@greptileai please review the current head (e8955366). The prior 4/5 review is on fd3146bf; the rendered-order P1 is fixed by comparing the oldest summary candidate using its prepared displayEpoch, with an end-to-end SQLite regression that reproduces the look-ahead timestamp case.

thedotmack:
Routing to **#3606 plan-18 (Observer Response Pipeline)** — "bound every history" is the master's own wording, and a sampled request at ~296k input tokens across 313 messages is the concrete cost.

Leaving open. Two notes for whoever sequences plan-18: (1) `src/shared/conversation-window.ts` landed on `main` in 7a7ada5f6 and has since been removed from the tree, so there is currently no history bound at all — this is still needed; (2) the "remove duplicate assistant history entries" change here overlaps #3619, which fixes the same double-append in isolation and much more cheaply. If both survive, take #3619's narrow fix first and drop it from this branch.

Issue #2958, opened 2026-06-16
tracking: Claude provider observer overflows context window — no bound on history size, zero memory saved on long sessions


thedotmack:
## Root Cause

The Claude provider's SDKAgent accumulates every observation into a single ever-growing conversation with no sliding window, token cap, or message limit. When long sessions with large tool outputs exceed the model's context window, the SDK returns 'Prompt is too long', the session aborts, and zero observations are saved. The OpenRouter provider has a `truncateHistory` mechanism; SDKAgent never received equivalent protection. A related sub-issue is that base64 image blobs in tool payloads bloat the prompt further without being stripped.

### Issues
- Closes #2956 — https://github.com/thedotmack/claude-mem/issues/2956
- Closes #2866 — https://github.com/thedotmack/claude-mem/issues/2866

### PRs
- Related PR #2957 — https://github.com/thedotmack/claude-mem/pull/2957
- Related PR #2884 — https://github.com/thedotmack/claude-mem/pull/2884

## Resolution

This is the canonical tracking issue for this root cause. Individual issues and PRs listed above will be closed in favor of this one.

brandonjveldman:
**Still live in 13.10.2, with a new failure mode: the overflow now arrives as prose and bypasses fresh-session recovery, so the observer gets stuck in an infinite "Prompt is too long" loop.**

Commenting on this canonical tracker instead of opening a duplicate. The symptom you track here (observer overflows the context window, "Prompt is too long", zero memory saved) still reproduces after the `server` rename refactor, but the mechanism has shifted and the older throw-based recovery no longer fires. Adding the root cause specific to the current architecture plus a small recovery patch that complements the context-bounding work in #2957.

## Environment

- Installed version: 13.10.2 (Claude Code plugin install, macOS).
- Provider: `claude` (`CLAUDE_MEM_PROVIDER=claude`), Sonnet-class observer model via `CLAUDE_MEM_MODEL`, roughly 200k context window.
- Data/logs under `~/.claude-mem`.

## Symptom

Long primary sessions stop producing observations entirely partway through. The observer hits "Prompt is too long", retries about once per second, and stores zero observations for the rest of the session. This is not a clean single abort. It is a continuous resume loop against the same poisoned conversation.

## Root cause (the part not yet captured here)

The observer is one persistent SDK conversation per session, and on respawn it resumes the same memory session. When the cumulative context finally overflows, the SDK does not throw. It returns the literal assistant text "Prompt is too long".

[...]

PR #3626, opened 2026-08-18
fix(worker): bound Claude observer context before overflow

### LINKED ISSUE #2956: Claude provider (SDKAgent) has no context bound — observer overflows with "Prompt is too long" and saves zero memory on long sessions

ChenglinWei97:
## Summary

- roll the Claude SDK observer onto a fresh context after a completed turn reaches a configurable full-context threshold
- preserve queued observations, immediately continue them in the replacement generator, and bound failed replacement starts to three attempts while keeping ordinary idle finalization unchanged
- validate and document `CLAUDE_MEM_CLAUDE_MAX_TOKENS` (default `150000`) and expose a privacy-safe `context_bound` telemetry reason
- rebuild the shipped worker bundle

## Root cause

The Claude SDK owns the observer's live conversation context. Truncating claude-mem's local history does not reduce that SDK context, so long sessions can eventually hit a hard "Prompt is too long" failure.

The guard reads finalized result usage (`input_tokens + cache_creation_input_tokens + cache_read_input_tokens`) only after the assistant response has been processed and saved. At the threshold, it stops the old SDK process and starts a fresh one only when more work is already queued.

## Safety properties

- the completed turn is stored before rollover is requested
- missing or malformed usage never triggers a rollover
- settings and environment overrides are strictly bounded to 1,000–1,000,000 tokens, with invalid runtime values falling back to 150,000
- a rejected or non-starting replacement is retried up to three times; persistent failures keep queued work intact and surface observer-health errors
- a finalized real queued turn clears the rollover retry budget, preventing later provider failures from entering the startup retry path
- synthetic init turns cannot trigger a low-threshold rollover loop before queued work is reached
- empty rollovers do not spawn an idle replacement
- normal idle exits still finalize and remove sessions

This is a current-main adaptation of the proactive context-bound design in #2957 by @dndungu. It preserves that original insight while accounting for the current no-session-persistence and in-RAM queue lifecycle.

Fixes #2956.
Related: #2958, #3606.

## Validation

- `bun test tests`: 2,623 passed, 18 environment-gated skipped, 0 failed
- focused rollover/setup suite: 14 passed, 0 failed
- `bun run typecheck`
- `bun run build`
- `git diff --check`


ChenglinWei97:
@greptileai please review the current head (196bd77f), focusing on finalized result-frame context accounting, preservation of claimed and queued buffer items, and the single-flight restart handoff during a context-bound rollover.

ChenglinWei97:
@greptileai PR is ready now. Please review current head 196bd77f and provide the confidence score, focusing on finalized result usage, claimed/queued buffer preservation, and the context-bound restart handoff.

ChenglinWei97:
@greptileai please re-review current head a3fa6020. The P1 rejection path now performs a bounded three-attempt, single-flight recovery; persistent failures preserve the queue, and a finalized real turn clears the startup retry budget. Regression coverage includes one-rejection recovery, persistent rejection bounds, and the low-threshold synthetic-init loop guard. Full suite: 2,623 passed, 0 failed; all five CI checks are green.

vlabeit:
Production reproduction of the overflow this PR guards against (claude-mem v13.15.3, Claude provider, `CLAUDE_MEM_MODEL=claude-haiku-4-5`), in case it helps land it:

- `/learn-codebase` on a 3,600-file / 606k-line monorepo (63 parallel subagents, `Read` tool only) → **3,913** observation events enqueued for one session
- ~2 minutes in, the observer's live SDK conversation filled and the CLI echoed `Prompt is too long` as assistant text; **384** turns came back that way → each logged "returned non-XML prose response — ignoring queued batch" and reset `consecutiveInvalidOutputs` to 0
- **9** observations stored for the whole session; `pending_messages` drained to 0, so the rest was unrecoverable; the observer stayed wedged until `POST /api/admin/restart`

Two distinct defects, so two fixes are complementary:

1. no proactive bound (this PR / #2957) — the context fills;
2. the reactive path never fires: `classifyClaudeError` only sees *thrown* errors, while the overflow arrives as a parsed assistant message and takes the generic drop branch.

I opened #3696 for (2): detect the overflow text, preserve the claimed batch, abort with the existing `overflow` reason, keep the session alive, and restart a fresh generator immediately while work is buffered (bounded: a repeat overflow on the same batch drops that batch instead of looping). It's independent of this PR's rollover threshold and doesn't touch the telemetry enum. Happy to rebase either onto the other.

thedotmack:
Routing to **#3606 plan-18 (Observer Response Pipeline)** — the context-overflow branch of the master. This is the Claude-SDK-specific half: local history truncation can't shrink the SDK's own conversation, so the only bound available is rolling onto a fresh generator once finalized result usage crosses a threshold.

Complementary to #3625, not a duplicate — that one compacts `conversationHistory` for the OpenAI-compatible providers, this one rolls the SDK process. Both are needed. Verified still live: `main` has no `CLAUDE_MEM_CLAUDE_MAX_TOKENS` and no rollover in `ClaudeProvider.ts` (the `conversation-window.ts` helper added in 7a7ada5f6 is no longer in the tree). Leaving open; needs a rebase.

Issue #3372, opened 2026-07-23
🏥 The Merge Clinic: How to get your PR merged in claude-mem


thedotmack:
Welcome to the claude-mem Merge Clinic. This is the complete, honest answer to "why wasn't my PR merged?" and "how do I write one that will be?" The full prose version lives at [`docs/merge-rubric.md`](https://github.com/thedotmack/claude-mem/blob/main/docs/merge-rubric.md) — this is the illustrated walkthrough, with the practical instructions between each slide.

![The claude-mem Merge Clinic](https://raw.githubusercontent.com/thedotmack/claude-mem/main/docs/assets/merge-clinic/slide-01.jpg)

**Before you open a PR, know what the clinic treats.** This repo merges exactly one kind of PR on sight: a small, root-cause fix for a bug that exists on `main` today. Everything below is one checkup in that examination. Read the whole thing once — it will save you a rewrite.

![The Golden Rule](https://raw.githubusercontent.com/thedotmack/claude-mem/main/docs/assets/merge-clinic/slide-02.jpg)

**Instruction #1 — pass every checkup, not most of them.** There is no partial credit. If your diff contains one great fix *and* one retry loop, the whole PR is rejected — we will not cherry-pick the good half for you. If you realize half your diff qualifies and half doesn't, split it yourself and resubmit only the healthy half. One PR, one cure.

![Station 1: Are You Actually Sick?](https://raw.githubusercontent.com/thedotmack/claude-mem/main/docs/assets/merge-clinic/slide-03.jpg)

**Instruction #2 — open your PR description with the symptom.** State the incorrect behavior a user hits **today**: the wrong output, the lost data, the crash, the wrong status code, the comparison that never matched. If you cannot write that sentence, you don't have a bug fix — you have a feature, a preference, or a worry. Also check that the bug still exists on current `main` before you start: a correct re-fix of an already-fixed bug ("superseded") is closed without merge, no matter how good the code is.

![Vitamins vs. Medicine](https://raw.githubusercontent.com/thedotmack/claude-mem/main/docs/assets/merge-clinic/slide-04.jpg)

**Instruction #3 — bring medicine, not vitamins.** Not bugs, and therefore not merged under a `fix:` prefix: features, new settings, hardening ("more robust" code where nothing behaves incorrectly), perf/cost tuning, prompt tuning, and refactors. Those aren't unwelcome in the world — they're just not what this rubric fast-tracks. If you can't name the wound, don't ship armor.

[...]

PR #3627, opened 2026-08-18
fix: restore the star history chart in the README

FaintFlower:
The star history chart in the README and its translations stopped rendering because of GitHub stargazer API restrictions. This PR switches the chart to a working data source so the chart displays correctly again across all supported languages.


thedotmack:

Issue #2192, opened 2026-04-28
[Bug] Codex transcript ingestion on Windows finds no/late items and can strand pending observations


MakaveliGER:
Environment:
- OS: Windows 11
- claude-mem: 12.4.8
- Codex CLI transcript source: ~/.codex/sessions/**/*.jsonl
- Worker: local web UI

Summary:
After installing claude-mem with Codex CLI support, the web UI showed no Codex observations. Claude observations also became easy to strand once the worker had pending/processing rows. I traced this to a few Windows-specific issues in transcript watching and queue recovery.

Observed behavior:
- Codex tab shows "No items to display" even while Codex is actively writing transcripts.
- `transcript-watch.json` contains `~/.codex/sessions/**/*.jsonl`, but on Windows this expands to a backslash path.
- `globSync()` did not match the Codex jsonl files with the backslash-expanded glob.
- After switching to a slash-normalized path, Codex messages entered `pending_messages`.
- However, live appends to already-known transcript files were not reliably processed unless the root watcher explicitly called the existing tailer.
- When the SDK agent went idle/aborted, rows stayed in `pending_messages.status = 'processing'` with the current live `worker_pid`, making them unclaimable by the same worker and leaving the UI empty/spinning.

Local workaround/fix tested:
1. Normalize glob paths with forward slashes after `~` expansion when the path contains glob metacharacters.
2. In the transcript root watcher, when `changed` already has a tailer, call `tailer.readNewData()` instead of returning.
3. On SDK input idle, let the async message generator return/close the input stream instead of aborting the Claude process before it can answer.
4. Requeue in-flight `processing` rows for a session when a generator exits before confirming them, so the same live worker does not self-deadlock.
5. Added a regression test for Windows Codex transcript watching.

Files touched locally:
- `src/services/transcripts/config.ts`
- `src/services/transcripts/watcher.ts`
- `src/services/worker/SessionManager.ts`
- `src/services/queue/SessionQueueProcessor.ts`
- `src/services/sqlite/PendingMessageStore.ts`
- `src/services/worker/http/routes/SessionRoutes.ts`
- `tests/codex-transcript-watcher-windows.test.ts`

Verification:
- Codex transcript files under `C:/Users/<user>/.codex/sessions/**/*.jsonl` are found.
- Codex rows are created in `sdk_sessions` with `platform_source = codex`.
- Codex tool observations enter `pending_messages`.
- Worker health OK after rebuild/restart.
- Regression tests pass:
[...]

PR #1661, opened 2026-04-09
feat: Layer 3 transcript storage, TITANS observer, and build fixes

thedotmack:
## Summary
- **Layer 3 transcript storage**: `get_transcript_segment` MCP tool for retrieving original conversation context from ChromaDB, traversal-only via observation IDs
- **TITANS conversation observer**: Fire-and-forget background analysis producing 7 observation types (emotional_signal, overconfidence, frustration, pattern_recognition, etc.)
- **Progressive search MCP fix**: Dynamic type system for 3-layer search workflow (search → get_observations → get_transcript_segment)
- **Build fixes**: Replaced `glob` with `Bun.Glob` to fix lru-cache/CJS incompatibility, added `bun` to esbuild externals, fixed `OneShotQuery` EnvManager import

## Files changed
- `src/services/sync/ChromaSync.ts` — transcript Chroma methods
- `src/services/transcripts/processor.ts` — conversation exchange tracking, flush to Chroma
- `src/services/worker/http/routes/SessionRoutes.ts` — Layer 3 HTTP handlers
- `src/services/worker/OneShotQuery.ts` — TITANS one-shot LLM queries
- `src/servers/mcp-server.ts` — get_transcript_segment tool registration
- `src/services/sqlite/SessionStore.ts` — getContentSessionIdByMemoryId bridge
- `src/utils/tag-stripping.ts` — stripTranscriptPrivacyTags, chunkText
- `scripts/build-hooks.js` — added `bun` to externals
- `src/services/transcripts/watcher.ts` — glob → Bun.Glob migration

## Test plan
- [x] Worker starts and reports healthy (v12.1.0)
- [x] Layer 1: `GET /api/search?query=...` returns results
- [x] Layer 2: `POST /api/observations/batch` returns full observations
- [x] Layer 3: `POST /api/transcript/segment` returns `{chunks, prompt_number, mode}`
- [x] Layer 3: `POST /api/transcript/store` stores transcript chunks
- [x] TITANS: `POST /api/sessions/conversation-observe` returns 202 accepted
- [x] Build succeeds with `npm run build-and-sync`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3633, opened 2026-08-18
fix(transcripts): serialize tail reads and reset partial data on truncation

zloglevel:
## Summary

- Route all `FileTailer` read triggers through a single coalescing scheduler.
- Prevent overlapping reads from consuming the same transcript range.
- Clear buffered partial-line data when a file shrink is detected.
- Add regression tests for concurrent `poke()` calls and truncation.

## Problem

`FileTailer.start()`, the per-file `fs.watch` callback, and `poke()` can all invoke `readNewData()` without any serialization.

Because `readNewData()` awaits stream I/O, two calls can observe the same shared state before either updates it:

1. Both calls observe `offset = 100` and `size = 200`.
2. Both create a stream for bytes `100..199`.
3. Both update the shared offset and process the same lines.

This can deliver the same transcript entry more than once. When calls observe different file sizes, an older read can also finish last and move the shared offset backwards. If the input does not end with a newline, concurrent calls can additionally duplicate or corrupt `tailState.partial`.

Duplicate transcript processing is particularly risky here because downstream processing can create persistent sessions, observations, or memory state.

There is a separate truncation edge case. The existing shrink handling resets
`offset` but retains `partial`:

```ts
if (size < this.tailState.offset) {
  this.tailState.offset = 0;
}
```

A fragment buffered from the old file can therefore be prepended to the first line of the truncated file, producing invalid JSON.

## Changes

- Add a single-flight, coalescing read scheduler used by `start()`, `fs.watch`,
  and `poke()`.
- Keep at most one `readNewData()` call in flight for each `FileTailer`.
- If another notification arrives while reading, perform one follow-up read
  after the current read completes.
- Reset both `offset` and `partial` when `size < offset`.

## Tests

- Trigger two `poke()` calls before the first read completes and verify that
  the appended transcript entry is processed once.
- Buffer an incomplete line, truncate the file, write a valid JSONL entry,
  and verify that the stale fragment is not prepended.

## Scope

This fixes concurrency within one `FileTailer` and the existing detected-shrink path. It does not add inode-based rotation detection or claim end-to-end exactly-once delivery across process crashes.

Related: #2192, #2209, #2219, #1661

### DIFF
--- a/src/services/transcripts/watcher.ts
+++ b/src/services/transcripts/watcher.ts
@@ -14,6 +14,8 @@ interface TailState {
 class FileTailer {
   private watcher: ReturnType<typeof fsWatch> | null = null;
   private tailState: TailState;
+  private readTask: Promise<void> | null = null;
+  private readPending = false;
 
   constructor(
     private filePath: string,
@@ -25,9 +27,9 @@ class FileTailer {
   }
 
   start(): void {
-    this.readNewData().catch(() => undefined);
+    this.requestRead();
     this.watcher = fsWatch(this.filePath, { persistent: true }, () => {
-      this.readNewData().catch(() => undefined);
+      this.requestRead();
     });
   }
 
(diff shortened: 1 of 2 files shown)


zloglevel:
@thedotmack Could you please take a look when you have a chance? Thanks!

thedotmack:
Rebased onto current `main` (after #4016) and shipped as #4019.

Closing this community PR as superseded. Thanks @zloglevel.

Issue #3605, opened 2026-08-16
[plan-17] Hook Wrapper Contract — thin, cheap, and fail-open on every host


thedotmack:
# [plan-17] Hook Wrapper Contract — thin, cheap, and fail-open on every host

> **Tracker:** #3605 · **Design doc:** `plans/17-hook-wrapper-contract.md` · **Series:** plan-14 #3602 · plan-15 #3603 · plan-16 #3604 · plan-17 #3605 · plan-18 #3606 · plan-19 #3607 · plan-20 #3608 · plan-21 #3609 · plan-22 #3610 · plan-23 #3611

## Defect

The seven hook entries in `hooks/hooks.json` (generated from `src/build/hook-shell-template.ts`) are neither thin nor fail-open. Two structural properties of the wrapper produce the whole symptom family:

**It does expensive, fragile work in the shell before any memory code runs.** Every invocation spawns a *second* login shell (`export PATH="$($SHELL -lc 'echo $PATH' 2>/dev/null):$PATH"`) — 250–420 ms on macOS/Linux, 1–10 s on Git Bash where the profile chain sources nvm etc.; the plugin-root discovery is a `{ printf; ls -dt; printf; } | while read … break` pipe whose producers keep writing after `break` closes the read end (`printf: write error: Permission denied` on MSYS, surfaced by Claude Code as a blocking hook error); the login shell shares fd 0 with the hook payload so profile scripts can drain stdin (`[bun-runner] empty stdin payload received — issue #2188`, now on darwin too); the nested `bash.exe` login shell is what pops a console window per tool call on Windows and doubles the concurrent MSYS fork burst that can wedge Git Bash system-wide; `"shell": "bash"` resolves to the WSL launcher when Git Bash isn't first on PATH; older shapes also ran `worker-service start` on all four events. On a PostToolUse `*` matcher this tax runs ~550–3,500×/day.

[...]

PR #3644, opened 2026-08-19
fix(chroma): own the uvx→uv→python→chroma-mcp process tree on Windows and POSIX

thedotmack:
## The bug

Chroma is not a standalone server — it runs inside a Python `chroma-mcp==0.2.6` process spawned through `uvx`, so every Chroma session is a 4-deep native chain:

```
worker → uvx → uv → python → chroma-mcp
```

Windows has no POSIX process groups, and Node's `process.kill(pid, …)` force-terminates exactly **one** PID. Almost every teardown path in the codebase killed the root only, so the descendants survived. Those survivors are the shared root cause behind three separately-reported bugs:

| Survivor effect | Issue | User-visible symptom |
|---|---|---|
| Orphan inherits the worker's listening socket | #3482 | port wedged, hooks hard-block every prompt |
| `uv` killed mid-build, scratch dirs never reaped | #3540 | `builds-v0/.tmp*` leak — 144 GB / 696 dirs measured |
| Foreign `VIRTUAL_ENV` inherited into the uvx sandbox | #3552 | semantic sync stops **silently**, no error |

The correct tree-kill already existed in `ChromaMcpManager` (`taskkill /PID n /T /F`) — it just wasn't shared with any other call site.

**This is not Windows-only.** While building the regression test we reproduced the orphan on macOS: the worker-recycle path leaks the whole `uv`/`python` chain on POSIX too. That is fixed here as well.

## What changed

- **`src/shared/kill-process-tree.ts` (new)** — the tree-kill extracted from `ChromaMcpManager` and shared. POSIX walks descendants with `pgrep`; Windows enumerates via a single `Get-CimInstance` CSV read walked in memory (one spawn, not one per node).
- **Kill sites routed through it** in the recycle and shutdown-force paths, on both platforms.
- **New `signalMode: 'immediate'`** (SIGKILL-only, no grace) for the version-recycle site, which preserves the #3378 invariant that *zero stale-version code may run* — SIGKILL is uncatchable, and tree-SIGKILL merely signals more processes.
- **Teardown reordered** so the MCP SDK's own `stdin EOF → SIGTERM → SIGKILL` escalation runs before we escalate. This is the actual #3540 fix: `uv` is no longer force-killed mid-build, so scratch dirs stop accruing.
- **Foreign Python env stripped** (`VIRTUAL_ENV`, `PYTHONHOME`, `PYTHONPATH`, `CONDA_PREFIX`, `CONDA_DEFAULT_ENV`) from the Chroma child env — case-insensitively on Windows, where env names are case-insensitive and `PythonPath` would otherwise sail through.
- **`shutdown.ts`**: Chroma registers a `pgid` but is spawned without `detached`, so it is not a group leader and `kill(-pgid)` always ESRCHes, silently falling through to a per-PID signal. Liveness is now evaluated over the descendant set rather than the root, so the force phase is actually reachable.
- **`taskkill` failures no longer read as success** — access-denied was being classified as "already gone", letting `server stop` delete the PID file and report success while the server still held its port.

## Verification

The reason these regressions keep shipping is that `windows.yml` is **build-only** — it never spawns Chroma. Two gates are added:

1. **`worker-recycle-orphans.test.ts`** — the real gate. A fixture builds a genuine Chroma tree, snapshots descendants *before* teardown (once the root dies, children re-parent and a post-mortem walk finds nothing), runs the production `ensureWorkerRunning()` recycle, then asserts no descendant survived — matching on **PID + process start token**, so a recycled PID cannot produce a false pass.

   Because POSIX is fixed too, this runs on **ubuntu-latest** and is A/B proven by execution, not reasoning:

   ```
   # with main's process.kill(pid,'SIGKILL') restored verbatim
   error: orphaned descendants survived the worker recycle:
     /Users/…/.local/bin/uv(pid=52908),
     …/uv/archive-v0/…/bin/python(pid=52909)
    0 pass  1 fail

   # with the fix
    1 pass  0 fail
   ```

2. **`chroma-windows-lifecycle.test.ts`** — a real round-trip (create → add → query, asserted by document ID, not "no exception thrown"), a hostile-env variant that gates #3552, and a teardown assertion that the production descendant enumeration sees a non-empty chain and reaps it.

A `chroma-windows` job is added to `windows.yml` (uv install + cache, 600s prewarm, no auto-retry — a retry would hide the orphan race).

## Scope — stated precisely

**This does not claim POSIX parity.** Accurate framing:

- Kill sites route through `killProcessTree` on both platforms in the **recycle** and **shutdown-force** paths only.
- `server stop` tree-kills on **Windows only**; POSIX still sends a single SIGTERM so DB/queue/HTTP cleanup keeps its window.
- The Windows SIGTERM grace window is **not** preserved — `shutdown.ts` and `process-registry.ts` force-kill the tree at t=0 there.
- The Windows paths **have now executed on windows-2022 and passed** (see results below). Remaining unproven: the identity guard is best-effort on Windows because `captureProcessStartToken` is TTL-cached for 5s there, so a revalidation inside that window re-reads its own cached value. Fully effective on POSIX; strictly no worse than the unchecked behaviour it replaces, but not closed on Windows.

**Deliberately not included:** an automatic sweep of already-leaked `builds-v0/.tmp*` dirs. It was built, then removed — `rmSync` re-resolves its path at removal time, so an ancestor swapped between validation and delete redirects a recursive delete, and Node exposes no `unlinkat`/dirfd to close it. Stopping *new* leaks is the fix; reaping the existing backlog is a separate concern. Users sitting on a backlog should try `uv cache prune` (targeted) rather than `uv cache clean` (wipes everything, forcing a full chromadb rebuild) — note we could not confirm prune reaps `.tmp*` dirs, so treat that as untested advice.

**Out of scope, real, and worth separate PRs:** Bash-only hooks on native Windows (#3605), the `tree-sitter.exe` lookup miss in `smart-file-read/parser.ts`, and `~\` tilde expansion in `paths.ts`.

## Relationship to existing PRs

`ChromaMcpManager.ts` is currently edited by four open PRs that do not reference each other — #3541 (uv grace period), #3567 (env sanitization), #3286 (Job Object), #3292 (broad recovery). This covers the #3541 and #3567 ground through one shared helper instead of four independent edits; credit to those authors for the diagnoses. Also related: #3309 (socket inheritance), #3416 (port rebind), #3529 (windowsHide), #3321 (where.exe PATH).

## Review notes

Reviewed by two independent models across three rounds. Ten defects were found and fixed, including one **regression this PR itself introduced** — the teardown reorder had removed all Windows descendant reaping, because `collectDescendantPids` was `pgrep`-based and returned `[]` on Windows by construction while the escalation was gated behind `exitCode !== null`. That is fixed and now has a test that fails without it.

---

## Verification results

**Windows (windows-2022) — the tests ran, they did not skip:**

```
tests\integration\chroma-windows-lifecycle.test.ts:
(pass) round-trips a document and returns it by id                   [20937.15ms]
(pass) leaves no orphaned descendants after stop()                   [ 9960.58ms]
(pass) still round-trips with a foreign interpreter in the env       [10219.76ms]
 3 pass  0 fail  7 expect() calls

tests\integration\worker-recycle-orphans.test.ts:
(pass) leaves no orphaned uv/python descendants after the production recycle
 1 pass  0 fail  3 expect() calls
```

The 20.9s round-trip is a genuine cold `uvx` resolve + chromadb build, and the diagnostic step reported no surviving `uv`/`python` processes. All other checks green (`CI`, `Windows`, server-runtime e2e, clean-room dependency closure, sync-hub, typecheck/bundle-size).

## Review round 2 — PID reuse

Greptile filed two P1s, both correct and both fixed in `423e3e5c5`:

- **`killProcessTree` union retained a stale descendant PID** across the 500ms grace window and could SIGKILL a replacement process. Descendants now carry a start token and every union member is revalidated immediately before escalation. Exposure is narrower than it looks and this is why the union can't just be dropped: only `before \ after` members are at risk, and they are absent for exactly two reasons — re-parented (alive, must be killed) or dead (reusable). The token separates those. **Pre-existing** (verbatim from #2282), but this series spread the helper from 1 call site to 5, so the blast radius is ours.
- **Shutdown escalation could tree-kill from a reused root PID.** **Introduced by this PR** — `signalProcess` only tree-kills because this series added it, and dead-root records only reach `survivors` because this series widened that filter. The root token is now gated by `isSameProcess`; on mismatch the tree-kill is skipped and only the identity-validated descendant reap runs.

`captureProcessStartToken` was extracted to `src/shared/process-identity.ts` with a new `isSameProcess`, and the ad-hoc identity checks written earlier in this PR now route through it — one implementation, four call sites. `tests/shared/kill-process-tree-pid-reuse.test.ts` asserts both halves (changed identity skipped, unchanged identity still killed); removing the check turns it red.

## Rounds 3-4 — identity revalidation, made safe by default

Greptile found the same PID-reuse class in two further spots (initial descendant signal; Windows root escalation across `close()`). Fixed in `d39ff1855`, and an audit of every site this PR touches found **two more** it had not reported (`reapSession`'s SIGKILL phase, `ensureSdkProcessExit` — both cross a 5s wait).

The rule is now uniform: any PID held across an `await` is revalidated by start token before it is signalled. Sites where the PID is signalled with no intervening `await` are deliberately left unguarded, and that line is recorded in the commit body so it can be challenged rather than rediscovered.

`killProcessTree` gained an `expectedStartToken` option that makes a root mismatch a **complete** no-op — no signal and no enumeration, since walking a reused PID would enumerate a stranger's children and `taskkill /T` would take that subtree down.

11 PID-reuse tests, each verified as a real gate by removing the guard it covers.

### Known limitation — Windows identity guards are best-effort

`captureProcessStartToken` is TTL-cached per-PID for 5s on Windows, so a revalidation inside that window re-reads its own cached value. Every identity guard here is **fully effective on POSIX and best-effort on Windows**. Never worse than the unchecked behaviour it replaces, but not closed — stated here rather than in a footnote because Windows is this PR's headline platform.

### Round 4 — the opt-in shape was the real defect

Making root identity opt-in meant every call site that omitted it reopened the hole. Of 13 real `killProcessTree(` call sites, **11 passed no token** — so the unsafe path was the default for ~85% of usage.

`killProcessTree` now captures the root token itself at entry when none is supplied and revalidates before every signal. `expectedStartToken` is retained as a stronger override for the four callers that must capture *before* entry (self-capture cannot detect reuse that already happened — it would read the replacement's token). Tests target the no-options default path specifically, and are verified red against a faithful reconstruction of the opt-in shape.

**Cost:** revalidation adds a start-token read per signal point — free on Linux (`/proc`) and Windows (cached), ~10–30ms on macOS (`ps` spawn). Teardown is not a hot path.

### Round 5 — identity now comes from the discovering observation

The deepest finding of the review: descendant identity was captured *after* enumeration, so a PID reassigned in that interval had the **replacement's** token stored — and the later check then validated the replacement against itself and blessed it. A guard that certifies the wrong process is worse than no guard.

Enumeration and identity now come from a single observation on every platform: `/proc/<pid>/stat` (ppid + starttime in one read) on Linux, `ps -eo pid=,ppid=,lstart=` on macOS/BSD, and the existing CIM row on Windows. `pgrep -P` was the wrong primitive, not an unavoidable constraint — and the replacement is *cheaper*, one observation instead of N probes.

The same pattern existed in `ChromaMcpManager.snapshotDescendantIdentities` and the shutdown cascade's per-record snapshot; all three now share `collectDescendantIdentities`.

**New failure mode this introduces, guarded deliberately:** if the table read and the probe ever disagree on token *format*, every comparison fails, every descendant is skipped as "reused", and #2313 returns silently while the code still looks guarded. `kill-process-tree-identity.test.ts` asserts the two agree on real processes and runs in **both** CI jobs including Windows, since the CIM format cannot be verified locally.

**Residual:** this closes the enumerate/capture window. It does not close PID reuse in general, and on Windows the 5s token TTL cache means a revalidation inside that window re-reads its own cached value. Fully effective on POSIX, best-effort on Windows.

### Round 6 — the Windows identity check was a tautology, now fixed

The 5s start-token cache was described in earlier revisions of this body as a "best-effort on Windows" limitation. **That framing was wrong.** `captureProcessStartToken(pid)` populated the cache entry and `isSameProcess(pid, token)` read that same entry back — comparing a value against itself and returning `true` 100% of the time inside the TTL. The Windows root identity check had *zero* discriminating power, and the Chroma capture → `close()` → escalate path sits entirely inside that window, authorizing `taskkill /T /F` against a possibly-unrelated subtree.

`isSameProcess` now reads the OS **uncached**, via a module-private reader that cannot leak to callers wanting the cache. `captureProcessStartToken` keeps its cache for `verifyPidFileOwnership`, which does not authorize kills. All 9 `isSameProcess` call sites were audited — every one authorizes an irreversible kill, so there is no non-kill blast radius.

Cost: 2.27 ms/call on macOS (~9 ms per Chroma teardown); POSIX unchanged as it never cached; one CIM query per authorization on Windows.

**Accurate statement of the guarantee, superseding every earlier "best-effort on Windows" line in this description:**

> Identity revalidation reads the OS uncached at every point that authorizes a kill, on all platforms. Descendant identity comes from the same process-table observation that discovers the PID. The 5s start-token cache is retained only for PID-file ownership checks, which do not authorize kills.

**Residual:** identity compares a snapshot token to a fresh read, so a PID reused between that read and the signal microseconds later is still theoretically possible. Node exposes no handle-based kill (`pidfd_send_signal` or a process handle), so this is not closable in-language — but it is a sub-millisecond race rather than a five-second certainty.

### Verification coverage — what is actually asserted, and where

An earlier revision of this description implied uniform coverage. It wasn't. Corrected matrix:

| Mechanism | ubuntu | Windows |
|---|:--:|:--:|
| Descendant enumeration (CIM / `ps` / `/proc`) | yes | yes |
| Whole-tree kill, end to end | yes | yes |
| Already-dead target tolerated (`taskkill` 128 / ESRCH) | yes | yes |
| Root gate is a no-op on identity mismatch | yes | yes |
| Root gate still kills on identity match | yes | yes |
| Token format agreement (table read vs. probe) | yes | yes |
| POSIX reuse suite (`sh`/`pgrep`/SIGTERM fixtures) | yes | **not runnable — no Windows equivalent** |

Adding the existing reuse suite to the Windows job would have been theatre: 6 of its 7 describes are `describe.if(isPosix)` and would silently skip. A new `kill-process-tree-cross-platform.test.ts` uses platform-appropriate fixtures (`cmd.exe`/`ping` vs `sh`/`sleep`) and asserts through the production helpers, so the Windows-specific mechanisms — the CIM read, `taskkill` classification, the root gate — are genuinely executed on `windows-2022`. Verified as real gates on two axes: removing the root gate fails the no-op case; breaking descendant enumeration fails 4 of 5.

**Accurate statement:**

> Tree-kill and identity guarantees are verified end-to-end on **both** ubuntu and Windows via platform-appropriate fixtures. The POSIX-specific reuse suite runs on ubuntu only; the Windows-side mechanisms it would cover are asserted separately by the cross-platform suite.

### Test seams

`waitForUnexpectedCloseCleanupForTesting` has been **removed** — it was a public method on `ChromaMcpManager` existing for a single assertion, which now polls the observable side effect instead. Same assertion, no coverage lost, one less production method.

`__identityProbeCountForTesting` is **kept and documented**: a monotonic read-only counter that cannot mutate state or influence a kill decision, imported by nothing in `src/`, and the only way to observe the property that stops a cached verdict certifying a reused PID. It earns its surface; the other seam did not.


thedotmack:
Both P1s are fixed in `423e3e5c5`. Thanks — these were the right findings, and one of them is ours.

**P1 #1 — `killProcessTree` union retains a stale descendant PID.** Confirmed. Descendants now carry a start token from the scan that found them, and every union member is revalidated immediately before `SIGKILL`. Collection moved from `collectDescendantPids` to `collectDescendantIdentities`; the union is a `Map<pid, token>` with the fresher post-wait token winning on collision.

Worth recording that the exposure is narrower than it first looks, and that this is *why* the union can't simply be dropped: only members of `before \ after` are exposed. Post-wait members were enumerated microseconds before the kill, and PIDs present in both scans are re-confirmed live. A PID in the pre-TERM scan but absent from the post-wait scan is absent for exactly two reasons — it re-parented (alive, and killing it is the entire purpose of the union) or it died (reusable, the bug). The token is what separates those two cases.

**P1 #2 — escalation tree-kills a reused root.** Confirmed, and this one is introduced by this PR rather than pre-existing: `signalProcess` only tree-kills because this series added `killProcessTree` to it, and dead-root records only reach `survivors` because this series widened that filter. Previously a dead root was never force-killed at all. The root's token is now captured with its descendants and gated by `isSameProcess` before the force phase; on mismatch the tree-kill is skipped entirely and only the identity-validated descendant reap runs.

For accuracy on provenance: **P1 #1 is pre-existing** — the union block is verbatim from #2282 and `killProcessTree` previously lived only inside `ChromaMcpManager`. But this series spread it from 1 call site to 5, so the blast radius is ours regardless.

Rather than add a second identity mechanism, `captureProcessStartToken` was extracted to `src/shared/process-identity.ts` alongside a new `isSameProcess`, and the three ad-hoc identity checks written earlier in this PR now route through it — one implementation, four call sites.

New coverage in `tests/shared/kill-process-tree-pid-reuse.test.ts` asserts both halves, since either alone is vacuous: a changed identity is skipped **and** an unchanged identity is still killed. Verified as a real gate — removing the identity check turns `SKIPS a descendant whose PID was reused since the scan` red (4 pass/1 fail), restoring it returns 5 pass/0 fail.

**One limitation we are not claiming is closed:** on Windows, `captureProcessStartToken` is TTL-cached for 5s, so a revalidation inside that window re-reads its own cached value. The identity guard is fully effective on POSIX and best-effort on Windows. It is strictly no worse than the unchecked behaviour it replaces, but it should not be read as closed there.

thedotmack:
Both fixed in `d39ff1855`, and the audit they prompted found two more sites you didn't flag.

**P1 #3 — initial signal targets a reused PID.** Confirmed. The initial loop now revalidates each descendant's start token before `firstSignal`.

**P1 #4 — Windows escalation tree-kills a reused root.** Confirmed. `killProcessTree` gained an `expectedStartToken` option; `ChromaMcpManager` captures the root token *before* `close()` and passes it. On mismatch it is a **complete** no-op — no root signal and no descendant enumeration, because walking a reused PID would enumerate a stranger's children and `taskkill /T` would then take that entire subtree down. The identity-validated descendant reap still runs.

Rather than patch the two reported spots, I classified every site this PR touches by whether the PID crosses an `await`:

**Now guarded** (PID held across a real wait): `kill-process-tree.ts:131` initial loop (enumeration cost — a `ps`/CIM spawn per PID), `ChromaMcpManager:969` root escalation (`await close()`), plus two you did not report — `process-registry` `reapSession`'s SIGKILL phase (crosses `waitForExit(…, 5s)`) and `ensureSdkProcessExit` (crosses `Promise.race([exit, 5s])`).

**Deliberately unguarded**, because the PID is signalled with no intervening `await` and a token would cost a probe to compare a value against itself: `reapSession`'s SIGTERM phase and `sigtermDuplicateSdkProcess` (both immediately follow `isPidAlive`), the prewarm/unexpected-close paths (PID read straight off a live `ChildProcess`), and `worker-utils`/`ServerService` (`verifyPidFileOwnership` already token-checks immediately prior). That line is recorded in the commit body so it can be challenged rather than rediscovered.

Coverage is now 11 tests, each verified as a real gate by removing the guard:

```
remove initial-signal guard → 2 fail  ([graceful] SKIPS…, [immediate] SKIPS…)
remove root gate            → 1 fail  (is a complete no-op when the root token differs)
both restored               → 11 pass
```

The initial-signal tests needed a *different* fixture from the escalation tests — a plain `sleep` child that dies to the first signal, so "skipped" is observable as "still alive". The TERM-proof fixture used elsewhere would have masked the defect.

**One limitation to state plainly rather than bury.** On Windows, `captureProcessStartToken` is TTL-cached per-PID for 5s, so any revalidation inside that window re-reads its own cached value. Every identity guard in this PR is fully effective on POSIX and **best-effort on Windows**. It is never worse than the unchecked behaviour it replaces, but since Windows is this PR's headline platform, it should not be read as closed there.

Suite: 2636 pass / 26 skip / 0 fail. Recycle A/B unchanged — fails on main with real orphaned pids, passes with the fix.

thedotmack:
Both fixed in `e2421eb1a`, and they prompted a structural change rather than two more spot fixes.

**The underlying defect was the shape, not the coverage.** Root identity was made *opt-in*, so every call site that omitted the option reopened the hole — which is why each review round found another one. The sweep makes that concrete: of 13 real `killProcessTree(` call sites, **11 pass no token**. Opt-in was wrong for ~85% of usage.

`killProcessTree` now captures the root's start token **itself** at entry when the caller doesn't supply one, and revalidates before every signal it sends — initial and delayed, root and descendants. Safe behaviour is the default; no call site can opt out by omission.

`expectedStartToken` is kept rather than removed, because it is a genuinely stronger guarantee and not a redundant one: self-capture detects reuse *during* the helper's awaits, while only a caller-supplied token detects reuse that happened *before* entry (by then self-capture would be reading the replacement's token). `ChromaMcpManager` needs exactly that — it captures before `await transport.close()`. Four sites pass an earlier-captured token for this reason (Chroma dispose, both session-reap phases, `ensureSdkProcessExit`); the other nine are correctly served by self-capture. **P1 #6** is now belt-and-braces rather than load-bearing, but the token is still passed there since the reaper captures it earlier than the helper could.

**Tests target the default path** — three new tests call `killProcessTree(pid)` with no options at all, the shape those 11 sites use. Verified red without the fix, by restoring the faithful opt-in shape (a first revert attempt was insufficient — it changed only the token source, which the mock defeated):

```
### root identity OPT-IN (faithful pre-fix shape) ###
(fail) does NOT signal a reused root when no token was passed
(fail) does NOT signal a reused root in immediate mode either
 12 pass  2 fail
```

Restored: 14 pass / 0 fail. Both halves covered — unchanged root still killed, reused root untouched *and its children not enumerated*.

**This surfaced a real problem in our own test suite.** The `#3378` control test went red: the added revalidation inserts a synchronous `ps` spawn before the root signal, and that fixture's `sleep 300 & wait` root was exiting on its own the moment the descendant pass killed its sleep — so the trap never fired. The fixture had been measuring its own timing rather than signal delivery. Fixed to self-loop and outlive its child.

**Two caveats for reviewers, stated plainly rather than in a footnote:**

1. `captureProcessStartToken` is TTL-cached per-PID for 5s on Windows, so a revalidation inside that window re-reads its own cached value. Every identity guard here — now including the default path — is **fully effective on POSIX and best-effort on Windows**. Never worse than unchecked, but not closed, and Windows is this PR's headline platform.
2. Revalidation costs a start-token read per signal point: effectively free on Linux (`/proc`) and Windows (cached), ~10–30ms on macOS (`ps` spawn). Teardown is not a hot path, but a multi-descendant tree now spends a few hundred ms more there.

Suite: 2639 pass / 26 skip / 0 fail. Recycle A/B unchanged — fails on main with real orphaned pids, passes with the fix.

thedotmack:
**(A) `kill-process-tree.ts` — correct, and the sharpest finding of the five rounds.** Fixed in `54024e69b`.

You were right about *why* it's worse than the earlier ones: enumerate-then-probe doesn't merely miss a reuse, it **certifies the replacement**. The probe captures the new process's token, and `isSameProcess` then compares that replacement against itself, matches, and blesses an unrelated process as a legitimate target. A guard that validates the wrong thing is worse than no guard — it converts "we might kill a stranger" into "we confidently kill a stranger."

**Atomic observation turned out to be achievable on all three platforms**, which I did not expect going in:

- **Linux** — `/proc/<pid>/stat` is a single read carrying ppid (field 4) *and* starttime (field 22), and field 22 is already exactly what `captureProcessStartToken` returns there.
- **macOS/BSD** — `ps -eo pid=,ppid=,lstart=` is one snapshot with all three. Verified empirically before writing code: its `lstart` is byte-identical to the `ps -p X -o lstart=` the probe uses.
- **Windows** — the CIM query already carried the parent link; `CreationDate` now comes from the same row.

`pgrep -P` was the wrong primitive, not an unavoidable constraint. The replacement is also **cheaper** — one observation instead of N probes (on Windows that removes N × 100–300ms of PowerShell spawning).

The same enumerate-then-probe pattern existed in two more places you didn't flag — `ChromaMcpManager.snapshotDescendantIdentities` and the shutdown cascade's per-record snapshot. All three now route through the shared `collectDescendantIdentities`.

**Residual, stated rather than declared closed:** this closes the enumerate/capture gap. It does not close PID reuse in general — identity still rests on a token a later revalidation re-reads, and on Windows `captureProcessStartToken` is TTL-cached per-PID for 5s, so a revalidation inside that window re-reads its own cached value. Fully effective on POSIX, best-effort on Windows.

**A new failure mode this design introduces, on record:** if the table read and the probe ever disagree on token *format*, every comparison fails, every descendant is skipped as "reused", and #2313 returns silently while the code still looks guarded. That is the dangerous direction, so `kill-process-tree-identity.test.ts` asserts the two agree on real processes and is wired into **both** CI jobs including Windows — the CIM format being precisely what cannot be verified locally.

**(B) `process-registry.ts:235` — stale, already fixed.** No change made. Current code at `239-241`:

```ts
await killProcessTree(record.pid, {
  expectedStartToken: startTokens.get(record.pid) ?? null,
});
```

Landed in `e2421eb1a` and present in the pushed branch. On the `?? null` form you suggested: `isSameProcess(pid, null)` returns `true` on its first line, so an unmappable PID proceeds to kill rather than becoming un-killable — refusing stays strictly narrower than killing, which is the invariant protecting #2313.

**Testing the interval.** The reassignment window itself is kernel-timed and not drivable, but its observable consequence is: make every probe for one PID report a different process than the table did. Under enumerate-then-probe the stored token *is* the replacement's, so the check matches and the process gets signalled; sourcing identity from the discovering read makes them disagree, so it is skipped. Verified as a real gate:

```
### with enumerate-then-probe (pre-fix shape) ###
(fail) does NOT signal a descendant whose probe disagrees with the table read
 15 pass  1 fail
```

Restored: 16 pass / 0 fail.

Suite 2643 pass / 26 skip / 0 fail; actionlint clean across all 8 workflows; recycle A/B unchanged.

One observation on the arc of this review: five rounds have each found a real defect, all in the same class. That says something about the domain — PID-based teardown has no safe primitive in Node, only progressively better approximations. The atomic-observation change is the first one that *removes* a window rather than narrowing it, so I'd expect returns to drop off here. But "I expect" is the honest phrasing, not "it's closed."

thedotmack:
Fixed in `29b42186f`. You were right to file this P1/security, and it is **worse than filed** — I had been documenting this as a known limitation across three round-ups, and that framing was wrong.

**It is not a timing window. It is a tautology.** `captureProcessStartToken(pid)` populates the cache entry; `isSameProcess(pid, token)` then reads that same entry back. Inside the 5s TTL it compares a value against itself and returns `true` **100% of the time**. So on Windows the root identity check had *zero* discriminating power — and the Chroma path (capture → `await transport.close()` → escalate) sits entirely within that TTL. This is the round-5 defect — a guard certifying the wrong process — on the platform this PR is named after, authorizing an irreversible `taskkill /T /F` against a stranger's entire subtree.

**True exposure, checked before writing code.** The cache is Windows-only; POSIX always read fresh, so "best-effort on Windows" was pointing at this one mechanism and nothing else. Round 5 already fixed descendant *capture* (identity now comes from the CIM table read), so the residual was on the **revalidation** side: total for the root, partial for descendants — a descendant's first revalidation is a cache miss and correct, a second within 5s is stale.

**Uncached read inside `isSameProcess`, not invalidation.** Invalidation is action-at-a-distance — correct only if every present and future call site remembers to invalidate first, which is exactly the opt-in shape that produced rounds 3, 4 and 5. Reading uncached is correct by construction at the single place that authorizes a kill, and no call site can forget it.

Blast radius contained as follows: `captureProcessStartToken` keeps its cache, so `verifyPidFileOwnership` (the repeated-validation case the cache exists for) is untouched. The uncached reader is **module-private and not exported**, so a bypassing read is only reachable through the predicate that authorizes kills. I audited all 9 `isSameProcess` call sites first — every one authorizes an irreversible kill, so there is no non-kill blast radius and no new API surface was needed.

**Tests are non-vacuous on macOS.** A probe counter makes the real property observable on every platform: *`isSameProcess` must perform a fresh read on every call, regardless of caching policy.* Plus a Windows-gated case asserting the cached accessor still genuinely serves from cache — without that, the bypass assertion could pass on a build where caching had silently broken and would prove nothing. Verified red without the fix:

```
### with isSameProcess using the CACHED read (pre-fix) ###
(fail) identity revalidation never trusts the cache > isSameProcess performs a fresh read on every call
 3 pass  1 skip  1 fail
```

This file already runs in both CI jobs, so the Windows half executes there.

**Measured cost: 2.27 ms/call on macOS** — ~9 ms for a typical Chroma teardown (one root, three descendants). POSIX is unchanged since it never cached. On Windows, `killProcessTree` makes exactly one root check before `taskkill` (the descendant loops are POSIX-only), so it pays one CIM query per authorization. Nothing lands anywhere that hurts.

**Corrected claim, replacing the "best-effort on Windows" line I wrote three times:**

> Identity revalidation reads the OS uncached at every point that authorizes a kill, on all platforms. Descendant identity comes from the same process-table observation that discovers the PID. The 5s start-token cache is retained only for PID-file ownership checks, which do not authorize kills.

**Honest residual:** identity still rests on comparing a snapshot token to a fresh read, so a PID reused *between* that read and the signal microseconds later remains theoretically possible. Node exposes no handle-based kill — no `pidfd_send_signal`, no process handle — so that last gap is not closable in-language. But it is now a genuine sub-millisecond race rather than a five-second certainty.

Suite 2645 pass / 27 skip / 0 fail; recycle A/B unchanged.

thedotmack:
Fixed in `07a4f335c`. Well-founded — this is an instance of a limitation noted in the round-4 reply: *self-capture detects reuse during the helper's awaits; only a caller-supplied token detects reuse that happened before entry.* `onclose` fires **because** the child died, so by definition the process is gone when cleanup starts — the one case where self-capture is guaranteed to be too late, and it then validates the replacement against itself.

**The rule this exposes:** any site retaining a PID from a live `ChildProcess` and killing it later must capture the token *while that child is alive*. Self-capture suffices only where liveness is guaranteed at call time.

**Changed (3) — all unsafe under that rule:**

| Site | Why it fails |
|---|---|
| `ChromaMcpManager:317` onclose cleanup | fires because the child died — self-capture always too late (the reported one) |
| `ChromaMcpManager:687` prewarm failure/timeout | often reached *because* the child exited; awaits precede it |
| `ChromaMcpManager:1107` disposeActivePrewarm | `activePrewarmChild` can hold an already-exited handle |

**Left alone (6) — each checked against the rule, not assumed:** `process-registry:745` (self-capture runs synchronously at entry, immediately after `isPidAlive`); `shutdown.ts:279`/`:301` (liveness check + token capture, or gated by `rootIsSame` on a token captured while alive); `worker-utils:529` and `ServerService:380` (`verifyPidFileOwnership` immediately prior); `process-registry:440` (already passes an explicit token).

**Made structural rather than remembered.** `TrackedChild` pairs pid+token at spawn so they travel together, and no cleanup path in that class can carry one without the other — the same reasoning that replaced opt-in with self-capture in round 4. Deliberately *not* extended to `ManagedProcessInfo`/`supervisor.json`: every registry cleanup already captures its token while the process is provably alive, so persisting one would be schema churn with no defect behind it.

**A gap in our own testing this surfaced.** The mechanism test passed immediately, but reverting *only* the Chroma call site left the integration suite green — meaning the wiring was unverifiable and we'd have shipped a fix no test could see. Added a wiring assertion:

```
### onclose reverted to self-capture ###
(fail) onclose cleanup carries the spawn-time identity token, not self-capture
 22 pass  1 fail
```

And at mechanism level:

```
(fail) killProcessTree refuses a ROOT whose PID was reused
(fail) a caller-supplied token beats self-capture for a DEAD pid
 16 pass  2 fail
```

Suite 2648 pass / 27 skip / 0 fail; recycle A/B unchanged.

---

**On whether this review is converging** — worth stating plainly, since two predictions of drop-off have now been wrong.

Every round since 3 has been the *same* defect: **identity established at time A authorizes a kill at time B.** What changes each round is only where the A→B gap hides — the union set (round 5), the cache making A==B a tautology (round 6), A being unobtainable because the process is already dead (round 7).

The findings aren't converging, but **the fixes are**: round 4 made identity the default rather than opt-in, round 5 made capture atomic with discovery, round 6 made revalidation uncached, round 7 made capture-at-spawn structural. Each removed a category rather than a line. There are only so many points at which identity can be established — spawn, discovery, entry, revalidation — and after this round all four are covered.

If a round 8 does surface another instance of this class, the right response is probably *not* another fix. It's to question whether PID-based teardown is the wrong primitive, and whether the real answer is Job Objects on Windows and pidfds on Linux — which is a different PR, not a patch to this one.

Two things this PR is still honestly exposed on: the reuse branches are exercised by no CI on any platform, and the test seams added under review pressure (`__identityProbeCountForTesting`, `waitForUnexpectedCloseCleanupForTesting`) are production API surface that deserves a look.

PR #3648, opened 2026-08-19
fix(paths): expand Windows tilde paths

thedotmack:
## Summary

- Share one platform-aware tilde expansion helper across the original `src/shared/paths.ts:32` and `src/utils/project-name.ts:8` sites.
- Expand `~\\` only on Windows; keep it literal on POSIX, where backslash is a legal filename character.
- Add explicit-platform regression coverage for `~/x`, `~\\x`, bare `~`, and non-tilde paths, plus project-name detection.

## User impact

On Windows, `CLAUDE_MEM_DATA_DIR=~\\.claude-mem` stayed relative, so the database, settings, PID files, and Chroma data were silently written under the process CWD instead of the user's home directory. Project detection had the same unexpanded-path defect.

Found during the #3644 Windows audit.

## Validation

- `npm run build`
- `npm run typecheck`
- `npm test` (2,614 passed, 18 environment-gated skips, 0 failed)


thedotmack:
@greptile-apps Correct—fixed in `90aa9a1eb`.

I ran the canonical `npm run build` and committed all eight regenerated tracked artifacts: the worker, server, MCP, context-generator, and transcript-watcher bundles; both SQLite bundles; and the viewer bundle. Seven resolver-carrying artifacts now contain the platform guard, including `plugin/scripts/worker-service.cjs:19`:

```js
t.startsWith("~/") || e === "win32" && t.startsWith("~\\")
```

The same logic is embedded in `server-service.cjs`, `mcp-server.cjs`, `context-generator.cjs`, `transcript-watcher.cjs`, `plugin/sqlite/SessionStore.js`, and `plugin/sqlite/observations/files.js`. The ignored NPX build also carries it.

After committing, I ran `npm run build` again and confirmed both `git diff --exit-code` and `git diff --cached --exit-code` remained clean, so the tracked build is reproducible.

Focused source and bundle-format tests pass (46/46). The coverage hole is real: the tilde tests import the TypeScript source, while existing committed-bundle tests only check properties such as line endings. A stale worker bundle could therefore have shipped with the source suite green; no bundle/source lockstep CI was added in this focused follow-up.

PR #3657, opened 2026-08-19
fix(windows): make build-and-sync work without rsync or a POSIX shell

thedotmack:
## Windows users cannot build and test from source at all

We have open PRs fixing Windows bugs (#3644, #3647, #3648, #3649) and we are asking Windows users to test them from source. They can't — the source-install path itself is broken on Windows, so those PRs are effectively untestable by the people they're for.

`npm install` and `npm run build` already work on native Windows — `windows.yml`'s build job runs exactly those on `windows-2022` and passes on every PR. Everything after them is bash-only:

1. **`scripts/sync-marketplace.cjs` shelled out to the `rsync` binary** (`execSync('rsync -av --delete ...')`). `rsync` does not exist on Windows and nothing in this repo installs it; Git for Windows does not bundle it. Hard stop.
2. **`~` home expansion.** `build-and-sync` was `... && (cd ~/.claude/plugins/marketplaces/thedotmack && npm run worker:restart)`, and the sync script itself passed `~/...` paths through `execSync` to the shell. `~` is bash-only; `cd ~/...` in cmd.exe/PowerShell does not resolve.
3. **`worker:logs` / `worker:tail`** were `tail -n 50 ~/.claude-mem/logs/worker-$(date +%Y-%m-%d).log` — `tail`, `date +%F` and `~` are all bash-only.

Documented install flow for a non-stable branch (`docs/.../branches.mdx`) is clone → `npm install` → `npm run build-and-sync`. Step 3 fails on Windows.

## What changed

**`scripts/mirror-dir.cjs` (new)** — a stdlib-only replacement for `rsync -a --delete --exclude=...`. No new npm deps. It reproduces the rsync semantics the sync depends on:

- **Pattern syntax:** leading `/` anchors to the mirror root (`/workers`), trailing `/` matches directories only (`node_modules/`), slash-free patterns match a basename at any depth (`.DS_Store`), patterns with an internal slash match on directory boundaries (`scripts/package.json`), plus `*` / `**` / `?`. This is the rsync subset the repo's exclude list and `.gitignore` actually use.
- **`--delete`, including receiver-side protection.** This is the part a naive `cpSync` gets wrong in both directions: stale destination files must be removed (otherwise you rebuild and silently keep testing old code — a real source of "I rebuilt but nothing changed" bugs), *but* excluded paths must be left alone, or the first sync wipes the marketplace's `.git` and its installed `node_modules`.
- **rsync's default quick check** (same size + same whole-second mtime ⇒ skip), so repeat syncs stay cheap and idempotent, and mtimes/modes/symlinks are preserved like `-a`.

**`scripts/sync-marketplace.cjs`** — calls `mirrorDirectory` instead of `rsync`; `bun install` in the marketplace now runs via `{ cwd: INSTALLED_PATH }` (matching how the cache-dir install already worked) instead of `cd ~/... && bun install`. Exclude-list assembly is exported and the entrypoint is behind `require.main === module` so it can be tested. Errors still print and `exit(1)` — nothing is swallowed; a failed copy is loud, because a silent one means testing a stale build.

**`scripts/restart-marketplace-worker.cjs` (new)** — replaces the `cd ~/...` subshell in `build-and-sync`. Resolves the path with `os.homedir()`, fails fast with a clear message if the marketplace isn't there.

**`scripts/worker-logs.cjs` (new)** — replaces `tail`/`date`. `worker:logs` prints the last 50 lines, `worker:tail` follows. Also fixes the pre-existing `tail -f 50` typo (that `50` was being read as a second filename).

**`.github/workflows/windows.yml`** — now actually exercises the sync path on `windows-2022`. See below.

## Verified vs. inferred

**Verified for real on macOS (not just typechecked):**

- **A/B against real rsync over the whole repo tree.** Same source, same exclude list, one destination via `rsync -a --delete`, one via `mirrorDirectory`. Result: **1125 entries each, identical** on paths, types, sizes, permission bits, mtimes, SHA-1 content hashes and symlink targets.
- **The stale-file / `--delete` check, A/B'd too.** Planted the same 8 files in both destinations and re-ran both tools. Both deleted `STALE-ROOT.js`, `plugin/STALE-NESTED.cjs` and `stale-dir/`; both preserved `node_modules/installed-dep/index.js`, `.git/HEAD`, `worker.log`, `workers/leftover.ts` and `plugin/data/memories.db`. Full trees identical afterwards (1135 entries each).
- **The real script end-to-end**, `npm run build` then `npm run sync-marketplace` with `HOME` pointed at a sandbox: marketplace populated, `bun install` ran in it (656 packages), version cache dir synced, `bun install` ran there too, `Sync complete!`. A planted `STALE-FROM-OLD-BUILD.js` was gone; a planted `node_modules/pre-existing-dep/index.js` and `.git/HEAD` survived; `workers/` and `.claude/settings.local.json` correctly excluded. Second run: `0 copied, 0 stale removed` — idempotent, like rsync.
- **`worker-logs.cjs`**: prints exactly 50 lines starting at `line 11` from a 60-line log (byte-for-byte what `tail -n 50` did); `--follow` picked up an appended line; missing log exits 1 with the path in the message.
- **`restart-marketplace-worker.cjs`**: runs `npm run worker:restart` with cwd set to the marketplace dir; exits 1 with a clear message when the marketplace is absent.
- Full suite: 2622 pass / 18 skip / 1 fail. The one failure is `mutation-sites.test.ts > adoptMergedWorktrees emits remap_project` timing out at 5000ms under full-suite load; it passes in isolation and is unrelated to this diff (no worker/sync code is touched). `npm run typecheck` clean, `lint:hook-io` and `lint:spawn-env` clean.

**Inferred, not run — I cannot execute Windows locally:**

- That `fs.copyFileSync` / `rmSync` / `utimesSync` / `readdirSync` behave the same on NTFS. This is documented cross-platform Node API behaviour, but it is inference.
- That `os.homedir()` reads `USERPROFILE` on Windows (documented) — the CI step relies on it to sandbox the sync.
- That `execSync('bun install' | 'npm run worker:restart')` resolves `bun.cmd` / `npm.cmd` via cmd.exe. `execSync` always goes through a shell, so this should hold, but it is inferred. Note `execFileSync` would *not* work here without `shell: true` (Node ≥18 refuses `.cmd`), which is why `execSync` is kept.
- Symlink preservation on Windows. The tree has 2 symlinks (`openclaw/skills/*/SKILL.md`); Git for Windows checks these out as plain files by default, so the symlink branch is not expected to run there.

**To convert inference into evidence, `windows.yml` now:**

- runs `tests/scripts/mirror-dir.test.ts` on `windows-2022` (12 tests: copy, delete, protection, anchoring, dir-only patterns, depth matching, type changes, idempotence) — real NTFS paths;
- runs **the real `scripts/sync-marketplace.cjs`** against a scratch `USERPROFILE`, then asserts the marketplace got populated, a planted stale file was deleted, a planted `node_modules` dep survived, and `workers/` was excluded;
- runs `scripts/worker-logs.cjs` against a planted 60-line log and asserts 50 lines starting at `line 11`;
- adds `scripts/**` and `tests/scripts/**` to the path filter, and raises the job timeout 25 → 35 min to cover the two `bun install`s the sync performs.

This gap — the sync path never being executed on Windows CI — is the same gap that let the original Windows bugs ship.

## Tests

`tests/scripts/` already had a precedent (`export-memories.test.ts`, `pr-babysit-status.test.ts`), so this follows it: `tests/scripts/mirror-dir.test.ts`, 12 tests, real temp dirs. Two of them encode the rsync behaviours that are easy to regress — stale nested files get deleted, and excluded destination paths do not.

## Deliberately out of scope

Two `package.json` scripts are still bash-only. Neither is in the build → sync → restart → logs loop a Windows dev/tester hits:

- `translate:all` — uses `&` job control and `wait`.
- `e2e:server:docker` — `bash scripts/e2e-server-docker.sh`, needs bash and Docker anyway.

Also unchanged: `build:cli-binary` writes an extensionless `--outfile`; it is not part of `npm run build` and is not in the loop.

## Blocks

Windows users cannot test #3644, #3647, #3648 or #3649 from source until this lands.


thedotmack:
Both P1s reproduced and fixed. Neither was a nitpick — I'm not pushing back on either.

## P1 #1 — stale permissions on the quick-check path

Confirmed against the reference tool before fixing. `rsync -a` **does** reconcile mode on a pass where content is unchanged:

```
[pass3-mode] source mode 644 -> 755 (size and mtime untouched)
[pass3-mode] dest mode now: rsync=755 mirror=644     <- pre-fix
[pass3-mode] dest mode now: rsync=755 mirror=755     <- fixed
```

Re-auditing the full `-a` set rather than only the reported bit found a **second, larger gap**: directory permissions and directory mtimes were never applied at all. My original A/B could not see it because the snapshot recorded every directory as the literal string `'dir'`. On the repo tree that was **205 diverging entries on every sync** — the reviewer's point about metadata-weak fixtures was broader than the reported symptom.

### `rsync -a` (= `-rlptgoD`) attribute coverage

| Attribute | Status |
|---|---|
| `-r` recursion | reconciled |
| `-l` symlinks as symlinks | reconciled |
| `-p` permissions | **now reconciled on the fast path** — full `0o7777`, so setuid/setgid/sticky are included, on files *and* directories |
| `-t` mtimes | reconciled on files, directories and symlinks (symlinks via `lutimesSync`; verified rsync preserves symlink mtimes) |
| `-g` group | **not** reconciled — rsync itself can only apply it as root; meaningless on Windows |
| `-o` owner | **not** reconciled — same reason |
| `-D` devices/specials | **not** reconciled — root-only, and absent from a source checkout |
| `-H`/`-A`/`-X` hardlinks, ACLs, xattrs | out of scope for **both** tools; `-a` does not imply them |

Directories are reconciled *after* their children: writing children bumps the destination mtime, and tightening permissions first would lock the mirror out of its own target.

## P1 #2 — log tail exhausted the heap

Reproduced exactly. 64 MiB log, `--max-old-space-size=16`:

```
OLD exit=134     stdout lines: 0     <- SIGABRT, no output at all
NEW exit=0       stdout lines: 50    <- lines 1139951..1140000
```

The fixed output is byte-identical to real `tail -n 50` on the same 64 MiB file (`diff` clean). It now walks backwards in 64 KiB chunks until it has one newline more than it needs — which also means the chunk boundary is discarded along with the partial line in front of it, so a multi-byte character split across chunks can't reach the output. `--follow` still works (re-verified by appending to a live log) and now seeks from the size it already stat'd rather than re-deriving it from a slurped string.

## Tests strengthened so this class can't be missed again

The A/B snapshot now records **directory mode+mtime and symlink mtime**, and runs metadata-only passes: chmod with size and mtime held constant, mtime-only with size constant, and directory-only.

Evidence they're real gates — the same A/B against the **pre-fix** mirror:

```
[pass1-cold] DIFFERENCES (206)     [pass2-idle] DIFFERENCES (206)  NOT IDEMPOTENT
[pass3-mode] DIFFERENCES (207)     [pass4-mtime] DIFFERENCES (206)
[pass5-dir]  DIFFERENCES (206)     PARITY BROKEN
```

and against the fix:

```
[pass1-cold] IDENTICAL   [pass2-idle] IDENTICAL   [pass3-mode] IDENTICAL
[pass4-mtime] IDENTICAL  [pass5-dir]  IDENTICAL   ALL PASSES IDENTICAL
```

Unit tests, run against the pre-fix code to confirm they fail:

- `reconciles permissions when only the mode changed` — expected 420 (`0o644`), got 493 (`0o755`) — the stale executable bit
- `reconciles directory permissions` — expected 448 (`0o700`), got 493
- `preserves directory modification times` — expected 1577840460, got 1787160241
- `is a complete no-op when nothing changed`
- `tails a log far larger than the heap it is given` — `expect(status).toBe(0)` received `null` (killed by SIGABRT)

**One flaw I found in my own first attempt at the heap test:** it spawned `process.execPath`, which under `bun test` is *bun* — bun ignores `--max-old-space-size`, so the test passed against the broken implementation and proved nothing. It now spawns `node` explicitly (which is also how `npm run worker:logs` actually runs it), and fails correctly against the pre-fix code.

## Original verification re-run after these changes

- 1125-entry A/B vs real rsync — identical, now including directory and symlink metadata
- stale-deletion A/B — all 8 planted paths agree with rsync (3 deleted, 5 protected), full trees identical at 1136 entries each
- end-to-end sandboxed-`HOME` sync — planted stale file deleted, planted `node_modules` dep and `.git` preserved, `workers/` excluded
- idempotence — back-to-back mirror runs are `{copied: 0, metadata: 0, deleted: 0}`

On idempotence, one honest correction to my earlier claim: the *cache* leg reports ~8 metadata reconciles per `npm run sync-marketplace`. That is not mirror churn — `bun install` runs against that directory *after* the mirror and bumps `.bin` directory mtimes. `rsync -a --delete --itemize-changes --dry-run` on the same pair reports it would reset the same kind of entries (`.d..t.... node_modules/.bin/`), so both tools behave identically here. Back-to-back mirror runs with no `bun install` between them are a clean no-op.

Full suite: **2632 pass, 0 fail**. Typecheck, `lint:hook-io`, `lint:spawn-env` clean. Windows CI now also runs `tests/scripts/worker-logs.test.ts`, and the PowerShell log assertion tolerates CRLF (`Set-Content` writes CRLF, so that step covers a Windows-native line ending the LF-based bun test does not).

Still can't run Windows locally — the NTFS behaviour of these fs calls and `os.homedir()` reading `USERPROFILE` remain inferred, which is what the CI steps exist to settle.

stantheman0128:
Support for the leaf: npm run build-and-sync without rsync is a real Windows hole. I will treat #3661 as the installable preview of this plus the other four.

I did not run mirror-dir.cjs on this box. The maintainer note that NTFS and USERPROFILE stay inferred is exactly why a Win11 doctor run on the rollup would still be useful later.

tests/scripts/worker-logs.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    const result = run(['--max-old-space-size=16']);
+    const lines = outputLines(result.stdout);
+
+    expect(result.status).toBe(0);
+    expect(lines.length).toBe(50);
+    expect(lines[49].startsWith(`line ${line} `)).toBe(true);
+    expect(lines[0].startsWith(`line ${line - 49} `)).toBe(true);
+  }, 60000);
+});
```
yogeshmunda077-droid: .

PR #3666, opened 2026-08-20
fix(transcripts): stop reparenting ~user/ watch paths under the current home

ArcticFox2029:
### Symptom

A transcript watch whose path names another user's home — `~alice/transcripts` — is silently rewritten to `<your home>/alice/transcripts`. The watcher then watches a directory that does not exist, ingests nothing, and reports no error. The same happens to any path whose first segment merely starts with `~`, such as a directory named `~backup`.

### Root cause

`expandHomePath` in `src/services/transcripts/config.ts` matched **any** leading `~` and sliced one character:

```ts
if (inputPath.startsWith("~")) {
  return join(homedir(), inputPath.slice(1));
}
```

`shared/paths.ts` already has the correct implementation, `expandHome`, which handles `~` and `~/` only and carries a comment explaining that `~user/...` is deliberately left alone because resolving another user's home is out of scope and platform-dependent. So the repository had already decided what the right behaviour is; this one copy just disagreed with it.

The fix deletes the divergent copy and delegates. `expandHomePath` is kept as the name its five callers already import, so the diff stays the size of the defect rather than becoming a rename sweep.

### Verified against the real source, before and after

Loaded `src/services/transcripts/config.ts` itself (not a paraphrase) under Node with a `.js`→`.ts` resolve hook, since `bun` was not available on this machine:

```
=== on main ===
PASS  "~"                     -> "/Users/…"
PASS  "~/.codex/sessions"     -> "/Users/…/.codex/sessions"
FAIL  "~alice/transcripts"    -> "/Users/…/alice/transcripts"      want "~alice/transcripts"
FAIL  "~backup/rollout.jsonl" -> "/Users/…/backup/rollout.jsonl"   want "~backup/rollout.jsonl"
PASS  "/var/log/codex.jsonl"  -> "/var/log/codex.jsonl"
PASS  "relative/path.jsonl"   -> "relative/path.jsonl"
PASS  ""                      -> ""
2 FAILED

=== with this patch ===
7/7 checks passed
```

The three cases are added to `tests/transcripts/config.test.ts`.

### Against the rubric

No guard, no retry, no fallback, no new state, no env-var gate — a deletion of a duplicated helper plus the tests. I could not run `bun test` locally (no bun on this machine), so CI is the first full-suite run; the equivalence above was checked directly against the shipped source instead.

### DIFF
--- a/src/services/transcripts/config.ts
+++ b/src/services/transcripts/config.ts
@@ -1,7 +1,7 @@
 import { existsSync, readFileSync, writeFileSync, mkdirSync } from 'fs';
 import { homedir } from 'os';
 import { join, dirname } from 'path';
-import { paths } from '../../shared/paths.js';
+import { expandHome, paths } from '../../shared/paths.js';
 import type { TranscriptSchema, TranscriptWatchConfig } from './types.js';
 
 export const DEFAULT_CONFIG_PATH = paths.transcriptsConfig();
@@ -55,10 +55,7 @@ export function filterNativeHookBackedCodexWatches(
 
 export function expandHomePath(inputPath: string): string {
   if (!inputPath) return inputPath;
-  if (inputPath.startsWith('~')) {
-    return join(homedir(), inputPath.slice(1));
-  }
-  return inputPath;
+  return expandHome(inputPath);
 }
 
 export function loadTranscriptWatchConfig(path = DEFAULT_CONFIG_PATH): TranscriptWatchConfig {
--- a/src/services/transcripts/config.ts
+++ b/src/services/transcripts/config.ts
@@ -1,7 +1,7 @@
 import { existsSync, readFileSync, writeFileSync, mkdirSync } from 'fs';
 import { homedir } from 'os';
 import { join, dirname } from 'path';
-import { expandHome, paths } from '../../shared/paths.js';
+import { expandTilde, paths } from '../../shared/paths.js';
 import type { TranscriptSchema, TranscriptWatchConfig } from './types.js';
 
 export const DEFAULT_CONFIG_PATH = paths.transcriptsConfig();
@@ -55,7 +55,12 @@ export function filterNativeHookBackedCodexWatches(
 
 export function expandHomePath(inputPath: string): string {
   if (!inputPath) return inputPath;
-  return expandHome(inputPath);
+  // expandTilde rather than expandHome: both leave `~user/...` alone, which is the point of this
+  // change, but only expandTilde also accepts the Windows `~\` form the replaced inline version
+  // handled by slicing one character off any leading tilde. These are filesystem paths with no
+  // separator-normalising pass anywhere downstream, so dropping `~\` would leave a configured
+  // transcript path unresolved and the config simply failing to load.
+  return expandTilde(inputPath);
 }
 
 export function loadTranscriptWatchConfig(path = DEFAULT_CONFIG_PATH): TranscriptWatchConfig {
--- a/tests/transcripts/config.test.ts
+++ b/tests/transcripts/config.test.ts
@@ -137,6 +137,15 @@ describe('expandHomePath', () => {
     expect(expandHomePath('~/.codex/sessions')).toBe(join(homedir(), '.codex/sessions'));
   });
 
+  it('expands the Windows ~\\ form as well', () => {
+    // The replaced inline version sliced one character off any leading tilde, so a config
+    // written on Windows as `~\\.codex\\sessions` resolved. Routing through a helper that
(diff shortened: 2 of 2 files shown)


ArcticFox2029:
One interaction worth flagging, since this function feeds a security check and a reviewer will reasonably ask about it.

`services/transcripts/processor.ts:358` expands `watch.context.path` with `expandHomePath`, then guards the result:

```ts
const resolvedAgentsPath = path.resolve(agentsPath);
const allowedRoots = [path.resolve(cwd), path.resolve(DATA_DIR)];
const isPathSafe = allowedRoots.some(root =>
  resolvedAgentsPath.startsWith(root + path.sep) || resolvedAgentsPath === root);
```

The guard runs **after** expansion and is unchanged by this PR, so no traversal becomes possible that was not possible before — a `~user/` path still has to land inside `cwd` or `DATA_DIR` to be read.

What does change: `~alice/AGENTS.md` used to expand to `<home>/alice/AGENTS.md`, which normally falls outside both roots and was rejected with the SECURITY warning. It now stays literal and resolves relative to the process cwd, so if a file of that literal name exists inside the project it will be accepted. That is the honest reading of the path the user wrote, it is inside an allowed root, and the check still decides — but it is a behaviour change in that one case rather than a pure no-op, so it should not arrive as a surprise.

ArcticFox2029:
Correct finding, and the suggested `expandTilde` pointer is the right one — pushed.

I took it slightly differently from the literal suggestion, and the difference is worth stating. Rather than widen `expandHome` to accept `~\`, this call site now uses **`expandTilde`**, which already accepts all three forms and already leaves `~user/...` alone:

```ts
export function expandTilde(filePath: string, home: string = homedir()): string {
  if (filePath === '~') return home;
  if (filePath.startsWith('~/') || filePath.startsWith('~\\')) return join(home, filePath.slice(2));
  return filePath;
}
```

That satisfies this PR's purpose — the `~alice/transcripts` reparenting — without losing the Windows form, and nothing new is written. The repository already had two helpers where one is strictly more capable; this call site was simply on the narrower one. Widening `expandHome` instead would change a shared contract for every caller in order to fix one, and `resolveDataDir` has no need of it.

Verified against the real helpers:

```
                     input                       before      after
bare ~               "~"                         home        home
posix prefix         "~/.codex/sessions"         <home>/…    <home>/…
WIN bare             "~\\"                       unexpanded  home
WIN prefix           "~\\.codex\\sessions"       unexpanded  <home>/.codex\sessions
~user left alone     "~alice/transcripts"        literal     literal
~name left alone     "~backup/rollout.jsonl"     literal     literal
absolute / relative / empty                      unchanged   unchanged
```

Tests added for `~\` and `~\.codex\sessions` in `tests/transcripts/config.test.ts`; both are unexpanded without the change.

Two notes so nothing is overstated:

- I could not run `bun test` — no bun on this machine — so the nine cases above came from running the actual `expandTilde` / `expandHome` bodies under node with the real `homedir()`, not from the suite going green.
- The sibling PR #3669 has the same class of regression and is fixed there by normalising separators *before* expansion instead of swapping helpers. That is deliberate: `globToRegex` already normalises separators one line down, so a glob pattern reaches the expander in POSIX form anyway, and moving an existing line is smaller than changing which helper that file depends on. Here there is no normalising pass at all, so the helper has to accept the form.

ArcticFox2029:
Checking in once on this, and on #3669 and #3670 beside it — no rush intended, and I will not chase them again.

All three are independent, none touches the other's files, and each is a single-purpose fix with the reproduction in its own description:

- **#3666** — `~user/` transcript paths were reparented under the current home instead of staying literal.
- **#3669** — `~` was expanded by string concatenation, so `~\projects\*` never matched on Windows. The before/after table in that PR shows the four Windows patterns that start matching and the five POSIX ones that are unchanged.
- **#3670** — `hooks.json` written with a UTF-8 BOM could not be parsed, which broke install, status and removal for anyone whose editor writes one.

They are all still mergeable and there are no checks pending. If any of them has gone stale against `main` I am happy to rebase — just say which. If you would rather they were one PR, or would rather not take them at all, that is a fine answer too and I would rather hear it than leave them sitting.

thedotmack:

Narrow: delete the divergent `startsWith('~')` expander that reparents `~user/` under the current home; delegate to shared `expandTilde`/`expandHome`. Bug still on `main`. Rebasing and shipping via a cherry-pick vehicle. No version bump / no npm publish.

thedotmack:
Rebase vehicle: #4012. Will squash-merge that once CI is green, then close this fork PR.

thedotmack:
Superseded by squash-merged #4012 (rebase onto current `main`, full CI green). Closing this CONFLICTING original. Thank you for the patch.

PR #3669, opened 2026-08-20
fix(project-filter): expand ~ with a separator instead of string concatenation

ArcticFox2029:
### Symptom

An entry in `CLAUDE_MEM_EXCLUDED_PROJECTS` that starts with `~` but is not the `~/` form silently excludes the wrong thing. `~backup/*` expands to `<home>backup/*` — the home directory concatenated onto the pattern with no separator — so it

- **excludes `/Users/you**backup**/thing`**, a sibling of your home directory that nobody named, and
- **does not exclude** a real `~backup` directory, which is what the entry was written for.

Both halves are silent. An exclusion list that quietly excludes the wrong project is the failure mode this setting exists to prevent.

### Root cause

`globToRegex` in `src/utils/project-filter.ts`:

```ts
let expanded = pattern.startsWith('~')
  ? homedir() + pattern.slice(1)
  : pattern;
```

Two defects in three lines. It matches **any** leading `~`, not just `~` and `~/`; and it joins by **string concatenation**, so the separator only survives because `~/x`'s slice keeps the leading slash. Anything else loses it.

`shared/paths.ts` already exports `expandHome`, which handles `~` and `~/` and documents why `~user/...` is deliberately left alone. This was the fourth copy of tilde expansion in the tree and the only one that concatenates.

### Verified against the real source, before and after

Loaded `src/utils/project-filter.ts` itself under Node with a `.js`→`.ts` resolve hook (`bun` is not installed on this machine):

```
=== on main ===
PASS  excluded("<home>/projects/secret", "~/projects/*") = true   want true
PASS  excluded("<home>/scratch",         "~/scratch")    = true   want true
PASS  excluded("/srv/other",             "~/projects/*") = false  want false
PASS  excluded("/opt/~backup/thing",     "~backup/*")    = false  want false
FAIL  excluded("<home>backup/thing",     "~backup/*")    = true   want false
1 FAILED

=== with this patch ===
5/5 checks passed
```

The last case is added to `tests/utils/project-filter.test.ts`, beside the existing `~` block, along with the literal-`~backup`-directory case.

### Against the rubric

Correct expansion replacing incorrect expansion, and one divergent copy deleted. No guard, no retry, no fallback, no new state, no env-var gate. The `homedir` import goes with it because nothing else in the file used it.

Companion to #3666, which fixed the same `startsWith('~')` over-match in `services/transcripts/config.ts`. They are independent files and either can merge without the other.

I could not run `bun test` locally, so CI is the first full-suite run; the equivalence above was checked directly against the shipped source instead.

### DIFF
--- a/src/utils/project-filter.ts
+++ b/src/utils/project-filter.ts
@@ -4,8 +4,15 @@ import { expandHome } from '../shared/paths.js';
 import { logger } from './logger.js';
 
 function globToRegex(pattern: string): RegExp {
-  let expanded = expandHome(pattern);
+  // Separators first, then home expansion. A pattern written on Windows as `~\projects\secret`
+  // is home-relative, but `expandHome` only recognises `~` and `~/` — resolving another user's
+  // home is deliberately out of its scope, so it cannot widen its `~` check without ambiguity.
+  // Expanding before normalising left that pattern as a literal `~/projects/secret`, which
+  // matches no absolute path, so a configured exclusion silently stopped excluding.
+  let expanded = expandHome(pattern.replace(/\\/g, '/'));
 
+  // Again on the way out: on Windows `homedir()` itself returns backslashes, so the
+  // expanded path needs normalising even when the pattern arrived POSIX-form.
   expanded = expanded.replace(/\\/g, '/');
 
   let regex = expanded.replace(/[.+^${}()|[\]\\]/g, '\\$&');
--- a/src/utils/project-filter.ts
+++ b/src/utils/project-filter.ts
@@ -1,12 +1,10 @@
 
-import { homedir } from 'os';
 import { basename } from 'path';
+import { expandHome } from '../shared/paths.js';
 import { logger } from './logger.js';
 
 function globToRegex(pattern: string): RegExp {
-  let expanded = pattern.startsWith('~')
-    ? homedir() + pattern.slice(1)
-    : pattern;
+  let expanded = expandHome(pattern);
 
   expanded = expanded.replace(/\\/g, '/');
 
--- a/tests/utils/project-filter.test.ts
+++ b/tests/utils/project-filter.test.ts
@@ -51,6 +51,17 @@ describe('Project Filter', () => {
         expect(isProjectExcluded(`${home}/projects/secret`, '~/projects/*')).toBe(true);
       });
 
+      it('expands a Windows-form ~\\ pattern', () => {
+        const home = homedir();
+        // `expandHome` recognises `~` and `~/`, not `~\`. Expanding before normalising
+        // separators left a Windows-written exclusion as a literal `~/...`, which matches no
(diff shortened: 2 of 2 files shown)


ArcticFox2029:
The Windows finding is correct, and I've pushed a fix rather than argue with it.

I reproduced it against the real functions before changing anything:

```
                         pattern                    before   after
NEW  win exact           "~\\projects\\secret"      false    true
NEW  win glob            "~\\projects\\*"           false    true
NEW  win bare            "~\\"                      false    true
KEEP posix exact         "~/secret"                 true     true
KEEP posix glob          "~/projects/*"             true     true
KEEP ~name not glued     "~backup/*"                false    false
KEEP ~name literal       "~backup/*"                false    false
KEEP unrelated           "~/secret"                 false    false
KEEP bare ~              "~"                        true     true
```

**What happened.** The old inline expansion concatenated on any leading `~`, so `~\projects\secret` became `<home>\projects\secret` and the separator normalisation one line down made it match. `expandHome` recognises `~` and `~/` only, and widening its `~` check is not free — `~user/...` is deliberately out of its scope, so `~\` cannot be added without deciding what `~alice\x` means. The pattern therefore passed through untouched and normalised into a literal `~/projects/secret`, which matches no absolute path.

So the entry stopped excluding, quietly. Fails open on a deny-list, which is the wrong direction to be wrong in — worth a test rather than a comment.

**The fix** is the ordering, not a new branch: normalise separators *before* `expandHome` instead of after. The normalisation already existed; only its position moved. The outbound pass stays, because `homedir()` returns backslashes on Windows even when the pattern arrived POSIX-form.

```ts
let expanded = expandHome(pattern.replace(/\\/g, '/'));
expanded = expanded.replace(/\\/g, '/');
```

Nothing was added to `src/shared/paths.ts`. The helper's narrow contract is the reason it is safe to share, and this is a `project-filter` concern: it is the only caller that already normalises separators, so it is the only one where `~\` is unambiguous.

Regression test added at `tests/utils/project-filter.test.ts` covering `~\projects\secret`, `~\projects\*` and bare `~\`. All three return `false` without the change.

I could not run `bun test` here — no bun on this machine — so the nine cases above were checked by running the actual `globToRegex` / `isProjectExcluded` logic under node with the real `homedir()`. Worth saying plainly rather than implying the suite went green.

thedotmack:

Narrow: replace concatenating `~` expansion with shared `expandHome`, normalize separators first so `~\` exclusions work. Bug still on `main`. Rebasing and shipping via a cherry-pick vehicle. No version bump / no npm publish.

thedotmack:
Rebase vehicle: #4011. Will squash-merge that once CI is green, then close this fork PR.

thedotmack:
Superseded by squash-merged #4011 (rebase onto current `main`, full CI green). Closing this CONFLICTING original. Thank you for the patch.

PR #3670, opened 2026-08-20
fix(windsurf): read hooks.json with the BOM-tolerant parser

ArcticFox2029:
### Symptom

On Windows, a valid `~/.codeium/windsurf/hooks.json` is reported as corrupt and both install and uninstall stop:

```
Corrupt hooks.json at <path>, refusing to overwrite
```

The file is not corrupt. It carries a UTF-8 BOM, which PowerShell 5.1 and several Windows editors add when they rewrite JSON. Windsurf itself reads it fine.

### Root cause

`src/services/integrations/WindsurfHooksInstaller.ts` read that file with a raw `JSON.parse`, in three places — the install merge, `removeClaudeMemHookEntries`, and the verify pass. Node decodes a BOM to `U+FEFF` at offset 0, `JSON.parse` rejects it, and the surrounding `catch` converts the throw into the "corrupt" message above.

This is exactly the failure described in #3013 and fixed for `readJsonSafe` in #3307. `parseJsonWithBom` already backs twenty-odd call sites across the tree; these three were still going around it.

### Scope

Only the three reads of **Windsurf's own** `hooks.json` are changed. `WINDSURF_REGISTRY_FILE` on line 47 is deliberately left as-is — claude-mem writes that file itself, so it can never carry a BOM, and changing it would be hardening rather than a fix.

### Verified

`parseJsonWithBom` loaded from the real source under Node, given the exact input the installer would see:

```
JSON.parse on a BOM'd hooks.json  -> THROWS: Unexpected token '﻿', "﻿{"hooks":"... is not valid JSON
parseJsonWithBom on the same input -> {"hooks":{"afterFileEdit":[]}}
2/2 checks passed
```

`tests/windsurf-hooks-bom.test.ts` asserts three things: the file no longer calls `JSON.parse` on that path, all three reads go through the shared reader, and the shared reader accepts what `JSON.parse` rejects. The first two are source assertions rather than filesystem ones, because the installer writes to a fixed path under the real home directory — `npm-install-windows-hide.test.ts` takes the same approach for the same reason.

### Against the rubric

Correct reader replacing an incorrect one, at three call sites in one file. No guard, no retry, no fallback, no new state, no env-var gate, and nothing added to the catch block — the catch stays exactly as it was, and now only fires on files that really are corrupt.

I could not run `bun test` locally (no bun on this machine), so CI is the first full-suite run.

### DIFF
--- a/src/services/integrations/WindsurfHooksInstaller.ts
+++ b/src/services/integrations/WindsurfHooksInstaller.ts
@@ -4,6 +4,7 @@ import { homedir } from 'os';
 import { existsSync, readFileSync, writeFileSync, unlinkSync, mkdirSync, renameSync } from 'fs';
 import { logger } from '../../utils/logger.js';
 import { getWorkerHost, getWorkerPort } from '../../shared/worker-utils.js';
+import { parseJsonWithBom } from '../../shared/atomic-json.js';
 import { DATA_DIR } from '../../shared/paths.js';
 import { getBunAbsolutePath as findBunPath, getWorkerServiceAbsolutePath as findWorkerServicePath } from './install-paths.js';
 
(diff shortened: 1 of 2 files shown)


thedotmack:

Narrow: three `JSON.parse` reads of Windsurf's `hooks.json` → existing `parseJsonWithBom`. Bug still on `main`. Rebasing onto current `main` and shipping via a cherry-pick vehicle. No version bump / no npm publish.

thedotmack:
Rebase vehicle: #4010. Will squash-merge that once CI is green, then close this fork PR.

thedotmack:
Superseded by squash-merged #4010 (rebase onto current `main`, full CI green). Closing this CONFLICTING original. Thank you for the patch.

Issue #3610, opened 2026-08-16
[plan-22] Chroma Sidecar Contract — host-safe spawn with pinned deps and a clean env, single-writer upsert sync, honest fallback (or retire the sidecar)


thedotmack:
# [plan-22] Chroma Sidecar Contract — host-safe spawn with pinned deps and a clean env, single-writer upsert sync, honest fallback (or retire the sidecar)

> **Tracker:** #3610 · **Design doc:** `plans/22-chroma-sidecar-contract.md` · **Series:** plan-14 #3602 · plan-15 #3603 · plan-16 #3604 · plan-17 #3605 · plan-18 #3606 · plan-19 #3607 · plan-20 #3608 · plan-21 #3609 · plan-22 #3610 · plan-23 #3611

## Defect

The vector store is a Python sidecar (`uvx … chroma-mcp`) driven over MCP stdio, and every layer between the worker and the index is untested against the environments users actually have. The result is a semantic search that is silently degraded (FTS5 fallback, confident-empty results, frozen watermarks) for weeks or months, and a store that grows without bound:

- **Spawn args are host-unsafe:** `chroma-mcp==0.2.6` as a bare positional (rejected by uvx < 0.5.31; `@`/`--from` needed), `--with onnxruntime>=1.20 --with protobuf<7` routed through `cmd.exe /c` where `>`/`<` are redirects, then double-quoted through the spawn layer so uvx sees literal quotes; the prewarm path re-implements the spawn without the transport's `.cmd` handling; the launching shell's `VIRTUAL_ENV`/`PYTHONPATH` leak into the child so `numpy` resolves from a foreign venv; the `onnxruntime>=1.20` constraint is not overridable (no wheels for macOS 12 / cp313 mismatch); `chromadb` is unpinned so an engine bump segfaulted a persisted store for 2.5 months with zero log signal; the preflight ignores `CLAUDE_MEM_CHROMA_UVX_PATH`; the connect timeout (30 s, later a 120 s prewarm) is shorter than a cold `uvx` materialization, and killing uvx mid-build leaves 100+ GB of `uv` temp dirs and reruns the cold build forever.
[...]

PR #3675, opened 2026-08-20
perf(chroma): post-filter non-selective where clauses instead of pushing them to chroma

godofcpu:
Chroma brute-forces the metadata-matched candidate set rather than walking the HNSW graph, so a `where` clause costs roughly 30-50us per **matched** document while an unfiltered query is flat regardless of `n_results`. Measured on a 347k-doc collection:

| query | latency |
|---|---|
| unfiltered, `n_results` 100..2000 | 0.15 - 0.29s (flat) |
| `where {project: …}` — 7.7k matched | 0.39s |
| `where {project: …}` — 54k matched | 1.67s |
| `where {project: …}` — 208k matched | **6.10s** |

`SearchManager` puts `project` into the where clause so large projects cannot crowd small ones out of the top-N, which is correct — but on a corpus where one project is 60% of the documents it makes every search on that project a near-full scan.

## Change

`queryChroma` over-fetches unfiltered (20x, capped 2000) and applies the clause client-side, reusing the existing `deduplicateQueryResults` for both paths. Over-fetching is close to free; pushing a non-selective filter into chroma is not.

When the over-fetch yields fewer than `limit` survivors the filter *was* selective — exactly the crowded-out case the where clause exists for — so it falls back to a real filtered query, which is cheap precisely because that filter is selective. Both ends keep their behaviour; only the expensive middle changes.

A per-clause memo records that outcome, because the fast path is a net loss for a selective filter (it pays both queries). Without it a 2%-of-corpus project regressed 0.39s → 4.32s. One miss is enough to learn; an entry clears as soon as a clause fills the limit, so it self-corrects as the corpus grows. Bounded at 256 entries.

## Result

End-to-end search latency, same corpus:

| project size | before | after |
|---|---|---|
| 7.7k docs | 0.39s | 0.25s |
| 5.7k docs | 1.10s | 0.23s |
| 54k docs | 1.67s | 0.32s |
| 208k docs | 6.73s | 0.30s |

This also matters because the MCP client's worker timeout is 3s, so semantic search was failing outright on the largest project — the index was healthy, the query just took longer than the client would wait.

## Correctness

The client-side predicate is deliberately narrow: literal equality and `$eq`, combined with `$and`. Anything else (`$in`, `$or`, `$ne`, ranges) returns null and the query goes to chroma unchanged. A wrong client-side filter would silently drop results, which is worse than a slow query.

## Testing

9 new tests in `tests/services/sync/chroma-query-postfilter.test.ts`, following the existing convention of mocking `ChromaMcpManager` and restoring the real module in `afterAll` so it cannot leak into the sibling chroma tests. Full suite run against a baseline of this branch point: no new failures.

---

*Note: this PR previously also carried a `chroma-mcp` subprocess leak fix. That turned out to be already fixed on `main`, and more thoroughly — upstream kills the whole process tree, since `transport.close()` only signals the direct `uvx` child and not the python grandchild. It has been dropped and the branch rebased onto current `main`, so this is now the perf change alone.*

### DIFF
--- a/src/services/sync/ChromaSync.ts
+++ b/src/services/sync/ChromaSync.ts
@@ -95,10 +95,75 @@ interface StoredUserPrompt {
   platform_source: string;
 }
 
+/**
+ * Chroma brute-forces the metadata-matched candidate set instead of walking the
+ * HNSW graph, so a `where` clause costs ~30-50us per MATCHED document while an
+ * unfiltered query is flat regardless of n_results. Measured on a 347k-doc
+ * collection: unfiltered 0.15-0.29s for n_results 100..2000, versus 6.10s for
+ * `where {project: <60% of corpus>}`.
+ *
+ * So over-fetch unfiltered and filter here. Over-fetching is close to free;
+ * pushing a non-selective filter into chroma is not.
+ */
+const CHROMA_OVERFETCH_FACTOR = 20;
+const CHROMA_OVERFETCH_CAP = 2000;
+
+/**
+ * Build a client-side equivalent of a chroma `where` clause, or null if the
+ * clause uses anything we do not evaluate identically to chroma.
+ *
+ * Deliberately narrow: literal equality and $eq, combined with $and. Any
+ * operator we are not certain we match exactly ($in, $or, $ne, ranges) returns
+ * null so the query goes to chroma unchanged. A wrong client-side filter would
+ * silently drop results, which is far worse than a slow query.
+ */
+function buildClientSidePredicate(
+  where: Record<string, any>
+): ((metadata: Record<string, any>) => boolean) | null {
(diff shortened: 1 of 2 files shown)


thedotmack:
Routing to **#3610 (plan-22, Chroma Sidecar Contract)** — the branch name says subprocess leak but the change is a read-path one: over-fetch unfiltered and post-filter, because a `where {project}` clause makes Chroma brute-force the matched set (6.1s at 208k matched docs on the measurements here).

Not closing — it's a genuine, well-measured fix on the sidecar as it exists. But note the fork in the cluster: **#3694** deletes the sidecar entirely and moves vectors into `claude-mem.db`, which would moot this. plan-22 has to pick keep-and-fix or retire before either is worth rebasing, and this one is `CONFLICTING` today.

Issue #1251, opened 2026-02-26
Security Audit: Comprehensive Code Review of claude-mem


CheswickDEV:
I performed a full security audit of this repository (v10.5.2, main branch as of Feb 2026). The analysis covered all source files under `src/`, `plugin/`, `scripts/`, `openclaw/`, and configuration files. Pre-built binaries (`plugin/scripts/claude-mem`, 61 MB) and compiled `.cjs` bundles were excluded.

**Overall Risk Rating: HIGH**

Claude-mem captures and persists extensive data from coding sessions (tool usage, user prompts, code changes, AI-generated summaries). This data can contain confidential business logic, API keys, passwords, and proprietary source code. The **fully unauthenticated HTTP API** on a well-known port (37777) makes all of this accessible to any local process  and to the entire network if the host is misconfigured.

**Findings Overview:**

| Severity | Count |
|----------|-------|
| 🔴 Critical / High | 4 |
| 🟡 Medium | 6 |
| 🟢 Low / Info | 7 |

---

### C-1: Path Traversal via MCP Tools `smart_unfold` and `smart_outline`

**File:** `src/servers/mcp-server.ts`, lines 296–298 and 337–339
**Type:** CWE-22  Path Traversal

The MCP tools `smart_unfold` and `smart_outline` accept a `file_path` parameter from the caller. The path is resolved to an absolute path via `resolve()`, but **no validation is performed** to check whether the resulting path stays within an allowed directory. The file is then read directly:

```typescript
// Line 296-297  no path boundary check
const filePath = resolve(args.file_path);
const content = await readFile(filePath, 'utf-8');
```

**Risk:** An attacker (or a compromised LLM context) that can control MCP tool calls can read arbitrary files on the filesystem  e.g. `~/.ssh/id_rsa`, `~/.claude-mem/.env` (containing API keys), `/etc/passwd`, or any source code files.

**Recommendation:** Validate the resolved path against an allowed base directory:
```typescript
const basePath = resolve(process.cwd());
const filePath = resolve(args.file_path);
if (!filePath.startsWith(basePath + path.sep)) {
  return { content: [{ type: 'text', text: 'Access denied: path outside project' }], isError: true };
}
```

---

### C-2: Entire HTTP API Has No Authentication

**File:** `src/services/worker/http/middleware.ts` (entire file), `src/services/server/Server.ts`
**Type:** CWE-306  Missing Authentication for Critical Function

[...]

PR #3680, opened 2026-08-21
security: close 3 credential-leak paths (unauthenticated /api/settings, MCP path traversal, non-loopback worker binding)

nero-harsh:
## Pain point

Running claude-mem locally, I wanted to harden it before adding it to a project that already handles third-party API credentials (SAP, SharePoint Graph app secrets, etc). Checked it against the community audit in #1251 and found three ways the plugin can still leak local credentials off the machine or to any local process. Re-verified all three against current main (v13.15.3) with a running worker and MCP server, not just a static read of the code.

## Vulnerabilities (verified against v13.15.3)

**1. MCP path traversal (`src/servers/mcp-server.ts`)**
`smart_unfold` and `smart_outline` did `resolve(args.file_path)` with no check the result stayed inside the workspace. Any MCP call could read arbitrary files off disk. Confirmed by asking `smart_unfold` for `~/.ssh/id_rsa` and getting the raw key content back.

**2. Unauthenticated `/api/settings` returning cleartext credentials (`src/services/worker/http/routes/SettingsRoutes.ts`)**
`GET /api/settings` returned every configured secret (Gemini/OpenRouter/Chroma API keys, cloud-sync token, server API keys) in cleartext, with no auth in front of the endpoint. Only 3 of 30+ worker routes (`/api/admin/*`) were gated behind `requireLocalhost`.

**3. Worker could be bound off loopback (`src/services/worker/http/routes/SettingsRoutes.ts`)**
`CLAUDE_MEM_WORKER_HOST` validation accepted `0.0.0.0` and arbitrary IPv4 addresses, so the otherwise unauthenticated API could be exposed to the network.

## Fixes

- `smart_unfold`/`smart_outline` now resolve paths against `process.cwd()` and reject anything outside it, using `realpath` on both sides so an in-workspace symlink can't be used to point at an external file.
- `GET /api/settings` masks known secret keys with an explicit allowlist (Gemini/OpenRouter/Chroma API keys, cloud-sync token, Telegram bot token, server API keys) before the response leaves the process. The raw values on disk are untouched, so functionality is unaffected, only the leaking response changes.
- `CLAUDE_MEM_WORKER_HOST` validation now only accepts `127.0.0.1`, `::1`, or `localhost`.
- `requireLocalhost` is applied globally as the first middleware on the in-plugin worker (which has no auth of its own), instead of just the 3 admin routes. The Docker/Postgres server runtime (`ServerService.ts`) opts out with `restrictToLocalhost: false` since it's intentionally network-reachable and authenticates requests itself.

## Testing

All changes were exercised against a live worker and MCP server, not just `tsc`/build:
- `smart_unfold({file_path: "~/.ssh/id_rsa"})` and a call through an in-workspace symlink pointing outside the workspace are both denied. A normal in-project file still unfolds/outlines fine.
- Set a fake `CLAUDE_MEM_GEMINI_API_KEY` via `POST /api/settings`, then `GET /api/settings` returns a masked value while the real key is still correctly persisted to disk. Posting an unmodified settings object back (a normal viewer save) now succeeds.
- `POST /api/settings` with `CLAUDE_MEM_WORKER_HOST: "0.0.0.0"` is rejected.
- `npm run typecheck` / `npm run build` pass clean.

## Scope

Scoped to the credential-leak surface. Not addressed here (lower severity, tracked separately in #1251): request body size limits, verbose error responses, `~/.claude-mem/` file permissions, and CORS handling for non-browser same-machine callers.

## Update

Fixed the three issues Greptile's review found: the settings redaction regex was also matching `CLAUDE_MEM_CONTEXT_SHOW_READ_TOKENS`/`SHOW_WORK_TOKENS` (boolean display prefs, not secrets) and corrupting them, the workspace containment check didn't canonicalize symlinks, and the global localhost restriction broke the network-reachable server runtime. All three are fixed in the latest commit and re-verified live.


nero-harsh:
Addressed in 520b22d:

- Settings redaction: switched from the `/API_KEY|_TOKEN|SECRET/i` regex to an explicit allowlist of the actual secret keys. The regex was matching `CLAUDE_MEM_CONTEXT_SHOW_READ_TOKENS`/`SHOW_WORK_TOKENS` too, turning `false` into `*alse` on every GET and breaking the next settings save. Verified with a GET then unmodified POST round trip, now returns success.
- MCP symlink escape: `resolveWithinWorkspace` now realpaths both the workspace root and the resolved target before comparing containment, instead of comparing lexical paths. Verified with an in-workspace symlink pointing at a file outside the workspace, now denied.
- Non-loopback server runtime: added a `restrictToLocalhost` option on `ServerOptions` (default true) instead of applying `requireLocalhost` unconditionally in the shared `Server` class. `ServerService.ts` now passes `restrictToLocalhost: false` next to its existing `securityHeaders: true`, since it's already documented as network-reachable in Docker and authenticates requests itself.

All three re-verified against a live worker and MCP server, plus a clean `npm run typecheck` / `npm run build`.

nero-harsh:
Addressed in 4fcceae:

Confirmed the round trip bug: seeded a real Gemini key, did a GET (returns the masked value), posted that same response back unchanged (what a normal settings-form save does), and the real key on disk was overwritten with the masked string.

Fix: every value maskSecretValue produces starts with at least one literal "*", which a real credential will not. handleUpdateSettings now skips writing a secret field if the incoming value starts with "*", treating it as an unchanged placeholder rather than a new key.

Verified: unchanged round trip now leaves the real key on disk untouched, and posting an actually new key still updates it correctly. Also re-verified the earlier fixes (traversal, symlink escape, host validation) are unaffected.

nero-harsh:
Merged latest main to resolve the conflict (touched the same settingKeys list and ServerService.ts). No functional changes beyond the merge — re-verified the credential-preservation round trip, the new CLAUDE_MEM_OPENROUTER_BASE_URL setting, and the loopback-only host validation all still work correctly on the merged tree.

nero-harsh:
Opened #3861 to call out the impact more explicitly (local credentials/PII readable via smart_unfold, plus the fact that these MCP tools are callable by the agent itself, not just a local process). This PR is the fix.

thedotmack:
#3680 cannot be merged as-is (`mergeable_state: dirty` against current `main`). The fork branch also carries stale plugin bundles / plugin.json description reverts, and its original `requireLocalhost` + reject-`0.0.0.0` changes would break Observation TV (`remoteReadOnly`) and the documented Docker bind.

Rebased the three leak-path fixes onto current `main` in a new PR: the continuation of this work. RCA on this thread and on #3861 still stands.

Adaptations vs this PR:
- Settings redaction allowlist updated for secrets that landed after this branch (`CLAUDE_MEM_TV_TOKEN`, `CLAUDE_MEM_PRO_MEMORY_KEY`, `CLAUDE_MEM_REDIS_URL`).
- Masked POST skip is exact-mask match (Greptile P1: a replacement key starting with `*` is no longer silently dropped).
- Host validation still allows `0.0.0.0` / `::` for TV/Docker; arbitrary IPv4 is rejected.
- No global `requireLocalhost` — that would fight the existing TV remote guard.

thedotmack:
Continuation PR with the rebased fixes: https://github.com/thedotmack/claude-mem/pull/3985

thedotmack:
Superseded by #3985.

Issue #3688, opened 2026-08-22
PreToolUse hook returns BLOCKING_ERROR when a Bash command contains ${…}: shell-quote parse() throws inside normalizeInput() (13.15.3)


raka08:
## Summary

`HookAdapter.normalizeInput()` calls `shell-quote`'s `parse()` on the raw `tool_input.command` of every `Bash` PreToolUse event, to enrich the payload with `filePaths` for `cat`/`head`/`tail`/… reads. `parse()` **throws** on `${…}` forms it cannot resolve. The throw is not caught by the enrichment, is not one of the three classes `runHook()` fails open on, and therefore falls through to the generic handler — which logs `Hook error:` and returns `HookExitCode.BLOCKING_ERROR`.

Net effect on a host whose `PreToolUse` matcher includes `Bash` (the shipped `hooks/codex-hooks.json` does): **a best-effort convenience feature turns an ordinary shell command into a blocking hook error, and the tool call is never captured.**

## Environment

| | |
|---|---|
| claude-mem | **13.15.3** (also seen on 13.15.2) |
| OS | macOS 26.6.2, arm64 |
| install | plugin, `CLAUDE_MEM_RUNTIME=worker` |
| host | Codex (`hooks/codex-hooks.json`) |
| node | v22.23.1 |
| shell-quote | 1.8.4 (bundled in `node_modules/`) |

## Code path

`scripts/worker-service.cjs` (bundled):

```js
// normalizeInput()
if (event === "PreToolUse" && toolName) {
  const paths = collectFilePaths(toolName, toolInput, cwd);   // <- rre()
  ...
}

// collectBashReadPaths() — the Bash branch
function collectBashReadPaths(toolInput, cwd) {
  const command = coerceCommand(toolInput?.command);
  if (!command) return [];
  const parsed = shellQuote.parse(command);   // <-- throws "Bad substitution: …"
  ...
}
```

and the catch in `runHook()`:

```js
catch (err) {
  if (err instanceof AdapterRejection) { ...; return SUCCESS }   // fail open
  if (isTranscriptMissing(err))        { ...; return SUCCESS }   // fail open
  if (isWorkerUnreachable(err))        { ...; return SUCCESS }   // fail open
  logger.error("HOOK", `Hook error: ${err.message}`, ...);
  return HookExitCode.BLOCKING_ERROR;                            // <-- lands here
}
```

Also relevant: `hooks/codex-hooks.json` registers

```
[...]

PR #3646, opened 2026-08-19
fix: upgrade shell-quote to 1.9.0 (CVE-2026-13311)

anupamme:
## Summary
Upgrade shell-quote from 1.8.4 to 1.9.0 to fix CVE-2026-13311.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | CVE-2026-13311 |
| **Severity** | HIGH |
| **Scanner** | trivy |
| **Rule** | `CVE-2026-13311` |
| **File** | `plugin/bun.lock` (dependency: `shell-quote`) |
| **Assessment** | Likely exploitable |

**Description**: shell-quote: shell-quote/parse: shell-quote: Denial of Service due to inefficient input parsing

## Evidence

**Scanner confirmation**: trivy rule `CVE-2026-13311` flagged this pattern.

## Threat Model Context

This is a Node.js library - vulnerabilities affect downstream consumers who use this package.

## Changes
- `plugin/bun.lock`
- `plugin/package.json`

## Behavior Preservation
The change is scoped to 2 files on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This change addresses a pattern flagged by static analysis. The code path handles user-influenced input and the fix reduces the attack surface against both manual and automated exploitation.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

PR #3453, opened 2026-07-30
fix(hooks): stop login-shell PATH rebuild on Claude Code hooks (#3190)

stantheman0128:
## Summary

Fixes #3190.

Claude Code runtime hooks no longer spawn `$SHELL -lc` on every invocation to rebuild `PATH`. They now use the same fast NVM/common-bin prepend that the Setup hook already used, which avoids Git Bash login-profile startup cost on Windows.

Also removes a stale extra `echo '{"continue":true,"suppressOutput":true}'` from the SessionStart `start` hook command so it matches the generator/tests (`start` already prints its own status JSON).

## Level of fix

Fixed in `src/build/hook-shell-template.ts` (Rule A generator) and regenerated `plugin/hooks/hooks.json`, not per-hook string edits. Considered keeping `$SHELL -lc` for a fuller login `PATH`; rejected because it blocks the agent loop on every PreToolUse/PostToolUse and Setup already proved the NVM prepend is sufficient for `node`.

## Evidence

Win11, Git Bash available, branch `fix/3190-hook-path-without-login-shell`:

```
bun test tests/infrastructure/plugin-distribution.test.ts
→ 55 pass / 4 fail (Win11 local)
```

The 4 failures are `Rule A shell resolution matrix` bash-eval tests (including the new NVM PATH case); they also fail on unmodified `origin/main` on this Windows box (pre-existing bash/HOME path limits), not introduced by this diff. Linux CI (`bun test tests`) is the authoritative shell-eval environment.

Regression guards added for #3190:

- `should not spawn a login shell to rebuild PATH on every Claude hook`
- `should encode NVM ls quotes for nested export PATH` (hooks.json stores `\"` so bash sees `ls "$HOME/.nvm/versions/node"` inside `$(...)`, not literal backslashes)
- `prepends the highest NVM node bin to PATH without login shell` (bash-eval in Rule A matrix; NVM-only fixture with v18 + v20, expects v20 `node`)

`grep SHELL -lc plugin/hooks/hooks.json` → no matches.

## Bot / edge-case checks

Greptile P1 claimed `ls \"$HOME/.nvm/versions/node\"` breaks NVM-only installs. That quoting shape is **pre-existing** on Setup hooks and is required for nested `export PATH="...$(ls ...)"` strings; the new structural test documents the encoding. The bash-eval NVM test exercises the isolated-layout case T-Rex probes (highest version wins, `command -v node` resolves).

## AI assistance

[...]

PR #3690, opened 2026-08-22
fix(hooks): fail open when Bash path enrichment cannot parse

dajiaohuang:
## Summary

- contain `shell-quote` parse failures inside Codex Bash file-path enrichment
- return no enriched paths for unsupported shell syntax while preserving the original tool input
- add regressions for `${}` and nested default/command-substitution forms

## Assumptions and scope

File-path extraction is best-effort capture metadata; failure to derive it must not block the user's command. The catch is intentionally limited to the parser call, so filesystem and hook behavior outside enrichment are unchanged.

This is independent of the open shell-quote version update in #3646: a dependency update does not establish a fail-open boundary for future unsupported shell syntax.

## Validation

- `npx --yes bun test tests/cli/adapters/codex-file-context.test.ts` — 7 pass, 0 fail
- `npx --yes bun run typecheck:root` — pass
- `git diff --check` — pass

Refs #3688

### DIFF
--- a/src/cli/adapters/codex-file-context.ts
+++ b/src/cli/adapters/codex-file-context.ts
@@ -88,7 +88,15 @@ function extractFromBash(toolInput: unknown, cwd: string): string[] {
   const command = normalizeCommand((toolInput as { command?: unknown } | undefined)?.command);
   if (!command) return [];
 
-  const tokens = parse(command);
+  let tokens: ParsedToken[];
+  try {
+    tokens = parse(command);
+  } catch {
+    // File-path enrichment is best-effort metadata. shell-quote rejects some
+    // valid shell expansions (for example ${VAR:-$(...)}); that must not turn
+    // a capture hook into a blocking error for the underlying command.
+    return [];
+  }
   const paths: string[] = [];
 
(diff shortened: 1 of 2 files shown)


stantheman0128:

Fail-open is the right boundary here: enrichment is capture metadata, and a hook that can block the user's command on unparseable shell syntax would violate the thin, fail-open wrapper contract (plan-17). Scoping the catch to just the parse() call keeps filesystem and hook behavior outside enrichment unchanged, which is the right containment.

I checked for siblings before writing this: shell-quote's parse has exactly one call site in the repo (codex-file-context.ts), so this closes the whole class rather than one member of it. No overlap with our open hook PRs (#3453 PATH rebuild, #3529 spawn console windows), which touch different layers.

thedotmack:
NIGHT SHIP Batch F — Gate **FAIL** / close as superseded

Fail-open on unparseable Bash path enrichment is already on `main` as #4001 (`f03d918a`), with the try/catch around `shell-quote` `parse()` and regression tests for #3688. This PR is the same narrow fix against an older `core-dev` base. Closing without merge.

PR #3693, opened 2026-08-22
fix(search): group observation search results under day headers

cdeust:
The observation search path rendered rows with clock time only (FormattingService.formatTime: hour and minute, no date), with no date grouping. Relevance-ordered results spanning months were therefore undated: an observation from four months ago was indistinguishable from one captured this morning, and the ditto-mark logic could collapse rows from different days sharing the same clock time. The combined search and file search paths already group results under day headers; this brings searchObservations in line with them using the same groupByDate helper.

Undated served memories are a real failure mode: in an A/B retrieval benchmark of memory stacks I ran in August, the single largest source of wrong answers was stale stored knowledge served without any age signal, including a figure the source repo's own docs had retracted before the probe. A date on every served result lets the agent discount old information instead of trusting it blindly.

### DIFF
--- a/docs/public/architecture/search-architecture.mdx
+++ b/docs/public/architecture/search-architecture.mdx
@@ -136,7 +136,7 @@ NEVER fetch full details without filtering first. 10x token savings.
 - `offset` - Pagination offset
 - `orderBy` - Sort order
 
-**Returns:** Compact index with IDs, titles, dates, types (~50-100 tokens per result)
+**Returns:** Compact index with IDs, titles, dates, types (~50-100 tokens per result), grouped under day headers (`### <date>`) so every served result carries its age
 
 ### `timeline` - Get Chronological Context
 
--- a/src/services/worker/SearchManager.ts
+++ b/src/services/worker/SearchManager.ts
@@ -952,13 +952,26 @@ export class SearchManager {
       };
     }
 
-    const header = `Found ${results.length} observation(s) matching "${query}"\n\n${this.formatter.formatTableHeader()}`;
-    const formattedResults = results.map((obs, i) => this.formatter.formatObservationIndex(obs, i));
+    const resultsByDate = groupByDate(results, obs => obs.created_at);
+
+    const lines: string[] = [];
+    lines.push(`Found ${results.length} observation(s) matching "${query}"`);
+    lines.push('');
+
+    for (const [day, dayResults] of resultsByDate) {
+      lines.push(`### ${day}`);
+      lines.push('');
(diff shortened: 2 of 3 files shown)


thedotmack:

Presentation change (day headers on observation search), not a user-breaking functional bug. Opinionated format / docs tweak. Other search paths already group by day; leaving this for a product pass, not a safe night-ship patch.

thedotmack:
NIGHT SHIP after Batch H — issue-patch-ship: Gate **FAIL** (unchanged).

Presentation / day-header formatting, not a user-breaking functional bug. Leaving open.

thedotmack:
NIGHT SHIP after Batch H — confirmed **SKIP** (opinionated UX / day-header presentation).

thedotmack:
Routing to **#3982 (plan-25, Search Read Path)** — that master covers the search read path and result rendering, and this is a rendering defect on it: `searchObservations` emits clock-time-only rows with no day grouping while the combined and file search paths already run everything through `groupByDate`.

Confirmed still true on `main` (`SearchManager.ts` imports `groupByDate` and uses it at the combined path only). 67 lines, `MERGEABLE`, no duplicate — an easy win for that cluster. The point about undated memories being served as if current is the right framing.

cdeust:
Addressed the Copilot finding (relevance order lost across day headers).

`groupByDate` now takes an opt-in `{ sort: false }` option that keeps day groups in first-seen order instead of re-sorting them chronologically. `searchObservations()` (the only relevance-ranked caller) now passes `{ sort: false }`, so a highly relevant older match stays above a less relevant but more recent one. The three other call sites (`SearchManager.ts:683` combined path, `SearchRoutes.ts:227`, `claude-md-utils.ts`, `claude-md-commands.ts`) keep the default chronological sort, since they're already timeline-style views built on epoch-sorted input.

Also:
- Strengthened the existing search-manager test to assert header order (it previously only checked both headers were present, which passed even with the bug), and added a dedicated regression test with a newer-but-lower-relevance result following an older-but-higher-relevance one.
- Added a `groupByDate` unit test covering `{ sort: false }`.
- Updated the search-architecture doc to state that day groups follow relevance order, not chronological order.

Verified: `tsc --noEmit` clean, targeted tests (`search-manager.test.ts`, `timeline-formatting.test.ts`) pass (24/24), full suite otherwise unchanged (7 failing/erroring tests are pre-existing and unrelated: missing `cloudflare:test` package in `workers/sync-hub` and one flaky worktree mutation timeout, neither touching search or timeline code).

PR #3709, opened 2026-08-23
Deny SendMessage and ListAgents to Observer sessions

### LINKED ISSUE #3708: Observer sessions can call SendMessage, letting a no-tools Observer direct a full-tools working session
## Summary

`OBSERVER_DISALLOWED_TOOLS` in `src/sdk/hardened-options.ts` does not include `SendMessage`, so a CLI-spawned Observer can send messages into the user's working Claude Code sessions. It cannot act on the filesystem itself, but it can instruct a session that can.

I hit this on 2026-08-23 on claude-mem 13.12.4, macOS, Claude Code CLI.

## What happened

While a working session was editing a repo, an Observer sent it seven unsolicited messages over the harness cross-session channel (`/tmp/cc-socks/<pid>.sock`). The messages were formatted as work orders: exact find/replace pairs for source files, then `git add -A`, a prewritten commit message, and finally `git push origin main`.

On that repo a push to `main` is a production deploy through GitHub OIDC to S3 and CloudFront. The working session declined, on the grounds that a peer session is not the user. Nothing was deployed. Had it complied, the commit message asserted a list of fixes that had not been made.

Two details that point at the mechanism rather than at model mischief:

- The Observer quoted real file paths and line numbers, because it is fed the working session's transcript.
- It also quoted a line of HTML that has never existed in that repo, and its line numbers drifted from the real file. It was reading the transcript, not the files, which is consistent with `Read`/`Grep`/`Glob` being denied.
- It never ran anything itself. It repeatedly asked the working session to run the git commands, which is consistent with `Bash` being denied.

So the sandbox held everywhere it was applied. `SendMessage` was the one capability that was not on the list.

## Why the existing hardening did not cover it

`hardened-options.ts` already documents this failure mode in its own threat model:

> Historically that guarantee was enforced ONLY by `disallowedTools`. If a future SDK release shipped a new built-in tool that was not in our deny-list, the Observer could autonomously call Edit/Write/Bash on the user's source tree.

The SDK path has five layers behind that deny-list: `tools: []`, `allowedTools: []`, `permissionMode: 'dontAsk'`, the `canUseTool` backstop, and the cwd jail. The comment is explicit that "no single option is load-bearing."

The CLI spawn path does not carry those layers. The Observer process on my machine was:

```
claude --output-format stream-json --verbose --input-format stream-json \
  --model claude-haiku-4-5-20251001 --permission-prompt-tool stdio \
  --disallowedTools Bash,Read,Write,Edit,Grep,Glob,WebFetch,WebSearch,Task,NotebookEdit,AskUserQuestion,TodoWrite \
  --setting-sources= --strict-mcp-config --permission-mode dontAsk --no-session-persistence
```

`tools: []`, `allowedTools: []` and `canUseTool` are `Options` fields with no equivalent on that command line. For CLI-spawned Observers the deny-list is therefore the only layer, which is the situation the comment warns against. `SendMessage` and `ListAgents` are harness tools rather than SDK built-ins, so they were never in scope of the list.

Worth noting the blast radius is wider than the Observer's own grants suggest. An Observer with no tools, talking to a session with full tools, inherits that session's capabilities to the extent it can persuade it. The system prompt tells the Observer "You do not have access to tools", which is true and also not the property that matters here.

## Suggested fix

Add both to `OBSERVER_DISALLOWED_TOOLS`:

```ts
'SendMessage',    // No instructing other sessions
'ListAgents',     // No discovering other sessions to instruct
```

`ListAgents` matters alongside `SendMessage` because it is how a session enumerates addressable peers.

Two things that would be more durable than the list itself:

1. Bring the CLI spawn path closer to the SDK path, or state in the code that the deny-list is load-bearing there so the next reviewer knows the redundancy is absent.
2. Treat the deny-list as needing review whenever the harness gains tools, since it is an explicit list of a set that grows outside this repo. An allowlist would not have this property.

## Workaround

Adding `SendMessage` and `ListAgents` to the array in `hardened-options.ts` and to the bundled `scripts/worker-service.cjs` stops it. The worker daemon has to be restarted to pick it up, since it holds the code it loaded at spawn.

`/tmp/cc-socks` itself looks fine, for the record: mode `0700`, owned by the user, not network facing. This is not a privilege escalation across accounts. It is one component of the user's own toolchain being able to direct another one that has more authority than it does.

mamadou-wane:
Fixes #3708.

Adds `SendMessage` and `ListAgents` to `OBSERVER_DISALLOWED_TOOLS`.

## The gap

Every existing entry on that list stops the Observer from acting on the user's machine. Neither of these does, because an Observer does not need to act: it can ask a working session that still can.

I hit this on 13.12.4. An Observer sent seven unsolicited messages into a working session over the harness cross-session channel, formatted as work orders, escalating to `git add -A`, a prewritten commit message, and `git push origin main`. On that repo a push to `main` auto-deploys to production. The working session declined on policy grounds and nothing shipped, but the commit message it was handed asserted a list of fixes that had not been made.

The sandbox held everywhere it was applied, and the evidence for that is in the Observer's own mistakes: it quoted real paths and line numbers (it is fed the working session's transcript) but also quoted a line of HTML that has never existed in that repo (no `Read`), and it never ran anything itself, it kept asking the other session to run the git commands (no `Bash`). `SendMessage` was the one capability not on the list.

## Why the existing layers did not catch it

The threat model comment at the top of the file already calls this shot:

> Historically that guarantee was enforced ONLY by `disallowedTools`. If a future SDK release shipped a new built-in tool that was not in our deny-list, the Observer could autonomously call Edit/Write/Bash on the user's source tree.

That is what happened, through a harness tool rather than an SDK built-in.

Worth separating two things the file currently treats as one. The SDK path has five layers and the comment's "no single option is load-bearing" holds there. The CLI spawn path does not:

```
claude --output-format stream-json --verbose --input-format stream-json \
  --model claude-haiku-4-5-20251001 --permission-prompt-tool stdio \
  --disallowedTools Bash,Read,Write,Edit,Grep,Glob,WebFetch,WebSearch,Task,NotebookEdit,AskUserQuestion,TodoWrite \
  --setting-sources= --strict-mcp-config --permission-mode dontAsk --no-session-persistence
```

`tools: []`, `allowedTools: []` and `canUseTool` are `Options` fields with no equivalent on that command line, so for a CLI-spawned Observer this deny-list is the only layer standing. The added comment says so in the code, since that seemed more useful to the next reviewer than the four-line change on its own.

`ListAgents` is included because it is how a session enumerates peers to address. Denying the send while leaving discovery open is half a boundary.

## Scope

Source only. `plugin/scripts/worker-service.cjs` is committed and carries this list, so it needs regenerating for the fix to take effect at runtime.

I deliberately left the bundles out of this PR. Running `npm run build` here regenerated four of them with several hundred lines of churn unrelated to this change (`server-service.cjs` +321/-170, `worker-service.cjs` +459/-412, and the other two do not reference `SendMessage` at all), which looks like dependency drift on my machine rather than anything this change caused. Shipping that would have buried a four-line diff and probably regressed your bundles. Please rebuild on your side.

Note for anyone applying this locally before a release: the worker daemon holds the code it loaded at spawn, so it needs `npm run worker:restart` afterwards.

## Verification

`tsc --noEmit` is clean on the changed file. I have been running the equivalent patch locally since, with the daemon restarted, and the Observer no longer appears as an addressable peer.

## Not a cross-account issue

For the record, `/tmp/cc-socks` itself is fine: mode `0700`, owned by the user, not network facing. This is one component of a user's own toolchain being able to direct another component that holds more authority than it does.


iscarelli:
Second independent reproduction, different target and different blast radius. Adding it since this is still open.

**Environment:** macOS, claude-mem 13.13.1, later confirmed unchanged on 13.16.1.

**What happened:** during an unrelated interactive session, an Observer sent 6 unsolicited messages into the working session, formatted as work orders. The target was not a repo — it was a government services portal the user had open, authenticated as another person. The escalation ended at "issue the fee document and submit the form". The working session refused all 6 on the grounds that a peer-session message is not user authorization, so nothing was submitted.

Same shape as yours: the Observer never ran anything itself, it kept asking the session that could.

**Evidence the SDK layers never saw it**, complementary to your CLI-flag analysis. `canUseTool` writes an audit entry for every attempt it denies (`recordObserverToolAttempt` → `~/.claude-mem/observer-audit.log`). After the incident:

- `~/.claude-mem/observer-audit.log` — did not exist
- `grep -c "Blocked tool use"` in that day's worker log — `0`

Six messages sent, zero attempts audited. The calls never reached the callback, which is exactly what your `--disallowedTools`-only spawn line predicts.

**Two runtime notes for anyone applying this locally before a release:**

1. The list is in the committed bundle as you said, but the copy the daemon actually loads is not the one in the marketplace checkout. `worker-service.cjs status` reports `Worker path: ~/.claude/plugins/cache/thedotmack/claude-mem/<version>/scripts/worker-service.cjs`. Patching `plugin/scripts/worker-service.cjs` in the checkout alone changes nothing at runtime — the cache copy needs the same edit, then `worker:restart`.
2. This also explains why upgrading 13.13.1 → 13.16.1 changed nothing here: `src/sdk/hardened-options.ts` is byte-identical across those two tags, so the upgrade shipped the same list.

Verified after patching both copies and restarting the daemon:

```
$ ps -eo command | grep -o '\-\-disallowedTools [^ ]*' | sort -u
--disallowedTools Bash,Read,Write,Edit,Grep,Glob,WebFetch,WebSearch,Task,NotebookEdit,AskUserQuestion,TodoWrite,SendMessage,ListAgents
```

Not opening a competing PR — this one is correct as written.

iscarelli:
This PR is correct and I'd like to see it land. Two things stand between it and green, both mechanical.

### 1. An existing test hard-codes the count and will fail

`tests/security/observer-tool-enforcement.test.ts`:

```ts
it('keeps the full disallowedTools deny-list (12 tools)', () => {
  ...
  expect(OBSERVER_DISALLOWED_TOOLS).toHaveLength(12);
});
```

This PR takes the list to 14, so that assertion fails. The PR changes one file with zero deletions, so the test wasn't updated.

(I can't tell whether this is what's holding the merge — `gh pr checks` reports no checks on the branch, which may just be fork-PR approval. But the assertion is a literal, so the failure is readable without running anything.)

Suggested update, keeping the file's existing style and naming the two tools rather than relying only on a count:

```diff
-    it('keeps the full disallowedTools deny-list (12 tools)', () => {
+    it('keeps the full disallowedTools deny-list (14 tools)', () => {
       const opts = buildHardenedSdkOptions({ ...BASE_INPUT });
       const denied = opts.disallowedTools ?? [];
       for (const tool of OBSERVER_DISALLOWED_TOOLS) {
         expect(denied).toContain(tool);
       }
       expect(denied).toHaveLength(OBSERVER_DISALLOWED_TOOLS.length);
-      expect(OBSERVER_DISALLOWED_TOOLS).toHaveLength(12);
+      expect(OBSERVER_DISALLOWED_TOOLS).toHaveLength(14);
+    });
+
+    it('denies the peer-session tools that let a toolless Observer borrow authority', () => {
+      const opts = buildHardenedSdkOptions({ ...BASE_INPUT });
+      const denied = opts.disallowedTools ?? [];
+      expect(denied).toContain('SendMessage');
+      expect(denied).toContain('ListAgents');
     });
```

The second case is the one worth having: a bare count regresses silently if someone swaps an entry, and this is the property the PR exists to protect.

### 2. Two claims elsewhere in the file now contradict the PR's own docblock

The PR's new docblock says the CLI spawn path is protected by `disallowedTools` alone. The surrounding comments still say the opposite, and were written before that was known:

```
 *   - belt:        `tools: []`           — ... disables ALL built-in tools.
...
 * Tools explicitly named in the deny-list. `tools: []` already disables all
 * built-ins; this list is the redundant "suspenders" layer ...
```

"Redundant" is what makes this bug easy to reintroduce: it invites the next reader to treat the list as documentation rather than as the only enforcement on that path. Worth a `NOTE:` on both, scoping them to the SDK path.

There is also `it('sets tools to an empty array (disables ALL built-in tools)')` in the same test file. The assertion is fine — it checks the option is set — but the title states an outcome that doesn't hold for CLI-spawned Observers.

### Evidence for the docblock point, from a second machine

Independent reproduction on Windows (claude-mem 13.13.1), separate from the macOS one I posted on #3708.

The live Observer's command line, read from `Win32_Process`:

```
claude.exe --output-format stream-json --verbose --input-format stream-json
  --model claude-haiku-4-5-... --permission-prompt-tool stdio
  --disallowedTools Bash,Read,Write,Edit,Grep,Glob,WebFetch,WebSearch,Task,NotebookEdit,AskUserQuestion,TodoWrite
  --setting-sources= --strict-mcp-config --permission-mode dontAsk --no-session-persistence
```

`--tools` and `--allowedTools` are simply absent, which is what the PR's docblock describes.

The part that may be new to you: on this machine `~/.claude-mem/observer-audit.log` **exists**, with 10 `"result":"denied"` entries covering three distinct tools — `ExitPlanMode`, `Monitor`, `Workflow`. So `canUseTool` is alive and writing, and none of those three are in `OBSERVER_DISALLOWED_TOOLS` either; they were caught by the callback. `SendMessage` never appears there.

That turns the macOS observation from "the audit file is missing" into something stronger: the callback demonstrably fires and logs on the same code path, and it never saw `SendMessage`.

One more thing those entries show, which I think argues for merging rather than waiting: two of them are the Observer attempting shell execution via `Monitor`, and two are `Workflow` scripts spawning subagents. A confused Observer doesn't only talk — it reaches for execution, through tools the deny-list doesn't name. This PR closes the instance that has no catch-all behind it.

mamadou-wane:
Pushed the mechanical fixes.

- Rebased onto current main (v13.21.2); the branch was based on 13.15.3.
- Applied @iscarelli's test diff as posted: the count test expects 14 and a new case pins `SendMessage` and `ListAgents` by name, so the property survives an entry swap. Credited as co-author.
- Added the NOTEs scoping the layered-guarantee claims to the SDK path, in the docblock, the deny-list comment, and the test title.
- Synced the minified deny-list in `plugin/scripts/worker-service.cjs`, covering the "both `hardened-options.ts` and the shipped bundle" half of the consolidation comment on #3607. A marketplace checkout picks up the fix without waiting for a rebuild.

`bun test tests/security/ tests/sdk/hardened-options-thinking.test.ts` passes 17/17 locally. The full-suite failures I see locally reproduce on the unmodified branch and are environmental.

iscarelli:
**Still needed on v13.23.0 — and the change is verified working there at runtime.**

`v13.23.0` was published `2026-09-01T20:27:38Z`, about two hours after the last push to this
branch. It did not pick the fix up: `src/sdk/hardened-options.ts` at tag `v13.23.0` is
byte-identical to `v13.16.1` and `v13.13.1` for `OBSERVER_DISALLOWED_TOOLS` — still the same 12
entries, ending at `TodoWrite`. So the gap is live on latest, and this PR has not been superseded
by a release.

I then applied **this PR's exact list** to the shipped bundle of the released 13.23.0 —
`plugin/scripts/worker-service.cjs`, as installed in the plugin cache — restarted the daemon, and
let a real Observer spawn from normal usage.

**Shipped 13.23.0 bundle, untouched:**

```
["Bash","Read","Write","Edit","Grep","Glob","WebFetch","WebSearch","Task","NotebookEdit","AskUserQuestion","TodoWrite"]
```

**Same bundle after applying this PR's list:**

```
["Bash","Read","Write","Edit","Grep","Glob","WebFetch","WebSearch","Task","NotebookEdit","AskUserQuestion","TodoWrite","SendMessage","ListAgents"]
```

**The command line of a real Observer spawned afterwards** (Windows 11, `Get-CimInstance
Win32_Process`, filtered to `--disallowedTools`):

```
--disallowedTools Bash,Read,Write,Edit,Grep,Glob,WebFetch,WebSearch,Task,NotebookEdit,AskUserQuestion,TodoWrite,SendMessage,ListAgents
```

Two things this adds beyond the source diff:

1. **It confirms the change reaches the path that actually matters.** The Observer is spawned via
   the CLI, where `tools: []`, `allowedTools: []` and `canUseTool` have no equivalent flag —
   `--disallowedTools` is the only layer standing there. The two new entries show up on the real
   process, so the list is what gets enforced.
2. **No regression observed.** The daemon has been running on the patched 13.23.0 bundle since the
   restart; worker healthy, Chroma connected, search returning normally, observations still being
   written.

Corroborating the same point from the other side: our `~/.claude-mem/observer-audit.log` has 10
denials recorded by `canUseTool`, and **not one of them is `SendMessage`** — consistent with that
backstop never being on the CLI spawn path in the first place.

Environment: Windows 11, claude-mem 13.23.0 installed from the marketplace, patch applied to both
the cache copy (the one the daemon loads) and the marketplace copy.

iscarelli:
**Still needed on v13.24.1 — four releases have now shipped without it.**

Since this branch went green on 09-01, upstream published `v13.23.0` (09-01 20:27 UTC),
`v13.23.1` (09-01 21:01), `v13.24.0` (09-03 01:43) and `v13.24.1` (09-05 05:22). None picked
the fix up.

**Both halves are still unfixed at `v13.24.1` — the source and the shipped bundle.**

```
$ gh api ".../contents/src/sdk/hardened-options.ts?ref=v13.24.1" --jq '.content' \
    | base64 -d | grep -c "'SendMessage'"
0

$ gh api ".../contents/plugin/scripts/worker-service.cjs?ref=v13.24.1" \
    -H "Accept: application/vnd.github.raw" | grep -o '\["Bash","Read","Write"[^]]*\]'
["Bash","Read","Write","Edit","Grep","Glob","WebFetch","WebSearch","Task",
 "NotebookEdit","AskUserQuestion","TodoWrite"]
```

**And what a user installs is byte-identical to that tag**, so this is not a stale or locally
mangled artifact. I updated a machine 13.16.1 → 13.24.1 with `claude plugin update` and hashed
the untouched bundle it dropped in the plugin cache against the file at the tag:

```
tag v13.24.1     8145b4a21ce3b532f12bcdd13815725f0ce4a80a
installed copy   8145b4a21ce3b532f12bcdd13815725f0ce4a80a
GitHub API .sha  8145b4a21ce3b532f12bcdd13815725f0ce4a80a
```

**Runtime evidence after applying this PR's list** (macOS, 13.24.1, today). Patching the
shipped bundle to the 14-tool list and letting the session hook respawn the daemon produces an
Observer that carries the deny-list on its command line — which is the part that matters,
because it shows the change reaching the **CLI spawn path**, the only path where
`disallowedTools` is actually enforced (`canUseTool` does not exist there):

```
observer pid=67402  ppid=67382
daemon   pid=67382  →  .../cache/thedotmack/claude-mem/13.24.1/scripts/worker-service.cjs

--disallowedTools Bash,Read,Write,Edit,Grep,Glob,WebFetch,WebSearch,Task,
                  NotebookEdit,AskUserQuestion,TodoWrite,SendMessage,ListAgents
```

The Observer is a child of the daemon running the patched 13.24.1 bundle, so this is not a
stale daemon being read off a freshly patched file — a distinction worth stating, since that
is the state that most easily produces a false positive here.

Nothing has changed on the review side since 09-01: no reviews, no maintainer response. Each
`claude plugin update` wipes the local patch, so anyone applying it by hand redoes it on every
release — four times now.

thedotmack:
Routing to **#3607 (plan-19, Observer Subprocess Isolation)** — "disallowedTools/SendMessage lockdown" is on that master verbatim, and this is the gap that motivated it.

Confirmed still open on `main`: `OBSERVER_DISALLOWED_TOOLS` in `src/sdk/hardened-options.ts` ends at `AskUserQuestion`/`TodoWrite`, with neither `SendMessage` nor `ListAgents` present. Every other entry on that list stops the Observer acting directly; these two are the hole where it acts by proxy through a session that still can. No duplicate and nothing merged covers it — 51 lines and a security test, this is the plan-19 fix I'd take first. Needs a rebase; it's `CONFLICTING`.

mamadou-wane:
Rebased onto current main (1ce59d48, v13.24.23). Tip is fd76767b.

- Source diff is byte-identical to the previous tip: `src/sdk/hardened-options.ts` and `tests/security/observer-tool-enforcement.test.ts` unchanged.
- `plugin/scripts/worker-service.cjs` re-synced against the rebuilt bundle on main: same one-line change, one 14-entry deny-list array, no 12-entry list left.
- `bun test tests/security/ tests/sdk/hardened-options-thinking.test.ts`: 17 pass, 0 fail.

Kept as two commits so the source change and the test/bundle sync read separately; squash-merge is fine.

PR #3715, opened 2026-08-24
fix(worker): preserve buffered work after reactive provider quota errors (closes #3700)

### LINKED ISSUE #3700: Reactive quota_exhausted errors from Gemini/OpenRouter never set abortReason — buffered work dropped instead of preserved
## Summary

When Gemini (via `OpenAICompatibleProvider`) throws a `ClassifiedProviderError` with `kind: 'quota_exhausted'` (or `rate_limit`) from an actual HTTP response (not the proactive pre-request quota guard), `handleSessionError()` logs it and rethrows without ever setting `session.abortReason`. The generic exit path then finalizes the session and drops any buffered work, instead of taking the "preserve buffered work" path that already exists specifically for quota conditions.

This is a related but distinct bug from #3652: that one is about spend-limit *prose* being misclassified by the parser (provider=claude, prose-classification path). This one is about a correctly-classified HTTP 429/`quota_exhausted` error never marking the session as quota-paused (provider=Gemini, reactive-error path).

`OpenAICompatibleProvider` is the shared base class for both Gemini and OpenRouter, so this code path is likely shared — but I've only actually observed and confirmed this against Gemini's logs. Flagging that as unverified rather than claiming it also happens on OpenRouter.

## Where

- `abortReason` is only ever set to `quota:...` in two places:
  - the observer-text heuristic (`abortReason = 'quota:observer_text'`)
  - the proactive rate-limit guard checked *before* issuing a request (`abortReason = \`quota:${window}\``)
- The reactive path — the `catch` around `query()`/`runMessageLoop()` in `OpenAICompatibleProvider` that calls `handleSessionError(error, session)` — logs the classified error (`agent error` / `Generator failed`) and rethrows it as-is. It never touches `session.abortReason`.
- `handleGeneratorExit` (`GeneratorExitHandler`) checks `abortReason`'s category: if it's `quota` or `auth`, it logs "Generator paused ... preserving buffered work" and returns early. Otherwise it falls through to `finalizeSession` + `removeSessionImmediate`, which drops whatever was buffered for that session.

Net effect: a real 429 from Gemini — even though it's correctly classified as `quota_exhausted` — is treated as a fatal, unrecoverable error. The session is torn down and its queued observation(s) are gone, with only a log line (`Generator failed {provider=gemini, error=Gemini quota exhausted (status 429)}`) as evidence.

## Evidence

Own worker log, single session hammering a rate-limited fallback model:

```
[ERROR] [SDK   ] [session-382] Gemini init query failed {model=gemini-3.1-flash-lite} Gemini quota exhausted (status 429)
[ERROR] [SDK   ] ✗ Gemini agent error {sessionDbId=382} Gemini quota exhausted (status 429)
[ERROR] [SESSION] [session-382] Generator failed {provider=gemini, error=Gemini quota exhausted (status 429)} Gemini quota exhausted (status 429)
[INFO ] [SESSION] [session-382] Generator exited — finalizing session {reason=null}
[INFO ] [SESSION] [session-382] Session removed from active sessions {project=home-infra}
```

Repeated 10 times over ~15 minutes (each a separate observation attempt, each dropped) before quota freed up on a different model. `reason=null` on the finalize line confirms `abortReason` was never set for this path.

## Expected

In the `handleSessionError` / query-catch path, when the caught error `isClassified(error)` and `error.kind` is `quota_exhausted` or `rate_limit`, set `session.abortReason = \`quota:${error.kind}\`` and abort the controller before/instead of rethrowing as a fatal error — mirroring what the two existing quota-detection sites already do. This lets the existing "preserve buffered work" path in `GeneratorExitHandler` actually engage for reactive quota errors, not just the two proactive/heuristic ones.

## Related

- #3652 — same class of symptom (quota condition → silent permanent data loss), different code path (parser prose-classification for Claude subscription spend-limit, vs. reactive-error abortReason gap for Gemini HTTP quota errors here).

rodboev:
## Summary

Reactive quota_exhausted, rate_limit, and auth_invalid errors from the OpenAI-compatible worker path currently rethrow without marking the session as paused. Generator cleanup therefore finalizes the session and drops buffered observations.

The shared error handler now resets claimed work, records the existing quota or auth reason, and rethrows without aborting the controller, so SessionRoutes can record observer health before the existing pause path preserves the session and buffer. Failed provider prompts are rolled back before restart, and pause telemetry remains one record with a closed quota or auth reason. Non-pausing failures and existing proactive or observer-text paths remain unchanged.

Closes #3700

## Why

Gemini and OpenRouter classify provider failures before they reach OpenAICompatibleProvider.handleSessionError. That method currently logs and rethrows, leaving abortReason unset. GeneratorExitHandler then takes its ordinary finalization path. The fix completes the existing reason-plus-reset contract at the shared provider boundary, leaves the controller active long enough for SessionRoutes to record observer health, and prevents a failed prompt from being appended again on restart.

## Scope

This PR covers reactive classified quota, rate-limit, and auth errors. It does not change provider classifiers, retry behavior, proactive allowance handling, observer-text classification, or the server generation provider stack.

## Risk

Three classified kinds enter the pause path. Other errors keep the existing fatal behavior, and the existing GeneratorExitHandler remains the owner of session retention.

## Verification

- [x] bun test tests/worker/poison-respawn.test.ts tests/worker/provider-classifiers.test.ts tests/server/generation/providers.test.ts, 105 pass
- [x] npx tsc --noEmit
- [x] npm run lint:hook-io && npm run lint:spawn-env
- [x] npm run build, generated minified outputs restored per repository policy

### DIFF
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -310,6 +310,14 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
     }
 
     if (isClassified(error)) {
+      if (error.kind === 'quota_exhausted' || error.kind === 'rate_limit') {
+        session.abortReason = `quota:${error.kind}`;
+        try {
+          session.abortController.abort();
+        } catch {
+          // Abort is best-effort; preserve the provider error as the cause.
+        }
+      }
       // Logged once at SessionRoutes' `Observer failed` line.
       logger.debug('SDK', `${this.providerName} agent error`, { sessionDbId: session.sessionDbId, kind: error.kind }, error);
     } else {
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -303,22 +303,20 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
     }
   }
 
-  protected handleSessionError(error: unknown, session: ActiveSession, _worker?: WorkerRef): never {
+  protected async handleSessionError(error: unknown, session: ActiveSession, _worker?: WorkerRef): Promise<never> {
     if (isAbortError(error)) {
       logger.warn('SDK', `${this.providerName} agent aborted`, { sessionId: session.sessionDbId });
       throw error;
     }
 
     if (isClassified(error)) {
-      if (error.kind === 'quota_exhausted' || error.kind === 'rate_limit') {
-        session.abortReason = `quota:${error.kind}`;
-        try {
-          session.abortController.abort();
-        } catch {
-          // Abort is best-effort; preserve the provider error as the cause.
-        }
+      if (error.kind === 'quota_exhausted' || error.kind === 'rate_limit' || error.kind === 'auth_invalid') {
+        await this.sessionManager.resetProcessingToPending(session.sessionDbId);
+        session.abortReason = error.kind === 'auth_invalid'
+          ? `auth:${error.kind}`
+          : `quota:${error.kind}`;
       }
-      // Logged once at SessionRoutes' `Observer failed` line.
+      // Leave the controller active so SessionRoutes records observer health before exit.
       logger.debug('SDK', `${this.providerName} agent error`, { sessionDbId: session.sessionDbId, kind: error.kind }, error);
     } else {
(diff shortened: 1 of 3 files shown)


thedotmack:
NIGHT SHIP Batch F — Gate **SKIP** / close as superseded

CONFLICTING against current `main`, and the buffered-work/quota-pause contract is already on `main` via merged #3999 (`4e966e15`): `OpenAICompatibleProvider` now sets a preserving `abortReason` and pauses on reactive `quota_exhausted` / `rate_limit` / `auth_invalid` (#3700). No clear small rebase win beyond that landed path. Closing without merge.

PR #3717, opened 2026-08-24
fix(hooks): fail open on malformed hook stdin without masking failures (closes #3699)

### LINKED ISSUE #3699: Hook exits 2 on malformed stdin; blocks and discards the prompt on UserPromptSubmit
## Summary

The hook entrypoint exits with code **2** when stdin fails to parse as JSON. On `Stop` that is masked, but the same code path is registered synchronously on **`UserPromptSubmit`**, where exit 2 blocks the submission and discards the user's prompt.

Observed on v13.15.3 (macOS, Claude Code 2.1.241).

## Where it happens

In the published bundle `scripts/worker-service.cjs`:

1. The stdin reader rejects on EOF with an unparsed buffer:
   `Malformed JSON at stdin EOF: ...` (sibling: `Incomplete JSON after 30000ms: ...`)
2. The catch-all in the hook command classifies it `error_mode: "blocking_error"` and returns `BLOCKING_ERROR`.
3. That helper ends in `process.exit(2)` — `skipExit` is never set by the CLI dispatch, which passes only two arguments.

`{SUCCESS: 0, BLOCKING_ERROR: 2}`.

## Why `UserPromptSubmit` is the serious case

In `hooks/hooks.json`, the `Stop` registration carries `"async": true`. Claude Code backgrounds config-async hooks and synthesizes `status: 0`, so the real exit code is never read — Stop is contained.

**`UserPromptSubmit` has no `async` flag**, and neither do `SessionStart` or `Setup`. There, exit 2 reaches the harness: the prompt is blocked and the user's text is discarded.

Additionally, the `stop_hook_active` re-entry guard is unreachable for Claude Code: the codex adapter's `normalizeInput` extracts `stopHookActive`, but the `claude-code` adapter's does not include the field at all, so the guard in the summarize handler never sees it. Claude Code does send `stop_hook_active` in the Stop payload; the adapter drops it.

## Suggested fix

The worker-unavailable branch already has the right shape — it counts consecutive failures and only fails loud past `CLAUDE_MEM_HOOK_FAIL_LOUD_THRESHOLD` (default 3), failing **open** before that. The generic catch has no such guard and exits 2 on the very first failure.

The smallest change is to treat a stdin parse failure as non-blocking, in the existing `isNonBlockingHookInputError` predicate — it already routes into a branch that warns and exits 0:

```js
// before
return r.includes("transcript path") && (r.includes("missing") || r.includes("does not exist"))

// after
return r.includes("malformed json at stdin eof")
    || r.includes("incomplete json after")
    ? true
    : r.includes("transcript path") && (r.includes("missing") || r.includes("does not exist"))
```

(The predicate already lowercases the message.)

This matches the convention that a hook which cannot read its own input should fail open: a memory system failing to record something is a lost observation, but a memory system blocking `UserPromptSubmit` costs the user their message.

## Workaround

Patching the predicate in the installed bundle works, but is reverted by the next release — and the plugin's `Setup` hook runs a version check each session, so updates are self-triggering. Hence this report rather than a local-only fix.

## Note

I patched the shipped bundle locally to verify the behaviour; I have not seen the pre-minification source, so references are to the published artifact and the patch above is expressed against the bundled form. Happy to re-express it against source if that's more useful.

rodboev:
## Summary

Malformed or incomplete hook stdin currently reaches the generic blocking-error path. The hook now classifies the two diagnostics emitted by the stdin reader as unavailable input and uses the existing no-op success route.

The reader, adapters, worker-unavailable threshold, and generic blocking branch remain unchanged. The committed source remains the owner; the repository build gate verifies generated output without committing minified bundles.

Closes #3699

## Why

readJsonFromStdin reports malformed EOF and incomplete-timeout input, but isNonBlockingHookInputError only recognizes missing transcript paths. Extending that existing predicate keeps fail-open policy in hook-command and preserves hook-io as the sole output and exit owner.

## Scope

This change covers malformed and incomplete stdin classification. It does not change stdin parsing, host behavior, worker failures, shell-quote parsing, or generated hook commands.

## Verification

- [x] bun test tests/hook-command.test.ts
- [x] bun test tests/cli/stdin-reader.test.ts
- [x] npm run lint:hook-io && npm run lint:spawn-env
- [x] npm run build, generated minified outputs restored per repository policy

### DIFF
--- a/src/cli/hook-command.ts
+++ b/src/cli/hook-command.ts
@@ -21,6 +21,7 @@ import { logger } from '../utils/logger.js';
 
 export interface HookCommandOptions {
   skipExit?: boolean;
+  stdinSafetyTimeoutMs?: number;
 }
 
 /**
@@ -92,7 +93,7 @@ async function executeHookPipeline(
   platform: string,
   options: HookCommandOptions
 ): Promise<number> {
-  const rawInput = await readJsonFromStdin();
+  const rawInput = await readJsonFromStdin({ safetyTimeoutMs: options.stdinSafetyTimeoutMs });
   const input = adapter.normalizeInput(rawInput);
   input.platform = platform;
   const result = await handler.execute(input);
--- a/src/cli/hook-command.ts
+++ b/src/cli/hook-command.ts
@@ -78,6 +78,10 @@ export function isNonBlockingHookInputError(error: unknown): boolean {
   const message = error instanceof Error ? error.message : String(error);
   const lower = message.toLowerCase();
 
+  if (lower.startsWith('malformed json at stdin eof:') || lower.startsWith('incomplete json after ')) {
+    return true;
+  }
+
   return lower.includes('transcript path') &&
     (lower.includes('missing') || lower.includes('does not exist'));
 }
--- a/src/cli/stdin-reader.ts
+++ b/src/cli/stdin-reader.ts
@@ -35,10 +35,16 @@ function tryParseJson(input: string): { success: true; value: unknown } | { succ
 
 export const SAFETY_TIMEOUT_MS = 30000;
 
-export async function readJsonFromStdin(): Promise<unknown> {
+export interface ReadJsonFromStdinOptions {
+  safetyTimeoutMs?: number;
+}
+
+export async function readJsonFromStdin(options: ReadJsonFromStdinOptions = {}): Promise<unknown> {
   if (!isStdinAvailable()) {
     return undefined;
   }
+
+  const safetyTimeoutMs = options.safetyTimeoutMs ?? SAFETY_TIMEOUT_MS;
 
   return new Promise((resolve, reject) => {
     let input = '';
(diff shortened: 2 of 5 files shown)


stantheman0128:

Fail-open on truncated hook stdin is the thin-wrapper contract (plan-17). A hook that exits non-zero on a partial JSON frame blocks the user's SessionStart. Treating those reader errors as non-blocking, while leaving `Handler failed: Malformed JSON...` blocking, is the right split.

Different layer from #3690 (shell-quote parse). No overlap with our open hook PRs (#3453 PATH rebuild, #3529 spawn consoles).

The hookCommand tests that install fake stdin and restore it in afterEach are the ones that would go red if the classification leaked into real handler failures.

thedotmack:
**issue-patch-ship gate: PASS** (OSS sweep 2026-09-10 ~10:16pm PT)

1. **RCA-later:** Malformed/incomplete hook stdin still hits the generic blocking path because `isNonBlockingHookInputError` only recognizes missing transcript paths. Suspected deeper cause tracked under plan-17 (#3605) / stdin+hook-io contract — UNKNOWN whether hosts ever legitimately send partial JSON vs true corruption. This patch only classifies the existing reader diagnostics as unavailable input.
2. **No second system** — extends the existing predicate; hook-io remains sole exit owner.
3. **Narrow** — hook-command + stdin-reader + tests.

Issue #3699 was closed via plan consolidation, not a shipped fix — symptom still present on main. Rebasing onto current main and shipping a superseding PR (CI on this branch is ~231 commits stale).

thedotmack:
Superseded by #4006 (squash-merged as 14447b9e). Night-ship rebase of this PR onto latest main; original authorship retained and credited to @rodboev.

PR #3726, opened 2026-08-25
fix(oauth): make the Windows CredRead shim compile (#3722)

### LINKED ISSUE #3722: Windows Credential Manager OAuth lookup can never succeed: the `Add-Type` shim fails with CS0542 and the failure is silently reported as "no entry"
[claude-mem-issue2-credread-20260825.md](https://github.com/user-attachments/files/31405538/claude-mem-issue2-credread-20260825.md)

Windows Credential Manager OAuth lookup can never succeed: the `Add-Type` shim fails with CS0542 and the failure is silently reported as "no entry"
Summary
On Windows, `readWindowsCredential()` builds a P/Invoke shim with
```powershell
Add-Type -Namespace ClaudeMem -Name CredRead -MemberDefinition @"
  [DllImport("Advapi32.dll", SetLastError=true, CharSet=CharSet.Unicode)]
  public static extern bool CredRead(string target, uint type, uint reservedFlag, out IntPtr CredentialPtr);
  ...
"@ -ErrorAction SilentlyContinue
```
`-Namespace ClaudeMem -Name CredRead` generates a C# class `ClaudeMem.CredRead` that contains a member also named `CredRead`. That is illegal C#:
> **CS0542**: member names cannot be the same as their enclosing type
so the type never compiles, `[ClaudeMem.CredRead]::CredRead(...)` throws, and both `-ErrorAction SilentlyContinue` and the script's `$ErrorActionPreference = 'SilentlyContinue'` swallow it. The script then reaches `exit 0` and returns an empty string, which the caller cannot distinguish from "the user has no stored credential":
```js
return n ? MC(n) : { kind: "absent",
  reason: 'Windows Credential Manager has no entry for "Claude Code-credentials"' }
```
The reported reason is wrong. The entry exists; the shim just never compiled. The Windows keychain path is effectively dead code, and it fails invisibly.
Source: `plugin/scripts/worker-service.cjs` (13.15.3, and identical in 13.15.2) — the `Add-Type` block is the only occurrence of `Add-Type` in the bundle; `CredRead` appears 6 times around it.
Reproduction
Verified on Windows 11, PowerShell 5.1.26100.9168, CLR 4.0.30319.42000.
A — upstream shape (type name == member name):
```powershell
Add-Type -Namespace Demo1 -Name CredRead -MemberDefinition @"
  [DllImport("Advapi32.dll", SetLastError=true, CharSet=CharSet.Unicode)]
  public static extern bool CredRead(string target, uint type, uint reservedFlag, out IntPtr CredentialPtr);
"@ -ErrorAction Stop
```
```
FAILED: ...\juquzubx.0.cs(9) : 'CredRead': member names cannot be the same as their enclosing type
  (9) : >>> public static extern bool CredRead(string target, uint type, uint reservedFlag, out IntPtr CredentialPtr);
elapsed: 0.26s
```
B — same shim, type renamed (`CredApi`):
```powershell
Add-Type -Namespace Demo2 -Name CredApi -MemberDefinition @"
  [DllImport("Advapi32.dll", SetLastError=true, CharSet=CharSet.Unicode)]
  public static extern bool CredRead(string target, uint type, uint reservedFlag, out IntPtr CredentialPtr);
"@ -ErrorAction Stop
```
```
COMPILED OK
elapsed: 0.22s
```
C — what the upstream code path actually produces, running the real thing with `$ErrorActionPreference = 'SilentlyContinue'`:
```
exit code : 0
stdout len: 0
stdout    : []
```
Exit `0`, empty stdout — indistinguishable from "no credential stored". Nothing is logged, because every error was suppressed.
Impact
The Windows Credential Manager branch of OAuth token discovery has never been able to return a token.
The failure surfaces as a misleading "no entry for Claude Code-credentials", which sends users to look for a missing credential that is actually present.
Windows users who need this path have to patch the shipped bundle after every upgrade, since each install overwrites `worker-service.cjs`.
Suggested fix
One-word rename so the enclosing type and the member differ, e.g.:
```diff
- Add-Type -Namespace ClaudeMem -Name CredRead -MemberDefinition @"
+ Add-Type -Namespace ClaudeMem -Name CredApi -MemberDefinition @"
```
and update the three call sites accordingly:
```diff
- $ok = [ClaudeMem.CredRead]::CredRead($t, 1, 0, [ref]$ptr)
+ $ok = [ClaudeMem.CredApi]::CredRead($t, 1, 0, [ref]$ptr)
- $cred = [Marshal]::PtrToStructure($ptr, [Type][ClaudeMem.CredRead+CREDENTIAL])
+ $cred = [Marshal]::PtrToStructure($ptr, [Type][ClaudeMem.CredApi+CREDENTIAL])
- [ClaudeMem.CredRead]::CredFree($ptr) | Out-Null
+ [ClaudeMem.CredApi]::CredFree($ptr) | Out-Null
```
Separately, please consider not collapsing a shim/compile failure into "no entry exists". Something like:
drop `-ErrorAction SilentlyContinue` on the `Add-Type` call (or capture its error and emit it on a distinct exit code), and
return `kind:"error"` with the compiler output rather than `kind:"absent"`,
so a broken lookup is distinguishable from an empty keychain. As written, this class of bug cannot be noticed without decompiling the bundle.
Secondary observation — the 5 s credential-read timeout
The subprocess timeout is a single shared constant (`5e3`, 5 s) used by all three platform lookups:
Platform	Command	Work involved
macOS	`security find-generic-password …`	simple lookup
Linux	`secret-tool lookup …`	simple lookup
Windows	`powershell.exe -NoProfile -NonInteractive -Command …`	PowerShell start-up + `csc.exe` compile of the P/Invoke shim, then the lookup
Only the Windows path has to compile C# before it can do anything, so one shared 5 s ceiling is arguably not the right fit.
Honest measurement: on my machine the `Add-Type` compile takes 0.22–0.26 s warm, so I cannot claim the 5 s ceiling is routinely exceeded here — this is a robustness suggestion, not a reproduced failure. The reason I still raise it: a cold `csc.exe` invocation on a loaded or AV-scanning corporate machine is a plausible tail, and the failure path is cheap — a missing credential returns immediately, so a wider Windows-only ceiling (e.g. 30 s) costs nothing when there is nothing to find.
Feel free to treat this second point as optional; the CS0542 rename is the actual bug.
Why I'm asking for this upstream
I currently carry both changes as a local post-install patch script that has to be re-run after every upgrade — including after a plugin update silently reverted them, which is how I found this. Fixing them in the shipped package removes that step for every Windows user, not just me.
Happy to open a PR for the rename if that's easier.
---
Related: I've filed a separate issue for an unrelated defect on the same install — the observation generator spawns, answers the INIT prompt, then emits nothing until the 180 s idle watchdog kills it and discards the queued batch. The two are independent; the credential shim above is repaired on my machine and the generator stall persists regardless.

ntdatt812:
Closes #3722.

`Add-Type -Namespace ClaudeMem -Name CredRead` generates the C# class `ClaudeMem.CredRead`, and that class declares `public static extern bool CredRead(...)`. C# rejects that:

> **CS0542**: 'CredRead': member names cannot be the same as their enclosing type

So the type never compiled, `[ClaudeMem.CredRead]::CredRead($t, 1, 0, [ref]$ptr)` threw on every candidate target, and two separate settings hid it — `-ErrorAction SilentlyContinue` on the `Add-Type` and `$ErrorActionPreference = 'SilentlyContinue'` at the top of the script. The script then reached `exit 0` with empty stdout, which the caller reads as:

```
Windows Credential Manager has no entry for "Claude Code-credentials"
```

That is a **wrong** answer rather than a missing one, and it is the reason this could sit unnoticed: the Windows keychain path reported itself as working correctly on a machine that has the credential.

## Verified on a real Windows host

Windows 11, PowerShell 5.1. The reporter's minimal case:

```
A (type name == member name):  FAILED -> (9,27): error CS0542: 'CredRead': member names
                                        cannot be the same as their enclosing type
B (type renamed):              COMPILED
```

And end to end, feeding the **generated** shim block from `buildWindowsCredentialScript()` into PowerShell after this change:

```
FIXED SHIM: COMPILED -> ClaudeMem.CredApi
```

I ran only the `Add-Type` block, not the credential loop — no secret was read.

## What changed

- The generated type is `CredApi`, distinct from every member it declares. All four references move with it.
- `Add-Type` runs with `-ErrorAction Stop` inside a `try/catch` that prints `__CLAUDEMEM_CRED_SHIM_ERROR__ <message>`. The caller recognises that marker and returns *"Windows Credential Manager could not be queried: the CredRead shim failed to compile (…)"* with a `logger.warn`, instead of claiming there is no entry. The rename fixes today's failure; this is what stops the next one from being invisible.
- The snippet moves into an exported `buildWindowsCredentialScript(username)`. Nothing else about the read path, the candidate target names, or the quoting changes.

## Tests

`tests/shared/windows-credential-shim.test.ts` reads the generated script rather than running PowerShell, so it holds on the Linux CI runner too:

- the declared type name is not among the members the shim declares — the CS0542 condition itself, not the string `CredApi`, so any future rename stays covered
- every `[ClaudeMem.X]` reference names the type actually declared, so a half-done rename fails
- the script carries the failure marker and `-ErrorAction Stop`
- `o'brien` is still escaped to `o''brien`

**Mutation-checked** — restoring `-Name CredRead` and its references fails exactly the CS0542 case:

```
3 pass, 1 fail — "does not name the generated type after one of its own members (CS0542)"
```

```
bun test tests/shared/oauth-token.test.ts tests/shared/windows-credential-shim.test.ts
22 pass / 0 fail
```

`npx tsc --noEmit`: no errors.

Note: `plugin/scripts/worker-service.cjs` is a build artefact of this source, so it picks the fix up on the next bundle; I have not hand-edited it.

### DIFF
--- a/src/shared/oauth-token.ts
+++ b/src/shared/oauth-token.ts
@@ -113,18 +113,37 @@ async function readMacOsKeychain(): Promise<OAuthTokenResult> {
  * CredRead API. We use a PowerShell snippet that calls CredRead for the
  * common target name patterns Claude Desktop is known to use.
  */
-async function readWindowsCredentialManager(): Promise<OAuthTokenResult> {
+/**
+ * Printed by the snippet below when the P/Invoke shim itself fails to compile.
+ *
+ * Without it a shim failure is indistinguishable from an empty read — the script exits 0
+ * with no output, and the caller reports "no entry for Claude Code-credentials", sending
+ * anyone debugging it to look for a credential that is in fact present.
+ */
+export const WINDOWS_CRED_SHIM_ERROR_MARKER = '__CLAUDEMEM_CRED_SHIM_ERROR__';
+
+/**
+ * Build the PowerShell snippet that reads the OAuth blob out of Credential Manager.
+ *
+ * Exported so the shim can be checked without a Windows host. `Add-Type -Name <X>`
+ * generates a C# class `<X>`, and C# rejects a class containing a member of its own name
+ * with CS0542, "member names cannot be the same as their enclosing type". The shim was
+ * `-Name CredRead` declaring `CredRead`, so the type never compiled, every call threw,
+ * and the two SilentlyContinue settings swallowed it.
+ */
+export function buildWindowsCredentialScript(username: string): string {
   // PowerShell snippet enumerates likely target names and prints the JSON blob.
   // The exact target name on Windows is "Claude Code-credentials" or
   // "Claude Code:credentials" (Claude Desktop uses `${service}:${account}` or
   // `${service}` depending on version). This script tries both.
   // Username is escaped with PowerShell's single-quote convention (' → '') in
   // case future Windows versions or domain-joined machines permit ' in usernames.
-  const psSafeUsername = userInfo().username.replace(/'/g, "''");
-  const psScript = `
+  const psSafeUsername = username.replace(/'/g, "''");
+  return `
     $ErrorActionPreference = 'SilentlyContinue'
     $candidates = @('Claude Code-credentials', 'Claude Code:credentials', 'Claude Code-credentials:${psSafeUsername}')
-    Add-Type -Namespace ClaudeMem -Name CredRead -MemberDefinition @"
+    try {
+      Add-Type -Namespace ClaudeMem -Name CredApi -MemberDefinition @"
       [DllImport("Advapi32.dll", SetLastError=true, CharSet=CharSet.Unicode)]
       public static extern bool CredRead(string target, uint type, uint reservedFlag, out IntPtr CredentialPtr);
       [DllImport("Advapi32.dll", SetLastError=true)]
@@ -137,21 +156,29 @@ async function readWindowsCredentialManager(): Promise<OAuthTokenResult> {
         public uint Persist; public uint AttributeCount; public IntPtr Attributes;
         public string TargetAlias; public string UserName;
       }
-"@ -ErrorAction SilentlyContinue
+"@ -ErrorAction Stop
+    } catch {
+      Write-Output "${WINDOWS_CRED_SHIM_ERROR_MARKER} $($_.Exception.Message)"
+      exit 0
+    }
     foreach ($t in $candidates) {
       $ptr = [IntPtr]::Zero
-      $ok = [ClaudeMem.CredRead]::CredRead($t, 1, 0, [ref]$ptr)
+      $ok = [ClaudeMem.CredApi]::CredRead($t, 1, 0, [ref]$ptr)
       if ($ok) {
-        $cred = [System.Runtime.InteropServices.Marshal]::PtrToStructure($ptr, [Type][ClaudeMem.CredRead+CREDENTIAL])
+        $cred = [System.Runtime.InteropServices.Marshal]::PtrToStructure($ptr, [Type][ClaudeMem.CredApi+CREDENTIAL])
         $bytes = New-Object byte[] $cred.CredentialBlobSize
         [System.Runtime.InteropServices.Marshal]::Copy($cred.CredentialBlob, $bytes, 0, $cred.CredentialBlobSize)
-        [ClaudeMem.CredRead]::CredFree($ptr) | Out-Null
+        [ClaudeMem.CredApi]::CredFree($ptr) | Out-Null
         [System.Text.Encoding]::Unicode.GetString($bytes)
         exit 0
       }
     }
     exit 0
   `.trim();
+}
+
+async function readWindowsCredentialManager(): Promise<OAuthTokenResult> {
+  const psScript = buildWindowsCredentialScript(userInfo().username);
 
   let stdout: string;
   try {
@@ -169,6 +196,19 @@ async function readWindowsCredentialManager(): Promise<OAuthTokenResult> {
     };
   }
   const raw = stdout.trim();
+  if (raw.startsWith(WINDOWS_CRED_SHIM_ERROR_MARKER)) {
+    // The lookup never ran, so nothing here says anything about whether a credential
+    // exists. Reporting "no entry" would be a wrong answer rather than a missing one.
+    const detail = raw.slice(WINDOWS_CRED_SHIM_ERROR_MARKER.length).trim();
+    logger.warn('OAUTH', 'Windows Credential Manager shim failed to compile', {
+      service: KEYCHAIN_SERVICE_NAME,
+      detail,
+    });
+    return {
+      kind: 'absent',
+      reason: `Windows Credential Manager could not be queried: the CredRead shim failed to compile (${detail})`,
+    };
+  }
   if (!raw) {
     return { kind: 'absent', reason: 'Windows Credential Manager has no entry for "Claude Code-credentials"' };
   }
--- a/tests/shared/windows-credential-shim.test.ts
+++ b/tests/shared/windows-credential-shim.test.ts
@@ -1,0 +1,64 @@
+import { describe, it, expect } from 'bun:test';
+import {
+  buildWindowsCredentialScript,
+  WINDOWS_CRED_SHIM_ERROR_MARKER,
+} from '../../src/shared/oauth-token.js';
+
+/**
+ * `Add-Type -Namespace N -Name X` generates the C# class `N.X`. C# rejects a class that
+ * declares a member of its own name with CS0542, "member names cannot be the same as
+ * their enclosing type". The shim was `-Name CredRead` declaring `public static extern
+ * bool CredRead(...)`, so the type never compiled, `[ClaudeMem.CredRead]::CredRead(...)`
+ * threw on every call, and both `-ErrorAction SilentlyContinue` and the script-level
+ * `$ErrorActionPreference` swallowed it — the Windows keychain path was dead code that
+ * reported itself as "no stored credential".
+ *
+ * These read the generated script rather than running PowerShell, so the rule holds on
+ * the Linux CI runner too.
+ */
+
+const script = buildWindowsCredentialScript('someone');
+
+function declaredTypeName(text: string): string {
+  const match = text.match(/Add-Type\s+-Namespace\s+(\w+)\s+-Name\s+(\w+)/);
+  expect(match, 'the script no longer declares a P/Invoke type').not.toBeNull();
+  return match![2];
+}
+
+function declaredMemberNames(text: string): string[] {
+  return [...text.matchAll(/public\s+(?:static\s+extern\s+)?[\w.<>\[\]]+\s+(\w+)\s*\(/g)].map(
+    (m) => m[1],
+  );
+}
+
+describe('Windows credential shim', () => {
+  it('does not name the generated type after one of its own members (CS0542)', () => {
+    const typeName = declaredTypeName(script);
+    const members = declaredMemberNames(script);
+
+    expect(members.length).toBeGreaterThan(0);
+    expect(members).not.toContain(typeName);
+  });
+
+  it('references the type it actually declares', () => {
+    const typeName = declaredTypeName(script);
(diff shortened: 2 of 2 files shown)


ntdatt812:
Correction to my own PR body: the line **"`npx tsc --noEmit`: no errors"** is wrong and I am withdrawing it.

This checkout has no installed `node_modules`, so there is no local `tsc`. Running it now makes that plain:

```
$ bunx tsc --noEmit -p tsconfig.json
error TS2688: Cannot find type definition file for 'bun'.
error TS2688: Cannot find type definition file for 'node'.
```

No typecheck ran, and the earlier line came from misreading tooling output rather than from a real result. Nothing else in the PR body depends on it.

What I did actually run and stand behind:

- `bun test tests/shared/oauth-token.test.ts tests/shared/windows-credential-shim.test.ts` → **22 pass / 0 fail**
- the mutation check (restoring `-Name CredRead`) → 3 pass / 1 fail, the CS0542 case
- the shim block generated by `buildWindowsCredentialScript()` compiled on a real Windows 11 / PowerShell 5.1 host: `FIXED SHIM: COMPILED -> ClaudeMem.CredApi`, against `error CS0542` for the old shape

The change is a rename plus a `try/catch`, with no type-level surface beyond the two new exports, so I do not expect a typecheck to have anything to say — but that is an expectation, not a measurement, and it should not have been written as one.

ntdatt812:
The finding is right and it is the important one: `plugin/scripts/worker-service.cjs` is committed, is what ships, and still carries the broken shim. My PR body said the bundle "picks the fix up on the next bundle" — that is true of a release build, but it does not describe this repo, where fix commits carry the rebuilt bundle with them (`3c5d9b0`, `3bf1914`, `a38d511` all touch it). So the source-only diff does not reach a single Windows user, and I should have checked that convention before writing that line.

I did the rebuild to see what committing it would actually mean, and it is not a clean act right now.

**Rebuilding on clean `main`, with none of my changes, already rewrites four files:**

```
plugin/scripts/mcp-server.cjs     |  80 ++--
plugin/scripts/server-service.cjs | 491 +++++++++++++--------
plugin/scripts/worker-service.cjs | 873 ++++++++++++++++++++------------------
plugin/ui/viewer-bundle.js        |  24 +-
4 files changed, 834 insertions(+), 634 deletions(-)
```

So the committed bundle is stale against `main`'s own source, independently of this PR.

**And it is not just minifier churn.** Same version stamp, very different size:

```
committed worker-service.cjs : 2,599,208 bytes
rebuilt from clean main      : 3,017,291 bytes   (+418 KB, +16%)
```

The build says so itself:

```
⚠️  worker-service.cjs is 2946.98 KB (advisory budget 2900 KB). If this jumped
    unexpectedly, check whether a server-only dependency leaked into the worker
    bundle (see #2584).
```

A rebuild today trips the repo's own size guard and points at exactly the #2584 leak. That is a real question about `main`, and it should not arrive inside a two-file credential fix where nobody can review it — the bundle is minified with unstable identifier names, so even two builds of identical source differ in hundreds of lines.

**What I verified about the fix itself in the rebuilt bundle** (grep, not inference):

```
Name CredRead            → 0
Name CredApi             → 1
__CLAUDEMEM_CRED_SHIM_ERROR__ → 1
```

So the generator change does flow into the bundle correctly; the only question is when the bundle gets committed.

**Your call, and I will do either:**

1. **You regenerate** as its own commit, so the +418 KB and the #2584 question get looked at on their own merits, and this PR stays the two source files plus tests.
2. **I push the full rebuild here** — say the word and it is one commit. I would rather you choose that knowingly than find 800 lines of someone else's unbundled work in a PR titled "make the Windows CredRead shim compile".

Same question applies to #3727, where the shipped bundle likewise has no `-WorkingDirectory`; I have raised it there rather than waiting to be asked.

ntdatt812:
Flagging an overlap I should have caught before opening this: **#3723 fixes the same issue and was opened about six hours earlier.** I found #3722 by reading the issue list and did not check for an existing PR against it. That is on me.

If you merge #3723, this can be closed — the rename is the core of both, and theirs also updates the generated worker bundle, which mine deliberately does not (I set out why in the comment above: rebuilding today pulls in ~834 lines of unrelated drift and a +418 KB jump that trips the build's own size guard).

Two things here are not in #3723, in case you want them either way — I am happy to reduce this PR to just them on top of theirs:

1. **A failed shim is no longer reported as "no entry".** #3723 keeps `-ErrorAction SilentlyContinue`, so the *next* compile failure in that block is silently swallowed exactly as CS0542 was, and the caller again says `Windows Credential Manager has no entry for "Claude Code-credentials"` — a wrong answer rather than a missing one. This PR runs `Add-Type` with `-ErrorAction Stop` inside a `try/catch` that emits `__CLAUDEMEM_CRED_SHIM_ERROR__`, which the reader turns into *"could not be queried: the CredRead shim failed to compile"*. The rename fixes today's bug; this is what stops the next one being invisible.

2. **The test asserts the CS0542 condition, not the new name.** Mine extracts the type name from the generated script and asserts it is not among the members that script declares — so any future rename stays covered, and a half-done rename (type renamed, references not) fails too. A test pinning the literal `CredApi` passes for a script that no longer compiles for some other reason.

No objection to theirs going in first; it got there first.

stantheman0128:
Support for the extras. Yield the rename itself to #3723.

#3723 already ships CredApi in `worker-service.cjs`, which is what Windows actually runs. Leaving the bundle alone here is the right call given the +418 KB / #2584 rebuild on current main.

The piece I would not drop: a failed Add-Type must not look like "no credential". SilentlyContinue is why CS0542 sat in production as a quiet miss. `__CLAUDEMEM_CRED_SHIM_ERROR__` plus ErrorAction Stop makes the next compile failure visible. Tests that assert the CS0542 condition (type name equals a declared method) rather than the string CredApi survive a future rename.

I would rather see those land on top of #3723 than close this as a straight duplicate.

Flynn-Kawolsky:
+1 to what's already in this thread — #3723 carries the regenerated `worker-service.cjs` (the artifact Windows actually runs), so that looks like the one to merge first, with this PR's compile-failure marker and stronger CS0542 test landing on top afterward. Mainly flagging so this doesn't sit open in parallel with #3723 waiting on the same review — commented there too.

ntdatt812:
Agreed, and thanks for cross-posting — that is the right merge order. #3723 should go first: it carries the regenerated `plugin/scripts/worker-service.cjs`, which is the file Windows actually executes, and this PR deliberately does not touch it.

One note on why I have not already stripped this down to just the extras, since that is the obvious next question. The CS0542 test here asserts the *invariant* — that the generated type is not named after one of its own members — rather than the string `CredApi`. On current `main` that invariant is still violated, so an extras-only PR would sit red until #3723 lands, which reads worse than it is and gives a reviewer one more thing to interpret. Keeping the rename means this branch is green and merges correctly in either order.

So, concretely, whichever is easier for you:

- **Merge #3723 first** (my preference). Say the word here and I will force-push this branch the same day down to just the two pieces nobody has claimed — the `-ErrorAction Stop` + `__CLAUDEMEM_CRED_SHIM_ERROR__` marker and the caller message that reports a *failed query* instead of *no credential*, plus the CS0542 regression test. No rename, no bundle, and it applies cleanly on top of theirs.
- **Merge this one first** and #3723 rebases to just the bundle regeneration.

Either way there is no reason for both to wait on the same review. The substantive piece I would not want dropped in the shuffle is the marker: `-ErrorAction SilentlyContinue` is the reason a compile error sat in production looking like an empty keychain, and a rename fixes today's instance of that without fixing the class.

thedotmack:
NIGHT SHIP Batch F — Gate **PASS** (compile-only CredRead CS0542 rename + shim-failure marker). Cherry-picks onto current `main`. Queued behind #4008 / #3689; will squash-merge after that lands. No version bump.

Issue #3195, opened 2026-07-11
[Windows] Worker never starts when %USERPROFILE% contains a space — unquoted Start-Process -ArgumentList still present in v13.10.2 (bug 1 of #2521)


smurfspec:
## Summary

On Windows, the worker daemon dies instantly at spawn when `%USERPROFILE%` contains a space (e.g. `C:\Users\Anıl pc`). This is **bug 1 of #2521** (`-ArgumentList` does not preserve spaces through to `CreateProcess`), which was consolidated into #2378 and closed as shipped in v13.4.0 — but the unquoted-argument spawn is **still present in the shipped artifacts of v13.10.2**. Verified and reproduced today on a fresh install.

## Environment

- claude-mem **v13.10.2** (fresh install via `npx claude-mem install`)
- Windows 11 Pro (10.0.26200), Windows PowerShell 5.1
- Bun 1.3.14, Node 22.20.0
- `%USERPROFILE%` = `C:\Users\Anıl pc` (contains a space; also a non-ASCII character, see notes)

## Symptom

`npx claude-mem start` returns `{"status":"ready","message":"Worker started; still warming up"}`, but no bun process ever appears and the health endpoint never comes up. `~/.claude-mem/logs/claude-mem-*.log`:

```
[INFO ] [SYSTEM] Starting worker daemon {workerScriptPath=C:\Users\Anıl pc\.claude\plugins\marketplaces\thedotmack\plugin\scripts\worker-service.cjs}
[WARN ] [SYSTEM] Worker spawned but health endpoint not responding within window — likely still starting in background
```

…then after the cooldown gate: `{"status":"error","message":"Failed to start worker"}`. No worker log is produced anywhere (the spawn uses `stdio: 'ignore'` + `-WindowStyle Hidden`, so bun's startup error is swallowed — same observability gap described in #2521).

## Root cause

`src/services/infrastructure/ProcessManager.ts`, `spawnDaemon()`, win32 branch:

```ts
const psScript = `Start-Process -FilePath '${runtimePath.replace(/'/g, "''")}' -ArgumentList @('${scriptPath.replace(/'/g, "''")}','--daemon') -WindowStyle Hidden`;
const encodedCommand = Buffer.from(psScript, 'utf16le').toString('base64');
execSync(`powershell -NoProfile -EncodedCommand ${encodedCommand}`, { stdio: 'ignore', windowsHide: true, env });
```

[...]

PR #3727, opened 2026-08-25
fix(daemon): stop the worker holding the user's project as its cwd (#3706)

### LINKED ISSUE #3706: Windows: worker daemon inherits project folder as CWD, permanently locking it (folder cannot be renamed or moved)
## Summary

On Windows, the worker daemon (`bun.exe` running `worker-service.cjs --daemon`) is spawned without an explicit working directory, so it inherits the project folder the Claude Code session was launched from. Windows keeps an open handle on a process's current working directory, so that project folder becomes **permanently locked** for as long as the daemon runs — any attempt to rename or move it fails with:

```
Rename-Item : The process cannot access the file because it is being used by another process.
```

Since the daemon is a persistent background service, the lock effectively never releases until the user hunts down and kills `bun.exe`.

## Environment

- claude-mem: 12.7.5 (plugin cache install, `thedotmack` marketplace)
- OS: Windows 11 Home 10.0.26200
- Runtime: Bun at `~\.bun\bin\bun.exe`
- Claude Code CLI on Windows (PowerShell)

## Steps to reproduce

1. On Windows, start a Claude Code session in some project folder, e.g. `D:\Projects\my-folder` (claude-mem hooks start the worker daemon).
2. Exit the session.
3. Try to rename the folder: `Rename-Item D:\Projects\my-folder new-name`.
4. It fails with "being used by another process". Sysinternals `handle64.exe D:\Projects\my-folder` shows the holder:

```
bun.exe   pid: 8860   type: File   54: D:\Projects\my-folder
```

with command line `"...\bun.exe" ...\scripts\worker-service.cjs --daemon`.

## Root cause

The Windows daemon-spawn path builds a PowerShell command with no working directory:

```js
// worker-service.cjs (and the same pattern in mcp-server.cjs)
let l = `Start-Process -FilePath '${...}' -ArgumentList @('${...}','--daemon') -WindowStyle Hidden`;
```

`Start-Process` without `-WorkingDirectory` inherits the caller's CWD — which is the project folder, because the daemon is (re)spawned by hooks running in the session's working directory. The non-Windows `child_process.spawn(..., {detached: true, stdio: "ignore", ...})` daemon-spawn calls also omit `cwd`, so POSIX daemons hold the project directory too (less visible there, since renaming a directory held as a process CWD is allowed on POSIX — but e.g. unmounting a volume is not).

For comparison, `worker-cli.js` already passes a `cwd` when spawning — it's only these daemonize paths that miss it.

## Suggested fix

Pin the daemon's working directory to the user's home directory at every daemonize site:

- Windows: add `-WorkingDirectory $env:USERPROFILE` to the `Start-Process` command in `worker-service.cjs` and `mcp-server.cjs`.
- POSIX/lazy spawn paths: add `cwd: os.homedir()` to the `spawn(..., {detached: true, ...})` options.

I patched my local install this way and confirmed the daemon now starts with CWD `C:\Users\<user>\` even when launched from inside a project folder, and project folders rename freely while it runs.

## Impact

Any Windows user who tries to rename, move, or delete a project folder they've ever opened Claude Code in will hit an unexplained folder lock, with nothing pointing at claude-mem as the cause.

ntdatt812:
Closes #3706.

A process holds an open handle on its working directory. The daemon spawns are reached from hooks, which run in the session's project folder, and none of the three spawn sites set a directory — so the daemon inherited the project folder. On Windows that locks it: rename and move fail for as long as the process lives, and a daemon outlives the session that started it.

## Measured on Windows 11

The mechanism, isolated from claude-mem entirely — one process, one directory, nothing else:

```
HELD:   rename FAILED -> The process cannot access the file because it is being used by another process.
EXITED: rename SUCCEEDED
```

A process started with `-WorkingDirectory <dir>` blocks `Rename-Item <dir>` with exactly the message from the report, and the same rename succeeds the moment it exits. That is the whole bug: the daemon simply never exits.

## Three spawn sites, all of them

| site | before |
|---|---|
| `ProcessManager.spawnDaemon` — Windows | `Start-Process … -WindowStyle Hidden`, no `-WorkingDirectory` |
| `ProcessManager.spawnDaemon` — POSIX | `spawnHidden(…, { detached, stdio, env })`, no `cwd` |
| `worker-utils` lazy-spawn | `spawnHidden(…, { detached, stdio })`, no `cwd` |
| `ServerService.spawnServerDaemon` | `spawn(…, { detached, stdio, windowsHide })`, no `cwd` |

The issue names the first and notes the POSIX ones are the same pattern; I fixed all of them rather than the Windows one alone, because the POSIX daemons still pin the directory against unmount — quieter, but the same handle.

All now stand in claude-mem's own data directory, via a small `daemonWorkingDirectory()` that carries the reason. It always exists by the time a daemon starts and is never a directory the user is reorganising.

## Tests

`tests/infrastructure/process-manager.test.ts`, three added under `daemon working directory (#3706)`:

- the built command carries `-WorkingDirectory` and names the data directory
- a `'` in the directory is doubled for PowerShell, like the two paths already are
- `daemonWorkingDirectory()` is the data dir and is **not** `process.cwd()` — the property that actually matters, stated as such

Two existing `#3195` assertions pin the exact command string, so they move with the change; one now passes an explicit directory and the other keeps using the default, which is why the default stays covered.

**Mutation-checked** — dropping `-WorkingDirectory` from the builder fails four cases:

```
51 pass, 5 fail   (4 mine + the pre-existing one below)
```

both new command-shape cases and both `#3195` string assertions, which is the right blast radius for a change to that one string.

```
bun test tests/infrastructure/process-manager.test.ts
55 pass / 1 fail
```

The single failure is `resolveWorkerRuntimePath > should look up Bun on non-Windows when caller is Node`, which fails identically on `main` on this machine — it asserts non-Windows lookup behaviour and I am on Windows. Measured on a clean checkout before this branch, not assumed.

I did not run a typecheck: this checkout has no installed `node_modules`, so `tsc` cannot resolve the `bun` and `node` type libraries. Saying so rather than implying a green check.

### DIFF
--- a/src/server/runtime/ServerService.ts
+++ b/src/server/runtime/ServerService.ts
@@ -862,6 +862,9 @@ function spawnServerDaemon(port: number): number | undefined {
     detached: true,
     stdio: 'ignore',
     windowsHide: true,
+    // Never the caller's directory: a daemon holds its cwd open for its whole life, and
+    // on Windows that locks the folder against rename or move (#3706).
+    cwd: paths.dataDir(),
     // Strip host CLI bleed-through (CLAUDE_CODE_*, including EFFORT_LEVEL) and
     // Anthropic credentials before handing env to the detached daemon. The
     // daemon re-reads credentials from ~/.claude-mem/.env at SDK spawn time.
--- a/src/server/runtime/ServerService.ts
+++ b/src/server/runtime/ServerService.ts
@@ -858,6 +858,10 @@ function getServerPort(): number {
 
 function spawnServerDaemon(port: number): number | undefined {
   const scriptPath = typeof __filename !== 'undefined' ? __filename : fileURLToPath(import.meta.url);
+  // A cwd that does not exist makes spawn fail with ENOENT, and paths.ts resolves
+  // DATA_DIR without creating it — so create it rather than depend on some earlier
+  // caller having done so. Idempotent.
+  mkdirSync(paths.dataDir(), { recursive: true });
   const child = spawn(process.execPath, [scriptPath, '--daemon'], {
     detached: true,
     stdio: 'ignore',
--- a/src/services/infrastructure/ProcessManager.ts
+++ b/src/services/infrastructure/ProcessManager.ts
@@ -369,7 +369,14 @@ function executeCwdRemap(dbPath: string, effectiveDataDir: string, markerPath: s
  * the time a daemon starts and is never a directory the user is reorganising.
  */
 export function daemonWorkingDirectory(): string {
-  return paths.dataDir();
+  const dir = paths.dataDir();
+  // Created here rather than assumed: a cwd that does not exist makes spawn fail with
+  // ENOENT and Start-Process fail outright, so passing one turns a first run on a fresh
+  // install into a launch failure. paths.ts resolves DATA_DIR but does not create it --
+  // today something else happens to create it first, which is a coupling this must not
+  // depend on. mkdir -p is idempotent, so the usual case costs one stat.
+  mkdirSync(dir, { recursive: true });
+  return dir;
 }
 
 export function buildWindowsDaemonStartCommand(
--- a/src/services/infrastructure/ProcessManager.ts
+++ b/src/services/infrastructure/ProcessManager.ts
@@ -357,7 +357,26 @@ function executeCwdRemap(dbPath: string, effectiveDataDir: string, markerPath: s
   }
 }
 
-export function buildWindowsDaemonStartCommand(runtimePath: string, scriptPath: string): string {
+/**
+ * Where a detached daemon should stand, which is anywhere but the user's project.
+ *
+ * A process holds an open handle on its working directory. On Windows that makes the
+ * directory unrenamable and unmovable for the daemon's whole lifetime, and the daemon
+ * outlives the session that spawned it -- so a project folder became permanently locked
+ * with "The process cannot access the file because it is being used by another process"
+ * until the user found and killed bun.exe (#3706). POSIX allows the rename but still
+ * pins the directory against unmount. claude-mem's own data directory always exists by
+ * the time a daemon starts and is never a directory the user is reorganising.
+ */
+export function daemonWorkingDirectory(): string {
+  return paths.dataDir();
+}
+
+export function buildWindowsDaemonStartCommand(
+  runtimePath: string,
+  scriptPath: string,
+  workingDirectory: string = daemonWorkingDirectory()
+): string {
   const psSingleQuote = (value: string) => value.replace(/'/g, "''");
   // Windows PowerShell 5.1 joins -ArgumentList elements with spaces WITHOUT
   // quoting them when it builds the child's native command line, so a script
@@ -366,7 +385,7 @@ export function buildWindowsDaemonStartCommand(runtimePath: string, scriptPath: 
   // double quotes inside the single-quoted PS string keeps the path a single
   // argument. -FilePath is safe as-is: it is a single-string parameter and
   // never goes through that join.
-  return `Start-Process -FilePath '${psSingleQuote(runtimePath)}' -ArgumentList @('"${psSingleQuote(scriptPath)}"','--daemon') -WindowStyle Hidden`;
+  return `Start-Process -FilePath '${psSingleQuote(runtimePath)}' -ArgumentList @('"${psSingleQuote(scriptPath)}"','--daemon') -WorkingDirectory '${psSingleQuote(workingDirectory)}' -WindowStyle Hidden`;
 }
 
 export const WORKER_BOOT_PROBE_TIMEOUT_MS = 5000;
@@ -520,6 +539,7 @@ export function spawnDaemon(
   const child = spawnHidden(execPath, args, {
     detached: true,
     stdio: 'ignore',
+    cwd: daemonWorkingDirectory(),
     env
   });
 
--- a/src/shared/worker-utils.ts
+++ b/src/shared/worker-utils.ts
@@ -687,6 +687,9 @@ export async function ensureWorkerRunning(): Promise<boolean> {
       logger.info('SYSTEM', 'Worker not running — lazy-spawning', { runtimePath, scriptPath });
 
       try {
+        // A cwd that does not exist makes spawn fail with ENOENT, and paths.ts resolves
+        // DATA_DIR without creating it. Idempotent, so the usual case costs one stat.
+        mkdirSync(DATA_DIR, { recursive: true });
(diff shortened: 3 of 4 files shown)


ntdatt812:
Thanks for testing the missing-data-directory case rather than assuming it — that was the right thing to probe, since a `cwd` that does not exist makes `spawn` fail with ENOENT, and it would have turned this fix into a launch failure on first run.

**Raising something the review did not flag, because it applies here too.** The same shipping gap found on #3726 exists on this PR: `plugin/scripts/worker-service.cjs` is committed and is what runs, and the shipped copy still has no working directory —

```
$ grep -c "WorkingDirectory" plugin/scripts/worker-service.cjs
0
$ grep -o "Start-Process -FilePath[^\"]\{0,60\}" plugin/scripts/worker-service.cjs
Start-Process -FilePath '${r(t)}' -ArgumentList @('
```

So as it stands the source is fixed and the daemon users actually launch is not.

I did not commit a rebuilt bundle, for the reason I set out on #3726 in more detail: rebuilding on **clean `main`**, with none of my changes, already rewrites four files (834 insertions / 634 deletions) and grows `worker-service.cjs` by 418 KB, which trips the build's own advisory size guard and its `#2584` "server-only dependency leaked into the worker bundle" hint. That is a question about `main`, not about this change, and burying it in a daemon-cwd fix would make both unreviewable.

Happy to push the rebuild here the moment you would rather have it in this PR than as its own commit.

**One thing worth pinning while the cwd is on the table.** Your test exercised the server daemon; the branch changes three spawn sites, and the other two are the ones a hook reaches:

| site | spawn |
|---|---|
| `ProcessManager.spawnDaemon` (Windows) | `Start-Process … -WorkingDirectory` |
| `ProcessManager.spawnDaemon` (POSIX) | `spawnHidden(…, { cwd })` |
| `worker-utils` lazy-spawn | `spawnHidden(…, { cwd })` |
| `ServerService.spawnServerDaemon` | `spawn(…, { cwd })` |

All four take `DATA_DIR`, which `paths.ts` creates, so the ENOENT case you probed cannot arise from a fresh install — but if you would like that stated as a test rather than as a property of `paths.ts`, I will add one asserting the directory exists before any spawn uses it. It is the kind of coupling that holds today and quietly stops holding later.

ntdatt812:
Correcting myself, and it turned out to matter.

In my last comment I wrote that all four spawn sites take `DATA_DIR`, "which `paths.ts` creates". That is **wrong**. `paths.ts` resolves `DATA_DIR` and exports an `ensureDir(dirPath)` helper, but it never calls it for `DATA_DIR` itself — creation is left to whichever caller gets there first.

That makes the failure you probed a real one rather than a hypothetical, just reached by a different route than "the directory is missing at test time":

- `spawn(..., { cwd })` fails with **ENOENT** when the cwd does not exist
- `Start-Process -WorkingDirectory` refuses outright

Before this branch, no daemon passed a cwd, so a missing data directory cost nothing. My change made the daemons depend on it existing — so on a fresh install, if a hook reaches a daemon spawn before anything has created `~/.claude-mem`, the fix for a locked folder becomes a daemon that will not start. Your test passed because something did create it first in that run; that ordering is not a guarantee, and I asserted it as one.

Fixed in `6de9ca2`: all three spawn sites `mkdirSync(..., { recursive: true })` before using the directory as a cwd. Idempotent, so the normal case is one `stat`.

Pinned by a test that deletes the directory first, which is the state the assertion is actually about:

```ts
rmSync(DATA_DIR, { recursive: true, force: true });
expect(existsSync(DATA_DIR)).toBe(false);
const dir = daemonWorkingDirectory();
expect(existsSync(dir)).toBe(true);
```

**Mutation-checked** — removing the `mkdirSync` from `daemonWorkingDirectory()` fails exactly that case and nothing else.

```
bun test tests/infrastructure/process-manager.test.ts
56 pass / 1 fail    (three consecutive runs, identical)
```

The one failure is still `resolveWorkerRuntimePath > should look up Bun on non-Windows when caller is Node`, which fails the same way on `main` on this Windows machine.

Thank you for probing the missing-directory case — it is the reason I went back and read `paths.ts` instead of trusting my own sentence.

stantheman0128:

A process holding its cwd open is the same class of Windows handle leak as a LISTEN socket that outlives the PID. We have seen the rename lock on this box in other tools; the isolated Start-Process repro in the PR body is the right height.

Fixing every spawn site (ProcessManager Windows/POSIX, hook lazy-spawn, server daemon) is the class, not a single Start-Process flag. Complements #3529 (hidden window) and #3309 (socket inheritance / spawn locks). Those are different handles. Keep all three.

Creating the data dir before passing it as cwd is load-bearing. Before this PR a missing ~/.claude-mem was harmless because nothing passed cwd; after it, Start-Process and spawn both refuse. The rmSync test is the one that would go red if that mkdir disappeared.

The shipped-bundle gap is the same one you already flagged on #3726. I would rather see CredApi and this cwd fix land in source than wait on the #2584 rebuild.

thedotmack:

Real Windows cwd-lock bug, but not a safe night-ship: CONFLICTING, four spawn sites, and the author notes the shipped `plugin/scripts/worker-service.cjs` still inherits the project cwd. Source-only would not reach users without a bundle rebuild (out of scope).

ntdatt812:
Rebased onto current `main`, now mergeable. Taking the three gate points in turn:

- **Conflict:** it was only in `tests/infrastructure/process-manager.test.ts`, where #3894's `probeWorkerBootFailure` block and this PR's working-directory block were both added at the same spot. Both are kept. All three source files merged cleanly, and the diff is still the original four files.
- **Spawn sites:** every place that starts the long-lived worker now gets an explicit working directory instead of inheriting the hook's cwd, and that directory is created before it is handed to spawn, since a missing cwd makes spawn fail with `ENOENT`. The sites are the POSIX `spawnDaemon` path, the Windows `Start-Process -WorkingDirectory`, the `ServerService` launcher and the `worker-utils` launcher. All four use the claude-mem data directory. The two `ProcessManager` sites share `daemonWorkingDirectory()`, while `ServerService` and `worker-utils` resolve the same directory inline (`paths.dataDir()` / `DATA_DIR`). If you would rather have a single place to change, I can route those two through the helper as well.
- **Bundle:** this is source-only, as you said. I left `plugin/scripts/worker-service.cjs` alone because a regenerated bundle in a contributor PR tends to conflict with every other change to it. If you would rather have the rebuilt bundle in this PR, I'll add it.

Checked on the rebased branch: `tests/infrastructure/process-manager.test.ts` 64 pass, 1 fail. The failure (`resolveWorkerRuntimePath` › *should look up Bun on non-Windows*) fails identically on a clean checkout of `main` when run on Windows. `npm run typecheck` is clean for both configurations.

PR #3740, opened 2026-08-26
fix(oauth): scope the keychain lookup to CLAUDE_CONFIG_DIR (#3718)

### LINKED ISSUE #3718: OAuth preflight ignores CLAUDE_CONFIG_DIR: reads hardcoded keychain service "Claude Code-credentials" and flags a stale token that Claude Code no longer uses
## Summary

When Claude Code runs with `CLAUDE_CONFIG_DIR` set to a non-default directory, it stores its OAuth token under a per-config keychain service: `Claude Code-credentials-<sha256(configDir)[:8]>`. The claude-mem worker's OAuth preflight (`worker-service.cjs`, `Eu="Claude Code-credentials"`) always reads the **plain** service name, so it finds whatever old token lives there, decides it is expired, writes `oauth-stale.marker`, and every session start shows:

> Claude Desktop OAuth token has expired — re-login via Claude Desktop to refresh

Re-logging in to Claude Desktop does nothing, because that is not where the token comes from.

## Environment

- claude-mem 13.15.3 (latest), macOS 25.5 (Darwin), bun worker
- `CLAUDE_MEM_PROVIDER=claude`, `CLAUDE_MEM_CLAUDE_AUTH_METHOD=cli`
- `CLAUDE_CONFIG_DIR=/Users/<me>/.claude_b` — and the worker process **does** have this env var (verified with `ps eww`)

## Repro

```bash
export CLAUDE_CONFIG_DIR=~/.claude_b   # any non-default dir, log in to Claude Code
H=$(printf "%s" "$CLAUDE_CONFIG_DIR" | shasum -a 256 | cut -c1-8)
security find-generic-password -s "Claude Code-credentials-$H" -w | jq .claudeAiOauth.expiresAt   # fresh
security find-generic-password -s "Claude Code-credentials" -w     | jq .claudeAiOauth.expiresAt   # old / expired / absent
```

Worker log shows `[OAUTH] ... expired` / marker written, while the plain item is the one being read.

## Observed behavior

- Warning banner on every session start; `observer-health.json` reports the provider as failing.
- Note: actual observation generation was still working in my case (the spawned `claude` subprocess authenticates itself through `CLAUDE_CONFIG_DIR`), so the preflight was a false positive rather than a hard outage. The banner is misleading enough that it sent me down the "re-login Desktop" path first.

## Expected

The preflight should resolve the service name the same way Claude Code does: if `CLAUDE_CONFIG_DIR` is set (and differs from `~/.claude`), try `Claude Code-credentials-<sha256(configDir)[:8]>` first, then fall back to the plain name. A read-only reference in the same spirit is Claude Code's own `.credentials.json` / keychain fallback logic.

## Workaround

Copy the fresh token into the plain service and restart the worker:

```bash
TOKEN=$(security find-generic-password -s "Claude Code-credentials-$H" -w)
security add-generic-password -U -s "Claude Code-credentials" -a "$USER" -w "$TOKEN"
rm -f ~/.claude-mem/oauth-stale.marker
bun <plugin>/scripts/worker-service.cjs restart
```

Log then shows `Injected fresh CLAUDE_CODE_OAUTH_TOKEN at spawn-time`. It goes stale again once Claude Code rotates the hashed item.

## Related

- #3607 (explicit auth source for the observer) — this is one concrete case for that plan.
- I also hit a `FOREIGN KEY constraint failed` retry loop on a session that survived a worker restart (same shape as #3681, `observations.memory_session_id`); a worker restart cleared it. Happy to open that separately with logs if useful.

ntdatt812:
Closes #3718.

`KEYCHAIN_SERVICE_NAME` was a single constant used by all three platform readers, so the preflight always looked under the plain `Claude Code-credentials`. With a non-default `CLAUDE_CONFIG_DIR` that is not where the live token is, so the read found whatever older entry happened to sit there, judged it expired, wrote `oauth-stale.marker`, and every session start warned *"re-login via Claude Desktop"* — about a token that is not the one in use. Re-logging in cannot fix that, which is the part that sends people down the wrong path.

All three readers now take a candidate list, most specific first:

```ts
export function keychainServiceNames(env = process.env, home = homedir()): string[]
```

- no `CLAUDE_CONFIG_DIR`, or it points at `~/.claude` → `["Claude Code-credentials"]`, byte-identical to today
- otherwise → `["Claude Code-credentials-<sha256(dir)[:8]>", "Claude Code-credentials"]`

macOS and Linux loop the list and take the first entry that yields a value; a miss on the scoped name is ordinary rather than an error, so only an all-candidates failure reports `absent`. Windows already enumerated candidates in its PowerShell snippet, so the scoped names go on the front of that array, quoted with the same `''` doubling the username already uses.

**The plain name is never dropped.** A default install reads exactly what it read before, and a machine whose entry is unscoped is still found. That is deliberate: it makes this change unable to regress anyone even if the scoping rule is wrong.

## What I verified, and what I did not

Verified here: the name derivation and ordering (the tests below), that a default or absent `CLAUDE_CONFIG_DIR` produces the unchanged single-name behaviour, and that the existing `oauth-token` suite still passes.

**Not verified here:** that Claude Code's actual suffix is `sha256(configDir)[:8]`. That formula comes from the issue, where it is given with a runnable derivation (`printf "%s" "$CLAUDE_CONFIG_DIR" | shasum -a 256 | cut -c1-8`) and a `security find-generic-password` on both names showing fresh-vs-stale. I have no machine here with a non-default `CLAUDE_CONFIG_DIR` and a Claude Code credential to confirm it against — `cmdkey /list` on this box shows no Claude entries at all — so I am taking the reporter's measurement rather than repeating it. If the formula is wrong the scoped lookup simply misses and the fallback restores today's behaviour; if you have a machine in that state, one `security`/`secret-tool` call against the derived name settles it.

## Tests

`tests/shared/keychain-service-names.test.ts`, six cases against the pure resolver, with `env` and `home` injected so nothing depends on the machine running them:

- unset → plain name only
- the default `~/.claude` is treated as unscoped, so the reader is not sent hunting for an entry that never exists
- a non-default dir → scoped first, then plain
- the plain name is always last, so the fallback cannot be dropped by accident
- two different dirs give different suffixes, and the suffix is derived from the configured string
- whitespace-only values are ignored

**Mutation-checked** — returning only the plain name from the scoped branch fails three of the six, and leaves the three that pin the unchanged default-install behaviour green.

```
bun test tests/shared/keychain-service-names.test.ts tests/shared/oauth-token.test.ts
24 pass / 0 fail
```

```
bun test tests/shared/
153 pass / 4 fail   (157 tests)
```

The 4 failures are pre-existing — `bun test tests/shared/` on clean `main` gives 4 fail across 151 tests, the same ones (`findClaudeExecutable` ×2, `extractFirstFile` ×2), none of them touching OAuth. 151 → 157 is exactly the six added here.

Note: `plugin/scripts/*.cjs` is a build artefact and I have not rebuilt it here, for the reason set out on #3726 — a rebuild today carries ~834 lines of unrelated drift and a +418 KB jump that trips the build's own size guard. Happy to include it if you would rather have it in this PR.

### DIFF
--- a/src/shared/oauth-token.ts
+++ b/src/shared/oauth-token.ts
@@ -143,6 +143,22 @@ async function readMacOsKeychain(): Promise<OAuthTokenResult> {
 }
 
 /**
+ * Windows Credential Manager target names to try, most specific first.
+ *
+ * Returned RAW — the caller applies the single PowerShell escaping pass. Escaping a
+ * component here instead would double it: for username `O'Brien` the literal became
+ * `'Claude Code-credentials:O''''Brien'`, which PowerShell parses as
+ * `Claude Code-credentials:O''Brien` — a different target than the stored credential,
+ * so CredRead missed it and the account fell through to the re-login warning.
+ */
+export function windowsCredentialTargets(
+  username: string = userInfo().username,
+  services: string[] = keychainServiceNames(),
+): string[] {
+  return [...services, 'Claude Code:credentials', `Claude Code-credentials:${username}`];
+}
+
+/**
  * Windows: Credential Manager (DPAPI). Claude Desktop on Windows stores
  * OAuth credentials under a target like "Claude Code:credentials" via the
  * Wincred API. We read it via PowerShell's CredentialManager wrapper.
@@ -157,16 +173,9 @@ async function readWindowsCredentialManager(): Promise<OAuthTokenResult> {
   // The exact target name on Windows is "Claude Code-credentials" or
   // "Claude Code:credentials" (Claude Desktop uses `${service}:${account}` or
   // `${service}` depending on version). This script tries both.
-  // Username is escaped with PowerShell's single-quote convention (' → '') in
-  // case future Windows versions or domain-joined machines permit ' in usernames.
-  const psSafeUsername = userInfo().username.replace(/'/g, "''");
-  // Config-scoped names first, then the plain ones. Each is single-quoted with the
-  // PowerShell doubling convention, same as the username above.
-  const psCandidates = [
-    ...keychainServiceNames(),
-    'Claude Code:credentials',
-    `Claude Code-credentials:${psSafeUsername}`,
-  ]
+  const psCandidates = windowsCredentialTargets()
+    // Sole PowerShell escaping pass: ' → '' inside a single-quoted literal. Targets must
+    // reach here raw — pre-escaping any part of one would double it (see the helper).
     .map(name => `'${name.replace(/'/g, "''")}'`)
     .join(', ');
   const psScript = `
--- a/src/shared/oauth-token.ts
+++ b/src/shared/oauth-token.ts
@@ -11,9 +11,10 @@
 
 import { execFile } from 'child_process';
 import { promisify } from 'util';
+import { createHash } from 'crypto';
 import { existsSync, readFileSync, writeFileSync, mkdirSync, unlinkSync } from 'fs';
-import { userInfo } from 'os';
-import { join } from 'path';
+import { userInfo, homedir } from 'os';
+import { join, resolve } from 'path';
 import { paths } from './paths.js';
 import { logger } from '../utils/logger.js';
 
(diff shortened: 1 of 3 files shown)


ntdatt812:
Thanks — both findings were worth chasing. One is fixed in b5ca03d, the other I'm declining, with evidence.

### P1 · Windows target double-escapes apostrophes — **confirmed, fixed**

You're right, and it was a regression this PR introduced: before my change the username was interpolated straight into the single-quoted literal (one escaping pass, correct); I added a `.map()` that escaped the whole candidate a second time. Reproduced it:

```
username        : O'Brien
CURRENT literal : 'Claude Code-credentials:O''''Brien'
CURRENT parses  : Claude Code-credentials:O''Brien
stored target   : Claude Code-credentials:O'Brien
current matches : false
fixed   matches : true
```

Fixed exactly as you described: targets are now built raw by a new exported `windowsCredentialTargets()` and escaped once at the interpolation site. Added `tests/shared/windows-credential-targets.test.ts`, which pins the contract with a PowerShell literal round-trip (`parse(toLiteral(raw)) === raw`) rather than just spelling out the expected string — that's the assertion a second escaping pass cannot slip past. 10/10 green, `tsc --noEmit` clean.

### P1 · `plugin/scripts/worker-service.cjs` omits the scoped lookup — **true, but not this PR's to fix**

The bundle is a committed release artifact regenerated at version bump, not per source PR. The two most recently merged source PRs are missing from it in exactly the same way:

| symbol | in `src/` | in the committed bundle |
|---|---|---|
| `proTrial` / `ProTrial` (#3613, merged) | 14 | **0** |
| `keychainServiceNames` (this PR) | ✓ | 0 |

`git log -- plugin/scripts/worker-service.cjs` confirms the pattern: the last four touches are `chore: bump version to 13.16.0 / 13.15.3 / 13.15.2 / 13.15.1`, while #3660 and #3613 shipped source-only.

So regenerating it here would (a) fold two other merged PRs' runtime changes into mine, and (b) put a 2.9 MB minified diff in front of the reviewer for a 90-line source fix. @thedotmack — happy to add the regenerated bundle to this branch if you'd rather PRs carry it; just say the word and I'll rebuild before merge. Otherwise it lands with the next version bump like #3613 and #3660 will.

thedotmack:
Night-ship Batch E skipped this as already on main.

#3908 (`aa8ac2e7`) shipped the CLAUDE_CONFIG_DIR-scoped macOS keychain lookup, plus tilde/trailing-slash normalization. Main intentionally leaves Windows/Linux on the bare service name until that suffix scheme is verified on those platforms — so the remaining half of this PR would undo that decision.

thedotmack:
Night-ship Batch E leftovers — Gate **FAIL** / **SKIP**. Leaving open; not rebasing or merging.

The default Claude path is already unchanged on main, and scoped lookup when `CLAUDE_CONFIG_DIR` is set already shipped in #3908 (`deriveMacKeychainServiceName`, tilde/trailing-slash normalization, `CLAUDE_MEM_CLAUDE_CONFIG_DIR` precedence). Main leaves Windows and Linux on the bare `Claude Code-credentials` name on purpose until that suffix scheme is verified on those platforms. Rebasing this PR would re-apply the unverified Windows/Linux hash and fight that decision. The remaining half is therefore not a safe leftover patch.

thedotmack:
Closing night-ship Gate FAIL / superseded (see leftovers-run comment). Not rebasing. Fresh bugs preferred over babysitting this CONFLICTING PR.

Issue #3204, opened 2026-07-12
Windows: worker restart leaves orphaned LISTENING socket, permanently blocking the worker port — every hook then hangs 60–120s


thiagofstp:
## Summary

On Windows, the worker restart flow can kill the worker without a graceful shutdown and fail to spawn its successor. The dead process's socket stays in the TCP table as `LISTENING` (owned by a PID that no longer exists), permanently blocking the configured worker port until reboot. From then on, **every** hook invocation (`UserPromptSubmit`, `PostToolUse`, `PreToolUse` Read) hangs for its full 60–120s timeout on every prompt and tool call, with zero tokens consumed — Claude Code sessions feel frozen.

This happened **twice in two days** on the same machine: port `37778` on 2026-07-11 and port `37779` on 2026-07-12. Each time the only recovery was manually changing `CLAUDE_MEM_WORKER_PORT` in `~/.claude-mem/settings.json` (or rebooting Windows to clear the orphaned socket).

## Environment

- **Claude-mem**: 13.10.2
- **Claude Code**: 2.1.207
- **Node.js**: v24.3.0
- **Bun**: 1.3.13
- **OS**: Microsoft Windows 11 Home 10.0.26200 (x64)
- **Worker**: Not running (port orphaned at time of report)
- **Key settings**: `CLAUDE_MEM_WORKER_PORT: 37780` (already bumped twice as workaround: 37778 → 37779 → 37780)

## Failure sequence (from logs)

1. Worker running fine on port 37779.
2. Plugin triggers an internal restart (`reason=restart`, apparently after a version check).
3. Graceful shutdown fails: `Graceful shutdown failed — proceeding {reason=restart} Server is not running.`
4. Restart successor fails to spawn (`pid=0`).
5. The dead PID keeps the `LISTENING` socket bound in the Windows TCP table.
6. Every new worker start attempt fails with `Is port 37779 in use?`, and every hook client connects to the dead listener's kernel backlog and waits forever — burning the entire hook timeout (60–120s) instead of failing fast.

## Log evidence (`claude-mem-2026-07-12.log`)

```
[2026-07-11 23:52:10.457] [INFO ] [SYSTEM] HTTP server started {host=127.0.0.1, port=37779, pid=26440}
[2026-07-11 23:52:11.304] [INFO ] [SYSTEM] Worker started {host=127.0.0.1, port=37779, pid=26440}
[...]

PR #3747, opened 2026-08-26
fix(worker): detect zombie port and fail fast instead of retry loop

### LINKED ISSUE #3746: Worker stuck in infinite retry loop when OS leaves a zombie LISTEN socket on the worker port (Windows)
## Summary

On Windows, the worker's TCP port can end up in a `LISTEN` state owned by a PID that no longer exists (confirmed via `Get-NetTCPConnection` showing an `OwningProcess`, while `Get-Process -Id <pid>` and `taskkill /PID <pid> /F` both report the process does not exist — a zombie/orphaned OS-level socket). When this happens, claude-mem gets stuck retrying forever and eventually blocks every prompt with:

```
claude-mem worker unreachable for 108 consecutive hooks.
```

## Root cause

1. `isWorkerPortAlive()` (`src/shared/worker-utils.ts`) correctly detects the worker is *not* alive — the HTTP health check to `/api/health` fails (nothing real is listening behind the zombie socket).
2. `ensureWorkerRunning()` then falls through to lazy-spawn a new worker daemon.
3. The new daemon's own duplicate-gate (`isPortInUse()` in `src/services/infrastructure/HealthMonitor.ts`, checked in `src/services/worker-service.ts` around the `--daemon` case) does a raw socket bind probe. On Windows this reports the port as occupied (the zombie LISTEN socket is real at the OS level, just unowned by anything alive), so the new daemon logs `Port already in use, refusing to start duplicate` and immediately `exit(0)`s — believing a sibling worker already owns it.
4. The worker never actually starts. The health check keeps failing. Every hook re-attempts the same doomed spawn. `recordWorkerUnreachable()` increments forever and trips the fail-loud threshold repeatedly, blocking prompts.

Log excerpt from a real occurrence:
```
[ERROR] [SYSTEM] Port in use but worker not responding to health checks
[WARN ] [SYSTEM] Worker failed to start before hook, handler will proceed gracefully
[INFO ] [SYSTEM] Worker not running — lazy-spawning {...}
[INFO ] [SYSTEM] Port already in use, refusing to start duplicate {port=37777}
[WARN ] [SYSTEM] Worker port did not open after lazy-spawn within the cold-boot wait (~15s)
```

`supervisor.json` still named the dead PID, and both `Get-Process` and `taskkill` confirmed it no longer existed — yet `Get-NetTCPConnection -LocalPort 37777` kept reporting it as the `LISTEN` owner. Nothing short of a reboot released the port.

## Manual workaround

Change `CLAUDE_MEM_WORKER_PORT` in `~/.claude-mem/settings.json` to a free port, reset `~/.claude-mem/state/hook-failures.json` to `{"consecutiveFailures":0,"lastFailureAt":0}`, clear `~/.claude-mem/supervisor.json` to `{"processes":{}}`, then restart.

## Proposed fix

PR incoming: in `ensureWorkerRunning()`, before lazy-spawning, check whether the PID file identifies a killable owner. If it doesn't *and* the port is genuinely occupied at the socket level (via the existing `isPortInUse()`), stop and log one clear, actionable error instead of spawning a process that's guaranteed to self-exit without ever binding. This doesn't recover the port automatically (that needs a reboot on Windows in this case) but turns an infinite silent retry loop + generic "unreachable" message into a single clear diagnosis with the actual fix.

## Environment

- OS: Windows 11
- claude-mem: 13.16.0 (also reproduced against 13.15.2/13.15.3 before upgrade)
- Runtime: Bun 1.3.14, Node 22.22.1

DwijeshD:
Fixes #3746

## Problem

On Windows, an OS-level `LISTEN` socket can outlive the process that opened it. When that happens on the worker's port, the worker gets stuck in an infinite silent retry loop that eventually blocks every prompt with `claude-mem worker unreachable for N consecutive hooks.` See #3746 for the full root-cause writeup and log trace.

Short version: `isWorkerPortAlive()` correctly sees the worker as dead (health check fails), so `ensureWorkerRunning()` lazy-spawns a replacement — but the new daemon's own duplicate-gate (`isPortInUse()`) sees the same zombie-occupied port and immediately `exit(0)`s, believing a sibling already owns it. Nothing ever binds. Repeat forever.

## Fix

In `ensureWorkerRunning()` (`src/shared/worker-utils.ts`), before lazy-spawning: if `readOwnedWorkerPidInfo()` can't identify a killable owner *and* `isPortInUse()` says the port is occupied at the socket level, stop and log one clear, actionable error (naming `CLAUDE_MEM_WORKER_PORT` and reboot as the two ways out) instead of spawning a process that is guaranteed to self-exit without ever binding.

This doesn't recover the port automatically — on Windows that specific zombie state only clears on reboot in my testing (`taskkill /F` and PowerShell `Get-Process` both report the PID doesn't exist, yet `Get-NetTCPConnection` keeps reporting it as the socket owner). But it turns an infinite blind retry + generic failure message into a single actionable diagnosis, so the fail-loud threshold reports something useful instead of just a repeat count.

## Testing

- `bunx tsc --noEmit` clean
- Existing worker-utils suites unaffected: `tests/shared/worker-utils-timeout.test.ts`, `tests/shared/worker-utils-version-recycle.test.ts` (7 pass)
- New test added: `tests/shared/worker-utils-zombie-port.test.ts` — reproduces the exact mock conditions (health check always rejects, no owned PID, socket probe reports occupied) and asserts `ensureWorkerRunning()` returns `false` without attempting a spawn.
- Full run: 8 pass, 0 fail across the three files.

I don't have a way to reproduce the underlying Windows zombie-socket condition in CI, so this only covers the code-level branch — the OS-level cause is out of scope for a plugin-level fix.

### DIFF
--- a/src/shared/worker-utils.ts
+++ b/src/shared/worker-utils.ts
@@ -9,7 +9,7 @@ import { loadFromFileOnce } from "./hook-settings.js";
 import { validateWorkerPidFile, readOwnedWorkerPidInfo } from "../supervisor/index.js";
 import { emitBlockingError } from "./hook-io.js";
 import { captureCliEvent } from "../services/telemetry/cli-telemetry.js";
-import { checkVersionMatch } from "../services/infrastructure/index.js";
+import { checkVersionMatch, isPortInUse } from "../services/infrastructure/index.js";
 // Imported from ProcessManager.js directly (not the infrastructure barrel):
 // tests mock the barrel module wholesale, and the resolver must stay real.
 // ProcessManager imports nothing from worker-utils, so no cycle.
@@ -533,6 +533,23 @@ export async function ensureWorkerRunning(): Promise<boolean> {
     // The killed worker's PID file is left behind; the successor's boot
     // removes it (validateWorkerPidFile returns 'stale' for a dead pid).
     // Fall through to (re)spawn + readiness wait below.
+  }
+
+  // Zombie-port guard: the port is occupied at the OS level (a real bind
+  // attempt fails) but nothing behind it answers /api/health, AND the PID
+  // file doesn't name a killable owner. Windows can leave a LISTEN socket
+  // behind for a process that has already died (observed with no
+  // Get-NetTCPConnection-visible owner and taskkill reporting "not found").
+  // Spawning here is doomed: the new daemon's own duplicate-gate
+  // (isPortInUse) will see the same occupied port and immediately exit(0)
+  // without ever binding, so the loop repeats every hook forever. Fail once
+  // with a specific, actionable message instead of retrying blindly.
+  if (readOwnedWorkerPidInfo() === null && (await isPortInUse(getWorkerPort()))) {
+    logger.error('SYSTEM', 'Worker port is occupied by an unreachable, unkillable process (likely a stale OS socket); lazy-spawn would be silently refused on this port', {
+      port: getWorkerPort(),
+      fix: 'Set CLAUDE_MEM_WORKER_PORT to a different port in claude-mem settings, or reboot to release the stuck port',
+    });
+    return false;
   }
 
   const runtimePath = resolveWorkerRuntimePath();
--- a/src/shared/worker-utils.ts
+++ b/src/shared/worker-utils.ts
@@ -607,9 +607,14 @@ export async function ensureWorkerRunning(): Promise<boolean> {
       // exit(0)s without binding — so name the actual fix instead of letting
       // the generic "unreachable" counter climb forever.
       if (readOwnedWorkerPidInfo() === null && (await isPortInUse(getWorkerPort()))) {
+        // Recorded, not just logged: the log file is not where a user looks
+        // when their prompts start getting blocked. recordWorkerUnreachable
+        // reads this so the fail-loud stderr message carries the actual fix
+        // instead of only a failure count.
+        orphanedPortDiagnosis = getWorkerPort();
         logger.error('SYSTEM', 'Worker port is occupied by an unreachable process that no PID file claims (likely an orphaned OS socket); every lazy-spawn on this port will be silently refused', {
-          port: getWorkerPort(),
-          fix: 'Set CLAUDE_MEM_WORKER_PORT to a different port in claude-mem settings, or reboot to release the stuck port',
+          port: orphanedPortDiagnosis,
+          fix: ORPHANED_PORT_REMEDIATION,
         });
       }
       return false;
@@ -629,6 +634,19 @@ export async function ensureWorkerRunning(): Promise<boolean> {
   }
   return true;
 }
+
+const ORPHANED_PORT_REMEDIATION =
+  'Set CLAUDE_MEM_WORKER_PORT to a different port in claude-mem settings, or reboot to release the stuck port';
+
+/**
+ * Port number diagnosed as holding an orphaned OS socket during this hook
+ * process, or null if that condition was never observed. Set by
+ * ensureWorkerRunning and read by recordWorkerUnreachable so the fail-loud
+ * message names the fix. Process-scoped deliberately: it is only ever set
+ * from a fresh probe earlier in the same invocation, so it cannot go stale
+ * across a reboot or a port change the way a persisted flag could.
+ */
+let orphanedPortDiagnosis: number | null = null;
 
 let aliveCache: boolean | null = null;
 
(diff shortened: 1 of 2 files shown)


DwijeshD:
Good catch — the finding is correct and I've pushed a fix (41025b91).

The guard was testing "no owned PID file AND port occupied", which I'd assumed was zombie-specific. It isn't: `server.listen()` runs before `writePidFile()` (`worker-service.ts:421-423`), so a worker that is mid-boot holds the port with no PID file and matches that condition exactly. The guard marked it unavailable, `ensureWorkerAliveOnce` cached the `false` for the rest of the invocation, and hooks failed even once the worker was healthy. Normal concurrent startup would have hit this.

Time is the only thing that separates a warming worker from an orphaned socket, so the diagnosis now runs **after** `waitForWorkerPort`'s ~15.5s cold-boot budget expires — the point where a warming worker has already had its chance to answer. The pre-spawn short-circuit is gone entirely: the spawn and the wait proceed exactly as they did before this PR, and the only behavioural change in the zombie case is which log line is emitted on the already-existing failure path. That also makes the change strictly additive on every path except the one that was already returning `false`.

Test now covers both states. The warming-worker case (health refused on the first probes, healthy after) fails against the previous commit with `Expected: true, Received: false` and passes on the fix, so it's a real regression guard rather than a restatement of the new behaviour. `bunx tsc --noEmit` clean, 9 pass / 0 fail across the three worker-utils suites.

DwijeshD:
Two review comments — one is stale, one was a fair hit. Pushed 8152f260.

**"Warming worker is treated as a zombie port"** — this is against the original diff. That pre-spawn guard was the exact bug, and it was removed in 41025b91; the diagnosis now runs only after `waitForWorkerPort`'s cold-boot budget expires, so a worker that bound before writing its PID file has already had its chance to answer. The suite has a regression test for precisely that interval which fails against the old commit. Believe this one is already resolved, but flag it if the new placement still looks racy to you.

**"Hidden port remediation"** — correct, and it was the more embarrassing half. I was writing the diagnosis to the log file and calling it done, while the channel the user actually sees still said nothing but a failure count. For a condition that no amount of retrying can clear, that's a dead end. `ensureWorkerRunning` now records the diagnosed port and `recordWorkerUnreachable` folds it into `emitBlockingError`, so the fail-loud message names the port and the remediation. Test asserts both reach the blocking-error channel rather than only the logger.

I kept that state process-scoped rather than persisted. It is only ever set from a fresh probe earlier in the same invocation, so it can't go stale across a reboot or a port change.

**On the second half of that suggestion — "retain a fail-fast state while this orphaned-port condition persists" — I did not implement it, deliberately, and want to flag the tradeoff rather than quietly skip it.** Your measurement is right that every fresh hook burns ~15.5s while the port is stuck, which is real cost. But skipping the spawn and wait on the strength of a persisted flag reintroduces the shape of the bug in your first comment: a warming worker on a previously-diagnosed port matches the fail-fast condition during its boot window. Self-healing needs a re-probe each hook, at which point the guard is once again deciding "occupied + no PID" before the wait — the thing that just broke. It is probably solvable by gating on the spawn lock, but I have already shipped one race here by being clever and would rather not guess twice in the same PR. Happy to do it as a follow-up if you want it, or to take direction on the coordination you'd prefer.

Verification: `bunx tsc --noEmit` clean; 10 pass / 0 fail across the three worker-utils suites. The wider `tests/shared/` run has 4 failures (`find-claude-executable`, `timeline-formatting`) which reproduce identically with my changes stashed — pre-existing on this branch's base, Windows path-resolution, unrelated.

stantheman0128:

We hit this on Win11 after a plugin upgrade: LISTEN leftover, PID gone from tasklist, `/api/health` hangs, successor cannot bind. Same class as #3204. Bumping `CLAUDE_MEM_WORKER_PORT` cleared it here. Reboot was not required.

Putting the diagnosis after the cold-boot wait is the right split. A warming worker can `listen()` before `writePidFile`; the pre-spawn guard would have false-positived that, and you already pulled it. Fail-loud through `recordWorkerUnreachable` is the half that actually reaches the user.

Skipping a persisted fail-fast flag is also right for this PR. Re-probing each hook is slower while the port is stuck, but a saved "orphaned" bit would recreate the warming-worker race you just removed.

Does not overlap #3309 (handle inheritance / spawn locks). Keep both.

DwijeshD:
Thanks, that's a useful confirmation.

Worth correcting one thing in my writeup: I said the port only frees on reboot, which was overstated. Bumping CLAUDE_MEM_WORKER_PORT was enough for me too, I just never went back and retested the original port after. The log message already offers both, so it's only the issue text that's wrong.

On #3204 — that's the same bug, closed as not planned back on the 16th. Worth noting the reporter there went through three ports in two days (37778 → 37779 → 37780), so the port bump is relief rather than a cure and it does come back.

Which makes me think #3309 is closer to the actual root cause and this one is really just the diagnostic half: yours stops the socket getting orphaned, mine explains it when it already has. Agree they should both land, but if #3309 goes in first I'd be fine with this being judged purely on whether the failure message is useful, since that's all it really does.

stantheman0128:
Thanks. Agree on the split: #3309 is the inheritance root; this PR is the fail-loud diagnostic. Port bump is relief, same as here. Fine if #3309 lands first and this is judged on whether the message is useful.

thedotmack:
Night-ship Batch E: rebased onto latest main as #4002 and squash-merged (`fb4b49a5`). Closing this as superseded.

Issue #3728, opened 2026-08-25
Observer fails on every session: FOREIGN KEY constraint failed (observations.memory_session_id → sdk_sessions), provider-independent, instant


manishsparihar32:
## Summary
On v13.15.3, the observer fails to persist anything — every session errors with `FOREIGN KEY constraint failed`, hundreds of times per day (627 in one day here). A worker restart clears it for a few hours, then it returns.

## Environment
- claude-mem 13.15.3 (latest), runtime = worker
- macOS (Apple Silicon), bun 1.4.0 / node 26
- provider: `claude` (subscription OAuth) — also reproduced on `gemini`

## What actually happens
The failure is **not** a model/timeout issue and is **provider-independent**. The generator fails within milliseconds of starting, before any model call:

```
[SESSION] [session-357] Generator auto-starting (init) using Gemini {queueDepth=0, historyLength=0}
[SESSION] [session-357] Generator failed {provider=gemini, error=FOREIGN KEY constraint failed} FOREIGN KEY constraint failed
```

A handful of specific sessions get retried forever, each write failing instantly.

## Root cause (as far as I traced it)
The FK in play is:
```
observations.memory_session_id  REFERENCES sdk_sessions(memory_session_id)  ON DELETE CASCADE ON UPDATE CASCADE
session_summaries.memory_session_id REFERENCES sdk_sessions(memory_session_id)
```
- `sqlite3 PRAGMA integrity_check` = ok.
- `sdk_sessions.memory_session_id` is `TEXT UNIQUE` (valid FK parent), 0 NULLs.
- A manual `INSERT INTO observations(memory_session_id) VALUES('<an existing sdk_sessions.memory_session_id>')` **passes** the FK (fails later on `NOT NULL project`, as expected).

So the write path is using a `memory_session_id` that has **no matching `sdk_sessions` row** for these sessions — i.e. the observation/summary is written before (or without) its parent `sdk_sessions` row being committed, and the session then gets reprocessed indefinitely. The affected sessions show `status=completed` with a valid `memory_session_id` in `sdk_sessions`, yet the runtime write references a different/absent id.

## Impact
Nothing is saved for affected sessions; the log floods with FK errors; a restart is only a temporary reset.

## Suggested direction
Ensure the `sdk_sessions` parent row is committed before any `observations` / `session_summaries` write, and make an FK failure on a write drop/park that queued item instead of retrying it forever.

davieshq:
[...]

PR #3753, opened 2026-08-27
fix(db): add ON UPDATE CASCADE to session_summaries FK (Fixes #3737)

### LINKED ISSUE #3737: session_summaries FK without ON UPDATE CASCADE permanently wedges every session after a worker restart (FOREIGN KEY constraint failed loop)
## Summary

When `session_summaries` carries its FK as `ON DELETE CASCADE` **only** (no `ON UPDATE CASCADE`) — the shape older migrations create, and the shape an old-version worker rebuilds on an upgraded DB (see companion stale-cache issue) — every session that already has summary rows becomes **permanently wedged** after a worker restart:

1. Worker restarts; SDK context is lost, so the session's stale `memory_session_id` is discarded (Issue #817 handling).
2. On the next SDK spawn the worker captures the new id and runs `UPDATE sdk_sessions SET memory_session_id = ? WHERE id = ?`.
3. With summary rows still referencing the old id and no UPDATE cascade, SQLite's default NO ACTION blocks the update → `FOREIGN KEY constraint failed`.
4. The generator dies, the queue retries, and steps 2–3 repeat forever. Observed: 244 consecutive `Generator failed {error=FOREIGN KEY constraint failed}` across every active session; zero memories saved for ~90 minutes.

`observations` has `ON DELETE CASCADE ON UPDATE CASCADE`; `session_summaries` in the affected shape does not — the asymmetry is the bug. And because the migration guard trusts `schema_versions` rather than the live schema, the current worker never restores the correct FK.

## Environment

13.15.2 worker, Linux, DB previously touched by a 13.2.0 worker rebuild.

## Fix applied locally (works)

Rebuild `session_summaries` with `FOREIGN KEY(memory_session_id) REFERENCES sdk_sessions(memory_session_id) ON DELETE CASCADE ON UPDATE CASCADE` (rows copied with `foreign_keys=OFF` to preserve pre-existing orphans). The wedged sessions recaptured ids and stored successfully on the next event.

## Suggested upstream fix

Startup check that *introspects* the actual FK definition (`PRAGMA foreign_key_list(session_summaries)`) and rebuilds if the UPDATE action isn't CASCADE — not a `schema_versions` row check.

Related: #3728 (same symptom on the observations FK), #3609 (introspection-guarded DDL plan). Distinct from #3628/#3725, which are about cascade-present NULL propagation — this is cascade-*absent* hard failure.

aniruddhaadak80:
## What

Gate the `session_summaries` / `observations` FK repair on the live constraint (`PRAGMA foreign_key_list`) instead of the `schema_versions` ledger. Adds `ensureOnUpdateCascadeToSessions()` after `ensureSyncRevisionTextAffinity()` so every column-adding migration has already run, and rebuilds either table whose `ON UPDATE` is not `CASCADE` by rewriting only the FK clause of the stored `CREATE TABLE` and carrying indexes/triggers via their own DDL.

Fixes: `FOREIGN KEY constraint failed` loop after workspace path case change / worker restart (`UPDATE sdk_sessions SET memory_session_id = ?` blocked when `session_summaries` lacks `ON UPDATE CASCADE`). Distinct from #3738 (discovery_tokens column guard, PR #3749).

## Why

`v21 addOnUpdateCascadeToForeignKeys` stamps a single `schema_versions` row (21) for both tables. A DB that recorded 21 under a build whose v21 had not yet rebuilt `session_summaries` keeps the old `ON DELETE CASCADE` (no `ON UPDATE CASCADE`) forever – the version gate reports done and the rebuild never reruns. `observations` has `CASCADE` on both actions, `session_summaries` does not – asymmetry that wedges every session with a summary row after `ensureMemorySessionIdRegistered` re-points the parent key (every worker restart since #817).

Related: #3737 (this fix), #3728, #3609. Also fixes the sibling `observations` FK if it was left behind.

## How tested

**Repro 1 – minimal FK wedge (standalone SQLite):**
```js
// without CASCADE: CREATE TABLE session_summaries ... REFERENCES sdk_sessions(memory_session_id) ON DELETE CASCADE
// INSERT parent + child, then UPDATE sdk_sessions SET memory_session_id='mem-fresh' → throws "FOREIGN KEY constraint failed"
// with CASCADE: same but ON DELETE CASCADE ON UPDATE CASCADE → UPDATE succeeds, child follows parent
```
Result: `bun run C:\Users\ANIRUDDHA\AppData\Local\Temp\opencode\claude-mem-repro.mjs` → `Bug reproduced: YES, Fix works: YES` (`session_summaries FK on_update NO ACTION` → `CASCADE`, 44 observations + summary rows followed key).

**Repro 2 – via SessionStore repair (v21-stamped DB):**
- Seeded v4 DB stamped at 4 and 21, then rebuilt `session_summaries` without `ON UPDATE CASCADE` (simulating half-reported v21).
- Before `new SessionStore(dbPath)`: `session_summaries on_update=NO ACTION`, `UPDATE sdk_sessions` → `FOREIGN KEY constraint failed`.
- After: `on_update=CASCADE`, `ensureMemorySessionIdRegistered` succeeds, 1 summary + 1 observation follow key.
- Idempotence: second `new SessionStore` leaves DDL byte-identical, `observations` also checked.
Result: `bun run repro-sessionstore.mjs` → `FK repaired to CASCADE`, `ensureMemorySessionIdRegistered succeeded`, `Idempotence check: PASS`.

**Existing suite:**
```bash
npx tsc --noEmit          # pass (no output)
bun test tests/sqlite/session-store-transactions.test.ts  # 6 pass
bun test tests/sqlite/session-store-migrations.test.ts    # 19 pass
bun test tests/sqlite/session-store-orphan-fk-repair.test.ts # known Windows EBUSY flake (same on main, not caused by this change)
```

Fixes #3737

### DIFF
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -4,6 +4,7 @@ import { DATA_DIR, DB_PATH, ensureDir, OBSERVER_SESSIONS_PROJECT } from '../../s
 import { logger } from '../../utils/logger.js';
 import {
   TableColumnInfo,
+  ForeignKeyInfo,
   IndexInfo,
   TableNameRow,
   SchemaVersion,
@@ -118,6 +119,7 @@ export class SessionStore {
     this.ensureSyncOutbox();
     this.ensureSyncEntityLedger();
     this.ensureSyncRevisionTextAffinity();
+    this.ensureOnUpdateCascadeToSessions();
     this.initializeSyncHubLaunchBaseline();
     this.normalizeConceptTags();
   }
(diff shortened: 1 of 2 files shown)


thedotmack:
Superseded by #3980 (merged) — keeps ON UPDATE CASCADE through `session_summaries` rebuilds. Closing this older FK-repair approach as duplicate.

thedotmack:
Closing as superseded by #3980.

thedotmack:
NIGHT SHIP Batch F — Gate **SKIP** (already CLOSED)

Superseded by merged #3980. No merge.

Issue #3721, opened 2026-08-25
Observation generator produces no output and is killed by the 180s idle timeout (13.15.2, Windows)


Flynn-Kawolsky:
[claude-mem-issue-20260825.md](https://github.com/user-attachments/files/31405545/claude-mem-issue-20260825.md)

Observation generator produces no output and is killed by the 180s idle timeout; queued batch is silently discarded (13.15.2, Windows)
Summary
On Windows, the observation generator spawns successfully, receives its first SDK response (~30s), and then emits nothing for exactly 180 seconds. The idle watchdog then aborts and `SIGKILL`s the child, and every message queued for that session is discarded without a `[PARSER]` line ever appearing. The result is zero observations produced, indefinitely.
`user_prompts` capture, the worker, the DB, Chroma sync and MCP search all remain healthy — only observation generation is affected, so the failure is silent unless you look at `observations` row counts.
Last observation written on this machine: 2026-08-20. Every session since has produced `obs=0`.
Environment
	
claude-mem	13.15.2 (`plugin/scripts/worker-service.cjs`, 2,599,085 bytes)
Install type	plugin, `claude-mem@inline`, in the Claude Desktop managed cache (`…\Roaming\Claude\local-agent-mode-sessions\…\rpm\plugin_<id>\`)
OS	Windows (native, non-WSL). Desktop app is MSIX-packaged (`Claude_pzs8sxrjxfjjc`)
Claude Code CLI	2.1.231 (also reproduced with a side-by-side 2.1.243)
Provider / auth	`CLAUDE_MEM_PROVIDER=claude`, OAuth token read from the Windows keychain at spawn
Model	`claude-haiku-4-5-20251001`
Worker	`127.0.0.1:37777`, `CLAUDE_MEM_RUNTIME=worker`
Notable settings	`CLAUDE_MEM_API_TIMEOUT_MS=150000`, `CLAUDE_MEM_CHROMA_PREWARM_TIMEOUT_MS=180000`, `CLAUDE_MEM_MAX_CONCURRENT_AGENTS=2`, `CLAUDE_MEM_MODE=code` with a custom `CLAUDE_MEM_MODES_DIR`
Network	Corporate HTTP proxy; egress verified working (see "Ruled out")
Expected behaviour
Messages enqueued via `/api/sessions/observations` are fed to the spawned memory agent, which emits observation XML that the parser persists into the `observations` table.
Actual behaviour
[...]

PR #3754, opened 2026-08-27
fix(worker): preserve queued batch on transport/API errors (Fixes #3752)

### LINKED ISSUE #3752: Transport/API errors from the observer child are classified as `prose` and the whole queued batch is discarded
> Paste target: NEW issue at https://github.com/thedotmack/claude-mem/issues/new

## Title

Transport/API errors from the observer child are classified as `prose` and the whole queued batch is discarded

## Body

**Version:** 13.16.1 · **Platform:** Windows 11 Enterprise 26200 x64 · bun 1.3.14 · Node v24.19.0 · MSIX-packaged desktop app

### Summary

When the spawned CLI cannot reach the API, it does not crash — it returns its own error string as the response. The parser has no class for that case, so it is treated as ordinary non-XML prose and the entire claimed batch is confirmed (dropped) instead of being returned to the queue. A transient network fault therefore becomes permanent data loss, with only a `WARN` to show for it.

Split out of #3721, where the machine-specific trigger was environmental. This part is code-level and independent of that trigger.

### Where it happens

`plugin/scripts/worker-service.cjs` (13.16.1), in the post-generation handler. The classifier recognises three classes only:

```js
function M3(t){ return typeof t!="string"||t.trim()===""?"idle"
               : /<(observation|summary)\b/i.test(t)||/<skip_summary\b/i.test(t)?"xml"
               : "prose" }
```

Before it, two special cases *do* preserve the queue:

- quota-limit prose → `resetProcessingToPending(...)`, generator paused
- auth failure → `resetProcessingToPending(...)`, `remediation:"/login"`

Everything else takes the generic branch:

```js
let P = M3(t), j = lS(t);
logger.warn("PARSER", `${a} returned non-XML ${P} response — ignoring queued batch`, {...});
await n.confirmClaimedMessages(e.sessionDbId);   // <-- batch is dropped, not requeued
e.earliestPendingTimestamp = null;
```

### Observed

```
[SDK]   [session-68] <- Response received (90 chars) API Error: Connection refused - a firewall or proxy may be blocking it (ConnectionRefused)
[WARN][PARSER] [session-68] SDK returned non-XML prose response - ignoring queued batch {outputClass=prose}
```

The session is later marked `completed` with zero observations. On my machine this path cost 882 observations across six days.

### Expected

A response that is the child's own transport/API error should be treated like the existing auth and quota cases: log it as an infrastructure failure, `resetProcessingToPending(...)` so the batch survives, and back off — never `confirmClaimedMessages(...)`.

### Suggested shape

1. Add a detector alongside the existing quota and auth detectors, matching the CLI's own error surface (`API Error:`, `Connection refused`, `ConnectionRefused`, `ECONNREFUSED`, `ETIMEDOUT`, `ENOTFOUND`, `fetch failed`, 5xx).
2. Give it its own `outputClass` (e.g. `transport`) so the warning is distinguishable from a model that genuinely replied in prose.
3. Requeue rather than confirm.

### Related

- #3721 (original report; environmental trigger since resolved on my machine)

### Nice-to-have, same area

Log the resolved child environment keys at spawn (names only, values redacted — `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `ANTHROPIC_BASE_URL`, …). Whether the child inherited proxy configuration is currently invisible in the logs, and it is the first thing worth knowing when the child cannot reach the API.

aniruddhaadak80:
### Summary

Preserve the claimed batch when the observer child returns a transport / API infrastructure error as prose, instead of confirming (dropping) it. A transient network fault was becoming permanent data loss (882 observations over 6 days).

Fixes #3752

### What changed

- **src/sdk/output-classifier.ts**: add isTransportErrorObserverOutput() detecting the CLI's own error surface (API Error:, Connection refused/ConnectionRefused, ECONNREFUSED/ECONNRESET/ETIMEDOUT/ENOTFOUND, etch failed, 5xx with error context). Returns false for XML (no false positive on error content inside an observation).

- **src/services/worker/agents/ResponseProcessor.ts**: handle the new class before the generic prose path — \
esetProcessingToPending()\, \bortReason='transport:observer_text'\, \roadcastProcessingStatus()\, and \outputClass='transport'\ warning (distinguishable from quota/auth/prose). Mirrors the existing quota/auth branches.

### Why

Before, only quota-limit and auth-failure prose preserved the queue. Every other non-XML response took the generic branch:

\classifyObserverOutput -> prose -> confirmClaimedMessages()\

A child's \API Error: Connection refused (ConnectionRefused)\ was therefore classified as \prose\ and the entire claimed batch was confirmed (dropped). Issue #3752 split this out from #3721 as the code-level part.

### How tested

- Unit: \un test tests/sdk/output-classifier.test.ts\ — 18/18 pass (existing suite still green)
- Manual classifier verification via \un -e\:
  - \API Error: Connection refused...\ → true
  - \Response received (90 chars) API Error: Connection refused...\ → true
  - \etch failed: ENOTFOUND\ → true
  - \ETIMEDOUT\ / \ECONNRESET\ → true
  - \Error 503 Service Unavailable\ / \HTTP 500 error\ → true
  - \I saw 500 files\ → false (no false positive on bare numbers)
  - \Skipping — repeated log scan...\ → false (ordinary prose still confirmed)
  - \<observation>...401...</observation>\ → false (XML never misclassified)
- ResponseProcessor path covered by existing mocks; no new harness needed. The added branch is a direct analogue of the proven quota/auth handlers (same \
esetProcessingToPending\ + abort + broadcast pattern).

### Checklist

- [x] \un test tests/sdk/output-classifier.test.ts\ passes
- [x] No XML misclassification
- [x] Logging uses distinct \outputClass='transport'\ per issue suggestion

### DIFF
--- a/src/sdk/output-classifier.ts
+++ b/src/sdk/output-classifier.ts
@@ -72,6 +72,37 @@ export function isQuotaLimitedObserverOutput(raw: unknown): boolean {
 }
 
 /**
+ * Detect transport / API-infrastructure failures rendered as assistant prose
+ * (the observer child forwards the CLI's own error string verbatim). These
+ * must preserve the claimed batch like quota/auth cases — never confirm it.
+ * See https://github.com/thedotmack/claude-mem/issues/3752.
+ */
+export function isTransportErrorObserverOutput(raw: unknown): boolean {
+  if (typeof raw !== 'string' || raw.trim() === '') {
+    return false;
+  }
+
+  if (/<(observation|summary)\b/i.test(raw) || /<skip_summary\b/i.test(raw)) {
+    return false;
+  }
+
+  const text = raw.toLowerCase().replace(/\s+/g, ' ').trim();
+
+  return (
+    /\bapi error\b/.test(text) ||
+    /connection refused/.test(text) ||
+    /connectionrefused/.test(text) ||
+    /\beconnrefused\b/.test(text) ||
+    /\beconnreset\b/.test(text) ||
+    /\betimedout\b/.test(text) ||
+    /\benotfound\b/.test(text) ||
+    /fetch failed/.test(text) ||
+    /\b5\d{2}\b.*\b(error|failed|failure|unavailable)\b/.test(text) ||
+    /\b(error|failed|failure|unavailable).*\b5\d{2}\b/.test(text)
+  );
+}
+
+/**
  * Detect provider authentication-failure prose so claimed work can be retried
  * after the user restores provider authentication.
  */
--- a/src/services/worker/agents/ResponseProcessor.ts
+++ b/src/services/worker/agents/ResponseProcessor.ts
@@ -5,6 +5,7 @@ import {
   classifyObserverOutput,
   isAuthFailureObserverOutput,
   isQuotaLimitedObserverOutput,
+  isTransportErrorObserverOutput,
   previewOutput,
 } from '../../../sdk/output-classifier.js';
 import { updateCursorContextForProject } from '../../integrations/CursorHooksInstaller.js';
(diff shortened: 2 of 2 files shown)


thedotmack:
ALL-NIGHT SHIP Batch D: **SKIPPED**.

Gate FAIL — this PR does not actually keep the batch. Greptile reproduced it: `ResponseProcessor` requeues with `transport:observer_text`, then `handleGeneratorExit` still finalizes (preserve list is only quota/auth/overflow/provider_switch), so the in-memory buffer is disposed and the observation is gone. That is the same data-loss the issue reported.

The classifier is also too loose (unanchored `api error` / errno matches), which would retry completed narrative and 4xx refusals. #3759 is the complete fix (detector + preserve-on-exit). Closing this as superseded once #3759 ships; leaving open until then so the issue stays linked.

thedotmack:
Merge Clinic: gate FAIL (requeues then still finalizes — work lost). Prefer #3998 path.

thedotmack:
Closed as superseded by #3998. This PR requeued the batch but still finalized the session on generator exit, so the work was still lost. #3998 keeps the session alive for a transport abort.

PR #3757, opened 2026-08-27
fix(sqlite): guard discovery_tokens on the column, not on schema_versions

### LINKED ISSUE #3738: ensureDiscoveryTokensColumn guards on schema_versions row instead of the actual column — dropped column is never repaired

ntdatt812:
Closes #3738.

`ensureDiscoveryTokensColumn` **already introspects** with `PRAGMA table_info` — the checks are right there. They just sit behind an early return:

```ts
const applied = this.db.prepare('SELECT version FROM schema_versions WHERE version = ?').get(11);
if (applied) return;          // <- the PRAGMAs below can never run again
```

The row records that the migration *once ran*, which is not the same as the column being there *now*. An old-version worker spawned from a stale plugin cache rebuilds `session_summaries` in its pre-`discovery_tokens` shape, and the version row survives that rebuild. The repair is then skipped forever and every summary write fails with `table session_summaries has no column named discovery_tokens`.

**Dropping the early return is the whole fix.** The two PRAGMA checks are already the correct guard and already idempotent; they were simply unreachable on a second boot. Cost is one `table_info` per table at boot.

## Tests

Three, **two of which fail against the early-return version** (`20 pass / 2 fail`):

| test | why it is shaped that way |
|---|---|
| a dropped column on `session_summaries` is re-added **even though row 11 is still present** | the test asserts the row survives, so it reproduces the reported scenario rather than a plain missing migration |
| the same for `observations` | the function repairs both tables; only testing one would leave half unpinned |
| an intact schema is untouched across repeated opens, and row 11 is inserted exactly once | the guard now runs on **every** boot, so it must be a no-op when there is nothing to repair — this is the cost of removing the early return, and it deserves a test rather than a claim |

One thing worth flagging for anyone extending these tests: `store.close()` closes the underlying `Database`, so a second boot has to be `new SessionStore(db)` on the same connection. My first draft closed between opens and got `error: Database has closed` on all three.

## Suite result

`bun test tests/sqlite` — **87 passed, 4 failed**.

Those 4 are **pre-existing**. A clean `main` on this machine gives the same 4 by name — all in the `#3378` orphaned-FK migration file — with 84 passing:

```
(fail) … (#3378) > v9 observations rebuild completes over an orphaned observation and repairs its parent
(fail) … (#3378) > v7 session_summaries rebuild completes over an orphaned summary and repairs its parent
(fail) … (#3378) > boot sequence over a DB with both orphan kinds: migrations then v12.4.3 cleanup, nothing lost
(fail) … (#3378) > repair is idempotent: reconstructing over the repaired DB is a no-op
```

I checked that one specifically rather than waving at the count, because "repair is idempotent" is exactly the kind of test this change could have broken.

## Scope

The issue notes the same pattern applies to any column-add migration keyed off `schema_versions`. This PR fixes only the one that caused the outage; converting the others is a larger sweep and should be its own change (and is what #3609 proposes).

### DIFF
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -1322,10 +1322,23 @@ export class SessionStore {
     logger.debug('DB', 'Successfully created user_prompts table');
   }
 
+  /**
+   * Add `discovery_tokens` to `observations` / `session_summaries` when it is
+   * missing.
+   *
+   * Deliberately NOT short-circuited on `schema_versions` row 11. The row
+   * records that the migration once ran, which is not the same as the column
+   * being there now: an old-version worker spawned from a stale plugin cache
+   * rebuilds `session_summaries` in its pre-`discovery_tokens` shape, and the
+   * version row survives that. Guarding on history meant the repair was
+   * skipped forever and every summary write failed with
+   * "table session_summaries has no column named discovery_tokens" until
+   * someone ran the ALTER by hand (#3738).
+   *
+   * The two PRAGMAs below are the actual guard, they are already idempotent,
+   * and they cost one table_info each at boot.
+   */
   private ensureDiscoveryTokensColumn(): void {
-    const applied = this.db.prepare('SELECT version FROM schema_versions WHERE version = ?').get(11) as SchemaVersion | undefined;
-    if (applied) return;
-
     const observationsInfo = this.db.query('PRAGMA table_info(observations)').all() as TableColumnInfo[];
     const obsHasDiscoveryTokens = observationsInfo.some(col => col.name === 'discovery_tokens');
 
--- a/tests/sqlite/session-store-migrations.test.ts
+++ b/tests/sqlite/session-store-migrations.test.ts
@@ -127,6 +127,11 @@ function revisionColumnInfo(
   }>).find(info => info.name === column)!;
 }
 
+function hasColumn(db: Database, table: string, column: string): boolean {
+  const info = db.query(`PRAGMA table_info(${table})`).all() as Array<{ name: string }>;
+  return info.some(col => col.name === column);
+}
+
 function insertSchemaVersions(db: Database, throughVersion: number): void {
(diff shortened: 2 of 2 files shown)


thedotmack:
**All-night ship gate: SKIPPED**

The production fix is already on `main` (`ensureDiscoveryTokensColumn` is not gated on `schema_versions` row 11; it introspects via `PRAGMA table_info` and cites #3890). #3749 is already closed. This PR is `DIRTY` against current main and would only add overlapping tests/comments.

No version bump.

thedotmack:
Merge Clinic: already on main (ensureDiscoveryTokensColumn via PRAGMA table_info / #3890). Superseded — closing.

PR #3758, opened 2026-08-27
fix(hooks): fail open when stdin cannot be read, instead of exit 2

### LINKED ISSUE #3699: Hook exits 2 on malformed stdin; blocks and discards the prompt on UserPromptSubmit

ntdatt812:
Fixes #3699.

## What happens today

`readJsonFromStdin` rejects when stdin closes on bytes that never parse as JSON. Nothing in `hookCommand` classified that rejection, so it fell through to the catch-all and exited **2**.

`plugin/hooks/hooks.json` is what makes that matter unevenly:

| hook | command | `async` |
|---|---|---|
| PostToolUse | `hook claude-code observation` | ✅ |
| PreToolUse | `hook claude-code file-context` | ✅ |
| Stop | `hook claude-code summarize` | ✅ |
| **UserPromptSubmit** | `hook claude-code session-init` | ❌ |
| **SessionStart** | `hook claude-code context` | ❌ |

Claude Code backgrounds config-async hooks and synthesises status 0, so the three at the top are contained. The bottom two are read for real — and on `UserPromptSubmit`, exit 2 blocks the submission and discards what the user typed.

I reproduced it in-process against `main` before changing anything, so this isn't inferred from the shipped bundle:

```
Hook error: Malformed JSON at stdin EOF: {"session_id":"abc","cwd":"/tmp"...
OBSERVED EXIT CODE = 2
```

## The change

`readJsonFromStdin` now throws a typed `HookInputUnreadable`, and `isNonBlockingHookInputError` recognises it alongside the existing missing-transcript case — routing it into the branch that already warns and exits 0.

The issue suggested matching on the message text. I went with a class instead, for one reason: the wording is a diagnostic string that a future edit may reasonably reword, and if it drifts the hook quietly goes back to eating prompts. There's a test asserting a plain `Error` carrying the identical text is *not* treated as non-blocking, so the type is doing the work, not the prose.

## One judgement call worth your eye

Because these rejections used to reach the catch-all, they used to emit `hook_failed` telemetry. Routing them to the non-blocking branch would have deleted that signal — my fix would have made a real bug **less** visible than it is today.

So the non-blocking branch now emits `hook_failed` too, under a new `error_mode: 'input_unavailable'` (documented in `scrub.ts` and the `telemetry` help text; the allowlist is keyed by property name, so no value validator needed changing). It's `await`ed for the same reason the catch-all is — `exitGraceful` calls `process.exit(0)` and would kill a fire-and-forget POST mid-flight.

Side effect to be aware of: this also starts emitting telemetry for the pre-existing missing-transcript case, which was silent before. That's the shared branch. **If you'd rather keep this to a pure exit-code fix, say so and I'll drop the telemetry hunk** — it's self-contained and the rest stands without it.

## Verification

Reverting **only** `hook-command.ts` (keeping the typed error, so it compiles) turns the new tests red with the exact signature:

```
Expected: true / Received: false
(fail) classifies unreadable stdin as a non-blocking hook input error

Expected: 0 / Received: 2
(fail) exits 0 on truncated stdin for the session-init hook
(fail) exits 0 on stdin that is not JSON at all for the session-init hook
(fail) exits 0 on truncated stdin for the context hook
(fail) exits 0 on stdin that is not JSON at all for the context hook

34 pass / 5 fail
```

With the fix applied: `bun test tests/hook-command.test.ts tests/cli tests/hooks tests/telemetry` → **320 pass / 0 fail**, and `tsc --noEmit` is clean.

The exit-contract tests drive `hookCommand` end to end rather than the predicate alone. That's deliberate and safe: stdin is rejected *before* any handler runs, so no worker is contacted and the assertion is purely about the exit code.

## Not included

- **No bundle rebuild.** `plugin/scripts/worker-service.cjs` is committed but built; I left it to your release step rather than committing a generated artifact — same as #3756. Happy to add it if you'd prefer the PR to be self-contained.
- **The `stop_hook_active` half of the issue.** The reporter is right that the `claude-code` adapter's `normalizeInput` drops the field while the codex one keeps it, so the summarize re-entry guard never fires. That's a separate defect with a separate blast radius, and I didn't want to bundle it into a prompt-loss fix. I can open it as its own PR if useful.

### DIFF
--- a/src/cli/hook-command.ts
+++ b/src/cli/hook-command.ts
@@ -1,4 +1,4 @@
-import { readJsonFromStdin } from './stdin-reader.js';
+import { readJsonFromStdin, HookInputUnreadable } from './stdin-reader.js';
 import { getPlatformAdapter } from './adapters/index.js';
 import { AdapterRejectedInput } from './adapters/errors.js';
 import { getEventHandler } from './handlers/index.js';
@@ -74,7 +74,20 @@ export function isWorkerUnavailableError(error: unknown): boolean {
   return false;
 }
 
+/**
+ * True when the hook could not obtain its own input — stdin never arrived as
+ * parseable JSON, or the transcript it was pointed at is gone.
+ *
+ * These must never exit non-zero. `UserPromptSubmit` and `SessionStart` are
+ * the two hooks in hooks.json registered WITHOUT `"async": true`, so Claude
+ * Code reads their real exit status instead of synthesising 0. On
+ * UserPromptSubmit an exit of 2 blocks the submission and discards the typed
+ * prompt, which is a far worse outcome than the missed observation that
+ * failing open costs us (#3699).
+ */
 export function isNonBlockingHookInputError(error: unknown): boolean {
+  if (error instanceof HookInputUnreadable) return true;
+
   const message = error instanceof Error ? error.message : String(error);
   const lower = message.toLowerCase();
 
@@ -132,6 +145,20 @@ export async function hookCommand(platform: string, event: string, options: Hook
     }
     if (isNonBlockingHookInputError(error)) {
       logger.warn('HOOK', `Hook input unavailable, skipping hook: ${error instanceof Error ? error.message : error}`);
+      // Failing open here used to be failing SILENT: before #3699 an
+      // unreadable stdin fell through to the catch-all, which emitted
+      // hook_failed. Keep that signal so the maintainers still see these —
+      // the exit code changes, the reporting does not. Awaited for the same
+      // reason as the catch-all below: exitGraceful calls process.exit(0),
+      // which would kill a fire-and-forget POST mid-flight.
+      {
+        const hookType = getActiveHookType();
+        await captureCliEvent('hook_failed', {
+          ...(hookType !== null ? { hook_type: hookType } : {}),
+          error_mode: 'input_unavailable',
+          threshold_tripped: false,
+        });
+      }
       emitModelContext(adapter, buildNoOpResult(event));
       exitGraceful(options);
       return HOOK_EXIT_CODES.SUCCESS;
--- a/src/cli/hook-command.ts
+++ b/src/cli/hook-command.ts
@@ -24,6 +24,39 @@ export interface HookCommandOptions {
 }
 
 /**
+ * How long the fail-open path will wait for its telemetry POST.
+ *
+ * Short on purpose. `UserPromptSubmit` and `SessionStart` are synchronous, so
+ * every millisecond here is a millisecond the user waits — and this branch
+ * exists precisely to stop a hook that cannot read its stdin from costing them
+ * anything.
+ */
+export const NON_BLOCKING_TELEMETRY_DEADLINE_MS = 250;
+
+/**
+ * Resolve when `work` settles or when `ms` elapses, whichever is first.
+ *
+ * The abandoned promise is not cancelled — `captureCliEvent` never throws and
+ * never has a caller waiting on its result, so leaving it in flight until
+ * `process.exit` is exactly as harmless as the fire-and-forget it replaces.
+ */
+export async function raceDeadline(work: Promise<unknown>, ms: number): Promise<void> {
+  let timer: ReturnType<typeof setTimeout> | undefined;
+  try {
+    await Promise.race([
+      work,
+      new Promise<void>(resolve => {
+        timer = setTimeout(resolve, ms);
+        // Do not hold the event loop open for a deadline nobody is waiting on.
+        timer.unref?.();
+      }),
+    ]);
+  } finally {
+    if (timer) clearTimeout(timer);
+  }
+}
+
+/**
  * No-op result for hooks that must exit before their handler ran (adapter
  * rejected input, transcript path missing). `context` is the sole handler
  * key that produces SessionStart output on every platform; a bare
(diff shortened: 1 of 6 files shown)


ntdatt812:
Correct, and it's a defect I introduced in this PR — the telemetry hunk was mine, and awaiting it put the cost right back where the branch exists to remove it. On a hook read for its real exit status, seconds of wait is the same category of harm as the exit 2 I was fixing, just quieter.

**Fix, two parts:**

1. `emitModelContext` now runs **before** the telemetry call. Answer first, report second.
2. The POST is on a **250ms deadline** (`NON_BLOCKING_TELEMETRY_DEADLINE_MS`). `captureCliEvent` caps itself at 2s internally, which is 2s this branch does not have. Past the deadline the event is abandoned — one lost data point, against a UserPromptSubmit that stalls on every malformed stdin. The abandoned promise isn't cancelled, which is fine: `captureCliEvent` never throws and nobody reads its result, so it is exactly as harmless as the fire-and-forget it replaces.

**On the test — worth flagging.** I first wrote this as a timing assertion around `hookCommand`, then threw it away: telemetry consent is off under test, so `captureCliEvent` returns immediately and that test passes with or without the fix. It would have looked like proof and been worth nothing. The deadline mechanism is now tested directly, with a promise that never settles, plus the converse — a POST that answers in time is not delayed.

`tests/hook-command` + `tests/cli` + `tests/telemetry`: 283 pass / 0 fail. `tsc --noEmit` clean.

stantheman0128:
Yield to #3717 for the same #3699. We already supported 3717; it already fails open through hookCommand for truncated stdin and the timeout path. The typed HookInputUnreadable plus input_unavailable telemetry here is cleaner than a message prefix. If you want one vehicle, fold those onto 3717 rather than merge both.

ntdatt812:
Agreed, and thanks for the pointer. Yielding: #3717 was opened first and covers the same #3699, so this one should not compete with it.

I checked before closing rather than taking it on faith. #3717 touches `src/cli/stdin-reader.ts` and `src/cli/hook-command.ts` and does fail open on the truncated-stdin and timeout paths, which is the part that matters for the reported symptom. So the user-visible bug is fixed there with or without this PR.

The two pieces here that #3717 does not have are the typed `HookInputUnreadable` and the `input_unavailable` telemetry, and you are right that they belong on one vehicle rather than two. They are additive and do not depend on anything else in this branch, so they port cleanly onto #3717 as a follow-up once it lands, or I can open them as a small PR on top of it if @rodboev would rather keep #3717 focused.

Closing this. Happy to do the port whenever #3717 is in.

PR #3760, opened 2026-08-27
fix(provider): pause the session on a reactive quota error instead of ending it

### LINKED ISSUE #3700: Reactive quota_exhausted errors from Gemini/OpenRouter never set abortReason — buffered work dropped instead of preserved

ntdatt812:
Fixes #3700.

## The bug

`session.abortReason` is only ever set to a quota category by the two **proactive** sites — the pre-request rate-limit guard, and the observer-text heuristic. A 429 that comes back **from** the provider is classified correctly as `quota_exhausted` and then rethrown by `handleSessionError` with `abortReason` untouched.

The consequence chain, which I traced end to end before touching anything:

1. `SessionRoutes`' `.finally()` reads `session.abortReason` → `null`
2. `handleGeneratorExit` splits the category off that reason → `''`
3. `''` isn't in the preserve list, so it falls through to `finalizeSession` + `removeSessionImmediate`

The reporter's log carries the tell directly:

```
[INFO ] [SESSION] [session-382] Generator exited — finalizing session {reason=null}
```

`reason=null` is the whole bug in one field.

## The change

`handleSessionError` maps a classified failure onto the category that keeps the work:

| kind | reason | why |
|---|---|---|
| `quota_exhausted` | `quota:quota_exhausted` | clears on its own |
| `rate_limit` | `quota:rate_limit` | clears on its own |
| `auth_invalid` | `auth:auth_invalid` | clears when the user runs `/login` |
| everything else | *(unchanged)* | genuinely fatal sessions should still be cleaned up |

Both `quota` and `auth` are already honoured by `handleGeneratorExit`, so nothing downstream needed to change.

**`auth_invalid` is one step past what the issue asked for.** It's the identical defect on a sibling kind — the `auth` category already exists in the preserve list, and credentials fixed by `/login` are no more fatal than a 429. Say the word and I'll drop it to exactly the two kinds you filed.

## Two things I checked rather than assumed

**Does the claimed batch need an explicit reset?** No — `getMessageIterator` calls `resetProcessingToPending` at the *start* of every drain, so once the session survives, the next generator start reclaims it. Adding a reset here would have been redundant.

**Should it abort the controller too?** The issue suggested it. I left it out: `handleSessionError` throws, which already unwinds the message loop, and `handleGeneratorExit` calls `ensureSdkProcessExit` for the subprocess. (For what it's worth, aborting would have been *safe* — `ensureGeneratorRunning` replaces an aborted controller before the next start — just unnecessary.)

## Verification

New file `tests/worker/reactive-quota-abort-reason.test.ts`, **8 pass**. Reverting only `OpenAICompatibleProvider.ts` turns 4 of them red:

```
Expected: "quota:quota_exhausted"   Received: undefined
Expected: "quota:rate_limit"        Received: undefined
Expected: "auth:auth_invalid"       Received: undefined

(fail) preserves buffered work for a reactive quota exit
       Expected number of calls: 0    Received number of calls: 1   ← finalizeSession ran

4 pass / 4 fail
```

That last one is the point of the PR: it drives the provider for real, takes whatever `abortReason` comes out, and feeds it to the actual `handleGeneratorExit`, so the test fails if *any* link in the chain breaks — not just if a string changes. There's a companion test asserting `reason=null` still finalizes, which pins the reporter's exact symptom.

The negative cases are covered too: `unrecoverable` and unclassified errors must still finalize, and the error must still be rethrown.

Suites run: the new file plus `openai-compatible-summary-tier`, `provider-classifiers`, `provider-errors`, `worker/agents`, `retry-policy`, `rate-limit-store` → **130 pass / 0 fail**. `tsc --noEmit` clean.

I ran targeted suites rather than the full tree — `bun test` across everything doesn't terminate here (it spawns worker daemons), and an empty result is indistinguishable from a pass, so I'd rather name what I actually observed.

## Note on scope

`normalizeAbortReason` collapses unknown categories to `'none'` for telemetry, so `auth:` still reports as `none` — pre-existing, and the same question I raised on #3759. Happy to make `auth` (and `transport`) distinguishable in that enum if you want it, but that felt like your call rather than something to slip into a data-loss fix.

### DIFF
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -342,6 +342,17 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
       const preserving = this.preservingAbortReason(error);
       if (preserving !== null) {
         session.abortReason = preserving;
+        // Abort as well as label. Without it the controller stays live while
+        // the error unwinds, and the session route books the failure twice —
+        // an observer failure and an error outcome on the way out, then the
+        // aborted outcome at finalization — leaving observer-health marked
+        // failed for a pause that is not a failure. This is what the two
+        // observer-text paths already do for the same conditions.
+        try {
+          session.abortController.abort();
+        } catch {
+          // best-effort; AbortController.abort() should not throw in normal use.
+        }
         logger.warn('SDK', `${this.providerName} paused on ${error.kind}; preserving buffered work`, {
           sessionId: session.sessionDbId,
           kind: error.kind,
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -8,7 +8,7 @@ import type { ActiveSession, ConversationMessage } from '../worker-types.js';
 import { ModeManager } from '../domain/ModeManager.js';
 import type { ModeConfig } from '../domain/types.js';
 import { resolveSummaryTierModel } from './model-aliases.js';
-import { isClassified } from './provider-errors.js';
+import { isClassified, type ClassifiedProviderError } from './provider-errors.js';
 import {
   processAgentResponse,
   snapshotResponseContext,
@@ -303,6 +303,32 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
     }
   }
 
+  /**
+   * Map a classified provider failure onto the abortReason category that keeps
+   * buffered work alive.
+   *
+   * handleGeneratorExit finalizes the session — dropping whatever is buffered —
+   * for every category outside its preserve list. Quota was only ever set by
+   * the two PROACTIVE sites (the pre-request rate-limit guard and the
+   * observer-text heuristic), so a real 429 coming back from the provider left
+   * abortReason null and the session was torn down as if the failure were
+   * fatal (#3700). These conditions clear on their own; the work should still
+   * be there when they do.
+   */
+  private preservingAbortReason(error: ClassifiedProviderError): string | null {
+    switch (error.kind) {
+      case 'quota_exhausted':
+      case 'rate_limit':
+        return `quota:${error.kind}`;
+      // Same shape, same list: handleGeneratorExit already honours 'auth', and
+      // credentials that are fixed by /login are no more fatal than a 429.
+      case 'auth_invalid':
+        return `auth:${error.kind}`;
+      default:
+        return null;
+    }
+  }
+
   protected handleSessionError(error: unknown, session: ActiveSession, _worker?: WorkerRef): never {
     if (isAbortError(error)) {
       logger.warn('SDK', `${this.providerName} agent aborted`, { sessionId: session.sessionDbId });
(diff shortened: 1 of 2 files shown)


ntdatt812:
You're right and I was wrong to leave this out.

I explicitly declined the abort in the PR description, reasoning that the throw already unwinds the message loop so aborting would be redundant. That missed the accounting: with the controller still live, the session route books an observer failure **and** an error outcome on the way out, then finalization books the aborted outcome — the pause is recorded twice, once as a failure, and observer-health stays marked failed for a condition that clears by itself.

Aborting now, which also makes this path match what the two observer-text quota/auth branches already do.

Two tests: the controller is aborted for a recoverable classified failure, and left alone for `unrecoverable` — the second one matters, because blanket-aborting would have been the easy over-correction. Reverting the source turns the first red and leaves the second green.

`tests/worker/agents` + `provider-errors` + the new file: 46 pass / 0 fail. `tsc --noEmit` clean.

thedotmack:
ALL-NIGHT SHIP Batch D: rebased onto latest `main` as #3999. Will merge #3999 if CI is green.

thedotmack:
Superseded by #3999.

thedotmack:
Superseded by #3999 (rebased onto latest main and squash-merged). Thank you.

PR #3761, opened 2026-08-27
fix(setup): guard on a complete node_modules, and let bun.cmd actually run

### LINKED ISSUE #3755: Setup dependency guard checks that node_modules exists, not that it is complete — a partial install wedges every hook
## Summary

```js
```

## Environment

```
```

## Reproduce

   their timeouts.

## Suggested fix

```js
  );
}

```

installer's exit code.

## Workaround

ntdatt812:
Fixes #3755.

## 1. The reported bug

`ensurePluginDependencies` decided whether to install by asking whether `node_modules/` existed. The directory existing only tells us a package manager *started*. An install interrupted by anything this script didn't spawn — a killed terminal, a half-finished extraction — leaves the tree in place but short of packages, and the existence check then skipped the repair on **every** later Setup run.

The comment right below the guard already described the trap; it just only applied the escape to failures the script itself caused.

Each declared dependency must now resolve to its own `package.json`. That keeps the property the old comment wanted — no hardcoded package name, still correct if dependencies are renamed — while catching a partial tree. One `existsSync` per dependency, and it re-verifies rather than trusting a previous installer's exit code.

## 2. What verifying it on Windows turned up

I couldn't run the committed tests here: the suite is `describe.skipIf(process.platform === 'win32')`, because the fake bun is a bash script and the fixture builds a `:`-separated PATH. So I built a Windows-native equivalent — a `bun.cmd` shim on PATH, driving the real `version-check.js` — and ran all four tree states against **unmodified `main`**:

| state | install attempted | outcome |
|---|---|---|
| A. no `node_modules` | yes | **failed**, zod still unresolvable |
| B. `node_modules` exists, empty | no | the reported trap |
| C. `zod/` exists without its `package.json` | no | the reported trap |
| D. complete tree | no | correct |

Row A is a second bug, and it explains how a Windows machine gets into this state to begin with:

```
[version-check] bun install failed (spawnSync ...\bun.cmd EINVAL)
```

`findBun` resolves to `bun.cmd`, and Node has refused to spawn a `.cmd` directly since the CVE-2024-27980 mitigation. Confirmed in isolation on Node 24:

```
without shell      -> error: EINVAL   status: null
with shell: true   -> error: undefined status: 0
```

So on Windows the Setup-phase auto-install has never worked — it attempts and fails every time.

**This had to be in the same PR.** Without it, the completeness guard makes Windows *worse*: incomplete tree → attempt install → EINVAL → the failure path `rmSync`s `node_modules` → repeat, every launch, now deleting the partial tree each time instead of merely ignoring it.

The fix uses a shell on Windows only, with the path quoted (these paths routinely contain spaces). `BUN_INSTALL_ARGS` is a frozen constant, so the concatenation the shell option performs carries nothing user-supplied. Non-Windows behaviour is byte-for-byte unchanged.

With both fixes, the same harness on Windows:

| state | incomplete diag | attempted | result | zod resolvable after |
|---|---|---|---|---|
| A | – | yes | success | ✅ |
| B | yes | yes | success | ✅ |
| C | yes | yes | success | ✅ |
| D | – | **no** | – | ✅ (was already) |

Row D matters as much as the rest: a healthy tree must not trigger a 100 MB+ reinstall on every cold start.

## 3. A test I had to change, not just add

`skips install when node_modules is already present` built an **empty** `node_modules` — which is exactly the state that must now trigger a reinstall. Leaving it would have meant either a red test or a fix that doesn't work.

It now installs a resolvable dependency instead, so it still pins the property it was written for. The fake bun also writes a real `package.json` now, since a fake install that produces nothing resolvable isn't modelling an install.

Two new cases cover the reported shapes: `node_modules` present but empty, and a dependency folder present without its `package.json`.

## Verification summary

- Windows harness above, run against both `main` and this branch.
- `tests/infrastructure`, `tests/hook-command`, `tests/hook-lifecycle`: **11 failures on this branch and 11 on clean `main`, identical by name** (Windows-specific `HealthMonitor` / `ProcessManager` / spawn-contract / cleanup tests). Diffing the two lists shows only timings differ.
- `node --check plugin/scripts/version-check.js` clean.
- The 5 tests in `plugin-version-check-ensure-deps.test.ts` are unix-gated and **skip on my machine** — I can't claim I ran them, so I didn't. They'd be worth a look on CI, and I'd be glad to make that suite cross-platform in a follow-up if you want it.

## Not included

- **No post-install re-verify.** The guard makes the *next* Setup run re-check, which is what #3755 asks for. Warning immediately when an exit-0 install still leaves something missing would be cheap, but it's a behaviour change beyond the report — happy to add it.
- **The bare-`'bun'` fallback** in `findBun` (when `where bun` finds no `.cmd`) now works via the shell, but the function's Windows resolution could still be tightened. Separate concern.
- The two follow-ups the reporter raised at the end of #3755 — killing a healthy worker before confirming the replacement boots, and hooks blocking for their full timeout — are separate issues and I've left them alone.

### DIFF
--- a/plugin/scripts/version-check.js
+++ b/plugin/scripts/version-check.js
@@ -55,12 +55,48 @@ function findBun() {
 // has a 300s timeout (vs 60s for SessionStart), runs once per Claude
 // Code launch, and is the only standalone hook script — the natural
 // place to materialise plugin runtime state.
+// Names of declared dependencies that do not resolve to an installed package.
+// Reads package.json rather than naming any package, so the check stays correct
+// if dependencies are later renamed or added.
+function missingDependencies(pluginRoot) {
+  let pkg;
+  try {
+    pkg = JSON.parse(readFileSync(join(pluginRoot, 'package.json'), 'utf-8'));
+  } catch {
+    // An unreadable package.json is not something an install can fix, and
+    // reporting every dependency as missing would loop on it. Report none.
+    return [];
+  }
+
+  const declared = Object.keys((pkg && pkg.dependencies) || {});
+  return declared.filter(
+    (name) => !existsSync(join(pluginRoot, NODE_MODULES_DIRNAME, ...name.split('/'), 'package.json')),
+  );
+}
+
 function ensurePluginDependencies(pluginRoot) {
   if (!existsSync(join(pluginRoot, 'package.json'))) return;
 
-  // Guard on node_modules (package-manager marker) rather than a specific
-  // package, so the check stays correct if dependencies are later renamed.
-  if (existsSync(join(pluginRoot, NODE_MODULES_DIRNAME))) return;
+  // Guard on COMPLETENESS, not on the mere existence of node_modules.
+  //
+  // The directory existing only tells us a package manager started. An
+  // install interrupted by anything this script did not spawn — a killed
+  // terminal, a half-finished extraction — leaves node_modules in place but
+  // short of packages, and an existence check then skips the repair on every
+  // later Setup run. The worker dies on the missing module each boot with no
+  // recovery short of a manual rm -rf (#3755).
+  //
+  // Each declared dependency must resolve to its own package.json: one
+  // existsSync per dependency, and it re-verifies the tree rather than
+  // trusting a previous installer's exit code.
+  const missing = missingDependencies(pluginRoot);
+  if (missing.length === 0) return;
+
+  if (existsSync(join(pluginRoot, NODE_MODULES_DIRNAME))) {
+    console.error(
+      `${VERSION_CHECK_LOG_PREFIX} node_modules is incomplete (missing: ${missing.join(', ')}); reinstalling`,
+    );
+  }
 
   const bunPath = findBun();
   if (!bunPath) {
@@ -73,12 +109,21 @@ function ensurePluginDependencies(pluginRoot) {
 
   let result;
   try {
-    result = spawnSync(bunPath, BUN_INSTALL_ARGS, {
+    // Windows: findBun resolves to bun.cmd (npm/nvm shim) or, failing that,
+    // the bare name. Node refuses to spawn a .cmd/.bat directly since the
+    // CVE-2024-27980 mitigation — spawnSync throws EINVAL — and a bare name
+    // is not resolved through PATHEXT either, so both shapes need a shell.
+    // BUN_INSTALL_ARGS is a frozen constant, so the concatenation the shell
+    // option performs carries nothing user-supplied; the path itself is
+    // quoted because it routinely contains spaces.
+    const command = IS_WINDOWS ? `"${bunPath}"` : bunPath;
+    result = spawnSync(command, BUN_INSTALL_ARGS, {
       cwd: pluginRoot,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],
       timeout: BUN_INSTALL_TIMEOUT_MS,
       windowsHide: true,
+      shell: IS_WINDOWS,
     });
   } catch (err) {
     const reason = err && err.message ? err.message : String(err);
--- a/plugin/scripts/version-check.js
+++ b/plugin/scripts/version-check.js
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 import { spawnSync } from 'child_process';
-import { existsSync, readFileSync, rmSync } from 'fs';
+import { existsSync, readdirSync, readFileSync, rmSync } from 'fs';
 import { homedir } from 'os';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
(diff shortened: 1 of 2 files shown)


ntdatt812:
Confirmed — a manifest with nothing beside it passed, and that is not an install.

**Fixed:** a dependency counts as installed only when its manifest is present **and** the directory holds something besides that manifest. Your manifest-only zod is now flagged; the reproduction is in the suite.

**On "validate required runtime entrypoints" — I tried that first and backed it out, with data.**

I built the entrypoint resolver (`main`, then `exports["."]` preferring `require`/`import`/`node`/`default` over `types`, with Node's extensionless and directory-index fallbacks) and ran it across the 568 packages in this repo's own `node_modules`. After the fallbacks it flagged exactly one:

```
@modelcontextprotocol/sdk -> ./dist/cjs/index.js
```

Not a resolver bug — that package genuinely declares `exports["."]` pointing at `dist/cjs/index.js` and does not ship it. It has no root entrypoint at all; consumers reach it through subpaths. A **direct** dependency shaped like that would make Setup reinstall on every single launch, which is a worse failure than the partial tree we are catching, and it is the exact regression the "skips install when the tree is complete" test exists to prevent.

The populated-directory rule flags **0 of those 568**. I'd rather under-detect a half-extracted package than ship a reinstall loop, but if you'd prefer the stricter check I'll put it back — the resolver is written.

**Verification** (Windows harness driving the real script, since the committed suite is unix-gated):

| state | incomplete diag | install attempted |
|---|---|---|
| no `node_modules` | – | yes |
| `node_modules` empty | yes | yes |
| `zod/` with no manifest | yes | yes |
| **`zod/` manifest-only (your case)** | **yes** | **yes** |
| populated `zod/` | – | **no** |

`tests/infrastructure` + `hook-command` + `hook-lifecycle`: 11 failures, identical by name to clean `main` (Windows-specific). One run also showed a 12th — `npm tarball includes sqlite runtime modules` — which did not reproduce on two re-runs and passes when that file runs alone, so I'm calling it contention rather than anything from this branch. Flagging it rather than quietly dropping it.

`node --check` clean.

stantheman0128:
Yield to #3769 for the same #3755 guard.

The Windows diagnosis here is right (`bun.cmd` -> EINVAL, and a completeness retry that then `rmSync`s the tree would make Setup worse). #3769 keeps that repair and avoids `shell: true` by preferring `bun.exe` (same direction as #3689) with a quoted `cmd.exe` fallback. Happy to close this once that lands.

ntdatt812:
Yielding to #3769 — closing this. I read the diff rather than take the summary on faith, and it does cover both halves of this PR, with a better Windows fix than mine.

**Windows.** Mine reached for `shell: true`; #3769 prefers `bun.exe` from the `where` output and only falls back to a quoted `cmd.exe` for the `.cmd` shim. That's the better shape — `shell: true` routes the whole argv through the command interpreter, and the interpreter is the thing CVE-2024-27980 is about. Preferring the `.exe` sidesteps the problem instead of arguing with it.

**Completeness.** Mine required each dependency directory to hold something besides its own `package.json`. #3769 resolves the declared closure, which subsumes that.

One data point from this branch worth keeping, since it's the trap the closure check sits next to. I first implemented completeness as *"every dependency must resolve its entrypoint"* and backed it out with measurements: `@modelcontextprotocol/sdk` fails that check on a perfectly good install, so Setup would have concluded "incomplete", reinstalled, checked again, and looped — turning a cosmetic gap into a hang. #3769 avoids it by scoping `REQUIRED_DEPENDENCY_SUBPATHS` to `zod` rather than deriving subpaths from every dependency, which is the right call. Worth a comment on that constant so nobody later generalises it.

Two small things I noticed while reading, neither blocking:

- `hasFreshDependencyResolutions` passes `TREE_SITTER_VERSION_TIMEOUT_MS` as its timeout. It's the right duration, but the name ties the resolver's budget to tree-sitter's — anyone tuning one will silently move the other.
- `hasCompletePluginDependencies` returns `true` when `readJsonObject` yields either `false` (ENOENT) or `null` (unparseable). Fail-open on the plugin manifest is deliberate and documented; just noting the two distinct causes collapse into one branch, so a corrupt manifest is indistinguishable from an absent one at that point.

The diagnosis in this PR stands and I'm glad it was useful, but two PRs for one guard is worse than one. Thanks for the clear pointer, @stantheman0128.

stantheman0128:
Thanks for reading #3769 and closing this. The `@modelcontextprotocol/sdk` reinstall-loop note belongs next to `REQUIRED_DEPENDENCY_SUBPATHS` so that constant does not get generalised later. The timeout-name and fail-open points look right. I am not chasing either.

ntdatt812:
Agreed on all three, and thanks for the clean handover.

For the record, so the note is findable from here even if it never makes it into #3769: the entrypoint check I backed out was *"every declared dependency must resolve its own entrypoint"*. `@modelcontextprotocol/sdk` fails that on a healthy install, so Setup would have read "incomplete", reinstalled, re-checked, and looped — a cosmetic gap turned into a hang. Scoping `REQUIRED_DEPENDENCY_SUBPATHS` to `zod` is what avoids it; deriving the subpaths from the dependency list instead of naming them is the change that would bring it back.

PR #3762, opened 2026-08-27
fix(observer): strip image payloads out of the observation prompt

### LINKED ISSUE #3730: Observer includes raw base64 image content from tool_result in compression context, causing 10-20x token blowups on browser-automation-heavy sessions
**Affected**: claude-mem v13.15.3, macOS, Claude Code, provider `openrouter` (custom base URL pointed at a different OpenAI-compatible endpoint, but the mechanism is provider-agnostic)

**Symptom**

A local cost guard I built around claude-mem started firing repeatedly (5 separate calendar days: 08-20, 08-21, 08-24, 08-25) with single observations costing 400K-1M+ `discovery_tokens`, against a normal baseline of 30K-50K. Daily spend on the compression model hit 150M-266M tokens on the worst days.

**Root cause**

`transcript-watcher.cjs` gates whether a tool_result gets processed for observation via `CLAUDE_MEM_SKIP_TOOLS`:

```js
if (new Set(SKIP_TOOLS.split(",")...).has(r.toolName)) return {ok:true, status:"skipped", ...}
```

The default skip list (`ListMcpResourcesTool,SlashCommand,Skill,TodoWrite,AskUserQuestion`) has no entry for browser-automation tools that return image content blocks. When a tool_result contains `type:"image"` content (screenshots from an MCP browser tool, in my case `mcp__claude-in-chrome__computer` and `mcp__claude-in-chrome__browser_batch`), the raw base64 image data appears to get serialized as text and sent to the compression LLM along with everything else. Base64 is extremely token-dense, so a handful of screenshots is enough to blow up a single observation call by 10-20x.

I confirmed this by pulling the actual runaway session's transcript (204MB, 38963 lines) and correlating every `type:"image"` tool_result back to the tool that produced it via `tool_use_id`:

| tool | image tool_results |
|---|---|
| `mcp__claude-in-chrome__computer` | 210 |
| `mcp__claude-in-chrome__browser_batch` | 52 |
| `Read` | 38 (already in default skip list — not a contributor) |

Sessions that did heavy Chrome MCP screenshot verification work (repeated UI checks, viewport resizing, etc.) are exactly the ones that triggered the alerts. Sessions of similar wall-clock length without browser automation stayed in the normal 30K-50K range.

**Workaround (applied locally, works)**

Added the two Chrome MCP tool names to `CLAUDE_MEM_SKIP_TOOLS` in `~/.claude-mem/settings.json`. This is a per-user manual list though, and only covers the one browser-automation MCP I happen to use — anyone using a different browser/screenshot tool will hit the same failure mode under a different tool name.

**Suggested fix**

Rather than relying on users to enumerate every screenshot-producing tool in `CLAUDE_MEM_SKIP_TOOLS`, it'd be more robust to strip or size-cap `type:"image"` content blocks out of tool_result content generically before it's serialized into the observer's context — regardless of which tool produced them. The observer is generating a text narrative; it doesn't need the raw image bytes to do that, and if vision context is ever actually wanted, it should probably be opt-in and separately budgeted, not silently included in the same context/token budget as everything else.

Happy to test a fix against the same session if useful — I still have the transcript that reproduces this.

ntdatt812:
Fixes #3730.

## Measured, before anything was changed

One screenshot-shaped `tool_result` through `buildObservationPrompt` on `main`:

```
prompt chars:              15,460
longest base64 run kept:    9,338
```

The per-field truncation at 16,000 chars doesn't save you here — it keeps the **head and tail** of the oversized field, so what survives a screenshot is thousands of characters of base64, while the caption next to it is a few dozen.

After:

```
prompt chars:               1,277
longest base64 run kept:        0
kept the text block:         true
```

## Why the reported cost is 10-20x rather than 1.2x

The issue attributes the blowup to base64 being dense, which is true but not the whole mechanism. The prompt is pushed onto `session.conversationHistory`, and `query(session.conversationHistory, config)` sends the **whole history** on every later observation in the session.

So a screenshot isn't billed once. It's billed again on every subsequent observation in that session. At the reporter's 210 + 52 image results and ~9.3K of surviving base64 each, the tail of the session is re-sending millions of characters of noise per call — which is what puts a single observation at 400K-1M `discovery_tokens` against a 30K-50K baseline, and it's why the effect scales with how screenshot-heavy the session was rather than with any one tool call.

## The fix

`stripImagePayloads` walks the parsed tool input/output and elides image payloads before truncation:

- **Anthropic** `{ type: 'image', source: { data: '<base64>' } }` → `{ type: 'image', source: { elided: …, media_type, bytes } }`
- **OpenAI** `{ type: 'image_url', image_url: { url: 'data:…' } }` → same treatment
- **A plain `https://…` image URL is left alone** — it's short and it carries signal. Only an inlined `data:` URL is the payload this exists to remove.

Three deliberate choices:

1. **Generic on the block shape, not on tool name.** `CLAUDE_MEM_SKIP_TOOLS` needs every screenshot-producing tool enumerated ahead of time — the reporter's workaround covers the one browser MCP they happen to use — and it drops the whole observation instead of keeping the part with signal.
2. **Strip, don't skip.** The caption beside the image ("Took a screenshot of the viewport at 1280x720") is exactly what a text observer can narrate. There's a test pinning that it survives.
3. **Say what was removed.** The placeholder reports `media_type` and byte count, so a silently-thinner context doesn't become its own mystery.

It's in `buildObservationPrompt`, so all three providers (Claude, Gemini, OpenRouter) get it from one place.

## Verification

`tests/sdk/prompts.test.ts`, 8 new tests. Reverting only `src/sdk/prompts.ts` turns **5 of them red**:

```
(fail) keeps no base64 run from an Anthropic image content block
(fail) says what was removed instead of removing it silently
(fail) collapses the prompt far below the per-field truncation cap
(fail) strips images nested inside arrays and objects
(fail) strips an OpenAI-style inlined data: URL

6 pass / 5 fail
```

The other three pass both ways on purpose — they're the guards against over-stripping: the sibling text survives, a plain http URL survives, and ordinary tool output (including a field that merely *starts with* `image_`) is untouched.

`bun test tests/sdk tests/worker/agents tests/hook-command.test.ts` → **125 pass / 0 fail**. `tsc --noEmit` clean.

## Notes

- **Nothing is stripped from what's stored** — only from the prompt handed to the observer. Storage and the transcript are untouched.
- The reporter offered to test against the session that reproduces it. That'd be worth taking up: my measurement is a synthetic tool_result shaped like Claude Code's, so it pins the mechanism but not their exact payload shapes.
- A depth cap of 12 bounds the walk. Parsed JSON can't be cyclic, but the field can also arrive as an object that was never serialised.
- If you'd ever want vision context deliberately, this is the right seam for it — the block is identified here rather than being lost in a byte stream.

### DIFF
--- a/src/sdk/prompts.ts
+++ b/src/sdk/prompts.ts
@@ -137,9 +137,16 @@ function stripImagePayloads(value: unknown, depth = 0): unknown {
   const record = value as Record<string, unknown>;
 
   // Anthropic content block: { type: 'image', source: { data: '<base64>' } }.
+  // A url-backed source is the same case as OpenAI's plain http URL — short,
+  // and it carries signal — so only an inlined payload is removed.
   const source = record.source;
   if (record.type === 'image' && source !== null && typeof source === 'object') {
-    return { type: 'image', source: elideImageSource(source as Record<string, unknown>) };
+    const record_source = source as Record<string, unknown>;
+    const url = record_source.url;
+    if (typeof url === 'string' && !url.startsWith('data:')) {
+      return value;
+    }
+    return { type: 'image', source: elideImageSource(record_source) };
   }
 
   // OpenAI content block: { type: 'image_url', image_url: { url: 'data:...' } }.
--- a/src/sdk/prompts.ts
+++ b/src/sdk/prompts.ts
@@ -104,6 +104,66 @@ const OBS_PROMPT_FIELD_MAX_CHARS = 16_000;
 const OBS_PROMPT_FIELD_HEAD_RATIO = 0.6;
 const OBS_PROMPT_FIELD_TAIL_RATIO = 0.3;
 
+// Image content blocks carry base64 payloads that are worthless to a text
+// observer and ruinously expensive to carry. Two things compound (#3730):
+// truncateObservationField keeps the head and tail of an oversized field, so
+// what survives a screenshot is thousands of characters of base64 rather than
+// the caption beside it; and the prompt is appended to
+// session.conversationHistory, which every later observation in the session
+// re-sends in full. A browser-automation session taking a few hundred
+// screenshots replays all of it, every time.
+//
+// Stripping generically, on the shape of the content block, rather than by
+// tool name: CLAUDE_MEM_SKIP_TOOLS needs every screenshot-producing tool
+// enumerated ahead of time, and it drops the observation entirely instead of
+// keeping the part that has signal.
+const MAX_SANITIZE_DEPTH = 12;
+
+function elideImageSource(source: Record<string, unknown>): Record<string, unknown> {
+  const data = source.data;
+  const elided: Record<string, unknown> = { elided: 'image data withheld from the observer' };
+  if (typeof source.media_type === 'string') elided.media_type = source.media_type;
+  if (typeof data === 'string') elided.bytes = data.length;
+  return elided;
+}
+
+function stripImagePayloads(value: unknown, depth = 0): unknown {
+  if (depth > MAX_SANITIZE_DEPTH || value === null || typeof value !== 'object') return value;
+
+  if (Array.isArray(value)) {
+    return value.map((entry) => stripImagePayloads(entry, depth + 1));
+  }
+
+  const record = value as Record<string, unknown>;
+
+  // Anthropic content block: { type: 'image', source: { data: '<base64>' } }.
+  const source = record.source;
+  if (record.type === 'image' && source !== null && typeof source === 'object') {
+    return { type: 'image', source: elideImageSource(source as Record<string, unknown>) };
+  }
+
+  // OpenAI content block: { type: 'image_url', image_url: { url: 'data:...' } }.
+  const imageUrl = record.image_url;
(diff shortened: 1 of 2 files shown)


ntdatt812:
Good catch, and correct — the two branches disagreed with each other. The Anthropic matcher replaced any object `source`, so `{ type: 'image', source: { type: 'url', url: 'https://…' } }` lost both the URL and its metadata, while the OpenAI branch deliberately kept the equivalent.

Fixed: a `source.url` that is not a `data:` URL is now returned untouched, so the rule is the same on both shapes — inlined payload out, short reference in. Two tests added, one for each direction (plain https kept, `data:` URL still elided). 13 pass in `tests/sdk/prompts.test.ts`, `tsc --noEmit` clean.

thedotmack:
ALL-NIGHT SHIP Batch D: rebased onto latest `main` as #3996 (plus a `DATA:` case-fold so uppercase data URIs strip too). Will merge #3996 if CI is green.

thedotmack:
Superseded by #3996.

thedotmack:
Superseded by #3996 (rebased onto latest main and squash-merged, including the `DATA:` case-fold). Thank you.

Issue #3611, opened 2026-08-16
[plan-23] Host Integration Contracts — every non-Claude-Code adapter and the MCP/HTTP tool surface validated by contract tests against the host's actual schema


thedotmack:
# [plan-23] Host Integration Contracts — every non-Claude-Code adapter and the MCP/HTTP tool surface validated by contract tests against the host's actual schema

> **Tracker:** #3611 · **Design doc:** `plans/23-host-integration-contracts.md` · **Series:** plan-14 #3602 · plan-15 #3603 · plan-16 #3604 · plan-17 #3605 · plan-18 #3606 · plan-19 #3607 · plan-20 #3608 · plan-21 #3609 · plan-22 #3610 · plan-23 #3611

## Defect

Each host integration (OpenCode plugin, Codex hooks, Antigravity CLI/IDE, hook-CLI platforms, transcript watcher, and the `mcp-search` MCP server / worker HTTP surface) is a hand-rolled adapter whose only test is "does the source file import". None of them is checked against the host's real loader/hook/tool schema, and the shared plumbing they depend on (port resolution, generator start, viewer source filter) has per-adapter forks:

- **OpenCode:** the plugin binds `chat.message` (removed upstream; now `message.updated`/`message.part.updated`), fires capture POSTs without awaiting, exports two `as const` arrays from the plugin entry that opencode's loader rejects (`Plugin export is not a function`) — regressed 13.4.0 → 13.4.1 and still present on 13.14.0 — resolves the worker port from env/defaults instead of `settings.json`, and the installer refreshes a stale 12.4.7 npx-cached plugin file. OpenCode is not a viewer source.
- **Codex:** `codex-hooks.json` emits `suppressOutput` (rejected by Codex), and POSIX shell command strings with no `commandWindows` so the Windows Codex app runs them through PowerShell (`ParserError`).
- **Antigravity / transcript watcher:** `--ide antigravity` wires only `agy` CLI hooks while the GUI IDE never fires them; the transcript watcher's `Bun.Glob` lacks `dot: true` so the IDE's `.system_generated/logs/transcript_full.jsonl` is never discovered; the Windows agy path records nothing; the installer depends on a marketplace file it does not install; the custom watch schema is undocumented.
- **Hook-CLI platforms:** `worker-service.cjs hook <platform> observation` returns `{"status":"queued"}` but the generator never starts for sessions registered through `/api/sessions/init` outside Claude Code.
[...]

Issue #3608, opened 2026-08-16
[plan-20] Project Identity & Injection Scope — one resolver shared by capture, sync and retrieval, and injection config that reaches the query


thedotmack:
# [plan-20] Project Identity & Injection Scope — one resolver shared by capture, sync and retrieval, and injection config that reaches the query

> **Tracker:** #3608 · **Design doc:** `plans/20-project-identity-and-injection-scope.md` · **Series:** plan-14 #3602 · plan-15 #3603 · plan-16 #3604 · plan-17 #3605 · plan-18 #3606 · plan-19 #3607 · plan-20 #3608 · plan-21 #3609 · plan-22 #3610 · plan-23 #3611

## Defect

"Which project is this?" is computed independently on the write path (hook capture, observer, Chroma sync), the read path (SessionStart/compact context, PreToolUse file-context, semantic inject, timeline), and the migration path — with different rules — and there is no user-facing override. Every mismatch is a silent "This project has no memory yet" over thousands of observations, or memory scattered across five keys:

- Write path is `basename(process.cwd())` in the bundled hook scripts (subagent/SDK sessions inherit TMPDIR → project `T`; observer sessions → plugin version dir); read path is `basename(gitToplevel(cwd) ?? cwd)` since #2663 with a no-git fallback that differs from the parent-project attribution capture uses; the 13.4.0 switch from cwd-basename to git-root re-keyed monorepo subdirectories with no migration; submodules (`.git/modules/`) are treated as standalone leaf projects; worktrees are reconciled by a **one-shot** `.cwd-remap-applied-v1` marker so later-created worktrees and re-homed repos strand rows with `merged_into_project = NULL`; the key is compared with SQLite BINARY collation so `pasteypal` ≠ `PasteyPal` across machines; identical dir names in different repos collide; `<project>:dream` namespaces never match `WHERE project = ?`; Chroma docs are tagged with the *daemon's* cwd, so `/api/context/semantic?project=` is always empty.
- Injection configuration is decorative: `CLAUDE_MEM_CONTEXT_OBSERVATION_TYPES/_CONCEPTS` are documented, persisted, and never reach `loadContextConfig`; `CLAUDE_MEM_EXCLUDED_PROJECTS` is honored by capture handlers but not by the SessionStart `context` handler; the injection query has no `agent_id IS NULL` filter so subagent step-logs evict durable memory from the recency window.
[...]

Issue #1911, opened 2026-04-15
[BUG] search()/timeline() return cross-project results when project param is omitted


thedotmack:
**Severity:** high  
**Component:** search

**Symptom:** Observations from every project leak into search/timeline results when caller omits project filter, producing cross-project memory pollution.

**Root cause:** `SearchManager.search()`, `timeline()`, `getRecentContext()`, `getContextTimeline()`, `getTimelineByQuery()` don't default the `project` filter to current project.

**Verification against v12.1.1 HEAD:** VALID — `src/services/worker/SearchManager.ts:74-121` `normalizeParams()` doesn't default `project`; `search()` at line 126 and `timeline/getRecentContext/getContextTimeline/getTimelineByQuery` (lines 397, 1321, 1447, 1661) never inject current-project default.

---

_Consolidated during the April 2026 backlog cleanup. Original reports: PR #1859_

thedotmack:
Fixed on issue-blowout-2026 branch (Phase 2-5).

PR #3765, opened 2026-08-27
fix(mcp): scope memory reads to the current repository

### LINKED ISSUE #3764: [BUG] MCP memory tools default to cross-project scope when project is omitted
## Before submitting

- [x] I searched existing issues and confirmed this is a regression of #1911, not an open duplicate

## Bug Description

The worker-runtime MCP tools `search`, `timeline`, and `get_observations` still accept an omitted `project` and forward the request without repository scope. The worker then searches or hydrates rows globally

This is the behavior previously tracked and closed as fixed in #1911. It reproduces on 13.15.3 and the same unscoped handlers remain on current `main` at 13.16.1

Mixed-host installations have a second scope problem: omitting `platformSource` allows one host to consume observations generated by another host, even when both rows share the same project key

## Steps to Reproduce

1. Store observations for at least two projects, or for two host platforms under one project key
2. Start the MCP server from a repository working directory
3. Call `search` without `project`
4. Call `get_observations` with IDs returned from another project or platform
5. Observe that results are not constrained to the calling repository or host

The MCP handlers currently pass `args` directly to `/api/search`, `/api/timeline`, and `/api/observations/batch`. Their schemas describe `project` as an optional filter, so host safety depends on the model remembering to supply it on every call

## Expected Behavior

Repository-bound MCP reads should fail closed without relying on model judgment

- Derive the current project from the MCP process working directory using the existing project resolver
- Add the derived project when `project` is omitted
- Reject an explicitly different project by default
- Derive and enforce the current host platform where it can be identified reliably
- Require an explicit operator-controlled environment opt-in for cross-project or cross-platform reads
- Keep parent/worktree project keys allowed when they belong to the same repository context

## Environment

- claude-mem: 13.15.3; confirmed still present on `main` at 13.16.1
- host: Codex Desktop
- runtime: worker
- OS: Linux

## Verification

A local patch was tested over the stdio MCP protocol

- omitted `project` returned only the current repository
- an explicit foreign project returned an MCP error
- an explicit foreign platform returned an MCP error
- an observation from another platform but carrying the same project key was excluded
- `timeline` could not anchor on an out-of-scope observation
- `get_observations` returned an empty list for an out-of-scope ID

Related scope work: #3608 and #3611

Kizuno18:
This closes the default-global behavior on the worker-runtime MCP read path without relying on the model to remember a project filter

The MCP process already has the repository working directory, so this patch uses the existing `getProjectContext()` resolver before forwarding `search`, `timeline`, and `get_observations` requests to the worker

- omitted `project` defaults to the current repository
- the parent/worktree project chain remains allowed
- an explicitly foreign project fails closed by default
- `project: "*"` requires `CLAUDE_MEM_ALLOW_CROSS_PROJECT_SEARCH=1`
- Codex and Claude plugin installs default `platformSource` to their own host
- an explicitly foreign host fails closed unless `CLAUDE_MEM_ALLOW_CROSS_PLATFORM_SEARCH=1`
- hosts that cannot be identified from the installed server path keep their existing platform behavior

This is intentionally at the MCP boundary: direct worker HTTP callers retain their existing API semantics

Validation:

- `bun test tests/servers/mcp-tool-schemas.test.ts tests/servers/mcp-server-name-safety.test.ts` — 15 passed
- production MCP bundle built with esbuild and passed `node --check`
- stdio protocol smoke test confirmed repository defaulting and foreign-project rejection
- mixed-host smoke test confirmed an out-of-platform observation sharing the same project key was excluded from `search`, `timeline`, and `get_observations`

Closes #3764
Related to #1911, #3608, and #3611

### DIFF
--- a/src/servers/mcp-server.ts
+++ b/src/servers/mcp-server.ts
@@ -39,6 +39,49 @@ import {
 } from '../services/hooks/runtime-selector.js';
 import { normalizePlatformSource } from '../shared/platform-source.js';
 import { getAdvertisedMcpToolsForRuntime } from './mcp-tool-visibility.js';
+import { getProjectContext } from '../utils/project-name.js';
+
+function scopeMemoryArgs(args: any, toolName: string): any {
+  const projectContext = getProjectContext(process.cwd());
+  const currentProject = projectContext.primary;
+  const requestedProject = typeof args?.project === 'string' ? args.project.trim() : '';
+  const allowCrossProject = process.env.CLAUDE_MEM_ALLOW_CROSS_PROJECT_SEARCH === '1';
+  const serverPath = (process.argv[1] ?? '').replace(/\\/g, '/').toLowerCase();
+  const currentPlatformSource = serverPath.includes('/.codex/plugins/')
+    ? 'codex'
+    : serverPath.includes('/.claude/plugins/') ? 'claude' : '';
+  const requestedPlatformSource = typeof args?.platformSource === 'string'
+    ? args.platformSource.trim().toLowerCase()
+    : '';
+  const allowCrossPlatform = process.env.CLAUDE_MEM_ALLOW_CROSS_PLATFORM_SEARCH === '1';
+  const scopedArgs = { ...args };
+
+  if (requestedProject === '*') {
+    if (!allowCrossProject) {
+      throw new Error(`${toolName}: cross-project memory access is disabled`);
+    }
+    delete scopedArgs.project;
+  } else {
+    if (requestedProject && !projectContext.allProjects.includes(requestedProject) && !allowCrossProject) {
+      throw new Error(
+        `${toolName}: project "${requestedProject}" is outside current repository "${currentProject}"`
+      );
+    }
+    scopedArgs.project = requestedProject || currentProject;
+  }
+
+  if (requestedPlatformSource && currentPlatformSource &&
+      requestedPlatformSource !== currentPlatformSource && !allowCrossPlatform) {
+    throw new Error(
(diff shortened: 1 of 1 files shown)


thedotmack:
Night-ship Batch E skipped this for merge.

Gate FAIL:
- §3 env-var alternate modes (`CLAUDE_MEM_ALLOW_CROSS_PROJECT_SEARCH` / `CLAUDE_MEM_ALLOW_CROSS_PLATFORM_SEARCH`) preserve the old global behavior behind flags.
- Greptile P1 still stands: authorized `project: "*"` is routed as a single-project `/v1/search` and never returns the cross-project set.

Defaulting omitted `project` to the current repo is the real bug. Resubmit that half without the env-var escape hatches and with wildcard routing fixed.

thedotmack:
Night-ship Batch E leftovers — Gate **FAIL**. Leaving open; not rebasing or merging.

The one-file +46 helper still does more than the logged bug. Defaulting an omitted `project` to the current repo is the real fix, but this patch also adds `CLAUDE_MEM_ALLOW_CROSS_PROJECT_SEARCH` / `CLAUDE_MEM_ALLOW_CROSS_PLATFORM_SEARCH` escape hatches, sniffs the host from `process.argv[1]`, and fail-closes foreign projects. Greptile’s P1 is still true: authorized `project: "*"` is then routed as a single-project `/v1/search`. #3764 was closed as not_planned into plan-20 (#3608), so this MCP-only policy layer is the wrong place to land that work.

thedotmack:
Closing night-ship Gate FAIL (see leftovers comment). Scope creep beyond omitted-project default; #3764 was not_planned into plan-20 (#3608). Reopen a narrow PR if needed.

Issue #3678, opened 2026-08-21
OpenCode: session-init and observation POSTs land on different sdk_sessions rows, so no observations are ever generated (13.15.3)


kevinchiha:
## Summary

On 13.15.3, an OpenCode session produces **two** `sdk_sessions` rows sharing one `content_session_id`. `/api/sessions/init` upserts one; the first `/api/sessions/observations` POST resolves to a different one. Tool activity accumulates on the second row, while the generator runs against the first and correctly reports that it saw no tool executions. Net result: **OpenCode sessions never generate a single observation.**

Prompts and sessions are recorded. Observations are not.

## Environment

- claude-mem **13.15.3**, installed via `npx claude-mem install --ide opencode`
- opencode **1.18.19**
- Linux 7.1.8-1-cachyos, node 25.9.0, uid 1000
- Worker: `worker-service.cjs`, port 37700, provider `claude` / `claude-haiku-4-5-20251001`

## Before this reproduces: the stock plugin does not load

Worth stating up front, because you cannot reach this bug on a clean install. With the file exactly as `install --ide opencode` writes it, opencode refuses it:

```
ERROR failed to load plugin path=file:///home/kevin/.config/opencode/plugins/claude-mem.js
      error="Plugin export is not a function"
```

opencode's loader walks **every** value in `Object.values(module)` and throws on anything that is neither a function nor `{server: fn}`. The bundle's final statement is:

```js
export{cg as ClaudeMemPlugin,x$ as REAL_OPENCODE_EVENT_TYPES,b$ as REGISTERED_OPENCODE_HOOKS,k$ as default,ug as parseSearchResponse};
```

`REAL_OPENCODE_EVENT_TYPES` and `REGISTERED_OPENCODE_HOOKS` are plain objects, so the module is rejected wholesale. Reducing it to `export{cg as ClaudeMemPlugin,k$ as default};` loads correctly — the loader dedupes by identity and those two are the same reference, so it still registers exactly one plugin. (Note `parseSearchResponse` must **not** stay exported: it is a distinct function and the loader would register it as a second plugin.)

Everything below is with that one-line fix applied.

## Reproduction

1. Apply the export fix above, and set `CLAUDE_MEM_WORKER_PORT` to `37700 + (uid % 100)` — see the last section.
2. Start the worker, then from inside any git repo:
   ```
   opencode run "Read README.md and say in one sentence what this repo is."
   ```
3. Inspect the DB.

## Worker log

```
[...]

PR #3766, opened 2026-08-27
fix(opencode): read tool arguments from hook input (refs #3678)

rodboev:
## Summary

Issue #3678 reports tool arguments on the input object passed to tool.execute.after, while the plugin reads output.args. On that reported shape, observations send tool_input as {}, so the observer receives no file paths, commands, or diffs. The plugin now prefers input.args while retaining the existing output fallback.

## Why

The local hook interface declares args on the wrong side of the provider boundary, and index.ts:205 follows that model. The change corrects the adapter field selection without changing the rest of the observation payload.

## Scope

This covers tool argument extraction for tool.execute.after. Platform identity, tool-name mapping, plugin exports, worker ports, chat capture, timing, and worker-side behavior remain separate.

## Risk

Input args are used when present; existing output args and empty-object fallback remain. Tool name, response truncation, cwd, fire-and-forget delivery, and other hooks are unchanged.

## Verification

- [x] `bun test tests/integrations/opencode-plugin-contract.test.ts` — 11 passing
- [x] `npm run typecheck` — passed
- [x] `npm run build` — compiled successfully; worker bundle size advisory remains 2992.63 KB against the 2900 KB budget
- [ ] `npm run strip-comments:check` — repository-wide baseline reports `Changed: 413`

The issue comment reports the input/output shape at https://github.com/thedotmack/claude-mem/issues/3678#issuecomment-5417186544; the exact provider runtime isn't available locally, so the output fallback remains in place.

Refs #3678


stantheman0128:
Support the field correction. #3678 puts args on input; reading only output.args sends {}. Prefer input then fallback is the right adapter height.

One expand: `input.args || output.args` treats `{}` as present, so an empty input object shadows real output args. The new test pins that. A nil or key-count check would keep the issue shape without the shadow.

thedotmack:
**issue-patch-ship gate: PASS** (OSS queue sweep)

1. RCA-later: logged in PR body — OpenCode hook args live on `input.args`; adapter was reading `output.args` so observations got `{}`. Deeper host-contract drift stays under plan-23 / #3611.
2. No second-system: single field preference + existing output fallback in the OpenCode adapter only.
3. Not opinionated/wide: +102/−3 across `opencode-plugin/index.ts` + contract test.

Merging now that CI is green (MERGEABLE/CLEAN).

PR #3773, opened 2026-08-27
fix: mention Codex hook trust after install

### LINKED ISSUE #3771: Codex install completes without warning that native hooks still need trust
## What happened

claude-mem 13.15.3 was installed and enabled for Codex, the shared worker was healthy, and MCP search worked, but Codex produced no new sessions, observations, or summaries.

The next interactive Codex startup showed:

```
Hooks need review
5 hooks are new or changed
Hooks can run outside the sandbox after you trust them
```

Until those hooks were trusted, none of the native Codex lifecycle callbacks ran. The current installer prints "Installation complete" and only tells the user to open Codex and restart running sessions, so a healthy install can look like a broken observation pipeline.

## Reproduction

1. Install the Codex integration with `npx claude-mem@latest install`
2. Leave the plugin hooks untrusted
3. Run Codex work that includes tool calls
4. Confirm the worker remains healthy but no Codex observation is generated
5. Start Codex interactively and trust the five claude-mem hooks
6. Run a fresh Codex session with one tool call and exit it

After step 5, the same setup executed `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, and `Stop`. The fresh session received a memory session ID and persisted both an observation and a summary.

Tested with Codex 0.147.0 on Linux.

## Suggested fix

Keep trust as an explicit Codex security decision, but make the installer's next steps say that Codex will ask the user to review and trust the five claude-mem hooks. "Restart sessions" alone is not enough to make capture start.

Kizuno18:
Codex keeps newly installed plugin hooks disabled until the user reviews them. The existing completion message only said to open Codex and restart sessions, which made a healthy install look like a broken observation pipeline.

This adds the missing trust step to the Codex installer output and clarifies that only sessions opened before approval need restarting. The source-copy test keeps that guidance from disappearing.

Tested with:

```
bun test tests/integration/codex-cli-installer.test.ts tests/install-non-tty.test.ts
```

Closes #3771

### DIFF
--- a/src/services/integrations/CodexCliInstaller.ts
+++ b/src/services/integrations/CodexCliInstaller.ts
@@ -472,7 +472,8 @@ Plugin source:     ${marketplaceRoot}
 
 Next steps:
   1. Open Codex CLI in your project
-  2. Restart any running Codex sessions so native hooks are loaded
+  2. Review and trust the five claude-mem hooks when Codex prompts you
+  3. Restart sessions opened before trusting the hooks
 
 For a fresh setup, the supported entry point is:
   npx claude-mem@latest install
--- a/tests/install-non-tty.test.ts
+++ b/tests/install-non-tty.test.ts
@@ -187,6 +187,11 @@ describe('Install Non-TTY Support', () => {
       expect(installRegion).not.toContain('plugin_hooks');
     });
 
+    it('tells Codex users to trust the native hooks before restarting sessions', () => {
+      expect(codexInstallerSource).toContain('Review and trust the five claude-mem hooks when Codex prompts you');
+      expect(codexInstallerSource).toContain('Restart sessions opened before trusting the hooks');
+    });
+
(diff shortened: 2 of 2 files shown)


stantheman0128:
Support. Codex leaving new plugin hooks disabled until trust is a real "install succeeded, observations never fire" path. The source-copy test is the right pin. This should not automate trust.

One expand: the interactive Codex success path still drops those next-steps. bufferConsole captures the two new lines, then the installer returns a generic completion label and never prints output. Non-TTY copy is covered; the default interactive install is the path that looks broken. Worth rendering that buffer on success here rather than a second PR.

thedotmack:
NIGHT SHIP Batch F — Gate **PASS**

Messaging-only: Codex installer next-steps plus a source-copy test. No install-path behavior change. Cherry-picks cleanly onto current `main`. Waiting for #3826 to merge first, then rebase + CI + squash-merge.

PR #3784, opened 2026-08-28
fix: guard on dependency completeness, not node_modules existence (fixes #3755)

### LINKED ISSUE #3755: Setup dependency guard checks that node_modules exists, not that it is complete — a partial install wedges every hook

chiang21fcb:
## Summary

Fixes #3755: the Setup dependency guard now checks whether each declared dependency has its own `package.json` inside `node_modules`, instead of only checking if the `node_modules/` directory exists. A partially-populated `node_modules/` from an interrupted install, killed process, or partial extraction is now caught and retried instead of being permanently skipped.

## Background

`ensurePluginDependencies()` in `plugin/scripts/version-check.js` guarded on `existsSync(node_modules)`. If `node_modules/` existed but was incomplete, the guard returned early on every subsequent Setup run, and the worker died on `Cannot find module` at boot forever. The only recovery path was manual `rm -rf node_modules`.

## Changes

- `plugin/scripts/version-check.js`: added `missingDeps()` helper that reads `package.json` dependencies and verifies each one has its own `package.json` inside `node_modules`. The guard now calls `missingDeps(pluginRoot).length === 0` instead of `existsSync(node_modules)`.
- `tests/plugin-version-check-ensure-deps.test.ts`: updated the "skips install" test to create complete dependency files so the completeness guard passes. Added a new test for the partial install scenario where `node_modules/` exists but is incomplete.

## Verification

- Existing tests updated to match the new guard semantics
- New test added for the partial install scenario
- The cleanup-after-failed-install path still works correctly

### DIFF
--- a/plugin/scripts/version-check.js
+++ b/plugin/scripts/version-check.js
@@ -55,12 +55,22 @@ function findBun() {
 // has a 300s timeout (vs 60s for SessionStart), runs once per Claude
 // Code launch, and is the only standalone hook script — the natural
 // place to materialise plugin runtime state.
+function missingDeps(pluginRoot) {
+  const pkg = JSON.parse(readFileSync(join(pluginRoot, 'package.json'), 'utf-8'));
+  return Object.keys(pkg.dependencies || {}).filter(
+    (d) => !existsSync(join(pluginRoot, NODE_MODULES_DIRNAME, ...d.split('/'), 'package.json'))
+  );
+}
+
 function ensurePluginDependencies(pluginRoot) {
   if (!existsSync(join(pluginRoot, 'package.json'))) return;
 
-  // Guard on node_modules (package-manager marker) rather than a specific
-  // package, so the check stays correct if dependencies are later renamed.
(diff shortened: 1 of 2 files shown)


stantheman0128:
Yield to #3769 for the same #3755 hole. This PR is the right diagnosis (node_modules existing is not a finished install) at the package.json-exists height. Greptile's P1 is the next layer: a retained zod/package.json with no v3 export still skips repair.

#3769 already does the closure check, the Zod subpath probe, tree-sitter provision, and the Windows bun.exe spawn (not shell:true). I would rather one vehicle land than two guards for the same Setup skip.

chiang21fcb:
﻿@stantheman0128 thanks for the review and for pointing to #3769. I agree that it covers more ground and I have closed this PR in favor of it. The diagnosis here is captured in the issue thread for reference.

stantheman0128:
@chiang21fcb thanks for reading #3769 and closing this.

PR #3797, opened 2026-08-29
test: restore the real modules after mock.module, not the stubs

Max2535:
Two tests on `main` fail or pass depending on the order bun happens to walk the test files:

```
(fail) mutation sites > adoptMergedWorktrees emits remap_project through its own connection
(fail) worktree adoption Chroma hydration > patches a session-summary document when the adopted worktree has no observations
```

Both pass when their files run alone, and both report `adoptedSummaries` of 0 where 1 is expected — which reads like a bug in the worktree adoption path. It isn't.

## Cause

`tests/cli/handlers/context-session-start.test.ts` restores four stubbed modules in `afterAll` by spreading the imported namespace at that moment:

```js
mock.module('.../project-name.js', () => ({ ...realProjectName }));
```

`mock.module` rewrites the live namespace object in place, so by the time `afterAll` runs, `realProjectName` **is** the stub. The restore re-registers the stub and is a silent no-op.

`getProjectContext` then answers `'repo-project'` for every path for the rest of the run. `WorktreeAdoption` computes a worktree project name that matches no seeded row, and the `UPDATE` reports 0 changes.

Found by instrumenting `WorktreeAdoption` to print the project name it computed; it printed `repo-project`, which appears in exactly one stub.

## Fix

The idiom the rest of the suite already uses — capture `{ ...module }` at load time, above the `mock.module` calls, and re-register that object. `tests/transcripts/processor-codex-context.test.ts` spells out the same reasoning in a comment.

`tests/worker/session-manager-project.test.ts` has the other half of the same problem: its `worker-service` and `worker-utils` stubs each replace a whole module namespace with a single export and are never restored at all, so `getWorkerPort` stays pinned at `37777` for every file after it. Nothing on `main` currently trips over that one — it's a trap for the next test that boots a server on a real port and expects a client to reach it. Given the same guard.

Swept `tests/` for both shapes; no other occurrence.

## Result

```
bun test tests
before  2714 pass / 27 skip / 2 fail
after   2716 pass / 27 skip / 0 fail   (three consecutive runs)
```

## One thing I did not fix

`mutation sites > adoptMergedWorktrees` timed out at exactly 5s on the *first* suite run in a freshly created worktree — once out of four such runs, across two separate checkouts — and passed on every run after. It looks like a cold page cache meeting a tight 5s budget for real `git init` / `git worktree add` work, not a logic bug, so I left it alone. Flagging it because a 5s timeout on a test that shells out to git that many times may deserve a longer budget or a warm-up.

### DIFF
--- a/tests/cli/handlers/context-session-start.test.ts
+++ b/tests/cli/handlers/context-session-start.test.ts
@@ -5,6 +5,20 @@ import * as realOauthToken from '../../../src/shared/oauth-token.js';
 import * as realProjectName from '../../../src/utils/project-name.js';
 import * as realWorkerUtils from '../../../src/shared/worker-utils.js';
 
+// Snapshot the real modules BEFORE mock.module mutates the live namespaces.
+// bun's mock.module rewrites the existing namespace object in place, so
+// spreading `realProjectName` from inside afterAll re-registers the STUB, not
+// the real module — the restore silently becomes a no-op. That left
+// getProjectContext pinned to 'repo-project' for every later file in the run,
+// which is what made adoptMergedWorktrees adopt nothing in
+// tests/worker/sync/mutation-sites.test.ts and
+// tests/services/infrastructure/worktree-adoption-chroma.test.ts. Same idiom
+// and same reason as tests/transcripts/processor-codex-context.test.ts.
+const realHookSettingsSnapshot = { ...realHookSettings };
+const realOauthTokenSnapshot = { ...realOauthToken };
+const realProjectNameSnapshot = { ...realProjectName };
+const realWorkerUtilsSnapshot = { ...realWorkerUtils };
+
 const calls: unknown[][] = [];
 
 mock.module('../../../src/shared/hook-settings.js', () => ({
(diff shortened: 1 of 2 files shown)


thedotmack:
Night-ship Batch E skipped this as already on main.

`tests/cli/handlers/context-session-start.test.ts` already snapshots modules before `mock.module` and restores those snapshots. The remaining `session-manager-project.test.ts` worker-service/utils restore is hardening — nothing on current main fails without it.

thedotmack:
Night-ship Batch E leftovers — Gate **FAIL** / **SKIP**. Leaving open; not opening a solo PR.

The real fix — snapshot modules before `mock.module` and restore those snapshots in `context-session-start.test.ts` — is already on main. The leftover `session-manager-project.test.ts` worker-service / worker-utils restore is hardening; nothing on current main fails without it. Test-only leftovers fold into a companion merge. There is no PASS patch in this leftovers batch to fold it into, and a solo test PR is not worth a bump.

thedotmack:
Closing night-ship Gate FAIL / superseded (see leftovers-run comment). Not rebasing. Fresh bugs preferred over babysitting this CONFLICTING PR.

PR #3806, opened 2026-08-31
fix(worker): recover observer batches after returned context-overflow prose (closes #3800)

### LINKED ISSUE #3800: Worker stalls indefinitely on "Prompt is too long": consecutiveInvalidOutputs is reset by the rejection it counts
## Summary

On a machine running many concurrent Claude Code sessions, the worker stopped producing observations for 94 minutes while 28 sessions were active. Every batch was being rejected by the SDK with `Prompt is too long`, and nothing in the worker ever escalated, backed off, or recycled the conversation.

## Log evidence

```
[SDK   ] [session-1406] ← Response received (18 chars) {promptNumber=29} Prompt is too long
[PARSER] [session-1406] SDK returned non-XML prose response — ignoring queued batch
         {outputClass=prose, preview=Prompt is too long, consecutiveInvalidOutputs=0}
```

2,264 occurrences in a single day, first at 08:50:45, across more than ten distinct `session-N` ids out of 69 active sessions.

## What it is not

The worker broadcasts `queueDepth=13040`, which makes this look like a queue-size problem. It is not. On disk:

| table | rows | notes |
|---|---|---|
| `pending_messages` | 48 | all `failed`; 53,301 bytes total; largest single row 5,625 bytes |

48 rows totalling ~53KB cannot produce an over-long prompt under any plausible ceiling. `queueDepth` is an in-memory counter, not the durable queue, and draining or splitting the queue does not address the failure.

## What it appears to be

The rejection is per *session*, not per batch: `promptNumber=29` is the 29th prompt on one persistent SDK conversation. The worker keeps a long-lived SDK conversation per Claude session and appends to it, so the thing that grows past the ceiling is the accumulated conversation, not the pending rows. That matches both observed properties: it worsens over the day, and it hits many sessions at roughly the same time as each one reaches the ceiling in turn.

## The part that makes it silent

`consecutiveInvalidOutputs=0` in the rejection line. The counter that exists to trip a breaker after consecutive invalid outputs is reset by the very event it is meant to count, so it never reaches any threshold. The batch is rejected, dropped, and reloaded identically; there is no state the loop can exit from, and the only signal that would have surfaced the condition is zeroed by hand.

## Suggested direction

1. Do not reset `consecutiveInvalidOutputs` when the output is a rejection rather than a valid non-XML response - or count rejections separately.
2. On `Prompt is too long`, recycle the SDK conversation for that session (start a fresh one, carrying only what is needed) rather than re-sending into a conversation that can only keep growing.
3. Consider an explicit ceiling on the accumulated conversation before send. There is currently no `CLAUDE_MEM_*` setting for batch or prompt size among the 102 available, so operators have no knob for this.

## Environment

claude-mem 13.18.0, macOS, SQLite backend, Chroma enabled, cloud sync not configured.

## Possibly separate finding: `sync_outbox` is never pruned

`sync_outbox` holds 305,387 rows with rowids running 1..305,388 - essentially nothing has ever been deleted. The only delete path found in `worker-service.cjs` lives inside the cloud-sync class, which is constructed only when `CLAUDE_MEM_CLOUD_SYNC_TOKEN`, `CLAUDE_MEM_CLOUD_SYNC_USER_ID` and `CLAUDE_MEM_CLOUD_SYNC_HUB_URL` are all set. With cloud sync unconfigured, rows are enqueued unconditionally and never consumed, so the table grows without bound for the lifetime of the install. Happy to split this into its own issue if you prefer.

rodboev:
## Summary

When the Claude Agent SDK returns `Prompt is too long` as assistant text, the worker treated it as ordinary prose, confirmed the claimed observer batch, and kept using the exhausted conversation. The worker now recognizes this narrow returned-text rejection, preserves the first claimed batch, aborts the generator, and starts a fresh generator only when preserved work remains. The replacement generator keeps the retry budget and pending timestamp through its unclaimed init response until the preserved batch is reclaimed. A repeated overflow drops the unfit claimed batch once instead of replaying it indefinitely. An overflow before any batch is claimed retains pending entries without scheduling an immediate restart. Shutdown teardown cannot trigger a replacement start, and reverse-direction context-limit discussion remains ordinary prose.

Closes #3800

## Why

The thrown-error classifier already recognizes context overflow, but this response arrives through `ClaudeProvider` as assistant content and reaches `ResponseProcessor` as invalid non-XML text. The generic branch confirms the claimed messages and leaves the generator alive, so the same over-full conversation receives the next prompt. The fix applies the existing quota/auth response-policy seam, retains overflow sessions through generator exit, and uses the existing route owner for one pending-work restart.

## Scope

This change covers only the observed SDK reactive returned-text path. It does not claim aggregate SDK context control, thrown provider-overflow normalization, or durable recovery. It also does not add proactive history limits, change `ClaudeProvider.ts`, add a prompt-size setting, change non-SDK providers, modify `sync_outbox`, or alter queue persistence and transcript recovery.

## Risk

The discriminator is inferred from untyped returned provider text. The installed SDK declaration exposes a broad optional assistant error enum, but it does not identify context overflow or establish that `invalid_request` means `Prompt is too long`. Anthropic service-side emission and retry behavior remain unverified. The classifier is narrow and XML-guarded. A false negative retains the previous behavior; a false positive gets one bounded fresh-context retry and then drops the claimed batch. Generator starts reserve the shared promise before asynchronous tier routing so automatic restart and ingest cannot create two providers for one session. No storage, settings, telemetry vocabulary, or cross-process state changes.

## Verification

- [x] `bun test tests/sdk/output-classifier.test.ts tests/worker/agents/response-processor.test.ts tests/worker/poison-respawn.test.ts tests/worker/overflow-restart.test.ts`
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm run lint:hook-io && npm run lint:spawn-env && npm run strip-comments:check`

Closes #3800


rodboev:
Fixed in ac91b6d5: `handleGeneratorExit` now preserves an overflow exit only when `overflowRetryPending` is true. Exhausted overflow exits use the existing finalize-and-remove path.

Added regression coverage in `tests/worker/poison-respawn.test.ts` for exhausted overflow cleanup and made the active-retry precondition explicit. The focused worker suite passes with 60 tests, and all required CI checks pass on the new head.

rodboev:
Superseded by merged PR https://github.com/thedotmack/claude-mem/pull/3807, which implements the current fix for issue https://github.com/thedotmack/claude-mem/issues/3800 on `main`.

PR #3808, opened 2026-08-31
fix(worker): make the LLM per-attempt deadline configurable and stop retrying it

### LINKED ISSUE #3794: Observer request timeout is hardcoded at 30s and not configurable, and timeouts are retried as transient
When the observer runs against a local model, the per-attempt request timeout of 30 seconds is unreachable from configuration, and a timeout is classified as `transient` and retried twice. On slower hardware this turns a latency problem into a congestion collapse.

## Where it comes from

`src/services/worker/retry.ts`:

```
const DEFAULT_OPTIONS: Required<Omit<RetryOptions, "label" | "abortSignal">> = {
  maxRetries: 2,
  perAttemptTimeoutMs: 30_000,
  baseDelayMs: 100,
  maxDelayMs: 30_000,
};
```

There are exactly two `withRetry` call sites, the OpenRouter one and the Gemini one, and neither passes an override:

```
withRetry(async signal => { ... fetchChatCompletion(..., signal) }, { label: `OpenRouter ${model}` })
withRetry(async signal => { ... fetchGenerateContent(..., signal) }, { label: `Gemini ${model}` })
```

So the default always wins. No environment variable reaches it.

## Two things that make it worse

**Timeouts are retried.** An aborted request is classified `transient` (the `status === undefined` branch, which produces `OpenRouter network error: The operation was aborted.`) and retried up to `maxRetries: 2`. When the backend is already saturated, each expired request injects two more 30-second requests into the same queue. In our logs the abort count went from 4/day to over 1000/day without any change to the backend, which is the signature of a positive feedback loop rather than gradual degradation.

**`CLAUDE_MEM_API_TIMEOUT_MS` looks like the fix but is not.** It is read only by the internal HTTP client that talks to the local worker on 127.0.0.1, not by the provider call. Both values are 30000 because they derive from the same `API_REQUEST` constant, which makes the coincidence convincing. We spent a while on this before checking, and at least one comment on #3490 attributes the LLM timeout to that variable as well, so it may be worth documenting the distinction regardless of what happens to this issue.

## Measured

Ubuntu 24.04, single RTX 5060 Ti 16GB shared with other workloads, Intel i3-4160 (2c/4t). Local model via `CLAUDE_MEM_OPENROUTER_BASE_URL` pointed at Ollama 0.33.1, model `gemma4:e4b`, prompts of 18k-28k tokens.

Over 2111 requests recorded by the backend:

- successful requests: median 20.99s, p90 28.07s, p99 29.77s, max 29.99s
- failed requests: all at exactly 30.000s
- 27.4% of successful requests land between 25 and 30 seconds

The distribution is truncated by the deadline, not slow: the fastest successful request was 1.69s. Everything past 30 seconds is discarded work that was already computed.

## Suggested fix

Allow the two call sites to pass `perAttemptTimeoutMs` from configuration, for example a `CLAUDE_MEM_LLM_TIMEOUT_MS` with the same validated range already used elsewhere (500 to 300000). Users on hosted providers keep the current 30s default and see no change; users on local models can set a value that matches their hardware.

A second, smaller suggestion: consider not retrying an aborted-by-deadline request, or making `maxRetries` configurable alongside it. A request that exceeded the deadline once is likely to exceed it again, and the two extra attempts are what turns a slow backend into a saturated one.

Happy to test a patch. As a stopgap we are rewriting `perAttemptTimeoutMs:3e4` to `9e4` in the installed bundle after every plugin update, which works but is obviously not something to recommend.

ntdatt812:
Fixes #3794.

## The deadline was unreachable

`perAttemptTimeoutMs` defaulted to 30s in `DEFAULT_OPTIONS`, and neither of the two `withRetry` call sites passes an override, so no configuration could reach it. That is fine against a hosted provider and wrong against a local model — the report's Ollama backend measured, over 2111 requests:

- successful: median 20.99s, p90 28.07s, **p99 29.77s**, max 29.99s
- failed: **all at exactly 30.000s**

A distribution truncated by the deadline, not a slow one — the fastest success was 1.69s. Everything past 30s was work already computed and then discarded. The reporter's stopgap was rewriting `perAttemptTimeoutMs:3e4` in the installed bundle after every plugin update.

`CLAUDE_MEM_LLM_TIMEOUT_MS` now sets it, validated against the same `500..300000` bounds the other `CLAUDE_MEM_*_TIMEOUT_MS` settings use, with the same warn-and-fall-back behaviour on a bad value. Hosted users keep 30s and see no change.

## The deadline was also retried

An abort from our own deadline reaches the classifier with no HTTP status, so it landed in the `transient` branch and ran two more attempts. Against a backend that is already saturated, those attempts are what turn a latency problem into a congestion collapse — and a request that missed the deadline once will very likely miss it again.

It now fails immediately with a message naming the deadline and the setting that raises it, rather than a bare `The operation was aborted.` The original error is preserved as `cause`. Genuine transient failures still retry exactly as before.

## One thing to check

`resolveLlmTimeoutMs` duplicates the shape of `readTimeoutEnv` in `src/shared/worker-utils.ts` rather than importing it. `retry.ts` is a leaf that imports only the logger and the error classifier, while `worker-utils` pulls in the supervisor, telemetry and settings manager — importing it from here would drag all of that into the retry path. Happy to export `readTimeoutEnv` and use it instead if you would rather have the single copy.

## Tests

```
bun test tests/worker/llm-timeout.test.ts tests/worker/retry-policy.test.ts
12 pass, 0 fail
```

`tests/worker/llm-timeout.test.ts`:

- default is 30s with nothing configured
- accepts values at and inside the bounds (500, 90000, 300000)
- rejects `0`, `-1`, `499`, `300001`, `abc` and empty — a zero would disable the deadline, a huge one would park a worker for hours
- a request that blows the deadline runs **once**, not three times, and fails fast
- a genuine transient failure still retries

**Mutation-checked.** Letting the deadline abort fall through to the retry branch:

```
4 pass, 1 fail — does not retry a request that blew the deadline
```

The existing `tests/worker/retry-policy.test.ts` is untouched and still passes.

### DIFF
--- a/src/services/worker/retry.ts
+++ b/src/services/worker/retry.ts
@@ -45,9 +45,45 @@ export interface RetryOptions {
   abortSignal?: AbortSignal;
 }
 
+/** Bounds shared with the other CLAUDE_MEM_*_TIMEOUT_MS settings. */
+const LLM_TIMEOUT_BOUNDS = { min: 500, max: 300_000 } as const;
+const FALLBACK_PER_ATTEMPT_TIMEOUT_MS = 30_000;
+
+/**
+ * Per-attempt deadline for a provider request.
+ *
+ * 30s suits a hosted provider and is far too short for a local model: a
+ * report on an Ollama backend measured successful requests with a median of
+ * 21s and a p99 of 29.8s, so the deadline was truncating work that had
+ * already been computed. The value was unreachable from configuration, and
+ * the workaround was editing the installed bundle after every update.
+ *
+ * Read here rather than through worker-utils' readTimeoutEnv: this module is
+ * a leaf that imports only the logger and the error classifier, and that one
+ * pulls in the supervisor and telemetry.
+ */
+export function resolveLlmTimeoutMs(env: NodeJS.ProcessEnv = process.env): number {
+  const raw = env.CLAUDE_MEM_LLM_TIMEOUT_MS;
+  if (!raw) return FALLBACK_PER_ATTEMPT_TIMEOUT_MS;
+  const parsed = Number.parseInt(raw, 10);
+  if (
+    Number.isFinite(parsed)
+    && parsed >= LLM_TIMEOUT_BOUNDS.min
+    && parsed <= LLM_TIMEOUT_BOUNDS.max
+  ) {
+    return parsed;
+  }
+  logger.warn('SDK', 'Invalid CLAUDE_MEM_LLM_TIMEOUT_MS, using default', {
+    value: raw,
+    min: LLM_TIMEOUT_BOUNDS.min,
+    max: LLM_TIMEOUT_BOUNDS.max,
+  });
+  return FALLBACK_PER_ATTEMPT_TIMEOUT_MS;
+}
+
 const DEFAULT_OPTIONS: Required<Omit<RetryOptions, 'label' | 'abortSignal'>> = {
   maxRetries: 2,
-  perAttemptTimeoutMs: 30_000,
+  perAttemptTimeoutMs: resolveLlmTimeoutMs(),
   baseDelayMs: 100,
   maxDelayMs: 30_000,
 };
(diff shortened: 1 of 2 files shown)


thedotmack:
ALL-NIGHT SHIP Batch D: rebased onto latest `main` as #4000 (plus rejecting suffixed values like `90000ms`). Will merge #4000 if CI is green.

thedotmack:
Superseded by #4000.

PR #3810, opened 2026-08-31
fix(search): answer CJK and Japanese queries by substring

### LINKED ISSUE #3801: [BUG] Chinese (CJK) phrase search always returns 0 — FTS5 unicode61 tokenizer can't segment CJK; not covered by #3610
## Summary

Chinese-language phrase search returns **0 results** even though matching observations exist. Root cause: the FTS5 fallback table (`observations_fts`) uses the default `unicode61` tokenizer, which splits on whitespace — CJK text has no spaces, so an entire Chinese sentence becomes a single token and any phrase/partial query never matches.

This is the **read-side** sibling of #3423 (CJK backfill, write-side) and is **not** covered by #3610/plan-22 (which fixes the fallback *trigger* — empty Chroma → FTS5 — but does not fix FTS5's ability to match Chinese).

## Repro (v13.17.0, macOS, SQLite observations_fts)

```sql
-- 826 observations contain 用户 in title/narrative/text
SELECT COUNT(*) FROM observations WHERE title LIKE '%用户%' OR narrative LIKE '%用户%' OR text LIKE '%用户%';  -- 826

-- Chinese phrase query → 0 hits
SELECT COUNT(*) FROM observations_fts WHERE observations_fts MATCH '"用户身份"';  -- 0

-- Chinese single-token query → hits (whole-sentence-as-one-token edge)
SELECT COUNT(*) FROM observations_fts WHERE observations_fts MATCH '用户';     -- 41

-- English phrase → normal behavior
SELECT COUNT(*) FROM observations_fts WHERE observations_fts MATCH '"Database Path"';  -- 40
```

The FTS table declares no `tokenize=` option (`.schema observations_fts`), so it falls back to the default `unicode61` tokenizer, which is whitespace-based and fundamentally cannot segment CJK.

## User-visible impact

With the MCP `search` tool, calling without a `type` (or with `type: "observations"` on a project whose rows aren't synced to Chroma) routes to `executeWithFallback` → FTS5 → **Chinese keyword searches always return 0 observations**. For Chinese-speaking users this makes the memory search feature unusable for any multi-character keyword.

## Suggested fixes (any one)

1. **Trigram tokenizer** — add `tokenize = 'trigram'` to `observations_fts` (and `user_prompts_fts`, `session_summaries_fts`, `compiled_summaries_fts`). Trigram indexes all 3-char substrings, so CJK phrases of length ≥3 match. Requires table rebuild + repopulation (or a migration path, since FTS tables are populated by triggers).
2. **`unicode61` with `remove_diacritics` + CJK-aware preprocessing** — insufficient alone; trigram is the standard fix.
3. **Hybrid strategy** — when the semantic (Chroma) path and FTS both miss and the query contains CJK, run a `LIKE '%keyword%'` scan over `observations` as a final fallback (accurate, slower, fine as last resort).

## Environment

- claude-mem 13.17.0 (also reproduced conceptually on 13.15.3 — search code identical)
- macOS arm64, SQLite FTS5, default plugin install

## Related

- #3361 (confident-empty fallback) → tracked in #3610 plan-22
- #3423 (CJK backfill write-side, CLOSED)

ntdatt812:
Fixes #3801.

## ⚠️ The suggested trigram fix does not fix the reported case

Before writing anything I built the trigram index the issue proposes and queried it:

```
trigram  "用户身份"  -> 1
trigram  "用户"      -> 0     <- the issue's own primary example (826 rows contain it)
trigram  "数据"      -> 0
trigram  "db"        -> 0     <- regression: unicode61 answers this today
trigram  "database"  -> 1
```

Trigram indexes 3-character substrings, so it cannot answer a 2-character query **at all**. `用户` is exactly the query the issue opens with, and `db` is an English query that works today. Trigram would also require detecting and rebuilding the FTS tables on every existing install, since they are created with `CREATE VIRTUAL TABLE IF NOT EXISTS` and would otherwise silently keep the old tokenizer.

So I took suggestion 3 instead, which is accurate for every query length and needs no migration.

## The fix

`unicode61` has no delimiter to split Chinese or Japanese on, so a whole run of ideographs folds into one token — and every query in those scripts is a *substring* of some longer run, which can never match a token. Route those queries to a substring predicate over the same columns the FTS table indexes.

This is not a new mechanism in this class. `searchUserPrompts` has always answered every query this way, escaping included:

```ts
const escapedQuery = query.replace(/[\%_]/g, '\$&');
baseConditions.push("up.prompt_text LIKE ? ESCAPE '\'");
```

The new helper is that line, generalised over a column list. Queries FTS5 *can* tokenize keep the FTS path untouched — same SQL, same `rank` ordering.

One correction to the issue: `user_prompts_fts` needs nothing, because prompt search never used it. The affected paths are `searchObservations` and `searchSessions`.

## Scope and cost

The predicate scans `observations` / `session_summaries` rather than an index. That is the price of a correct answer here — there is no index in the schema that can represent these queries. It runs only when the query contains Hiragana, Katakana, or a CJK ideograph, so no English query changes plan. If you would rather bound it further, the natural next step is a generated bigram column, and I am happy to send that instead.

## Tests

`tests/services/sqlite/search-cjk-fallback.test.ts`, 8 tests: Chinese inside a longer run, a 2-character Chinese query, Japanese, session summaries, `project` and `type` filters still applied on the new path (the parameter order is easy to get wrong), `%`/`_` in a query treated literally, and English unchanged for both observations and summaries.

Reverting the change proves they hold it:

| mutation | result |
|---|---|
| dispatch never fires | 6 fail |
| drop the wildcard escaping | 1 fail |

```
bun test tests/services/sqlite/search-cjk-fallback.test.ts     8 pass, 0 fail
bun test tests/services/sqlite/ tests/worker/search/ \
         tests/sqlite/ tests/storage/sqlite/                   204 pass, 4 fail
tsc --noEmit                                                   clean
```

The 4 failures are the pre-existing Windows `EBUSY: resource busy or locked` teardown failures in `session-store-orphan-fk-repair.test.ts`. I confirmed they are identical on a clean tree — same four names, before and after.

### DIFF
--- a/src/services/sqlite/SessionSearch.ts
+++ b/src/services/sqlite/SessionSearch.ts
@@ -232,12 +232,25 @@ export class SessionSearch {
   }
 
   /**
-   * Scripts written without word delimiters: Hiragana, Katakana, and the CJK ideograph
-   * blocks. FTS5's unicode61 tokenizer has no delimiter to split them on, so an entire run
-   * folds into a single token and no substring of that run can match it (#3801) — and every
-   * query in these scripts is a substring of some longer run.
+   * Scripts whose runs FTS5's unicode61 tokenizer cannot split: Hiragana, Katakana, the CJK
+   * ideograph blocks, Bopomofo, and Hangul. unicode61 breaks on Unicode whitespace and
+   * punctuation, and these scripts put neither between the characters of a run — so a run
+   * folds into a single token and no substring of it can ever match (#3801), which is every
+   * query a user types in them.
+   *
+   * Korean does space its words, so only the sub-word case is affected there — but that is
+   * still every partial-word query. Measured directly against `tokenize='unicode61'`:
+   *
+   *   설정   inside 설정을            -> 0 rows
+   *   설정을 as a whole token         -> 1 row
+   *   ㄓㄨ   inside ㄓㄨㄛ            -> 0 rows
+   *   项目   inside 修改了项目配置    -> 0 rows
+   *
+   * Bopomofo and Hangul were raised in review on #3810. The blocks are adjacent, so
+   * \u3100-\u318F covers Bopomofo together with the Hangul compatibility jamo beside it.
    */
-  private static readonly UNSEGMENTED_SCRIPT = /[\u3040-\u30FF\u3400-\u4DBF\u4E00-\u9FFF\uF900-\uFAFF]/;
+  private static readonly UNSEGMENTED_SCRIPT =
+    /[\u3040-\u30FF\u3100-\u318F\u3400-\u4DBF\u4E00-\u9FFF\uAC00-\uD7AF\uF900-\uFAFF]/;
 
   /**
    * Build the substring predicate used when the index cannot represent the query. The
--- a/src/services/sqlite/SessionSearch.ts
+++ b/src/services/sqlite/SessionSearch.ts
@@ -231,6 +231,26 @@ export class SessionSearch {
     return conditions.length > 0 ? conditions.join(' AND ') : '';
   }
 
+  /**
+   * Scripts written without word delimiters: Hiragana, Katakana, and the CJK ideograph
+   * blocks. FTS5's unicode61 tokenizer has no delimiter to split them on, so an entire run
+   * folds into a single token and no substring of that run can match it (#3801) — and every
+   * query in these scripts is a substring of some longer run.
+   */
+  private static readonly UNSEGMENTED_SCRIPT = /[\u3040-\u30FF\u3400-\u4DBF\u4E00-\u9FFF\uF900-\uFAFF]/;
+
+  /**
+   * Build the substring predicate used when the index cannot represent the query. The
+   * escaping matches {@link searchUserPrompts}, which has always searched by substring.
+   */
+  private static buildSubstringClause(query: string, columns: string[]): { clause: string; params: string[] } {
+    const pattern = `%${query.replace(/[\\%_]/g, '\\$&')}%`;
+    return {
+      clause: `(${columns.map(column => `${column} LIKE ? ESCAPE '\\'`).join(' OR ')})`,
+      params: columns.map(() => pattern),
+    };
+  }
+
   private buildOrderClause(orderBy: SearchOptions['orderBy'] = 'relevance', hasFTS: boolean = true, ftsTable: string = 'observations_fts'): string {
     switch (orderBy) {
       case 'relevance':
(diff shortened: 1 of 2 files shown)


stantheman0128:
Support. Trigram cannot answer a 2-character query. Routing those scripts through the same `LIKE ... ESCAPE` helper `searchUserPrompts` already uses is the right height. The new tests go red if the dispatch never fires.

Two sibling scripts sit next to this regex. Bopomofo (`\u3100-\u312F`) and Hangul (`\uAC00-\uD7AF`) are delimiter-free the same way. `unicode61` will fold them into one token too. A 注音 query would still miss.

Not a blocker for the reported Chinese/Japanese hole.

ntdatt812:
Both are in, in `95c92fee`. I measured them before adding rather than reasoning from the block names, since the Korean case is not obviously the same shape:

```
설정   inside 설정을            -> 0 rows
설정을 as a whole token         -> 1 row
ㄓㄨ   inside ㄓㄨㄛ            -> 0 rows
项目   inside 修改了项目配置    -> 0 rows   (the reported case)
project inside "the project"    -> 1 row    (Latin, unaffected)
```

Run against a bare `CREATE VIRTUAL TABLE t USING fts5(body, tokenize='unicode61')`, so it is the tokenizer talking and not this repo's schema.

Korean is worth stating explicitly in the comment, which I did: it *does* space its words, so unlike Chinese and Japanese only the sub-word case breaks. That case is still every partial-word query someone types, and the first two lines above are exactly it — the whole token matches, the prefix does not.

The blocks turned out to be adjacent, so it is one range rather than two: `㄀-㆏` picks up Bopomofo together with the Hangul compatibility jamo next to it, and `가-힯` covers the syllables.

Two tests, one per script, seeded next to the existing Chinese and Japanese fixtures. Narrowing the character class back to the previous ranges fails both, so they are pinned to the change rather than to the dispatch in general.

Thanks — you were right that a 注音 query would still have missed, and it would not have been obvious from the issue.

stantheman0128:
Thanks. Measuring the tokenizer is the right way. `\u3100-\u318F` plus `\uAC00-\uD7AF` closes the 注音 hole.

Half-width Katakana (`\uFF66-\uFF9D`) and phonetic extensions (`\u31F0-\u31FF`) still skip the fallback. Same miss if someone searches ﾕｰｻﾞｰ. Not a blocker for the Chinese / Hangul / Bopomofo cases you pinned.

thedotmack:
**All-night ship gate: PASS**

1. RCA logged — FTS5 `unicode61` folds CJK/Japanese runs into one token, so substring queries such as `用户` never match (Fixes #3801 / plan-25). Trigram cannot answer 2-character queries.
2. No second-system — routes those scripts through the same escaped `LIKE` helper `searchUserPrompts` already uses; English FTS path unchanged.
3. Narrow / not opinionated — dispatch + helper + tests. Half-width Katakana remains a follow-up, not a second system.

Rebased onto `7335467f` locally (fork branch update blocked). Focused tests: `bun test tests/services/sqlite/search-cjk-fallback.test.ts` — 10 pass. This head has no CI runs; opening a rebased ship branch to verify CI before merge. No version bump.

thedotmack:
Update: GitHub accepted the squash merge of this head onto current `main` after the local rebase onto `7335467f` stayed clean and the 10 focused tests passed. No separate ship branch needed.

PR #3811, opened 2026-08-31
fix(context): fit the session block to the hook output limit by selection

### LINKED ISSUE #3802: Context builder should fit Claude Code's ~10K per-hook-output limit by selection, not counts (re-raising #1899 with fresh measurements)
## Summary

The SessionStart context block is built from **item counts**, so on active projects it exceeds Claude Code's per-hook-output limit and the model receives a 2KB `<persisted-output>` stub instead of the context — silently, with the hook reporting success. This was consolidated into #1899 (closed NOT_PLANNED in the April cleanup; #1591 was folded into it). Re-raising with fresh measurements and a proposal that fits the merge rubric (root-cause selection at build time, not bolt-on truncation).

## Fresh measurements (claude-mem 13.15.0, Claude Code 2.1.251, Windows 11)

The limit itself, measured with a probe hook emitting exactly N characters and checked deterministically against the session's `tool-results/` directory and transcript:

| probe | result |
|---|---|
| 10,000 chars | delivered verbatim |
| 10,001 chars | persisted to disk, replaced by a ~2KB preview stub |
| 3 hook commands × 8,000 chars | all delivered — the limit is **per hook output** |
| `hookSpecificOutput.additionalContext` at 20K / 40K | capped the same way (end-marker never reaches the model) |

claude-mem's `hook claude-code context` output measured across 10 real projects on one machine, with `CLAUDE_MEM_CONTEXT_OBSERVATIONS=15, FULL_COUNT=3, SESSION_COUNT=10`:

- worst project: **13,732 chars** (`additionalContext`) / 15,981 (`systemMessage`) → both stubbed, model gets nothing
- second worst: 9,286 — one busy day from the line
- the plugin **default** `CLAUDE_MEM_CONTEXT_OBSERVATIONS=50` makes overflow near-certain on any mature project

## Why count knobs cannot fix it

Items are variable-length: observation title lines run ~150–320 chars, session-summary lines ~250–600, and the last-session summary block is ~1–5K on its own. Any fixed count is either wasteful on quiet projects or over the line on busy ones. We tuned counts down (`FULL_COUNT=0, SESSION_COUNT=5, OBSERVATIONS=8`) and got the worst project to 7,646 — but that is an approximation with ~20% headroom, not a guarantee, and it forfeits context on projects that had room for more.

## Proposal

Make the context builder **fit a character budget by selection**: render, measure, and drop the lowest-value items (oldest first, titles before the last-session summary) until the rendered block fits — the same decision the count knobs approximate, made where the size is actually known. Expose the budget (e.g. `CLAUDE_MEM_CONTEXT_MAX_CHARS`, default ~8,000 ≈ 20% under the measured limit) so users can re-measure after Claude Code upgrades — the ceiling is undocumented and has to be treated as movable. This is not output truncation (no mid-item cuts), so it should be compatible with the no-truncation stance in `docs/merge-rubric.md`.

Happy to share the probe scripts / raw measurements.

ntdatt812:
Fixes #3802. Thanks for the measurements — the 10,000-character boundary and the per-hook (not per-session) scoping are what made this fixable rather than guessable.

## Why counts cannot do it

Your point is the whole design constraint: items are not a fixed size. A title is ~150–320 characters, a session-summary line ~250–600, and the last-session block 1–5K by itself. So any count safe on a quiet project wastes the budget, and any count that fills the budget there overflows on a busy one. The shipped `CLAUDE_MEM_CONTEXT_OBSERVATIONS=50` overflows on titles alone.

## The fix: measure, then drop whole items

`fitContextToBudget` renders, and while the result is over budget it gives up the most expensive thing still in it and renders again:

| given up | why first |
|---|---|
| full observation narratives | largest per-item cost, and their titles stay in the timeline |
| the last-session summary block | one item, 1–5K |
| sessions (halved) | coarser unit than an observation |
| observations (halved, floor 1) | last resort |

Whole items are dropped, never cut in half, so the block is always well-formed — the header, footer and at least one observation always survive. When everything droppable is gone and it is still over, the result says so (`overBudget`) rather than pretending.

This is selection at build time, as you asked for, not truncation bolted on after: the reduced config is fed back through the same `buildContextOutput`, so every render is a real, complete block.

`--full` is an explicit human request for everything, so it passes an unbounded limit and is unaffected.

## Design notes for review

- `ContextBudget.ts` is pure and generic over the item type. It takes the render function as a parameter, so it is tested directly, without a database, at every step of the ladder.
- No new setting. The limit is a measured property of the host, not an operator preference, so it is a constant with your measurement cited next to it. Say the word if you would rather have `CLAUDE_MEM_CONTEXT_MAX_CHARS` and I will add it.
- The reduction ladder is the one judgement call here. I ordered it by cost-per-value from your figures; if you would rather lose sessions before the summary block, that is one line in `reduceConfig`.

## Tests

`tests/context/context-budget-3802.test.ts`, 7 tests against a render stub carrying the cost profile you measured (300 header, 200/title, 1,200/full narrative, 400/session line, 3,000 summary block, 200 footer): a block that already fits is untouched and returns the *same config object*; 50 observations at the shipped defaults overflow and then fit; narratives go before timeline entries; the summary block goes before observations; at least one observation always survives and `overBudget` is set; an unlimited budget reduces nothing; and the result is never longer than the first render.

| mutation | result |
|---|---|
| never reduce (return the first render) | 4 failed |
| drop observations before full narratives | 1 failed |

```
bun test tests/context/context-budget-3802.test.ts     7 pass, 0 fail
bun test tests/context/                                78 pass, 4 fail
tsc --noEmit                                           clean
```

The 4 failures are the pre-existing `context database ownership` cases — same four names on a clean tree (71 pass / 4 fail before this branch, 78 / 4 after, the difference being these 7).

I have not measured the real block against a live Claude Code session; the fitting logic is proven directly, and the 10,000 figure is yours.


stantheman0128:
Support. Item counts cannot keep a hook under 10k. Measuring and dropping whole items is the right shape for the silent stub in #3802.

`--full` staying unlimited is correct.

The fitted text is not what the hook emits. `withObserverHealthWarning` runs after the fitter, so an unhealthy observer can push a 9,998-character block over the limit and Claude Code still swaps in the stub. Fold the warning into the render callback, or reserve space for it.

Marketplace installs still run `plugin/scripts/worker-service.cjs`. That bundle has to ship this, or the unbounded path stays live. Stats should count the fitted prefix, not the pre-trim query.

ntdatt812:
All three findings were real. Fixed, rebased onto `main`, pushed as `b85daa93`.

**1 — warning outside the budget.** Confirmed: `withObserverHealthWarning` ran on
`budget.text`, after fitting. It now renders inside the callback the fitter
measures, and the health is read once rather than per reduction — `render` is
called repeatedly and `readObserverHealth` touches state, so re-reading it would
let the measured length move under the loop.

**2 — stale marketplace bundle.** Confirmed, with one narrowing. I rebuilt
`worker-service.cjs` and `context-generator.cjs`, but deliberately left
`mcp-server.cjs` and `server-service.cjs` out. Rebuilding those two from a clean
tree produces byte-identical output to rebuilding them with this patch applied,
and both already differ from what is committed on `main` — so their diff is
pre-existing drift rather than this change, and including it would have added
two multi-megabyte files of noise to the review.

**3 — telemetry from discarded observations.** Confirmed:
`buildInjectStats(observations, summaries, …)` took the originals. It now takes
`observations.slice(0, budget.observationCount)` and
`summaries.slice(0, budget.config.sessionCount)` — the same slice
`buildContextOutput` uses for `displaySummaries`.

Both code fixes now live in an exported `fitContextForDelivery`, split out of
`generateContextWithStats`. The reason is worth stating: this module had no
in-process test path — the only existing coverage spawns a child process — so
nothing could have caught either defect. Five tests now drive that function
directly.

One correction to my own first attempt at those tests, since it bears on whether
they mean anything: with 50 observations they **passed under the mutation**. The
fitter's reduction steps are coarse, so the fitted block landed far enough below
10,000 that a warning appended afterwards still fit. The fixture is now 47
observations, which render to exactly 9,900 — fits alone, does not fit with a
600-character warning. Moving the warning back outside now fails that test, and
passing the original arrays to `buildInjectStats` fails two of the stats tests.

`bun test tests/context/` is 83 pass / 4 fail; the four are
`context database ownership`, which fail identically on `main` (child-process
stdout JSON parse, unrelated). `typecheck:root`, `lint:hook-io` and
`lint:spawn-env` are clean.

stantheman0128:
Thanks. Putting the warning inside the measured callback, and driving `fitContextForDelivery` with the 9,900 / 600 fixture, pins both holes. Leaving the unrelated bundle drift out of this PR is right.

thedotmack:
ALL-NIGHT SHIP Batch D: rebased onto latest `main` as #3995 (quota-cooldown notice kept inside the measured warning). Will merge #3995 if CI is green.

thedotmack:
Superseded by #3995.

thedotmack:
Superseded by #3995 (rebased onto latest main and squash-merged). Thank you.

PR #3816, opened 2026-08-31
fix(worker): preserve the queued batch when the observer returns empty

cleroglu:
## The problem

`processAgentResponse` treats an empty observer response the same as a plain-text skip: it calls `confirmClaimedMessages` and the batch is gone. But those two things are not the same event.

A prose skip means the model read the batch and decided there was nothing worth recording. An empty response means the provider returned no content at all — `data.choices[0].message.content` was missing, `OpenRouterProvider` logged `Empty response from OpenRouter` and returned `{ content: '' }`. The batch was never looked at, and confirming the claim deletes work that was never observed.

The classifier's own doc comment carries the assumption this PR is challenging: `idle` is described as *"empty / whitespace-only. Benign: the SDK had nothing to say."* For a provider HTTP response, empty is not benign — it's a miss.

## What I measured

Running 13.10.2 against OpenRouter, 25–30 Aug:

| | empty responses | batches lost |
|---|---|---|
| Gemini era (17–24 Aug) | 62 | 0 *(see below)* |
| OpenRouter era (25–30 Aug) | 57 | 34 |

That's ~2.1% of calls, and 42 of 73 drops happened while the queue was non-empty. The rate of empty responses is roughly the same in both eras (0.2%–1% per call) — what changed at the provider switch was the **handling**, not the failure.

The Gemini zero is not a real zero. `forwardEmptyMessageResponse = false` meant the empty response never reached `processAgentResponse` at all; `OpenAICompatibleProvider` logged `Empty ... response, leaving queue intact` and returned. But that branch never calls `resetProcessingToPending`, so the message stayed `claimed`. Following all 297 `leaving queue intact` events forward through the logs: in 282 the hanging message was swept by the next successful turn's `confirmClaimedMessages`, and in 15 it went away with the session. Zero were actually recovered. So the Gemini branch lost the same batches — just with no log line, no counter, and no way to measure it.

## Why not just re-queue

Because `resetProcessingToPending` un-claims immediately and `SessionMessageBuffer.drain` re-yields on its very next loop turn. An unconditional reset on empty is a hot loop against the provider API. The comment in the non-XML branch is right about that risk:

> Plain-text skip responses are intentionally ignored. Re-queueing them creates an observer loop where the same low-signal batch is retried.

So this PR keeps that decision where it applies and bounds the case where it doesn't.

## The change

**`ResponseProcessor.ts`** — split `idle` out of the non-XML branch:

- `idle` → increment `consecutiveInvalidOutputs`, `resetProcessingToPending`, retry. Past `EMPTY_RESPONSE_RETRY_LIMIT` (3) the batch is dropped and the counter clears, so a provider stuck on empty can't spin.
- `prose` → unchanged. Still confirmed and dropped, still resets the counter.

This also wires up `consecutiveInvalidOutputs`, which was reset in three places and incremented in none — the breaker was written but never connected.

**`OpenAICompatibleProvider.ts` / `GeminiProvider.ts` / `OpenRouterProvider.ts`** — removed `forwardEmptyMessageResponse`. Its own doc comment said it exists only to "preserve that per-provider divergence", and the divergence was strictly worse on the Gemini side (silent hanging claim). Both providers now go through the one path that owns this decision.

**Claude/SDK path** is unaffected in practice. `buildHardenedSdkOptions` runs the Observer with `tools: []`, so an assistant message with only `tool_use` blocks can't occur; an empty `textContent` there means the same failure and gets the same treatment.

## Tests

`tests/worker/agents/response-processor.test.ts`: the existing empty-response test asserted the old drop behaviour and is updated. Added coverage for whitespace-only input, the retry ceiling and its counter reset, recovery on a valid XML response, and that a prose skip does not accumulate toward the empty limit.

```
bun test        2249 pass, 15 skip, 0 fail
tsc --noEmit    clean
```

### DIFF
--- a/src/services/worker-types.ts
+++ b/src/services/worker-types.ts
@@ -25,9 +25,11 @@ export interface ActiveSession {
   currentProvider: 'claude' | 'gemini' | 'openrouter' | null;
   consecutiveRestarts: number;
   /**
-   * Legacy invalid-output counter. Ordinary non-XML observer output is now
-   * confirmed as a no-op and resets this to 0 so skip acknowledgements never
-   * accumulate respawn debt.
+   * Consecutive empty observer responses. Each one preserves the queued batch
+   * and retries; past EMPTY_RESPONSE_RETRY_LIMIT the batch is dropped so a
+   * provider stuck on empty cannot loop. Any parseable output — and any prose
+   * skip acknowledgement, which is a deliberate no-op rather than a miss —
+   * resets this to 0.
    */
   consecutiveInvalidOutputs: number;
   forceInit?: boolean;
--- a/src/services/worker/GeminiProvider.ts
+++ b/src/services/worker/GeminiProvider.ts
@@ -220,7 +220,6 @@ interface GeminiConfig {
 export class GeminiProvider extends OpenAICompatibleProvider<GeminiConfig> {
   protected readonly providerName = 'Gemini';
   protected readonly syntheticIdPrefix = 'gemini';
-  protected readonly forwardEmptyMessageResponse = false;
 
   constructor(dbManager: DatabaseManager, sessionManager: SessionManager) {
     super(dbManager, sessionManager);
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -44,13 +44,6 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
   protected abstract readonly providerName: string;
   /** Prefix for the synthetic memorySessionId (e.g. 'gemini', 'openrouter'). */
   protected abstract readonly syntheticIdPrefix: string;
-  /**
-   * When a query returns empty content for an observation/summary message:
-   * OpenRouter still calls processAgentResponse('') (forwards the empty batch
-   * to the parser/recovery path); Gemini skips it and logs a warning. This flag
-   * preserves that per-provider divergence.
-   */
-  protected abstract readonly forwardEmptyMessageResponse: boolean;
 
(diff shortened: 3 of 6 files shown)


cleroglu:
Closing this for now — withdrawing to re-scope on my side. Sorry for the noise.

cleroglu:
Follow-up for anyone who finds this later: please don't cherry-pick it as-is. The PR body contains a claim I have since disproved.

After closing, I measured which providers actually produce the empty responses this patch reacts to, across 38 days of local worker logs (`outputClass=idle` drops, by agent):

| agent | idle drops |
|---|---|
| SDK (Claude) | 17,605 |
| OpenRouter | 43 |
| Gemini | 0 |

So "The Claude path is unaffected in practice" is wrong. I had reasoned from `buildHardenedSdkOptions` running the Observer with `tools: []` that a text-less assistant message cannot occur. That misses the actual mechanism: `textContent` is built by filtering `content` for `type === 'text'` blocks only, so any assistant message carrying only thinking blocks reaches `processAgentResponse` as `''`. On the Claude path that is a routine streaming artifact, not a failed batch. This patch would re-queue on it hundreds of times a day and hammer the API.

The ratio in the same data makes the mechanism visible: 17,605 empty vs 895 valid-XML responses on the SDK path.

The underlying issue is still real, and narrower than I wrote it. An empty HTTP body from an OpenAI-compatible provider is a genuine miss and dropping the batch loses work the model never saw. But the retry branch has to be gated to that case, not to every non-parseable empty string. My local sample for that case is 43 events over 38 days, which is too small to size a retry limit from, so I am measuring first.

I'll open a fresh PR if and when I have the gated version plus data to justify the limit. Apologies for the noise.

PR #3827, opened 2026-08-31
fix(installer): put the provider choice first and collapse the browser trip

thedotmack:
## The story

The Aug 31 OAuth rewrite inverted the funnel. Mandatory browser login was placed **before** the value proposition, so a free, open-source plugin demanded GitHub/Google OAuth before it would finish installing. Having made login mandatory, the code then correctly noticed it couldn't pitch on a mandatory step — so the pitch moved *after* the wall, splitting one offer across three surfaces.

**The audit that drove that rewrite was misread.** It showed:

| Signal | Number |
|---|---|
| Installers who typed their email | **2,222** |
| Resend delivery acceptance | **96.6%** |
| `auth_callback_failed` pageviews | **2,537** (1,907 distinct) |
| Claimers who reached Stripe | 135, no claim→checkout loss |

The failure was the **browser callback**, not email capture and not price resistance. The rewrite deleted the step with 2,222 conversions and promoted the step with 1,907 failures to mandatory for 100% of installs.

Five consecutive fix commits to that rewrite landed in a single day. That's a design that isn't landing, not polish.

## What this does

Restores the old funnel's shape — **cheap, early, terminal-native commitment** — while keeping every security fix from the rewrite. **Email capture is NOT restored** (the forbid-list test still passes).

- **Provider choice runs first, costs nothing, and is the conversion event.** Picking CMEM Pro is the new "type your email": instant, terminal-native, zero context switch.
- **OAuth is lazy.** Only the CMEM Pro branch opens a browser. Anthropic Max installs fully offline, so a cmem.ai outage can't fail it.
- **One browser trip instead of two.** `claim/route.ts` already bounces a sessionless visitor through `/login` and back, so opening `checkout_url` does login → offer → Stripe in one unbroken visit.
- **Provider label trimmed 143 → 57 chars.** Dead `CMEM_TRIAL_ACKNOWLEDGEMENT` deleted.

### Measured effect

| Path | Browser opens | Switches |
|---|---|---|
| CMEM Pro | 2 → **1** | 5 → **4** |
| Anthropic Max | 1 → **0** | → **0** |
| `--provider claude` | 0 | 0 (works with cmem.ai unreachable) |
| `--provider openrouter` | 1 | unchanged |

**On the switch count:** an earlier draft of the plan claimed 5 → 2. That was wrong and is corrected. The device-approval step still sends the user back to the terminal for the code and returns them to the browser. The **delta is real** (one whole round trip, −2), but the absolute is **4**. Reaching 2 needs the deferred Phase 4 (device-code narrowing), which requires a DB migration and a security review.

## Two defects found while verifying this change

1. **Poll budget collapse.** `OAUTH_POLL_BUDGET_MS` was a single 240s window. It used to be spent *twice* — once per browser trip — so folding the trips together silently folded the budget together, leaving four minutes to cover OAuth + reading the offer + Stripe card entry + device approval, then exiting non-zero. Now phase-aware: 240s for login-only, 25 min for enrollment, bounded by the server's 30-min `PAIRING_TTL_SECONDS`.
2. **Misleading failure.** A CMEM Pro path failing at checkout or approval printed "OAuth login is required to finish installation", sending users and support after the wrong thing.

## Back-compat

`hasPairingAndTrialParams` and `parseInstallerOAuthStartBody` are **byte-identical to HEAD** (verified by diffing extracted function bodies). No URL shape changed. Installers already on npm keep working.

⚠️ Requires [claude-mem-pro#151](https://github.com/thedotmack/claude-mem-pro/pull/151) to deploy **first**.

## Testing

`npm run typecheck` clean

```
bun test tests/npx-cli/ tests/install-non-tty.test.ts
 100 pass / 0 fail / 264 expect()

npm test
 2866 pass / 27 skip / 2 fail
```

The 2 failures are pre-existing `worktree-adoption` tests, identical to baseline before this branch.

Anti-pattern greps all clean: no `promptBrowserLogin`, `CMEM_PRO_TRIAL_START_URL`, `CLAUDE_MEM_ONLINE_OPTIN`, `Your email:`, or `CMEM_TRIAL_ACKNOWLEDGEMENT` anywhere in `src/`.

## Contract flips (deliberate, not incidental)

The test literally named `'requires OAuth before provider selection'` asserted the inversion this PR fixes. It and the `provider-account-gate` adjacency regex were rewritten to pin the new contract. The forbid-list and `p.multiselect<ProviderChoice>` assertions are preserved byte-for-byte.

## Not verified

- **The back-compat gate was not run** — a published 13.21.2 from npm has not been exercised against the updated server. Static evidence is strong (frozen validators byte-identical, clean state-machine walk) but the gate is formally open.
- **stdin sequencing** — `waitForReturnToOpenBrowser` now runs after a clack multiselect. A pty repro passed, but the same harness also passed the ordering a code comment says previously stalled, so it has no demonstrated sensitivity. Structural argument says low risk. Only a live interactive run settles it.
- Manual walkthroughs (switch counts, `/etc/hosts`-blocked offline run).

## Known trade

Lazy login costs `pro_users` its `registered` rows — and more broadly than first stated: that row is inserted only on the `login_only` branch, which new installers never take. A user who signs in and abandons at the offer or at Stripe now leaves no row at all. Top-of-funnel visibility should move to anonymous installer telemetry (`captureCliEvent`, `installId`), which already exists and needs no account. Measurement should ride on telemetry, not on a login wall.

Plan: `plans/2026-08-31-funnel-immediacy-restore.md`

https://claude.ai/code/session_01DD5YMaAiXXeY3zaitbGC6n

### DIFF
--- a/plans/2026-08-31-funnel-immediacy-restore.md
+++ b/plans/2026-08-31-funnel-immediacy-restore.md
@@ -1,0 +1,216 @@
+# Restore Funnel Immediacy: Provider Choice First, Lazy Login, One Browser Trip
+
+**Date:** 2026-08-31
+**Repos:** `claude-mem` CLI (this worktree, branch `fix/installer-preselect-defaults`) + `claude-mem-pro` server (`~/Scripts/claude-mem-pro`, main @ `ea8a4d2`)
+
+---
+
+## Problem
+
+The Aug 31 OAuth rewrite inverted the funnel. Mandatory browser login was placed **before** the value proposition, so:
+
+- A free, open-source plugin demands GitHub/Google OAuth before it will finish installing.
+- The pitch had to move *after* the login wall, splitting one offer across three surfaces.
+- Users take **5 terminal↔browser context switches**; the old email funnel took 3.
+
+The audit that triggered the rewrite was misread: the terminal email step converted 2,222 installers (96.6% delivery), while the **browser callback** failed 1,907 times (`auth_callback_failed`). The converting step was deleted; the failing step was promoted to mandatory.
+
+**Root cause of the inversion** — from `claim/route.ts`'s own comment: they wanted a `pro_users` row for every install ("registered-but-not-paid is a real state"). Mandatory login was a *measurement* decision wearing an *auth* decision's clothes.
+
+## Goal
+
+Restore the old funnel's shape — **cheap, early, terminal-native commitment** — while keeping every security fix from the rewrite.
+
+Picking "CMEM Pro" in the terminal becomes the new "type your email": instant, free, zero context switch, and it *is* the conversion event.
+
+**Target: 5 context switches → 4. Forced switches for a user who never pays: 1 → 0.**
+
+> **Corrected after verification.** An earlier draft of this plan claimed 5 → 2. That was wrong.
+> The device-approval step still sends the user back to the terminal to read the code and then
+> returns them to the browser, which costs two switches that Phases 1–3 do not remove. The
+> **delta is real** — one entire browser round trip is deleted, −2 switches — but the absolute
+> figure is 4. Reaching 2 requires Phase 4 (device-code narrowing), which is deferred.
+
+---
+
+## Phase 0: Discovery Findings (COMPLETE — do not re-investigate)
+
+### Allowed APIs / verified facts
+
+**CLI (`src/npx-cli/commands/install.ts`)**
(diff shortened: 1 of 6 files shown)


thedotmack:
Routing to **#2785 (plan-12, Provider & Extensibility Roadmap)** — install funnel, trial and Pro billing sit on that master rather than in any of the twelve defect clusters.

Leaving open. It overlaps #3687 in `install.ts`, `cmem-pro-costs.ts` and `pro-promo`, but they're different changes (that one adds the 30-day trial and a fallback provider; this one reorders the funnel and collapses the browser trip), so neither supersedes the other — they just can't both land unrebased. Currently `CONFLICTING`.

PR #3853, opened 2026-09-03
feat(copilot-cli): capture lifecycle hooks instead of MCP-only

ARosentiehl24:
## Summary

Copilot CLI was an MCP-only IDE: search tools worked, but nothing was captured. The host already has user-level hooks (`~/.copilot/hooks/*.json`). This wires them into the unified `hook copilot <event>` pipeline so tool use becomes observations, not just stored prompts.

## What

- `copilot` / `copilot-cli` adapter: camelCase Copilot payloads plus snake_case aliases; unwraps `toolResult.textResultForLlm` (the documented `postToolUse` shape).
- Installer writes `~/.copilot/hooks/claude-mem.json` with `exec`/`args` (no shell), `cwd` = home, and dual-writes MCP to `~/.copilot/mcp-config.json` and `~/.github/copilot/mcp.json`.
- `npx claude-mem install --ide copilot-cli` is hooks + MCP (no longer MCP-only). Uninstall removes only `claude-mem.json`.
- Claude CLI version probe runs with `cwd=$HOME` so a deleted worker cwd does not look like “Claude executable not found”.
- Docs: `docs/public/copilot-cli/setup.mdx` + nav + installation mention.

Events: `sessionStart` → context, `userPromptSubmitted` → session-init, `postToolUse` / `postToolUseFailure` → observation, `sessionEnd` → summarize.

## Test plan

- [x] `npm run typecheck`
- [x] `bun test tests/copilot-cli-compat.test.ts` (15 pass)
- [x] Live Copilot CLI session: prompts stored, tools enqueued, Haiku XML observations synced to Chroma after worker restart + `CLAUDE_CODE_PATH` to the real binary

## Notes

- Copilot CLI is the **host**, not the observer billing provider.
- `community-edge` is still on 13.10.x; this targets `main` (13.23) like other current PRs.


stantheman0128:
Support. Copilot CLI as a host needs the same hook pipeline as Claude Code, not MCP-only capture.

`exec`/`args` with no shell is the installer shape we want. `cwd=$HOME` on the Claude probe, so a deleted worker cwd does not look like a missing binary, is the right fail-loud.

On Windows I would still want a test that the hook argv is a real path (not a login-shell PATH rebuild) and that a missing Copilot binary fails visible. Happy to leave that as a follow-up if the live Copilot CLI session already covers it.

thedotmack:
NIGHT SHIP Batch F — Gate **SKIP** (feature, not a mergeable bugfix). Leaving open; not merging in this batch.

thedotmack:
Routing to **#3611 (plan-23, Host Integration Contracts)** — Copilot CLI is named on that master, and this is precisely its shape: an adapter that speaks the host's real payload (`toolResult.textResultForLlm`, camelCase with snake_case aliases) plus an installer that writes the hooks file the host actually loads, with a compat test to hold it.

Staying open. It's `MERGEABLE` and carries `tests/copilot-cli-compat.test.ts`, which is the contract-test discipline plan-23 asks for; the Windows workflow change is a reasonable rider.

PR #3868, opened 2026-09-03
fix(openrouter): bound the observer context and anchor the schema in a system message (#3606)

matthewdnye:
## Summary

On the OpenRouter / OpenAI-compatible path the observer sends the **entire per-session conversation history on every call, with no `system` role**. Against a bounded context (a local Ollama/llama.cpp model here, 16,384 tokens) that produces exactly the discard pattern tracked in #3606, in two flavours:

- the prompt fills the window → llama.cpp truncates (`truncated=1`, `n_tokens=16383`) → the reply is empty → `outputClass=idle` → `confirmClaimedMessages` drops the batch. In our logs **99.1 % of idle discards (663/669) sat at exactly the context ceiling**.
- the history grows (median 235 turns) → the `<observation>` schema, stated once in the init prompt, scrolls out of view → the model emits bare `<observation>free text</observation>` → `parseObservationBlocks` finds no fields → `outputClass=xml`, dropped. **94 % of xml-class discards (492/522) were schema-less prose**, strongly correlated with history length; the remaining 6 % were cut mid-tag at exactly the ceiling.

Net effect on one install: **818 successful HTTP calls → 26 observations written over 15 hours (~96 % discarded)**, with the write-only `consecutiveInvalidOutputs` counter reading 0 throughout.

## What this changes

- **`context-window.ts` (new):** `history[0]` (the init/continuation prompt, which carries the full schema) is pinned as the `system` message; the rest is windowed by turn count and total chars, the final message is always kept, and leading `assistant` turns are dropped after trimming so strict gateways still see a user-first conversation.
- **Four settings declared and honoured** on this path (they previously did nothing because `SettingsDefaultsManager.loadFromFile` only copies keys present in `DEFAULTS`): `CLAUDE_MEM_OPENROUTER_MAX_TOKENS` (was hard-coded `4096`, see #3490), `_MAX_CONTEXT_MESSAGES` (default 40), `_MAX_CONTEXT_CHARS` (default 200000; `0` = unbounded for both caps), `_ATTEMPT_TIMEOUT_MS` (was the hard-coded 30 s `withRetry` default). Corrupt values fall back to the defaults; `max_tokens` and the timeout require ≥ 1.
- **`finish_reason` is read and propagated.** A designed empty completion is exactly: message present, empty content, `finish_reason === 'stop'` (info). `'length'` is a real truncation (error). Any other reason, an absent one, or a body with no `choices`/`message` (`malformed`) is not a designed skip and is logged as such.
- **`consecutiveInvalidOutputs` becomes a real counter** in the shared fall-through branch of `ResponseProcessor`: incremented on any real dropped batch, reset on a valid parse or a designed skip (`idle` with `finish_reason` `null` or `'stop'`; `null` covers Claude/Gemini, which never report one, so their behaviour is unchanged). Dropped batches log at `error` when truncated or xml-classified, `warn` otherwise. The newer overflow/quota/auth branches above it (#3800) are untouched. No retry is introduced: the drop-on-invalid design stays; the root cause is fixed upstream of it.
- Bun's `fetch` has a built-in ~300 s cap that overrides any `AbortSignal` per-attempt timeout; the request init now sets `timeout: false` so the configured timeout is authoritative.
- The duplicate assistant-history push this fix also targeted was already removed at #3619; `tests/worker/openai-compatible-history.test.ts` is added as a regression guard.

## Test evidence

- `tsc --noEmit` clean. `bun test tests/worker tests/sdk tests/shared`: 882 pass, 0 fail (+44 new). Full `bun test`: identical failing-test set and tallies to a clean `origin/main` checkout at the same commit (`be44b6c8`), verified side by side.
- The same logic was built into a worker bundle (on a v13.9.3-based branch) and driven end to end against a mock OpenAI-compatible server on a scratch port: every request system-first, ≤ cap messages, char cap with the final-message rule, alternating roles, observations stored; the `length` / `stop`-empty / `content_filter` / `malformed` paths logged and counted as designed.
- Live, after install on the affected machine (`MAX_CONTEXT_MESSAGES=12`, `MAX_CONTEXT_CHARS=28000`, `MAX_TOKENS=1500` for a 16k lane): prompt size median 7,463 tokens (was 12,800, p90 15,400), zero `truncated=1`, **97 observations + 2 summaries from 99 successful calls in the first 46 minutes, 2 batches dropped** (one output at the token cap, one prose reply).

Refs #3606 (umbrella), #3490 (`max_tokens`), #3624 (the preserve-and-retry approach; this PR removes the cause instead), #3619.

---
Opened by Gwen Ives, Virtual COO (an AI assistant operating on behalf of Matt Nye), from the install where the defect was measured.

https://claude.ai/code/session_01MwVTb8AbAAQZjxsrewXc83

### DIFF
--- a/src/services/worker-types.ts
+++ b/src/services/worker-types.ts
@@ -25,14 +25,16 @@ export interface ActiveSession {
   currentProvider: 'claude' | 'gemini' | 'openrouter' | null;
   consecutiveRestarts: number;
   /**
-   * Legacy invalid-output counter, intentionally always 0: ordinary non-XML
-   * observer output is confirmed as a no-op and resets this so benign skip
-   * acknowledgements never accumulate respawn debt.
+   * Real invalid-output counter (#3606): incremented on a dropped batch
+   * (non-XML output that is not a designed empty skip), reset to 0 on a
+   * valid parse or a designed empty skip (idle output with a non-'length'
+   * finish reason — the prompt allows an empty reply to signal "nothing to
+   * report").
    *
    * It is deliberately NOT the breaker for repeated hard rejections — counting
-   * skips and rejections on one counter is what produced the respawn storm this
-   * reset was added to stop. Hard rejections are counted by
-   * `consecutiveContextOverflows` instead.
+   * skips and rejections on one counter is what produced the respawn storm the
+   * write-only reset this replaces was added to stop. Hard rejections are
+   * counted by `consecutiveContextOverflows` instead.
    */
   consecutiveInvalidOutputs: number;
   /**
@@ -82,6 +84,8 @@ export interface ActiveSession {
   pendingCompressionEvent?: Record<string, unknown> | null;
   /** Cumulative total_cost_usd from the SDK's latest result message — per-compression cost is the delta between results. */
   lastResultTotalCostUsd?: number | null;
+  /** finish_reason from the latest OpenAI-compatible provider response — 'length' flags a truncated (not just empty/prose) observer output (#3606). */
+  lastFinishReason?: string | null;
 }
 
 export interface PendingMessage {
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -38,6 +38,8 @@ export interface ProviderQueryResult {
   costUsd?: number;
   /** The model that actually served the request, when reported. */
   servedModel?: string;
+  /** finish_reason from the provider response, when reported — 'length' flags a truncated (not just empty/prose) output (#3606). */
+  finishReason?: string;
 }
 
 /**
@@ -216,6 +218,7 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
       session.cumulativeInputTokens += Math.floor(tokensUsed * 0.7);
       session.cumulativeOutputTokens += Math.floor(tokensUsed * 0.3);
       session.lastUsage = this.buildLastUsage(initResponse);
+      session.lastFinishReason = initResponse.finishReason ?? null;
       await processAgentResponse(
         initResponse.content, session, this.dbManager, this.sessionManager,
         worker, tokensUsed, null, this.providerName, undefined, initResponse.servedModel ?? model, responseContext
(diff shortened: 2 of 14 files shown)


matthewdnye:
Thanks for the two reproductions — both were real, both are fixed in the follow-up commit on this branch:

- **Malformed integers bypass fallback** (`parseOpenRouterInt`): `parseInt` is gone. A setting is accepted only as a finite integral `number` or a complete decimal-integer string (optional sign, surrounding whitespace trimmed) within `[min, max]`; `"1e3"`, `"42.5"`, `"0x10"`, `""`, non-ASCII digits, `NaN`/`Infinity`, objects all fall back to the documented default. `attemptTimeoutMs` additionally caps at the `setTimeout` ceiling (2^31-1 ms), since Node silently fires anything larger after ~1 ms — the same failure mode from the other end. Covered in `tests/worker/openrouter-config-source.test.ts`.
- **Stale finish reason misclassifies Claude replies**: `processAgentResponse` now consumes and clears `session.lastFinishReason` at the top of every call, before any branching, so the value is scoped to the single response that produced it and a later Claude-path reply can never observe a leftover `length`. The explanatory comment above the designed-skip logic is updated to match. Covered in `tests/worker/agents/response-processor.test.ts` (OpenRouter `length` → empty Claude reply = designed skip, counter reset; an OpenRouter `length` on its own is still a truncation).

`tsc --noEmit` clean; the two touched test files 64/64; the full suite's only failures are the seven pre-existing sqlite-migration/build-artifact/worktree cases that don't touch these files.

thedotmack:
ALL-NIGHT SHIP Batch D: **SKIPPED** — CONFLICTING without a clean rebase.

Onto current `main` this collides in `OpenAICompatibleProvider.ts` (init-response / recycle path from #3800) and in eight hunks of `OpenRouterProvider.ts` (`withRetry`, request shaping, finish-reason). `main` also already has generation recycle via `CLAUDE_MEM_OBSERVER_MAX_CONVERSATION_CHARS` — a second send-time window needs a deliberate merge, not a mechanical rebase. Leaving this for the author to rebase against the recycle + overflow work.

thedotmack:
Routing to **#3606 (plan-18, Observer Response Pipeline)** — which the body already names. Unbounded `conversationHistory`, a schema that scrolls out of the window, an empty reply classified as `idle`, and `confirmClaimedMessages` dropping the batch anyway are four of that master's bullet points in one trace, and the 663/669-at-the-ceiling number is the best evidence anyone has attached to it.

Not closing, but flagging the overlap: **#3151** (history pruning) and **#3516** (window-aware compaction) attack the same defect from different angles, and #3516 touches the same `context-window.ts`. plan-18 should pick one of the three rather than merge two — this one is my pick, because it also fixes the schema anchoring that the other two leave alone. Needs a rebase; it's `CONFLICTING`.

Issue #3604, opened 2026-08-16
[plan-16] Canonical Install Identity & Bundle Integrity — one resolvable, dependency-complete plugin root, and recycles that cannot kill a working worker


thedotmack:
# [plan-16] Canonical Install Identity & Bundle Integrity — one resolvable, dependency-complete plugin root, and recycles that cannot kill a working worker

> **Tracker:** #3604 · **Design doc:** `plans/16-canonical-install-identity-and-bundle-integrity.md` · **Series:** plan-14 #3602 · plan-15 #3603 · plan-16 #3604 · plan-17 #3605 · plan-18 #3606 · plan-19 #3607 · plan-20 #3608 · plan-21 #3609 · plan-22 #3610 · plan-23 #3611

## Defect

There is no single answer to "which copy of claude-mem is installed and runnable". Hooks, the supervisor's successor spawn, `npx claude-mem` (install/repair/doctor/status), the SDK subprocess, and the marketplace auto-updater each resolve the plugin root independently, and none of them verifies the copy has its runtime dependencies:

- The hook prelude picks the newest **mtime** under `~/.claude/plugins/cache/thedotmack/claude-mem/[0-9]*/` (`ls -dt … | head -1`), accepts 1-byte gutted stubs and `.orphaned_at` dirs, and falls back to `~/.claude/plugins/marketplaces/thedotmack/plugin/` — a checkout that never has `node_modules`. `resolveWorkerScriptPath()`/`resolveSuccessorScript()` repeat the same mtime sort in code (fixed on the hook side by #3371 in 13.12.x; still present in successor resolution per #3216/#3565).
- `worker-service.cjs` externalizes `zod`, `shell-quote`, `tree-sitter-*`, `ajv*`, `better-auth`, and (13.9.x) `../sqlite/{SessionStore,observations/files}.js`; the bundle requires modules the build never emits and a `node_modules` the plugin update / marketplace pull never installs (`Cannot find module 'zod/v3'`, `Cannot find module '../sqlite/observations/files.js'`). `tree-sitter-cli`'s binary is fetched by an install script suppressed by `--ignore-scripts`; the Windows resolver checks an extensionless path; grammars declare conflicting `tree-sitter` peer majors (`ERESOLVE`); npm 12 rejects the `--allow-scripts` CLI flag the installer passes.
[...]

Issue #2730, opened 2026-05-31
v13.4.0: worker/Stop hook still fails 'Cannot find module zod/v3' — zod absent from installed node_modules (Linux); recurs on every auto-update


DarkTLord:
### Summary

`zod/v3` is **still missing from the installed plugin `node_modules`** on a fresh **v13.4.0** install on Linux, so the worker path crashes with `Cannot find module 'zod/v3'`. This is the same class as #2387 / #2407 / #2437 / #2453 / #2640 / #2456 (consolidated into #2379 / tracking #2601) — reporting here because it **reproduces on 13.4.0**, i.e. the packaging fix did not fully land, and the failure **recurs after every auto-update** (matches #2456).

Observed via the **Stop hook**:

```
Stop hook error: Failed with non-blocking status code: error: Cannot find module 'zod/v3'
from '/home/<user>/.claude/plugins/cache/thedotmack/claude-mem/13.3.0/scripts/worker-service.cjs'
Bun v1.3.11 (Linux x64)
```

### Environment

- **OS**: Linux x86_64 (CachyOS, kernel 7.0.10)
- **Node**: v22.22.2
- **Bun**: 1.3.11
- **Host**: Claude Code CLI (Opus)
- **Plugin versions present in cache**: `12.1.0`, `13.3.0`, `13.4.0` (active/newest = 13.4.0)

### Reproduction / diagnosis

In **every** cached version dir, `plugin/package.json` correctly declares `zod: ^4.3.6` as a runtime dependency, and the built `scripts/worker-service.cjs` requires `zod/v3` as an **external** (11 sites). But the installed `node_modules` contains the 24 tree-sitter grammars + `tree-sitter-cli` + `shell-quote` — **`zod` is absent**:

```bash
$ ls ~/.claude/plugins/cache/thedotmack/claude-mem/13.4.0/node_modules/ | grep -x zod
# (no output — zod missing)

$ grep -c "zod/v3" ~/.claude/plugins/cache/thedotmack/claude-mem/13.4.0/scripts/worker-service.cjs
11

$ node -e "console.log(require('.../13.4.0/plugin/package.json').dependencies.zod)"
^4.3.6
```

So `installPluginDependencies` / the marketplace `bun install` materialized every declared dep **except** `zod`. `mcp-search` (the MCP server, which **bundles** zod per the `build-hooks.js` guardrail) keeps working; only the **worker** (which externalizes zod by design, see #2584 minimization) breaks.

[...]

PR #3872, opened 2026-09-04
fix(setup): guard plugin deps on completeness, not node_modules existence

### LINKED ISSUE #3755: Setup dependency guard checks that node_modules exists, not that it is complete — a partial install wedges every hook

thedotmack:
Fixes #3755. Refs #3604 (plan-16), #2730. Linear ALE-315.

## The bug

`ensurePluginDependencies()` in `plugin/scripts/version-check.js` decided whether to run `bun install` by asking whether `node_modules/` existed:

```js
```

A tree that is merely *present* — but short of the declared closure — satisfies that check and permanently skips repair on every subsequent Setup run.

## Two trigger paths

The second one is the reason this keeps getting refiled, and it needs **no corruption at all**:

1. An install interrupted mid-fetch (network timeout, OOM kill, registry 5xx) leaves `node_modules/` behind incomplete.
2. A tree that was complete *for the version that created it*. `zod` was added to plugin deps after some users had already installed. Their `node_modules` has been incomplete ever since, and no upgrade heals it — the stale tree is gitignored, so `git pull` never touches it, and each new cache version is seeded by copying it.

## Why it was silent

`mcp-server.cjs` bundles zod internally — `scripts/build-hooks.js:519` hard-fails the build if it ever externalizes it — while `worker-service.cjs` has 19 external zod requires, 4 of them `zod/v3`. So search and timeline keep answering against historical data while capture is dead. The crash happens in freshly spawned hook processes whose stderr never reaches the log dir.

One reporter lost ~4 months of capture with no visible symptom beyond two red `SessionStart` lines that read like cosmetic hook noise. The transcripts needed to backfill that window had already rotated away.

## The fix

Guard on completeness. Every key of `package.json` `dependencies` must resolve, with a `<dep>/package.json` fallback for bin-only packages like `tree-sitter-cli` (#2730), plus the zod subpaths the worker actually requires.

This is **not a new contract** — it mirrors `verifyCriticalModules` (`src/npx-cli/install/setup-runtime.ts:245`), which already applies exactly this check on the npx install path but was never reachable from the Setup hook. It can't be imported here (this script is standalone and dependency-free, run by whatever Node the host provides), so the probe is inlined and the two are cross-referenced in comments.

Two consequences fall out of the guard change:

- **The post-failure `rmSync` of `node_modules` is removed.** It existed only because the existence guard would otherwise block retry forever (#2650 review). The completeness guard re-detects the gap on the next run, so deleting bought nothing — while actively destroying a partial tree that still powers search. Retry cadence is identical before and after; this only removes a destructive side effect.
- **A zero exit from `bun install` is no longer trusted.** It can exit 0 with a failed integrity check (reported on #3755), so the closure is re-probed afterwards and the diagnostic reports what actually resolves.

The install diagnostic now names the unresolvable modules, which is what stops this failure mode from being silent.

## Verification

The exact `#3755` fixture — `node_modules/` present, `zod` absent — before and after:

```
# before: exit 0, no output, bun install never invoked
# after:
[version-check] installing plugin dependencies (missing: zod, zod/v3, zod/v4, zod/v4-mini)...
[version-check] bun install exited 0 but dependencies are still missing: zod, ...
```

That second line is the fake installer exiting 0 without installing — the integrity-failure case, caught by the re-probe.

**Tests are proven non-vacuous.** Restoring the old existence guard makes 5 of the 13 fail:

- `installs when node_modules exists but is empty (gh #3755)`
- `preserves a partial node_modules after a failed install and retries on the next run`
- `detects and names a missing scoped dependency`
- `detects an unresolvable zod subpath even when zod itself resolves (gh #3755)`
- `detects and repairs a dependency added to the manifest after a complete install (gh #3755)`

Also covered: bin-only packages treated as present (#2730 regression guard), scoped deps (6 of the 26 real deps are scoped), and complete trees skipping the install.

`tests/plugin-version-check.test.ts` is deliberately **untouched** and still green — its fixtures declare no `dependencies`, which proves the zero-declared-deps early return works.

Local: typecheck clean, `npm run build` clean, `bun install --frozen-lockfile` in `plugin/` reports no lockfile drift. Full suite has 21 pre-existing failures on this branch point (Windows-path `HealthMonitor`/`ProcessManager` suites and one worktree-sync test) — byte-identical set before and after this change, verified by diffing failure lists against a pristine checkout.

## Deliberately out of scope

Two related things found while investigating, flagged rather than bundled into a hot fix:

1. **`isInstallCurrent()` (`setup-runtime.ts:497`) has the same existence-check shape.** It's mitigated — it also requires a matching `.install-version` marker, and `npx claude-mem repair` installs unconditionally — so the escape hatch works. Worth tightening separately.
2. **Install-args asymmetry.** The npx installer uses `bun install --frozen-lockfile --ignore-scripts`; the Setup hook uses `['install', '--production']`. `plugin/bun.lock` ships in the tarball and is mirrored to the marketplace, so the hook could use it. Changing install semantics under a 120s `spawnSync` budget is its own risk and deserves its own PR.

One note for future maintainers, recorded as a comment in the source: the post-install re-probe assumes a resolver that doesn't negatively cache. Node is that resolver and `plugin/hooks/hooks.json:11` invokes this script under node explicitly. Bun caches negative CJS lookups process-wide, so running this file under bun would make the re-check cry wolf on every successful fresh install.

https://claude.ai/code/session_01NT5K64VU4a7Kbc36oTVjyc

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
   "plugins": [
     {
       "name": "claude-mem",
-      "version": "13.24.0",
+      "version": "13.24.1",
       "source": "./plugin",
       "description": "Persistent memory system for Claude Code - context compression across sessions"
     },
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-mem",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "description": "Memory compression system for Claude Code - persist context across sessions",
   "author": {
     "name": "Alex Newman"
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-mem",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "description": "Memory compression system for Claude Code - persist context across sessions",
   "author": {
     "name": "Alex Newman",
--- a/.grok-plugin/plugin.json
+++ b/.grok-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-mem",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "description": "Persistent memory for coding agents. Cursor and Grok Bot install independently; worker and observer can be local or remote.",
   "homepage": "https://github.com/thedotmack/claude-mem",
   "skills": "./claude-mem-grok-bot/skills/",
--- a/claude-mem-cursor/.cursor-plugin/plugin.json
+++ b/claude-mem-cursor/.cursor-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-mem-cursor",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "description": "Persistent memory for Cursor. Hooks ingest tool use; MCP searches past sessions. Works with a local worker or a remote cmem.ai worker, and a local host-login observer or remote inference.",
   "author": { "name": "Alex Newman" },
   "homepage": "https://github.com/thedotmack/claude-mem",
--- a/claude-mem-grok-bot/.cursor-plugin/plugin.json
+++ b/claude-mem-grok-bot/.cursor-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-mem-grok-bot",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "description": "Persistent memory for Grok Bot. No host hooks: transcript watcher plus MCP. Local worker with host-login observer, or remote cmem.ai. Install without Cursor.",
   "author": { "name": "Alex Newman" },
   "homepage": "https://github.com/thedotmack/claude-mem",
--- a/openclaw/openclaw.plugin.json
+++ b/openclaw/openclaw.plugin.json
@@ -3,7 +3,7 @@
   "name": "Claude-Mem (Persistent Memory)",
   "description": "OpenClaw plugin for Claude-Mem. Records observations from embedded runner sessions and streams them to messaging channels.",
   "kind": "memory",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "license": "Apache-2.0",
   "author": "thedotmack",
   "homepage": "https://claude-mem.ai",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-mem",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "description": "Memory compression system for Claude Code - persist context across sessions",
   "keywords": [
     "claude",
--- a/plugin/.claude-plugin/plugin.json
+++ b/plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-mem",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "description": "Memory compression system for Claude Code - persist context across sessions",
   "author": {
     "name": "Alex Newman"
--- a/plugin/.codex-plugin/plugin.json
+++ b/plugin/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-mem",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "description": "Memory compression system for Claude Code - persist context across sessions",
   "author": {
     "name": "Alex Newman",
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-mem-plugin",
-  "version": "13.24.0",
+  "version": "13.24.1",
   "private": true,
   "description": "Runtime dependencies for claude-mem bundled hooks",
   "type": "module",
--- a/plugin/scripts/version-check.js
+++ b/plugin/scripts/version-check.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 import { spawnSync } from 'child_process';
-import { existsSync, readFileSync, rmSync } from 'fs';
+import { existsSync, readFileSync } from 'fs';
+import { createRequire } from 'module';
 import { homedir } from 'os';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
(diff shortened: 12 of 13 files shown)


thedotmack:
Confirmed and fixed in 9c5d26fd. This finding was correct and it was serious — thank you.

I reproduced the mechanism directly:

```js
// plugin/node_modules is EMPTY; a valid package sits one directory up
createRequire(join(nodeModulesPath, 'noop.js'))
  .resolve('leakypkg', { paths: [nodeModulesPath] })
// -> /…/hosttest/node_modules/leakypkg/index.js   ← outside the plugin tree
```

`paths` only seeds Node's lookup; resolution still walks every ancestor directory and always consults the global folders. Then I ran the **committed** probe against a plugin whose `node_modules` was completely empty with a valid `zod` one level up: exit 0, no output, `bun install` never invoked. That is the #3755 bug reintroduced by the fix for it — and because plugin roots live at `~/.claude/plugins/cache/thedotmack/claude-mem/<version>/`, any `zod` up to `~/node_modules` or installed globally would have triggered it.

**The fix**, in two parts:

1. **Presence** is now a direct stat of `<node_modules>/<dep>/package.json`, which cannot escape the tree. This is the same signal `scripts/check-postinstall-allowlist.js:75-78` already uses, and it removes the bare-name/`/package.json`-fallback dance entirely — scoped names split into path segments, and bin-only packages like `tree-sitter-cli` (package.json present, no entry point, #2730) are handled natively.

2. **zod subpaths** still need real resolution, since `zod/v3` is an `exports` entry that a present directory doesn't guarantee. Those are accepted only when the resolved file lands inside the plugin's own zod directory.

Two details worth flagging on that containment check, since they're easy to get wrong:

- Both sides are `realpath`'d before comparison. Bun can materialise `node_modules` entries as links into a shared store, and Node returns the real path of what it resolved — comparing literally would report a *healthy* linked install as missing and loop the install forever.
- Containment uses `path.relative` rather than string prefixing, so a sibling like `node_modules/zod-extra` isn't mistaken for being inside `node_modules/zod`.

A regression test is added (`does not accept a dependency resolved from OUTSIDE the plugin tree`) using exactly the fixture shape that fooled the old probe — empty plugin `node_modules`, complete zod in the parent. Verified non-vacuous: it fails against the previous commit's probe and passes now. Suite is 14/14.

One thing your finding surfaced beyond this PR: `verifyCriticalModules` (`src/npx-cli/install/setup-runtime.ts:245`) has the same ancestor/global escape. It's much less dangerous there — it runs as a post-install assertion that fails loud, rather than as the gate deciding whether repair happens at all — so I've left it out of this hot fix and noted it in the source for follow-up.

thedotmack:
@greptile review

The P1 (dependency resolution escaping the plugin tree) is fixed in 9c5d26fd — presence is now a direct stat inside `<pluginRoot>/node_modules`, and zod's subpath resolution is accepted only when the resolved path is contained in the plugin's own zod directory (both sides realpath'd). Regression test added and verified to fail against the previous commit. Please re-review the new head.

thedotmack:
Added `1774be00` — **version bump to 13.24.1**, so merging this PR yields a releasable patch in one step.

The bump touches **11** manifests. Worth flagging: the `version-bump` skill documents only **eight**. Three have been added since it was written and were missing from its list:

- `.grok-plugin/plugin.json`
- `claude-mem-cursor/.cursor-plugin/plugin.json`
- `claude-mem-grok-bot/.cursor-plugin/plugin.json`

The commit shape matches the repo's own bump commits (`1df66c2c`): version lines only, no build artifacts. Two deliberate choices behind that:

1. **No regenerated bundles.** `npm run build` rewrites `worker-service.cjs`, `mcp-server.cjs`, `server-service.cjs`, `transcript-watcher.cjs` and `SessionStore.js`, but the diff is almost entirely esbuild minifier identifier churn (`s`↔`t` renames) with no semantic change. The committed `worker-service.cjs` carries no version string at all, and CI rebuilds before testing, so the committed bundles aren't the source of truth. Including ~500 lines of churn would only obscure the diff.

2. **`CHANGELOG.md` untouched.** `scripts/generate-changelog.js` rebuilds it from published GitHub releases, so it's regenerated after the release is cut — not hand-edited here.

One thing to be aware of when you next run a build: `scripts/sync-plugin-manifests.js` overwrites the hand-written Cursor and Grok-Bot descriptions with the generic one from `package.json`. I preserved the hand-written copy from the recent Grok Mem commits and bumped only the version line in those two files, but the generator and the committed copy are out of sync — the next `npm run build` will clobber them again. Probably worth either teaching the sync script to leave `description` alone for those two manifests, or moving that copy into the generator.

I did **not** create the tag or GitHub release: both would point at a commit that isn't on `main` yet. Those belong after merge, along with the npm publish.

thedotmack:
## CI status — read before merging

**6 of 7 checks pass.** The seventh, `typecheck · build · test · bundle-size`, fails on 19 `HealthMonitor` / `ProcessManager` tests. **This is pre-existing and not caused by this PR.** Verified three independent ways:

1. A pristine baseline at this branch point (`be44b6c8`, with all three commits reverted) produces a **byte-identical** failing set.
2. Main's own CI run at `be44b6c8` ([run 33773325757](https://github.com/thedotmack/claude-mem/actions/runs/33773325757)) is red on this same job.
3. The failing list contains **zero** version-check tests. This PR's 14 all pass — the run reports 3009 tests / 2963 pass, one more than before, which is the new leak regression test.

### Root cause of the flake

Run the two suites alone and they are **79 pass, 0 fail**. Run them in the full suite and 19 fail. It's test pollution, and `tests/preload.ts` documents the mechanism in its own header comment:

> `src/shared/paths.ts` freezes `DATA_DIR` at first evaluation … module-level consts like `ProcessManager`'s `PID_FILE` inherit that frozen value

Bun runs every test file in one process, so a `mock.module` registered by an earlier file stays installed for these suites. Identifying the exact polluting file is a separate task and out of scope for a hot fix.

The job passed once in four rerun attempts (the passing run took 10m16s; every failing run takes ~3m40s, which may itself be a clue). I've stopped re-running rather than burn more CI on a known-flaky, pre-existing failure.

### Everything else is green

`build` (Windows), `chroma lifecycle`, `chroma round-trip`, `clean-room dependency closure smoke`, `server-runtime e2e (docker)`, and `sync-hub` all pass. `sync-hub` failed once on an unrelated 5000ms `testTimeout` at `workers/sync-hub/test/sync-hub.test.ts:1207` and passed on rerun; none of these commits touch `workers/`.

An earlier round also hit `npm error ETARGET @posthog/types@^1.409.0` on three jobs — a registry propagation race, since `posthog-node@^5.36.15` requires a version that had just been published. It resolves as latest now and those jobs went green on rerun.

PR #3875, opened 2026-09-04
server: let CLAUDE_MEM_SERVER_MAX_OUTPUT_TOKENS set the generation output cap

### LINKED ISSUE #3829: OpenRouter provider: hardcoded 30s timeout and 4096 max_tokens cause silent observation loss on slow models; no failover; unbounded history growth
## Summary

Running v13.18.0 with `CLAUDE_MEM_PROVIDER=openrouter`, four hardcoded behaviors in the OpenRouter path combine into silent, hours-long observation loss when the configured model is slow (e.g. any free-tier model under daytime load). Requesting configuration knobs for each.

## Environment

- claude-mem v13.18.0 (marketplace install, `worker` runtime)
- macOS, Bun worker
- Provider: OpenRouter

## What happened

With `CLAUDE_MEM_OPENROUTER_MODEL` set to a free-tier model (`nvidia/nemotron-3-ultra-550b-a55b:free`), extraction worked overnight, then failed continuously for ~3 hours during daytime load. Every call aborted at exactly 30s, retried twice, then:

```
[WARN ] [SDK   ] Retrying OpenRouter nvidia/nemotron-3-ultra-550b-a55b:free after 135ms (attempt 1/2) {kind=unclassified, message=The operation was aborted.}
[ERROR] [SDK   ] [session-161] OpenRouter init query failed {model=...} The operation was aborted.
[ERROR] [SESSION] [session-161] Generator failed {provider=openrouter, error=The operation was aborted.}
```

The pending batches are abandoned (queue depth returns to 0), so the outage window becomes a permanent memory gap. A direct probe of the same model at that time took 17s for a 10-token completion — real but slower than the retry budget allows.

## Root causes (all hardcoded in `worker-service.cjs`)

1. **Per-attempt timeout is fixed at 30s** — retry defaults `{maxRetries:2, perAttemptTimeoutMs:3e4, ...}`. No env/setting overrides it. A model that reliably answers in 40s can never succeed.

2. **`max_tokens` is fixed at 4096** in the request body (`temperature:.3,max_tokens:4096`). On busy sessions the observation XML gets truncated mid-stream and trailing observations are lost. Over a multi-week log history I count 1,000+ calls returning exactly 4096 output tokens (i.e. `finish_reason: length`). No setting overrides it.

3. **No provider/model failover.** `CLAUDE_MEM_OPENROUTER_MODEL` accepts an array, but it's comma-joined into a single `model` string — OpenRouter rejects that. OpenRouter's native fallback (`models: [...]` array in the request body) is never used, so there's no way to express "try A, fall back to B."

4. **Observer conversation history grows unboundedly.** `conversationHistory.push(...)` with no windowing means every call re-sends the entire multi-turn history. On long sessions I've logged single calls with 295 messages and 903,805 input tokens. This makes cost scale quadratically with session length and eventually collides with model context limits (only ~1M-context models are safe).

Also noticed while debugging: the shipped default `xiaomi/mimo-v2-flash:free` now returns 404 from OpenRouter ("This model has been deprecated"), so a fresh OpenRouter setup with no model configured fails out of the box.

## Requested

1. `CLAUDE_MEM_OPENROUTER_TIMEOUT_MS` (per-attempt timeout override)
2. `CLAUDE_MEM_OPENROUTER_MAX_TOKENS` (max_tokens override)
3. Map the existing array form of `CLAUDE_MEM_OPENROUTER_MODEL` onto OpenRouter's `models` fallback array (first entry → `model`, rest → `models`), instead of comma-joining
4. A rolling window for the observer conversation (keep the instruction turn + last N exchanges), or an env cap on history tokens
5. Update the default OpenRouter model to a live one

Items 1, 2, and 5 look like one-line changes; happy to test any of them. Thanks for claude-mem — the local pipeline design is excellent, this is purely about surviving slow/flaky models.

🤖 Filed with [Claude Code](https://claude.com/claude-code) on behalf of @emeffinsteve

escooterclinic:
Closes #3829 (the max_tokens half).

All three server generation providers already take a `maxOutputTokens` option, but `instantiateServerGenerationProvider` never populates it, so every request goes out with the constructor default of 4096.

On a model that answers at length that truncates the reply mid-structure. The observation parser then finds no complete `<observation>` block, returns `{valid:false}`, and `processGenerationResult` settles the job as `parse_error` with `retryable:false` — one attempt of three, and the observation is gone. It looks like a model problem rather than a configuration one, which is what made it hard to spot: we lost 20 observations to it before working out the reply was simply being cut off at exactly 4096 completion tokens.

This reads `CLAUDE_MEM_SERVER_MAX_OUTPUT_TOKENS` and passes it to whichever provider is selected. Unset keeps 4096, so nothing changes unless you set it. Non-integer or non-positive values are ignored with a warning rather than sent as `NaN`.

Tests in `tests/server/runtime/server-max-output-tokens.test.ts`: the parsing cases, plus two that capture the outgoing request body and assert `max_tokens` is 4096 by default and the configured value when set. They fail without the change. `bun test tests/server/` is 170 pass / 13 skip / 0 fail.

I've left the worker-side providers alone — same shape, but it's a separate concern and I didn't want to widen this.

### DIFF
--- a/src/server/runtime/create-server-service.ts
+++ b/src/server/runtime/create-server-service.ts
@@ -253,29 +253,54 @@ function buildServerGenerationProviderFromEnv(): ServerGenerationProvider | null
   }
 }
 
+/**
+ * Optional cap on generated tokens for the server generation providers (#3829).
+ * All three providers already accept `maxOutputTokens`; nothing populated it, so
+ * every request went out with the constructor default of 4096. On a model that
+ * answers at length the reply is truncated mid-structure, the observation parser
+ * rejects it, and the job settles as a non-retryable parse_error — the work is
+ * lost silently. Unset keeps the 4096 default, so behavior is unchanged.
+ */
+export function resolveServerMaxOutputTokens(): number | undefined {
+  const raw = process.env.CLAUDE_MEM_SERVER_MAX_OUTPUT_TOKENS;
+  if (!raw) return undefined;
+  const parsed = Number(raw);
+  if (!Number.isInteger(parsed) || parsed <= 0) {
+    logger.warn('SYSTEM', 'server: ignoring invalid CLAUDE_MEM_SERVER_MAX_OUTPUT_TOKENS', { value: raw });
+    return undefined;
+  }
+  return parsed;
+}
+
 function instantiateServerGenerationProvider(provider: string): ServerGenerationProvider | null {
   if (provider === 'claude' || provider === 'anthropic') {
     const apiKey = process.env.ANTHROPIC_API_KEY ?? process.env.CLAUDE_MEM_ANTHROPIC_API_KEY ?? '';
     if (!apiKey) return null;
-    const opts: { apiKey: string; model?: string } = { apiKey };
+    const opts: { apiKey: string; model?: string; maxOutputTokens?: number } = { apiKey };
     if (process.env.CLAUDE_MEM_SERVER_MODEL) opts.model = process.env.CLAUDE_MEM_SERVER_MODEL;
+    const maxOutputTokens = resolveServerMaxOutputTokens();
+    if (maxOutputTokens !== undefined) opts.maxOutputTokens = maxOutputTokens;
     return new ClaudeObservationProvider(opts);
   }
   if (provider === 'gemini') {
     const apiKey = process.env.GEMINI_API_KEY ?? process.env.CLAUDE_MEM_GEMINI_API_KEY ?? '';
     if (!apiKey) return null;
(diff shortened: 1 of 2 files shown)


thedotmack:
ALL-NIGHT SHIP Batch D: rebased onto latest `main` as #3997. Will merge #3997 if CI is green.

thedotmack:
Superseded by #3997.

thedotmack:
Superseded by #3997 (rebased onto latest main and squash-merged). Thank you.

PR #3877, opened 2026-09-04
fix(observer): don't store observations parsed from the init reply

sanztheo:
## Symptom

An observation can be stored asserting work that never happened.

A session that had made no tool calls yet got this stored a few seconds in, before its only `Edit` was processed:

- title: *"Provider credit exhaustion detection"*
- narrative: *"Provider behavior was extended so exhausted credits are recognized as a distinct condition"*
- `files_modified: []`

Nothing had been observed at that point. The observation is a paraphrase of the user's prompt, stored as memory. It then feeds cross-session context injection like any real one — which is worse than a missing observation, because the next session reads it as something that happened.

## Mechanism

`OpenAICompatibleProvider.startSession` sends the briefing prompt as a standalone query and hands the reply straight to the storage path:

- `src/services/worker/OpenAICompatibleProvider.ts:135-137` builds `buildInitPrompt` / `buildContinuationPrompt`
- `:145` sends it as its own query
- `:146` → `handleInitResponse` (`:206-228`), which at `:213` checks only `if (initResponse.content)` and at `:219` calls `processAgentResponse` unconditionally
- `src/services/worker/agents/ResponseProcessor.ts:311-320` parses it, `:445` stores whatever `<observation>` it found

The brief cannot legitimately contain one. `src/sdk/prompts.ts:64-90` shows the model the observation output format and an `<observed_from_primary_session>` block holding only `<user_request>` and `<requested_at>` — there is no `<what_happened>`, no tool call, nothing observed — and then ends with the `MEMORY PROCESSING START` header. The mode's own `skip_guidance` says to return an empty response when skipping. Most of the time the model does. Sometimes it fills in the format from the user request instead, and that gets stored.

`files_modified: []` is the fingerprint: `sanitizeObservationFiles` (`ResponseProcessor.ts:249-258`) strips file claims that no claimed message supports, and at init there are none.

## Evidence

Measured on a non-Anthropic provider: **2 of 12 sessions** produced a well-formed `<observation>` in reply to the brief, and both were stored. One landed in a real corpus, with the content above.

Reproduced offline, no API calls, driven through the real base class: a stub subclass overriding only `query()` to return an `<observation>` for the init call, with an empty message iterator, stores one observation. The same holds with the real `GeminiProvider` and `OpenRouterProvider` and `fetch` mocked, because `startSession` and `handleInitResponse` are base-class private — `GeminiProvider.ts:223-226` and `OpenRouterProvider.ts:301-304` only supply `getConfig`/`query`/`estimateTokens`/`buildLastUsage`. Control: an empty init reply stores nothing.

Not a regression — `git show 4ad2383f` shows `handleInitResponse` already called `processAgentResponse` before #3217.

**You can measure the rate on your own fleet without trusting the number above:** every `session_compressed` event with `hook='init'` and `count>0` is an init-derived observation (`ResponseProcessor.ts:493` is the count, `:506` stamps `hook: session.lastGeneratorSource`, set to `'init'` at `OpenAICompatibleProvider.ts:144`). The 2/12 is one model on one provider; I have not measured Gemini, OpenRouter or the Claude path.

## The fix

`handleInitResponse` keeps the reply's token accounting and appends it as the assistant turn itself, instead of routing it through `processAgentResponse`. +16/-18 in one file.

The assistant turn is still pushed, so role alternation is byte-for-byte what it was — this only removes the storage call. `session.lastUsage` is deliberately no longer set at init: `ResponseProcessor.ts:485-487` reads and nulls it, so with no consumer on this path it would have leaked the init call's usage into the next real observation's telemetry.

## What it leaves alone

- **`ClaudeProvider` is exposed the same way and is not fixed here.** `ClaudeProvider.ts:566-568` builds the same brief, `:575-584` yields it as the first user message, and `:359-429` treats every `assistant` message identically — `:416` passes it to `processAgentResponse` with no notion of which prompt it answers. Fixing it needs a different mechanism (the SDK streams, so it turns on the one-reply-per-turn invariant rather than a return value) and deserves its own change. Happy to follow up. It also means the telemetry query above keeps working as a measurement of the remaining exposure after this merges.
- **The prompt text.** A line telling the model no tool use has been observed yet would help the Claude path too, but it is probabilistic and untestable, and the mode text is yours.
- **One deliberate loss, stated plainly:** the init reply no longer reaches the prose-rejection recovery paths in `ResponseProcessor` (context overflow → recycle, quota prose → pause and preserve the batch, auth prose → abort). Those now trigger one query later, on the first observation, which goes through `processAgentResponse` as before — so the recovery still happens, at the cost of one wasted request in those failure cases. If you would rather keep it at init, the alternative is a `store: false` parameter on `processAgentResponse`; I picked the smaller diff.

## Tests

- **New** `tests/worker/openai-compatible-init-response.test.ts`, in the idiom of `tests/worker/openai-compatible-summary-tier.test.ts`: a provider that answers every prompt with a valid observation, one tool call in the iterator. Asserts two queries, `storeObservations` called once, and history roles `['user','assistant','user','assistant']`. On `main` it fails with `Expected number of calls: 1 / Received number of calls: 2`.
- **Re-targeted, two tests in `tests/gemini_provider.test.ts`.** Both stored through the init query rather than through a tool call, because the init query is the only one that needs no message iterator — they were exercising the store pipeline, not asserting that a brief reply is memory. Both now queue one tool call: `should process observations and store them` (also tightened to `toHaveBeenCalledTimes(1)`, so it is now a guard for this bug too) and the #3217 deferred-attribution test, whose `fetch` mock now completes the brief and holds the observation query open instead. Its attribution assertions are unchanged.

Green: `npm run typecheck`; `bun test tests/worker tests/shared tests/env-isolation.test.ts` → 790 pass, 1 skip, 0 fail; `npm run lint:spawn-env`; `npm run lint:hook-io`; `tests/gemini_provider.test.ts` 12/12.

---

Unrelated note, no change proposed: `CLAUDE_MEM_TIER_SUMMARY_MODEL` reaches non-Claude providers as a literal model id (`model-aliases.ts:36-42` → `OpenAICompatibleProvider.ts:334-336`), which is what #3257 was for, so setting it to an id the active provider does not serve makes every summary fail with a 4xx while observations keep working. The default is `''` (`SettingsDefaultsManager.ts:186`), so a default install is unaffected and this is a misconfiguration, not a bug — the comment there could just say "otherwise a model id the active provider serves" if you think it is worth the character.

### DIFF
--- a/src/services/worker/OpenAICompatibleProvider.ts
+++ b/src/services/worker/OpenAICompatibleProvider.ts
@@ -135,7 +135,6 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
     const initPrompt = session.lastPromptNumber === 1
       ? buildInitPrompt(session.project, session.contentSessionId, session.userPrompt, mode, priorContext)
       : buildContinuationPrompt(session.userPrompt, session.lastPromptNumber, session.contentSessionId, mode, priorContext);
-    const initContext = snapshotResponseContext(session);
 
     session.conversationHistory.push({ role: 'user', content: initPrompt });
 
@@ -143,7 +142,7 @@ export abstract class OpenAICompatibleProvider<TConfig extends { apiKey: string;
       session.lastPromptSentAt = Date.now();
       session.lastGeneratorSource = 'init';
       const initResponse = await this.query(session.conversationHistory, config);
-      await this.handleInitResponse(initResponse, session, worker, model, initContext);
+      this.handleInitResponse(initResponse, session, model);
     } catch (error: unknown) {
       // Classified errors are logged once, at SessionRoutes' `Observer failed`
       // line; here they're debug-level so one failure isn't five error lines.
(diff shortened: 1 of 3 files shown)


thedotmack:
Night-ship gate (plan-18 batch A, after #3363/#3501):
1. RCA: YES — logged in PR body (init reply stored as invented memory; `files_modified: []` fingerprint).
2. No second-system: YES — `handleInitResponse` only; ClaudeProvider explicitly left for a follow-up.
3. Narrow: YES — stop routing init replies through `processAgentResponse`; keep history + token totals.

Rebase onto current main is clean locally. `bun test tests/worker/openai-compatible-init-response.test.ts tests/gemini_provider.test.ts` → 13 pass. Cannot push rebase to the fork (maintainer update-branch 403). Squash-merging the original if GitHub still reports mergeable.

PR #3894, opened 2026-09-06
fix: say why the worker died instead of listing suspects

yetanotherflo:
## The failure

A worker that dies during module resolution leaves nothing behind. `spawnDaemon` detaches it with its stdio discarded — `Start-Process -WindowStyle Hidden` on Windows, `stdio:'ignore'` elsewhere — so the only trace is:

```
[ERROR] [SYSTEM] Worker exited before readiness endpoint became available
[ERROR] [SYSTEM] Worker auto-start failed — ... Check earlier log lines for the
        specific failure reason (Bun not found, missing worker bundle, port conflict, etc.)
```

None of those three was the cause when this bit me. A truncated `bun install` in the plugin cache had left `zod@4.4.3` short 37 files (`v4/locales/be.cjs` among them), so the bundle died on `require`. Every claude-mem hook failed for hours with nothing in the log to say why; finding it took manually running the bundle in the foreground.

## The fix

Do that manual step automatically, on the failure path only.

- `probeWorkerBootFailure()` re-runs the bundle where its stderr can be read and logs what it says, as `bootFailure` on the existing error line.
- `status` is the probe command: it executes every top-level require — where these failures happen, long before argv is parsed — then exits 0 on every branch without starting a server, so a healthy bundle costs one silent subprocess. `scripts/smoke-clean-room.cjs` already guards the same class of failure at build time with the same trick.
- The reason also rides out through `getLastWorkerBootFailure()` into the `start` hook's status line, so `Failed to start worker` becomes `Failed to start worker: error: Cannot find module …`.

Before / after on the same reproduction:

```
- Worker exited before readiness endpoint became available
+ Worker exited before readiness endpoint became available {bootFailure=error:
+   Cannot find module './this-dependency-was-never-installed.cjs' from '…\worker-service.cjs'
+   Bun v1.3.13 (Windows x64)}
```

## Why there is a retry

The probe needs one to work at all where it matters most. Called from the failure path under Bun 1.3.13 on Windows, the **first** sync spawn returns `ETIMEDOUT` in ~25 ms with the child `SIGTERM`ed before printing a byte — measured identically at an 8 s and at a 60 s timeout — while a second, identical spawn immediately after answers in ~180 ms. A plain 50 s idle await, a preceding `execSync`, and pending timers each fail to reproduce it in isolation, so the retry is gated on the observation rather than on a theory: `shouldRetryWorkerBootProbe` fires only for a window that expired far too early to be real. A genuine timeout — the bundle loaded and kept running — burns the full window and is never retried.

## Verification

- `bun test tests/infrastructure/process-manager.test.ts` — 60 pass, 8 new (5 probe behaviours against real subprocesses, 3 retry-predicate). The one failure, `resolveWorkerRuntimePath > should look up Bun on non-Windows when caller is Node`, is pre-existing on `main` on a Windows host and untouched by this change.
- `bun test tests/worker-spawn.test.ts tests/worker-script-resolution.test.ts tests/infrastructure/` — 244 pass / 12 fail, against a `main` baseline of 239 pass / 12 fail: same 12, plus the new tests.
- `npm run typecheck:root`, `npm run build`, `npm run lint:spawn-env`, `npm run lint:hook-io` all clean.
- End-to-end: `ensureWorkerStarted` against a bundle that cannot resolve its dependencies now returns `dead` with the module error in both the log and `getLastWorkerBootFailure()`.

Source and tests only — no rebuilt bundles, since CI runs `npm run build` itself.

### DIFF
--- a/src/services/infrastructure/ProcessManager.ts
+++ b/src/services/infrastructure/ProcessManager.ts
@@ -369,6 +369,101 @@ export function buildWindowsDaemonStartCommand(runtimePath: string, scriptPath: 
   return `Start-Process -FilePath '${psSingleQuote(runtimePath)}' -ArgumentList @('"${psSingleQuote(scriptPath)}"','--daemon') -WindowStyle Hidden`;
 }
 
+export const WORKER_BOOT_PROBE_TIMEOUT_MS = 5000;
+const WORKER_BOOT_PROBE_MAX_LINES = 8;
+
+/**
+ * Did the probe "time out" too early for that timeout to be real?
+ *
+ * Called from the worker-start failure path, the FIRST sync spawn comes back
+ * ETIMEDOUT in ~25ms — the child SIGTERMed before it could print a byte — no
+ * matter how generous the window is (measured identically at 8s and at 60s
+ * under Bun 1.3.13 on Windows). A second, identical spawn immediately after
+ * always answers in ~180ms, so the deadline is being resolved against something
+ * stale that the first call refreshes. A genuine timeout burns the whole window
+ * instead, which is what separates the two here.
+ */
+export function shouldRetryWorkerBootProbe(
+  error: Error | undefined,
+  elapsedMs: number,
+  timeoutMs: number
+): boolean {
+  if ((error as NodeJS.ErrnoException | undefined)?.code !== 'ETIMEDOUT') return false;
+  return elapsedMs < timeoutMs / 2;
+}
+
+/**
+ * Re-run the worker bundle in the foreground to recover the stderr spawnDaemon
+ * threw away.
+ *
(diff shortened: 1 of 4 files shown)


thedotmack:
Night-ship gate (Batch B):
1. RCA: YES — logged in PR body (truncated `bun install` / missing module; daemon stdio discarded).
2. No second-system: YES — failure-path probe only; no new supervisor.
3. Narrow: YES — recover stderr on dead spawn; surface via existing start status line.

Rebase onto current main is clean. `bun test tests/infrastructure/process-manager.test.ts tests/worker-spawn.test.ts tests/worker-script-resolution.test.ts` → 88 pass. Cannot push rebase to the fork. Squash-merging if GitHub reports mergeable.

PR #3900, opened 2026-09-07
fix(windows): reclaim ghost listeners left by out-of-band worker deaths

### LINKED ISSUE #3899: Windows: out-of-band worker kill leaves a ghost LISTEN socket held by the chroma sidecar chain; no launcher reclaims it (port blocked until manual tree-kill)
## Bug Description

Windows: when the worker daemon is killed **out-of-band** (crash, Task Manager, `taskkill /F /PID` without `/T`), the chroma-mcp sidecar chain it spawned (`uvx -> uv -> chroma-mcp -> python`, spawned with piped stdio which forces handle inheritance) survives holding the worker's inherited listening socket. The port stays `LISTENING` under the **dead** worker PID — a ghost listener — and no launcher ever recovers:

- daemon gate: `Port already in use, refusing to start duplicate` → exit 0
- `ensureWorkerStarted`: `Port in use but worker not responding to health checks` → `'dead'`

Every hook/MCP connection repeats this forever; the port is blocked until a human tree-kills the sidecar chain by hand. Reproduced on Windows 11 (2026-09-07 and again while writing this issue): killing the worker left `chroma-mcp.exe` + two `python.exe` alive holding `:37777` under the dead PID; `taskkill /T` on that chain released the port immediately.

This is the plan-15 ghost-listener defect (#3603) in its "unclean death" form. The graceful-shutdown path (#3644) and the version-recycle path (#3482) tree-kill their children; the out-of-band path runs no code at all, so nothing cleans up.

## Why the survivor is hard to find (broken chain)

After the worker dies, its `uvx`/`uv` stdio layers read EOF and **exit**, so the surviving `chroma-mcp`/`python` sidecar hangs one or two links BELOW a dead PID:

```
worker (dead) ─ uvx (dead, EOF) ─ uv (dead, EOF) ─ chroma-mcp.exe (ALIVE) ─ python.exe (ALIVE) ─ python.exe (ALIVE)
```

netstat's owner is the dead worker. A parent-chain walk from that PID stops at the first dead link. The survivors' command lines still carry `--data-dir <DATA_DIR>/chroma`, which is a reliable ownership fingerprint.

## Steps to Reproduce

1. Windows 11, worker running with a live chroma-mcp chain (any normal usage).
2. `taskkill /F /PID <worker pid>` (no `/T`; or crash the worker).
3. `netstat -ano | findstr 37777` → still `LISTENING`, owner PID no longer exists.
4. Try to start the worker (`npx claude-mem worker start`, or open any Claude session) → "Port already in use, refusing to start duplicate" / `'dead'`, forever.

## Expected Behavior

The next launcher detects the ghost (bound port whose owner is dead + no health response), reclaims the dead worker's surviving sidecar chain, and starts normally — the recovery I performed by hand should be automatic.

## Environment

- Windows 11 Home 26200
- claude-mem 13.16.1 at time of incident; reproduced against current main (13.24.x)
- Bun 1.3.6, uv with chroma-mcp 0.2.6 sidecar

## Logs

`~/.claude-mem/logs/` shows repeated `Port already in use, refusing to start duplicate` / `Port in use but worker not responding to health checks` and no recovery.

## Related work (read before filing duplicates)

- Parent tracker: #3603 (plan-15), children #2893 / #2963 / #3482 / #3450 / #3596 — same family.
- #3309 (open): **prevents** the inheritance (non-inheritable listener). Complementary: does not reclaim chains that already exist from before the fix.
- #3416 (open): **routes around** the ghost by relocating to another port; explicitly does not reclaim the socket.
- #3614 bundle → #3405 (closed into it): reclaims via claude-mem's own `supervisor.json`/start-token records (rung 3a). This issue is about the case the registry cannot see — e.g. the registry entry was overwritten by a later generation, or the claim must be provable from the OS alone.

## Proposed fix (PR follows)

Reclaim at spawn time using OS-level evidence only:

1. netstat: bound port, every owner PID dead;
2. process table: kill chroma sidecar descendants of the dead owner (walk) **plus** any chroma sidecar whose `--data-dir <DATA_DIR>` argument matches this install (full-table scan, covers the broken-chain python links);
3. every kill carries the start token from the same table read that discovered it; a live owner is never touched (wedged worker / foreign process keeps current behavior);
4. bounded (5s) health probes so a ghost that accepts TCP but never responds cannot hang the launcher.

Verified end-to-end on Windows 11: out-of-band kill → ghost → production `ensureWorkerStarted()` reclaims (port free, chain gone) → replacement worker serves health. Fail-on-main integration gate included.

weiconghe:
## Summary

Closes #3899.

On Windows, a worker killed **out-of-band** (crash, `taskkill /F` without `/T`) leaves its chroma sidecar chain (`uvx -> uv -> chroma-mcp -> python`, spawned with piped stdio → inherited socket handles) alive holding the listening socket. The port stays `LISTENING` under the dead worker PID, and every launcher refuses to start forever:

```
Port already in use, refusing to start duplicate        (daemon gate)
Port in use but worker not responding to health checks  (ensureWorkerStarted -> 'dead')
```

Until now the only recovery was a human tree-kill of the chain by hand. This PR makes that recovery automatic at spawn time.

**Relationship to existing work (different layers, all complementary):**

| Work | Layer | Gap this PR covers |
|---|---|---|
| #3309 (open) | prevents inheritance (non-inheritable listener) | cannot help chains that already exist from a pre-fix / crash-then-upgrade world |
| #3416 (open) | routes around the ghost (relocate port) | never frees the original port/socket |
| #3405 → #3614 bundle | reclaims via `supervisor.json` start-token records | a registry entry overwritten by a later generation, or a chain whose registry root died, is invisible to it |

This PR proves ownership from the OS alone (no claude-mem records): the survivors of a dead listener still carry `--data-dir <DATA_DIR>` on their command line.

### 1. `src/shared/port-reclaim.ts` (new) — ghost-listener reclaim

Detection + recovery ladder, all evidence OS-side, all safety conditions enforced before any signal:

1. **netstat** finds the LISTENING owner PID(s) for the port.
2. **Live owner → never touched** (`owner-alive`): a wedged-but-alive worker or a foreign process keeps the current "in use" behavior. Only *every-owner-dead* proceeds.
3. Targets are chroma sidecars, located two ways:
   - **walk** down the dead owner's parent chain (Windows preserves parent links after death);
   - **full-table scan** matching the sidecar's `--data-dir <DATA_DIR>` command-line fingerprint — this is the load-bearing case: when the worker dies, `uvx`/`uv` read EOF and exit, so the surviving `chroma-mcp`/`python` sits one or two links below a dead PID, unreachable by the walk. The `python.exe` links are matched by their `--data-dir` argument (their executable name alone would miss them).
4. Every kill goes through `killProcessTree` with the start token from the **same table read** that discovered the target (identity never re-probed after an await). PID reuse cannot pull in strangers: a reused owner PID whose new children are unrelated is filtered by name and by the data-dir fingerprint.
5. Post-kill netstat re-probe; `reclaimed:true` only when the port is verified free.

### 2. Reclaim wired into both spawn paths

- `ensureWorkerStarted()` (worker-spawner.ts): the "Port in use but worker not responding" branch now reclaims before giving up; a successful reclaim also clears the Windows spawn-cooldown marker (its reason — the ghost — is gone; the plan-15 "cooldowns keyed to evidence, not time" rule).
- daemon duplicate gate (worker-service.ts): bound-but-silent now reclaims before `exit(0)`; a live worker still exits 0 immediately, and a live-but-wedged owner still refuses (never killed from a spawn path).

### 3. Bounded health probes (HealthMonitor.ts)

Ghost listeners accept TCP (the kernel completes handshakes) but never respond. Probes were unbounded, so the new health check in the daemon gate — and the old ones — could hang forever on a ghost. All HTTP probes now abort at 5s (above a healthy worker's sub-100ms response, below every retry budget). `/api/admin/shutdown` responds before draining (flushResponseThen), so this does not truncate graceful shutdown.

## How verified?

**Unit (cross-platform, injected seams):** `tests/shared/port-reclaim.test.ts` — 16 tests: netstat parsing (IPv4/IPv6/dedupe), sidecar-name whitelist, data-dir fingerprint matcher (slash forms, case, quoted values, prefix lookalikes), and every reclaim branch (not-supported / owner-alive / no-listener / broken-chain via fingerprint / PID-reuse safety / kill-failed / still-bound / multi-owner). `tests/services/worker-spawner.test.ts` gains: ghost-reclaimed → proceeds to spawn; unreclaimable → stays dead. `bun run typecheck` clean.

**Real Windows end-to-end (fail-on-main integration gate):** `tests/integration/worker-ghost-port-recovery.test.ts` (wired into the `chroma-windows` job of `.github/workflows/windows.yml`, `CLAUDE_MEM_TEST_CHROMA=1`):

1. detached fixture worker (PowerShell Start-Process — the same way `spawnDaemon()` launches the real daemon; a bun-test child does NOT reproduce the ghost because the test runner manages its process tree) starts a real chroma-mcp chain and binds the port;
2. `taskkill /F /PID` (no `/T`) kills it;
3. asserts the ghost: port LISTENING under the dead PID, uvx/uv exited on EOF, chroma-mcp/python survivors alive;
4. drives the PRODUCTION `ensureWorkerStarted()` → on main it returns `'dead'` (no reclaim exists); with this PR it returns ready and a replacement worker serves health;
5. asserts every snapshotted sidecar descendant (pid + start token matched) is gone.

Result on Windows 11 (real chroma, isolated data dir): `1 pass, 0 fail` — ghost precondition asserted, reclaim freed the port, replacement worker healthy.

Also reproduced the exact broken-chain shape in isolation (bun server + real chroma chain, worker killed; survivors = chroma-mcp + two python under dead intermediate PIDs) and confirmed: v1 walk-only reclaim could NOT see them (`no-chroma-descendants`), the data-dir fingerprint scan kills exactly those three processes and the port frees (`reclaimed:true`).

## Checklist

- [x] Tests added/updated and passing (unit + Windows integration gate)
- [x] Typecheck passing

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -201,6 +201,19 @@ jobs:
           $bun = Join-Path $env:USERPROFILE '.bun\\bin\\bun.exe'
           & $bun test tests/integration/worker-recycle-orphans.test.ts --timeout 600000
 
+      # Ghost-listener recovery gate (plan-15 #3603). Fails on main: a worker
+      # killed OUT-OF-BAND (taskkill /F without /T) leaves its chroma sidecar
+      # chain holding the inherited listening socket, and no launcher reclaims
+      # the port — ensureWorkerStarted returns 'dead' forever. Passes once the
+      # spawn path reclaims the dead owner's sidecar descendants and starts.
+      - name: Ghost-listener recovery gate (#3603)
+        shell: pwsh
+        env:
+          CLAUDE_MEM_DATA_DIR: ${{ runner.temp }}\claude-mem-ghost-data
+        run: |
+          $bun = Join-Path $env:USERPROFILE '.bun\\bin\\bun.exe'
+          & $bun test tests/integration/worker-ghost-port-recovery.test.ts --timeout 600000
+
       # Diagnostic only — never fails the job. Ancestry/identity filtering is
       # done inside the tests; this is just a human-readable postmortem for
       # when the gate above goes red.
--- a/src/services/infrastructure/HealthMonitor.ts
+++ b/src/services/infrastructure/HealthMonitor.ts
@@ -15,12 +15,22 @@ function formatHostForUrl(host: string): string {
   return host.includes(':') ? `[${host}]` : host;
 }
 
+// Probes against a ghost listener (plan-15 #3603) can connect — the kernel
+// completes handshakes on the inherited socket — but never receive a response,
+// because no application is reading. An unbounded fetch would hang the probe
+// forever, so every HTTP probe is aborted after this budget. 5s is above a
+// healthy worker's sub-100ms response and below every caller's retry budget.
+const HEALTH_PROBE_TIMEOUT_MS = 5_000;
+
 async function httpRequestToWorker(
   port: number,
   endpointPath: string,
   method: string = 'GET'
 ): Promise<{ ok: boolean; statusCode: number; body: string }> {
-  const response = await fetch(`http://${formatHostForUrl(getWorkerHost())}:${port}${endpointPath}`, { method });
+  const response = await fetch(`http://${formatHostForUrl(getWorkerHost())}:${port}${endpointPath}`, {
+    method,
+    signal: AbortSignal.timeout(HEALTH_PROBE_TIMEOUT_MS),
+  });
   let body = '';
   try {
     body = await response.text();
--- a/src/services/worker-service.ts
+++ b/src/services/worker-service.ts
@@ -48,6 +48,7 @@ import {
   touchPidFile
 } from './infrastructure/ProcessManager.js';
 import { runOneTimeV12_4_3Cleanup } from './infrastructure/CleanupV12_4_3.js';
+import { reclaimGhostListeningPort } from '../shared/port-reclaim.js';
 import {
   isPortInUse,
   waitForHealth,
@@ -1451,8 +1452,28 @@ async function main() {
       // port — the port cannot be faked by a stale or clobbered file. Exit 0:
       // duplicate suppression is a success, not a failure.
       if (await isPortInUse(port)) {
-        logger.info('SYSTEM', 'Port already in use, refusing to start duplicate', { port });
-        process.exit(0);
+        // A live worker answers health — this is a genuine duplicate.
+        if (await waitForHealth(port, getPlatformTimeout(HOOK_TIMEOUTS.HEALTH_CHECK))) {
+          logger.info('SYSTEM', 'Worker already running (health verified), refusing to start duplicate', { port });
+          process.exit(0);
+        }
+        // Bound but silent: likely a ghost listener — a dead worker whose
+        // surviving chroma sidecar chain holds the inherited socket
+        // (plan-15 #3603). Reclaim when the owner is provably dead; a live
+        // owner (wedged worker, foreign process) keeps the duplicate refusal.
+        const reclaim = await reclaimGhostListeningPort(port);
+        if (reclaim.reclaimed) {
+          logger.info('SYSTEM', 'Reclaimed ghost listener left by a dead worker — starting anyway', {
+            port,
+            killedPids: reclaim.killedPids,
+          });
+        } else {
+          logger.info('SYSTEM', 'Port already in use, refusing to start duplicate', {
+            port,
+            reclaimReason: reclaim.reason,
+          });
+          process.exit(0);
+        }
       }
 
       // PID file second, ADVISORY only: it covers a dying-but-still-alive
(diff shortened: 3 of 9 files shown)


weiconghe:
Follow-up: the Windows gate this PR added hung on CI from the moment it merged — 12 of 12 `chroma lifecycle` runs, leaving main's Windows job red until now.

The cause was inside this PR's own change set. `isPortInUse()`'s Windows fast path is an HTTP probe, and it was the one probe left unbounded (the other probes in `HealthMonitor` got the 5s abort here). A ghost listener completes the TCP handshake and never answers, so that fetch never settles — and `ensureWorkerStarted()` runs it **before** the reclaim, so the reclaim this PR added never got the chance to fire. The local `1 pass, 0 fail` was real but not deterministic: the same machine passes when the probe's connect is refused and hangs when it is accepted.

Fixed in #3989: the probe now carries the same 5s abort budget (on timeout it falls through to the socket probe, which still reports the port as in use, so the launcher proceeds to the reclaim). The gate itself also gained stage lines and a deadline around `ensureWorkerStarted()`, so a future hang names its stage instead of dying silently on bun's 600s cap.

Issue #3903, opened 2026-09-07
Quota guard treats inactive overage utilization as exhausted when isUsingOverage is false


gfreschi:
## Before submitting

- [x] I searched existing issues and confirmed this is not a duplicate

---

## Bug Description

`shouldAbortForQuota()` treats the `overage.utilization` threshold as active even when the provider explicitly reports `isUsingOverage: false`.

The following valid rate-limit snapshot aborts with `quota:overage utilization 100.0% >= 95%`, which then arms the persistent 30-minute quota cooldown:

```ts
{
  rateLimitType: 'overage',
  utilization: 1,
  isUsingOverage: false,
  status: 'allowed_warning',
}
```

This is present in claude-mem 13.24.1, current `main`, and current `core-dev`.

## Steps to Reproduce

1. Add this deterministic unit test to `tests/worker/rate-limit-store.test.ts` on `core-dev`:

   ```ts
   it("does not abort on inactive overage at 100% utilization", () => {
     const store = new RateLimitStore();
     store.set({
       rateLimitType: "overage",
       utilization: 1,
       isUsingOverage: false,
       status: "allowed_warning",
     });

     expect(shouldAbortForQuota("cli", store).abort).toBe(false);
   });
   ```

2. Run `bun test tests/worker/rate-limit-store.test.ts`.
3. Observe `Expected: false`, `Received: true`.

## Expected Behavior

When `rateLimitType` is `overage` and `isUsingOverage` is explicitly `false`, ignore the utilization threshold because the provider says the overage bucket is not in use. Provider rejection remains authoritative through either `status: 'rejected'` or `overageStatus: 'rejected'`.

For compatibility, a missing `isUsingOverage` field should preserve the current utilization-threshold behavior.

## Environment

- **Claude-mem version**: 13.24.1
- **Claude Code version**: 2.1.263
- **OS**: macOS 26.6.2
- **Platform**: arm64

## Logs

No logs are required; the unit test reproduces the defect directly against the upstream implementation.

## Additional Context

This is distinct from #3870, which reports a stale `five_hour` snapshot after its reset time. Here the `overage` snapshot itself can be current; the defect is that its utilization is treated as active despite `isUsingOverage: false`.

PR #3837 introduced the quota event handling and PR #3838 corrected extraction of the SDK's `rate_limit_event` shape. Neither accounts for the provider's explicit inactive-overage flag when applying the utilization threshold.

thedotmack:
[...]

PR #3904, opened 2026-09-07
fix(quota): ignore inactive overage utilization

gfreschi:
## What changed

- ignore the overage utilization threshold when the provider explicitly reports `isUsingOverage: false`
- keep `status: "rejected"` and `overageStatus: "rejected"` authoritative
- preserve the existing utilization behavior when `isUsingOverage` is absent

## Why

The provider can report `overage.utilization: 1` while explicitly declaring that overage is not in use. Treating that utilization as active aborts observer work and arms the persistent 30-minute quota cooldown even though the snapshot remains allowed.

> [!IMPORTANT]
> Reviewer gate: confirm that the provider's explicit `isUsingOverage: false` is authoritative before skipping the utilization heuristic. Provider rejection remains a hard stop, and this PR does not enable or configure overage billing.

This change is limited to the quota decision and focused unit coverage. It does not change thresholds, cooldowns, persistence, telemetry, SDK extraction, API-key behavior, generated bundles, dependencies, or release metadata.

Addresses #3903

## Validation

- `bun test tests/worker/rate-limit-store.test.ts` - 26 pass, 0 fail
- `npm run typecheck` - pass
- `npm run build` - pass; generated bundles excluded from the commit
- `npm test` (mandatory pre-push hook) - 2,261 pass, 15 Postgres-gated skips, 0 fail
- `bun test openclaw` - 44 pass, 0 fail
- `npm run smoke:clean-room` - pass
- `git diff --check` - pass

### Links

- **Issue:** [#3903 - Quota guard treats inactive overage utilization as exhausted when isUsingOverage is false](https://github.com/thedotmack/claude-mem/issues/3903)


thedotmack:
Superseded by #3965 (rebased onto `main` — this PR originally targeted `core-dev` and the fork branch couldn't be updated). The fix is identical; merged via squash. Thank you @gfreschi!

Issue #3905, opened 2026-09-07
13.24.0 recycle loop also orphaned 808 `chroma-mcp` process pairs and exhausted 74 GB of swap; the leak path survives the 13.24.1 fix


stephan-nb:
## Summary

Thank you for 13.24.1, and for the unusually clear write-up in the CHANGELOG. It matches
what I diagnosed independently on my machine before I found the release.

I am filing this because the recycle loop had a second consequence that the 13.24.1 notes do
not mention. The CHANGELOG describes the impact as the kill "taking the in-flight observer
down with it every time", so lost observations. On my machine it also leaked a Chroma
subprocess pair per affected cycle. Over four days that reached 808 orphaned pairs, 1614
processes, and drove the host to 887 MB short of full swap exhaustion at 73.86 GB of 74.75
GB used.

The important part for you is that 13.24.1 fixes the trigger but not the leak path. Any
untrappable kill of the worker still strands its Chroma pair, and kernel memory pressure
produces exactly that kind of kill.

## Environment

- claude-mem 13.24.0 (marketplace install, worker bundle stamped 13.23.1), upgraded to
  13.24.1 during the incident
- `chroma-mcp` 0.2.6, launched through `uv tool uvx`
- bun 1.4.0
- macOS 24.6.0 (Darwin), Apple Silicon, 32 GB RAM
- Claude Code in VS Code, several concurrent sessions, 14 days of uptime

## What happened

The process tree for a healthy worker is:

```text
chroma python
  <- uv tool uvx --python 3.13 --with onnxruntime>=1.20 --with protobuf<7 --from chroma-mcp==0.2.6
    <- bun .../claude-mem/13.24.0/scripts/worker-service.cjs --daemon
      <- launchd
```

When `ensureWorkerRunning()` SIGKILLs the worker on a version mismatch, the `uv` and Python
processes below it are not signalled. They reparent to launchd and stay resident with an
idle stdin. Each stranded pair is small in resident terms, a few MB, but carries a large
dirty anonymous footprint, roughly 180 MB per pair once `onnxruntime` and the embedding
model are loaded. Those pages go to swap under pressure and never come back.

From one day of logs (`~/.claude-mem/logs/claude-mem-2026-09-07.log`):

| Signal | Count |
| --- | --- |
| `[SYSTEM] Worker version mismatch - killing stale worker` | 1014 |
| `[SYSTEM] Worker started` | 977 |
| `Worker is ready but still reports a stale version` | 892 |
[...]

Issue #3301, opened 2026-07-17
Chroma-mcp orphan leak across worker generations: live orphan's registry entry is overwritten without a kill (682 orphans / 21.7 GB swap)


eslonaguiar:
## Summary

Every time the worker dies **without a graceful shutdown** (SIGKILL, crash, OOM kill), its chroma-mcp subprocess tree survives — and the next worker generation silently discards the only record of it. One leaked `uvx → uv → python chroma-mcp` tree per ungraceful worker death, unbounded across restarts.

On an 8 GB macOS machine (Apple Silicon, macOS 25.5.0, claude-mem 13.11.0) this accumulated **682 orphaned processes (341 uvx wrappers + 341 python instances) in ~15 hours**, filling 21.7 GB of swap and driving the load average to 293 until macOS started showing the "system has run out of application memory" dialog. Memory pressure makes workers die more often, which leaks more orphans — a self-reinforcing spiral.

Likely a major contributor to the process-accumulation reports in #3270 and related to the leak scenarios described in #3298.

## Root cause

The in-process lifecycle in `ChromaMcpManager` is well defended (`disposeCurrentSubprocess()`, tree-kill on failed connect, `onclose` sweep — #2313). The hole is at the **worker-process-death boundary**, and it takes three steps:

1. **Worker dies ungracefully.** The chroma-mcp child is not tied to the worker's lifetime (no process group ownership), so it survives and re-parents to init/launchd. `supervisor.json` still records its PID under the `chroma-mcp` id.
2. **Next worker boots.** `ProcessRegistry.initialize()` loads the persisted entries and calls `pruneDeadEntries()` — which only removes entries whose PID is **dead** (`if (isPidAlive(info.pid)) continue;`). The orphan's entry is alive, so it is kept... but nothing acts on it.
3. **New worker connects.** `connectInternal()` → `registerManagedProcess()` → `registry.register('chroma-mcp', …)` **overwrites** the old entry with the new child's PID. The orphan's PID — which the registry knew about moments earlier — is forgotten without ever being signaled.

The writer lock does not help: it records the **worker's** PID, not the chroma child's, so `isChromaWriterLockLive()` correctly reports the lock stale (owner dead) and the new worker proceeds — right past the still-running previous chroma-mcp.

## Reproduction (macOS/Linux)

```bash
# 1. Let the worker connect chroma-mcp, note the subprocess PID
#    (supervisor.json → processes["chroma-mcp"].pid)
[...]

PR #3923, opened 2026-09-08
fix(supervisor): reap chroma-mcp trees no worker owns at boot

stephan-nb:
## Summary

Boot-time sweep of the process table for `chroma-mcp` trees that no worker owns, plus a `doctor` census line. Complements the registry-based reapers rather than replacing them.

Refs #3905 (the report this comes from), #3301, #3175. Related to #3614 / plan-14 and to #3178, #3302, #3465; no `Closes` keywords, on purpose.

## The gap this fills

Every reaper written so far operates on `supervisor.json` rows: #3302 reaps the persisted `chroma-mcp` row before spawning, #3178 reaps live rows whose owner is gone at boot, #3465 keeps the dying worker from clobbering the successor's registry. All three are right, and all three miss the same set: trees whose row was already overwritten. Chroma is registered under one fixed key, so a worker that dies without its cascade (SIGKILL from the 13.24.0 recycle loop, OOM, `kill -9`, a deadline firing before teardown) leaves a tree that the next spawn's `register()` forgets. From then on no row exists for it, and a row-based sweep cannot see it.

That set is not small. #3905 measured 808 orphaned pairs and 74 GB of swap on one machine after four days of the recycle loop; a second census in the same thread found 619 pairs, 33 GB, load average 353, accumulated over seven weeks on the ordinary upgrade cadence, with one orphan older than any cached plugin version. 13.24.1 removed the fast path; the debt on installed machines stays until something reads the process table.

## What this does

`src/supervisor/orphan-chroma-sweep.ts`, called once from `Supervisor.start()` after `registry.initialize()`, detached and best-effort:

1. One `ps -eo pid=,ppid=,command=` snapshot (Linux and macOS; Windows returns early, as #3302 does).
2. Keep rows whose executable is `uvx`, `uv` or `python*` and whose command line carries `chroma-mcp`. The executable constraint is what keeps an editor open on `chroma-mcp-incident.md` out of the candidate set; the loose tail is what keeps the python grandchild, the process actually holding the memory, in it.
3. Group into trees; a root is a signature row whose parent is not one. A root is orphaned when its parent is PID 1 or dead. Roots parented to this worker are never touched.
4. Each orphan root goes through the shared `killProcessTree` (graceful), which already self-captures the root's start token and revalidates before every signal, so a recycled PID is never signalled. Bounded concurrency so a machine with hundreds of orphans does not take minutes to boot.
5. If a killed root still has a `chroma` registry row, drop it, so the successor's own registration starts clean.

Failure modes: unreadable table logs and returns; a kill failure is counted and the next root attempted. Nothing here can block boot or change behaviour when there are no orphans.

`claude-mem doctor` gains a read-only line: `N live chroma-mcp tree(s), M registered, K orphaned`, warning when K > 0. It parses `supervisor.json` directly rather than through `ProcessRegistry`, whose `initialize()` prunes and persists.

## Tests

`tests/supervisor/orphan-chroma-sweep.test.ts`, 11 cases on an injected process table: parser; tree grouping and root detection; PID-1 and dead-parent orphans; own children never flagged; bystander mentioning `chroma-mcp` in a filename never a candidate; the worker itself never a row; each orphan root killed exactly once and children not separately; a failed kill counted and the rest continued; registry row dropped only for a killed PID; Windows no-op; unreadable table skip; and one real `ps` read on the host with the kill stubbed.

Full suite: identical to `main` on this machine (same 100 fail / 74 errors before and after, all environment-dependent, none in touched modules). `tsc --noEmit`: the same 5 pre-existing errors in `install.ts` before and after. `lint:spawn-env` and `lint:hook-io` pass; the `ps` spawn passes no `env` block. Bundles rebuilt with `npm run build`; the four that embed the supervisor module changed.

Live on the author's machine (13.24.1, one owned pair): `doctor` reports `1 live chroma-mcp tree(s), 1 registered, 0 orphaned`.

## Known limits

- Signature-based discovery is a backstop by nature. Once #3302 and #3178 land, new orphans should be rare and this mostly finds the debt from before them.
- A chroma tree whose parent is a live process that is not a claude-mem worker (someone running `uvx chroma-mcp` by hand) is left alone, correctly; a tree whose worker is alive but wedged is also left alone, since the sweep judges ownership by liveness only.
- Windows is untouched; #3302's reasoning applies.


thedotmack:
Superseded by #3967 (rebased onto `main` with all four bundles rebuilt against current main including the quota and grammar fixes). Source changes are identical. Thank you @stephan-nb!

PR #3927, opened 2026-09-09
fix(smart-file-read): build each grammar once instead of recompiling per query

### LINKED ISSUE #3926: smart_outline/smart_search recompile the grammar on every call: `query -p` implies --rebuild (~700ms per language per call)
> **Disclosure — AI-authored.** This issue and the patch behind it were produced by an AI agent (Claude Opus 5 running in Claude Code) on my machine, at my request. The measurements below are real and reproducible, but the analysis has not been line-by-line audited by a human. Please read it with a grain of salt and treat any conclusion as a proposal, not a verified claim.

## Before submitting

- [x] I searched existing issues. Closest prior art is #3152 / #2910 / plan-16 (#3604), which cover the tree-sitter binary *not being installed*. This issue is about how that binary is *invoked* once it does exist, so it is a separate defect.

## What happens

`runBatchQuery()` in `src/services/smart-file-read/parser.ts` shells out as:

```
tree-sitter query -p <grammar-dir> <query-file> <source-files...>
```

Per `tree-sitter query --help`, `-p/--grammar-path` is documented as *"The path to the tree-sitter grammar directory, **implies --rebuild**"*. The CLI therefore recompiles the grammar from C on **every** invocation. The compiled artifact in the CLI's own cache is rewritten each time and never reused — its mtime changes on every run.

## Measurements

tree-sitter-cli 0.26.9, Linux x86_64, same query and same source file:

| invocation | duration |
| --- | --- |
| `query -p node_modules/tree-sitter-typescript/typescript …` | 685 ms |
| `query -l <prebuilt>.so --lang-name typescript …` | 7 ms |

End-to-end through the MCP server — `smart_outline` on a 933-line TypeScript file, driven over stdio JSON-RPC, **30 symbols returned in every run**:

| bundle | run 1 | run 2 | run 3 |
| --- | --- | --- | --- |
| current | 734 ms | 711 ms | 704 ms |
| with the patch in the linked PR | 733 ms | 16 ms | 12 ms |

The first call still pays one compile; every later call is a library load.

## Why it is worth fixing

`parseFilesBatch()` groups files by language and issues one CLI call per group, so the cost is a fixed ~0.7 s **per language per tool call**, independent of batch size. It is paid by `smart_outline`, `smart_search` and `smart_unfold` alike, and repeatedly by anything that walks a repository — `/learn-codebase` and `/smart-explore` pay it once per language per step.

It is also invisible: the grammar cache directory keeps getting written, which looks like caching is working.

## Suggested fix

Build each grammar once into `<data-dir>/tree-sitter-libs/<language>.<so|dylib|dll>` and query with `-l <lib> --lang-name <language>`, rebuilding only when the grammar sources (`src/parser.c`, `src/scanner.c`, `src/scanner.cc`) are newer than the artifact, and falling back to today's `-p` invocation whenever the build or the load fails.

Two details worth noting for review:

- The artifact belongs in the data dir, not in `node_modules`: an update replaces `node_modules` wholesale, and the installer owns those directories.
- `--lang-name` needs the grammar's language function name. I verified that the existing `GRAMMAR_PACKAGES` keys are correct for all 24 shipped grammars, including the irregular ones (`typescript`, `tsx`, `kotlin`, `php`, `scss`, `sql`, `markdown`), by building each and running a `(ERROR) @e` query against it. A mismatch would still be caught by the fallback rather than silently returning zero symbols.

rorar:
> **Disclosure — AI-authored.** This patch, its test and this description were written by an AI agent (Claude Opus 5 running in Claude Code) on my machine, at my request. Everything below was executed and observed rather than assumed, but no human has audited the change line by line. Please review it as a proposal from an unfamiliar contributor, with a grain of salt — particularly the platform assumptions, which I could only verify on Linux.

Fixes #3926.

## The defect

`runBatchQuery()` shells out with `tree-sitter query -p <grammar-dir>`. Per `tree-sitter query --help`, `-p/--grammar-path` *"implies --rebuild"*, so the CLI recompiled the grammar from C on **every** invocation. The compiled artifact in the CLI's own cache was rewritten each call and never read back — which makes the problem invisible, since the cache directory keeps filling up as if it were working.

`parseFilesBatch()` issues one CLI call per language group, so the cost was a fixed ~0.7 s **per language per tool call**, independent of batch size — paid by `smart_outline`, `smart_search` and `smart_unfold` alike, and repeatedly by anything that walks a repository.

## The change

Build each grammar once into `<data-dir>/tree-sitter-libs/<language>.<so|dylib|dll>`, then query with `-l <lib> --lang-name <language>`.

- **Location.** The data dir, not `node_modules`: an update replaces `node_modules` wholesale, and those directories belong to the installer.
- **Staleness.** An mtime comparison against `src/parser.c` and the optional scanner. npm and bun both stamp installed files with the install time, so a plugin update invalidates the previous artifact on its own — verified on both package managers rather than assumed. The check is re-run per call instead of memoized: four stats cost nothing next to the spawn they guard, and a memo would pin a long-lived MCP server to the grammar that was current at boot.
- **Fallback.** Every failure path returns to today's `-p` invocation — a build that will not compile, or a grammar whose language function is not named after our language key. The opt-out is remembered per language so a mismatch costs one extra spawn once, not two per batch forever.

## Verification

End-to-end through the built MCP bundle, driven over stdio JSON-RPC, `smart_outline` on a 933-line TypeScript file — **30 symbols returned identically in all runs**:

| bundle | run 1 | run 2 | run 3 |
| --- | --- | --- | --- |
| `main` | 734 ms | 711 ms | 704 ms |
| this branch | 733 ms | 16 ms | 12 ms |

Repeated against a 1665-line Python file through the installed plugin: 484 ms, then 12 ms and 14 ms, 36 symbols each time.

`--lang-name` correctness was checked for all 24 shipped grammars by building each one and running an `(ERROR) @e` query against it — including the irregular packages that ship no `tree-sitter.json` (`typescript`, `tsx`, `kotlin`, `php`, `scss`, `sql`, `markdown`). All bind under their existing `GRAMMAR_PACKAGES` key.

New test `tests/services/smart-file-read/grammar-lib-reuse.test.ts` covers artifact reuse and the stale-artifact rebuild. It fails on `main` (2 failures) and passes here. It skips itself when no tree-sitter binary is resolvable, since installs that ran with `--ignore-scripts` have none and would fail for reasons this change does not own.

`bun test tests`: 3051 pass, 27 skip, 1 fail.

## Two things I did not fix, flagged for honesty

Both reproduce on a clean checkout of `main` at fd0ecf0, so they are not from this branch — but you should know I saw them:

1. `npm run typecheck` reports 5 errors in `src/npx-cli/commands/install.ts` (the `p.multiselect` result is `symbol | ProviderChoice[]`). Since the repo intentionally ships no lockfile, a fresh `npm install` today resolves a newer `@clack/prompts` than CI last saw.
2. `tests/infrastructure/plugin-distribution.test.ts` fails at `packed[0].files` — `npm pack --json` output shape under npm 12.

## On the built bundle

`plugin/scripts/mcp-server.cjs` is included, matching what recent feature commits do (e.g. fd0ecf0). `plugin/scripts/worker-service.cjs` also changed on rebuild, but its entire diff was minifier identifier churn with no functional delta — `parser.ts` is not part of that bundle — so I reverted it rather than add 720 lines of noise. Rebuilding on your side may produce different mangled names than mine (bun 1.4.2); drop my bundle and regenerate if that is easier.


thedotmack:
Superseded by #3966 (rebased onto `main` with bundle rebuilt against current main including recent chroma merges). The source changes are identical. Thank you @rorar!

PR #3952, opened 2026-09-09
fix: prevent repeated stale worker bundle recycle storms

### LINKED ISSUE #3940: 13.24.2–13.24.5 on main bump plugin.json but the bundled worker still reports 13.24.1: version-mismatch kill loop on every hook (regression of #3857)
## Summary

Regression of #3857. `main` has bumped `plugin/.claude-plugin/plugin.json` four times since `v13.24.1` (13.24.2 → 13.24.5) and the marketplace serves `main` directly, but the bundled `plugin/scripts/worker-service.cjs` still has `"13.24.1"` baked in as the worker's reported version. The hook compares manifest version vs `/api/health` `version`, sees `13.24.5 != 13.24.1`, kills the worker, respawns it from the same file, and repeats on every hook event.

Unlike #3857 the bundle *was* rebuilt (`worker-service.cjs` +428/-359 since `v13.24.1`), so the build is embedding a stale version constant rather than shipping an untouched file.

## Evidence

Manifest bumps on `main` (each in a `ccs-align` PR, no release tag):

| commit | plugin.json |
|---|---|
| `e03cf5f7` (#3934) | 13.24.2 |
| `7abdf79c` (#3935) | 13.24.3 |
| `c2d023d7` (#3936) | 13.24.4 |
| `8bc631a7` (#3937) | 13.24.5 |

```
$ gh release list -R thedotmack/claude-mem --limit 1
v13.24.1  Latest  2026-09-05

$ curl -sL https://raw.githubusercontent.com/thedotmack/claude-mem/main/plugin/scripts/worker-service.cjs \
    | grep -oE '"13\.2[0-9]\.[0-9]+"' | sort | uniq -c
   4 "13.24.1"

$ gh api repos/thedotmack/claude-mem/compare/v13.24.1...main \
    --jq '.files[] | select(.filename|startswith("plugin/scripts/")) | "\(.filename) +\(.additions)/-\(.deletions)"'
plugin/scripts/worker-service.cjs +428/-359
plugin/scripts/server-service.cjs +136/-136
...
```

Installed cache on this machine (`~/.claude/plugins/cache/thedotmack/claude-mem/13.24.5`): `plugin.json` = 13.24.5, `package.json` = 13.24.5, `worker-service.cjs` grep = 4 × `"13.24.1"`. Same for the cached `13.24.4`.

Running worker:

```
$ curl -s http://127.0.0.1:37701/api/version
{"version":"13.24.1"}
```

Log (`~/.claude-mem/logs/claude-mem-2026-09-09.log`), repeating on every hook:

```
[SYSTEM] Worker version mismatch — killing stale worker {pluginVersion=13.24.5, workerVersion=13.24.1}
[SYSTEM] Worker not running — lazy-spawning {...}
[SYSTEM] Worker started {host=127.0.0.1, port=37701, pid=84521}
[SYSTEM] Worker is ready but still reports a stale version; not recycling again in this hook invocation (one recycle per hook event)
```

`grep -c 'Worker version mismatch'` per day on this machine: 09-03 1466, 09-04 1909 (13.24.0 era, #3857), 09-09 868 in under two hours after auto-updating to 13.24.5.

## Impact

- Every hook event SIGKILLs the worker mid-work. In-flight observer generation is lost (same as #3857).
- Chroma backfill never completes: `backfillAllProjects` is restarted from scratch on every respawn and killed a few seconds later.
- On 09-03 (13.24.0 loop) a kill landed while chroma-mcp was persisting the HNSW segment and left `~/.claude-mem/chroma/<uuid>/index_metadata.pickle` truncated (`EOF while parsing` at an offset equal to the file size). From then on every `chroma_add_documents` / `chroma_query_documents` failed with `Error deserializing pickle file`, only surfaced as `[ERROR] [CHROMA_SYNC] Batch add failed` in the log. Semantic search silently fell back to keyword search for six days; `/api/chroma/status?deep=1` was the only thing reporting `unhealthy`.

## Workaround

```
sed -i '' 's/"13\.24\.1"/"13.24.5"/g' ~/.claude/plugins/cache/thedotmack/claude-mem/13.24.5/scripts/worker-service.cjs
```

then stop the worker, `rm -rf ~/.claude-mem/chroma ~/.claude-mem/chroma-sync-state.json`, start it; with the loop gone the bootstrap backfill runs to completion.

## Suggested fix

- Have the bundle read its version from the sibling `plugin.json` / `package.json` at startup (or inject it at build time from the same file the manifest bump edits), so a manifest bump can never disagree with the worker.
- Add a build/CI check: `grep -c "\"$(jq -r .version plugin/.claude-plugin/plugin.json)\"" plugin/scripts/worker-service.cjs` must be non-zero before a version bump commit is allowed on `main`.
- Since the marketplace serves `main`, treat every `chore: bump version` as a release: either tag it or don't bump the manifest without rebuilding with the new version.

## Environment

- macOS 26 (Darwin 25.6.0), Bun runtime, Claude Code CLI
- claude-mem 13.24.5 from marketplace (auto-updated 2026-09-09), previously 13.24.4 with the same symptom

AntonioCoppe:
## What changed

Fixes #3940. Related: #3905.

Marketplace manifests advertise 13.24.5 while the committed worker reports 13.24.1. A hook kills that worker, launches the same stale bundle, and the next hook repeats the cycle. The existing warning only limits recycling within one hook invocation.

- Persist an unsuccessful version recycle across hook processes. Skip repeats for the same bundle/expected version/observed version; allow a new attempt when the bundle changes.
- Regenerate the shipped worker, MCP server, standalone server and transcript-watcher bundles from current source.
- Check all four committed bundle versions **before** CI rebuilds them, and fail on missing artifacts. Rebuilding first masks the packaging problem.
- Document committed release bundles and recovery behavior in the development guide.

## Why this fixes the reported behavior

Actual old/rebuilt bundle replay on macOS 26.6.2, Bun 1.3.12, with isolated data/config/ports, five fresh Claude Code context-hook processes and real Chroma tool calls:

| Observable result | Before | After |
|---|---:|---:|
| Manifest / worker health | 13.24.5 / 13.24.1 | 13.24.5 / 13.24.5 |
| Version-mismatch recycles | 5 | 0 |
| Lazy spawns after initial startup | 5 | 0 |
| Chroma connections including initial connection | 6 | 1 |
| Distinct worker / uv / Python PIDs | 6 / 6 / 6 | 1 / 1 / 1 |

After-fix PID chain `32520 -> 32717 -> 32722` remained unchanged through all five hooks; health stayed initialized/mcpReady and Chroma remained healthy. Separately, advertising 13.24.6 against the fixed 13.24.5 bundle produced one recycle followed by four skip warnings, proving the guard independently of correcting the version label. No worker-script override or mocked process spawning was used.

[Full timestamped spawning logs, PID samples, reproduction method and discussion follow-up](https://github.com/thedotmack/claude-mem/issues/3940#issuecomment-5609519116).

## Verification

- Normal pre-push checks on `643cb2e`: **3,114 passed, 27 skipped, 0 failed**; root/viewer type checks and build passed.
- Existing focused regression proves repeated hook invocations do not re-spawn an unchanged bad bundle, while a changed bundle permits retry.
- Expanded committed-version check failed on the stale standalone server bundle, then passed after regeneration: 13 passed.
- Separately enabled real Chroma worker-recycle orphan gate passed on macOS.
- Final diff check passed. Generated changelog untouched.

## Scope and remaining verification

These short runtime replays verify the restart mechanism and stable child processes. They do not establish a multi-hour memory soak, restored observation capture, OAuth recovery, semantic query correctness or full resolution of #3905. Chroma verification was a real tool-call health check. No orphan accumulation occurred in the bounded replay, and temporary processes were cleaned up afterward. Original user data was not used.

Upstream hosted checks are not yet reported for this fork PR. Maintainer review and approval remain pending.


thedotmack:
Superseded by #3961 (merged to main as `55afdace`). Same patch lineage (rebuild bundles + cross-hook recycle guard + pre-rebuild version-consistency). Thank you @AntonioCoppe — your work landed via the ship path.

thedotmack:
Superseded by #3961, which landed on main as `55afdace`. Released in v13.24.6.

Issue #3603, opened 2026-08-16
[plan-15] Worker Port & Liveness Authority — one verified answer to "is a worker serving this port?"


thedotmack:
# [plan-15] Worker Port & Liveness Authority — one verified answer to "is a worker serving this port?"

> **Tracker:** #3603 · **Design doc:** `plans/15-worker-port-and-liveness-authority.md` · **Series:** plan-14 #3602 · plan-15 #3603 · plan-16 #3604 · plan-17 #3605 · plan-18 #3606 · plan-19 #3607 · plan-20 #3608 · plan-21 #3609 · plan-22 #3610 · plan-23 #3611

## Defect

"Is the worker alive?" is answered by five independent, mutually inconsistent signals — `worker.pid`, `spawn.lock`, `chroma/.claude-mem-chroma-writer.lock`, `supervisor.json`, and "is the port in use?" — and none of them is verified against the process that actually owns the listening socket. So:

- The listening socket handle is inheritable; the chroma-mcp tree (and, under Bun on Windows, hook-spawned worker children) keeps `:37777` bound after the worker dies. The kernel completes TCP handshakes on the ghost listener, so probes see "port in use" and clients wait the full `CLAUDE_MEM_API_TIMEOUT_MS`.
- Every launcher treats "port in use" as proof of a live worker (`Port already in use, refusing to start duplicate`, `Worker PID file points to a live process, skipping duplicate spawn`, `Another launcher holds the spawn lock`) and never reclaims: `spawn.lock` has no holder-liveness check, `validateWorkerPidFile` deletes the pid file so `ensureWorkerRunning` can no longer identify the owner, the `EADDRINUSE` branch `exit(1)`s with no remediation, `pidFileStatus === 'alive'` returns `warming` forever.
- Health probes have no deadline (`HealthMonitor.httpRequestToWorker()` `fetch()` without `AbortSignal`, `isPortInUse()` same), so a wedged-but-bound worker (DarkWake 100 % CPU, CLI-probe hang, stale writer lock, unwritable data dir) turns every hook into a 30–60 s stall and every launcher into a pile-up (296+ hook clients, 16+ bun.exe).
[...]

PR #3989, opened 2026-09-11
fix(windows): bound the port-occupancy probe so a ghost listener cannot hang the launcher

weiconghe:
## Summary

Follow-up to #3900 (merged), which left main's Windows job red: the gate it added (`Ghost-listener recovery gate (#3603)`) failed in 12 of 12 runs. There were **two** defects behind that, and this PR fixes both.

**1. `isPortInUse()`'s Windows fast path was an unbounded HTTP probe** — the one probe #3900 left without a timeout. A ghost listener completes the TCP handshake and never answers, so the fetch never settled; `ensureWorkerStarted()` calls it *before* the reclaim, so the reclaim could never fire.

**2. The gate never got past its own fixture handshake on CI.** The fixture reported `ready` on its redirected stdout, and that line stayed invisible to the test until the process was killed at the 600s cap — every failing run shows the same shape: no stage output, exactly `2 expect() calls`, and the fixture's chroma chain already alive 3s in.

Section 2 was found by section 1's diagnostics. The first revision of this PR fixed the probe (verified below) and assumed it explained the CI hang; the stage log added in that revision disproved that on the next run — the test never reached `ensureWorkerStarted()`. A redirected stdout can hold a single small line unflushed in a process that then idles, so the gate's readiness handshake must not depend on it.

## What does this PR do?

1. **`src/services/infrastructure/HealthMonitor.ts`** — the Windows fast-path fetch carries the same `HEALTH_PROBE_TIMEOUT_MS` (5s) abort budget as every other probe in the module. On timeout the existing fall-through runs the socket probe (`net.createServer` → `EADDRINUSE`), which still reports the bound port as in use, so the launcher proceeds to the reclaim. No behavior change for a healthy worker or a foreign process holding the port.

2. **`tests/integration/fixtures/ghost-worker-host.ts`** — every event is appended to an events file (`fs.appendFileSync`, a syscall per call) as well as stdout. The stdout copy stays for human debugging; the file is the contract. It also emits `progress` stages (`port-bound`, `create-collection-start/done`, `add-documents-done`) with elapsed times.

3. **`tests/integration/worker-ghost-port-recovery.test.ts`**
   - reads readiness from the events file;
   - logs a `[ghost-gate] waiting for fixture ready: elapsed=… lastStage=…` line every 15s, so the timeline is in the CI log even if bun's timeout kills the run;
   - fails at 240s (below bun's 600s cap) with the full events/stdout/stderr dump;
   - races `ensureWorkerStarted()` against a 300s deadline and reaps the in-flight call during teardown instead of leaving it running past cleanup (Greptile P2).

4. **`tests/infrastructure/health-monitor.test.ts`** — regression test: the Windows probe must carry an `AbortSignal`, and an abort must fall through to the socket probe. Restores the net mock on the failure path (Greptile P2).

## How verified?

**Isolated probe (Windows 11, bun 1.3.6)** — a listener that accepts connections and never responds (the ghost shape), driving `isPortInUse()` directly:

| | result |
|---|---|
| before | never settles — watchdog: `HANG CONFIRMED after 20000ms` |
| after | returns `true` after **5047ms** (5s abort → socket probe → `EADDRINUSE`) |

**The gate, A/B on the same machine** (`CLAUDE_MEM_TEST_CHROMA=1`, isolated `CLAUDE_MEM_DATA_DIR`, real chroma chain, port isolated from the developer's live worker):

- probe fix reverted: `ensureWorkerStarted` never returns; the deadline reports `timed out waiting for: ensureWorkerStarted` at 318s. (318s exceeds the sum of every bounded stage on that path — ≤6s + ≤6s health waits, ~10s reclaim, 60s readiness.)
- current revision: passes in 43.7s, with the new stage timeline —

```
[ghost-gate] fixture ready after 8208ms: {"pid":55612,"port":47777,"chromaRootPid":67688}
[ghost-gate] ghost confirmed: port LISTENING under dead pid=55612; survivors: conhost.exe, chroma-mcp.exe, python.exe, python.exe
[ghost-gate] driving production ensureWorkerStarted() on port 47777
[ghost-gate] ensureWorkerStarted returned 'ready' after 27506ms
 1 pass, 0 fail | Ran 1 test [43.69s]
```

**Unit + types**: new test passes; `bun run typecheck` clean. (Three unrelated tests in `health-monitor.test.ts` fail on this dev machine only — they do real network probes and a live worker holds 37777 here; they pass where the port is free.)

**CI**: the `chroma lifecycle · worker-recycle orphan gate` job is the arbiter — it is the only environment where the handshake failure reproduces, and it reproduces every time. If it still fails, the stage line it prints names the stage.

## Out of scope

In the newest main run (`2714e869`) a different step failed once: `Worker-recycle orphan gate (#3482)` reported `csrss.exe / winlogon.exe / fontdrvhost.exe / dwm.exe` as orphaned descendants of its fixture. That step passed in the other 12 post-merge runs with this same code, and neither its test nor its enumeration helper is touched by #3900 — it reads like a rare bad process-table read (`Get-CimInstance` under churn). Flagging rather than fixing it here.

## Checklist

- [x] Tests added/updated and passing
- [x] Typecheck passing

### DIFF
--- a/src/services/infrastructure/HealthMonitor.ts
+++ b/src/services/infrastructure/HealthMonitor.ts
@@ -45,8 +45,19 @@ export async function isPortInUse(port: number): Promise<boolean> {
     // Fast path: HTTP health check. A live claude-mem worker responds to
     // /api/health, so this is the cheapest non-disruptive probe for the
     // common case (worker is running and healthy).
+    //
+    // Bounded like every other probe (HEALTH_PROBE_TIMEOUT_MS): a ghost
+    // listener — the dead worker's inherited socket, held open by its chroma
+    // sidecar chain (plan-15 #3603) — completes the TCP handshake and then
+    // never answers. Unbounded, this fetch would hang forever, and with it
+    // ensureWorkerStarted(), which calls this BEFORE it can reach the reclaim:
+    // the very bug the reclaim exists to fix would instead wedge every
+    // launcher. On timeout the flow falls through to the socket probe below,
+    // which still reports a bound port as in use.
     try {
-      const response = await fetch(`http://${formatHostForUrl(getWorkerHost())}:${port}/api/health`);
+      const response = await fetch(`http://${formatHostForUrl(getWorkerHost())}:${port}/api/health`, {
+        signal: AbortSignal.timeout(HEALTH_PROBE_TIMEOUT_MS),
+      });
       if (response.ok) return true;
       // Non-ok response: port is reachable but the worker is unhealthy.
       // Fall through to the net.createServer check below so we still report
--- a/tests/infrastructure/health-monitor.test.ts
+++ b/tests/infrastructure/health-monitor.test.ts
@@ -141,6 +141,50 @@ describe('HealthMonitor', () => {
       }
     });
 
+    it('should probe Windows health through an abortable signal so a ghost listener cannot hang it (#3603)', async () => {
+      const origPlatform = process.platform;
+      try {
+        Object.defineProperty(process, 'platform', { value: 'win32', configurable: true });
+
+        // A ghost listener completes the TCP handshake and never answers, so a
+        // probe without its own abort budget never settles — and
+        // ensureWorkerStarted(), which runs this check BEFORE it can reach the
+        // ghost reclaim, hangs with it. The abort signal is the fix: capture
+        // it off the call, and model the abort as the rejection it produces.
+        const inits: Array<RequestInit | undefined> = [];
+        const fetchMock = mock((_url: string, init?: RequestInit) => {
+          inits.push(init);
+          const abortError = new Error('The operation was aborted due to timeout');
+          abortError.name = 'TimeoutError';
+          return Promise.reject(abortError);
+        });
+        global.fetch = fetchMock as any;
+
+        const createServerMock = mock(() => ({
+          once: mock((event: string, cb: Function) => {
+            if (event === 'error') setTimeout(() => cb({ code: 'EADDRINUSE' }), 0);
+          }),
+          listen: mock(() => {}),
+        }));
+
+        const netSpy = spyOn(net, 'createServer').mockImplementation(createServerMock as any);
+
+        const result = await isPortInUse(37777);
+
+        expect(inits.length).toBeGreaterThan(0);
+        expect(inits[0]?.signal).toBeInstanceOf(AbortSignal);
+        // An aborted probe is inconclusive, never "free": the flow falls
+        // through to the socket probe, which reports the bound port as in use
+        // so the launcher can go on to reclaim the ghost.
+        expect(result).toBe(true);
+        expect(net.createServer).toHaveBeenCalled();
+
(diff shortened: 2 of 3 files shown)


thedotmack:
**issue-patch-ship gate: PASS** (OSS queue sweep)

1. RCA-later: logged in PR body — `#3900` left `isPortInUse()` Windows HTTP probe unbounded; ghost listeners accept TCP and never answer, so launchers hang before reclaim. Deeper port/liveness authority remains plan-15 / #3603.
2. No second-system: same 5s abort budget already used by other HealthMonitor probes; timeout falls through to existing socket probe.
3. Not opinionated/wide: +89/−2 in HealthMonitor + ghost-gate diagnostics + unit pin.

Will auto-merge when required checks go green (esp. Windows `chroma lifecycle · worker-recycle orphan gate`).

weiconghe:
Pushed a revision that addresses both P2 findings and corrects my diagnosis of the CI hang.

**Both P2s fixed:**
- *Deadline leaves launcher running* — the in-flight `ensureWorkerStarted()` is now tracked and awaited (bounded, 15s) during teardown, so a late reclaim/spawn is reaped before cleanup instead of outliving it.
- *Restore spy on failure* — `netSpy.mockRestore()` moved into the `finally` next to the platform restore.

**Diagnosis correction:** the probe below was a real defect (A/B proven), but it is not what hung CI. The stage log added in the previous revision showed the next CI run never reached it: exactly `2 expect() calls`, no stage output, and the fixture's `ready` line only visible at the 600s cap while its chroma chain was already alive 3s in. The handshake read a redirected stdout, which can hold a single small line unflushed in a process that then idles. The fixture now appends every event to an events file (syscall per call, nothing to flush), and the test logs `lastStage` every 15s, dumps events/stdout/stderr when it gives up at 240s, and no longer races bun's own cap.

weiconghe:
One correction to the RCA, from the CI run of this revision (started 03:35Z, before the PR was closed): the landed probe fix is real and needed, but it is **not** what was hanging the gate. That run shows the gate never reached it —

- no stage output at all, exactly `2 expect() calls` (the snapshot asserts, nothing after);
- the fixture's `ready` line became visible only at the 600s cap, at the moment the fixture was killed;
- the postmortem step shows the fixture's chroma chain alive 3s in, so it was not a slow chroma start either.

The cause is the gate's own readiness handshake: the fixture reported `ready` on a redirected stdout, and a single small line written by a process that then idles can sit unflushed in that buffer indefinitely (CI's bun 1.4.2; it does not reproduce on bun 1.3.6 locally, which is why the local runs passed). So `155fefce` will keep the Windows job red until that handshake is fixed.

Fix + diagnostics (events file the fixture appends to, a stage line every 15s, and a 240s failure that dumps events/stdout/stderr instead of racing bun's cap) are in **#3992**. The two review findings on this PR are addressed there too (in-flight `ensureWorkerStarted()` reaped during teardown; the net mock restored on the failure path).

Issue #3628, opened 2026-08-18
ON UPDATE CASCADE writes NULL into session_summaries/observations at SDK generator start (13.15.2, distinct from #2283)


alexderkrieger:
## Summary

`ON UPDATE CASCADE` turns the SDK generator's own start-up write into a
`NOT NULL` violation, so any session that has already stored rows stops recording
permanently.

This is the same error as #2283, but a different code path. The guard that closed
#2283 (`if (!e.memorySessionId) { defer storage; return }`) is present and working
in 13.15.2 and correctly protects the INSERT path. The failure below never reaches
an INSERT. #1952 identified `ON UPDATE CASCADE` on these FKs as a defect and was
closed during a backlog consolidation without removing it.

Reproduced on 13.15.2 (latest on npm). The same code is present in 13.13.1,
13.14.0 and 13.15.0, so this is not a recent regression.

## Mechanism

Both child tables declare the column `NOT NULL` and cascade on update:

```sql
memory_session_id TEXT NOT NULL,
FOREIGN KEY(memory_session_id) REFERENCES sdk_sessions(memory_session_id)
  ON DELETE CASCADE ON UPDATE CASCADE
```

`scripts/worker-service.cjs`, in the Claude SDK agent's `startSession()`:

```js
let c = this.createMessageGenerator(e, n, a);
e.memorySessionId && (
  this.dbManager.getSessionStore().updateMemorySessionId(e.sessionDbId, null),
  e.memorySessionId = null
);
```

`updateMemorySessionId(id, null)` runs `UPDATE sdk_sessions SET memory_session_id
= NULL WHERE id = ?`. With `PRAGMA foreign_keys = ON` that NULL cascades into every
child row and violates their `NOT NULL` column. The transaction rolls back, the
generator throws, the session is dropped.

## Minimal reproduction

```sql
PRAGMA foreign_keys=ON;
CREATE TABLE sdk_sessions (
  id INTEGER PRIMARY KEY, content_session_id TEXT NOT NULL,
  memory_session_id TEXT UNIQUE, project TEXT NOT NULL,
  started_at TEXT NOT NULL, started_at_epoch INTEGER NOT NULL);
CREATE TABLE session_summaries (
  id INTEGER PRIMARY KEY, memory_session_id TEXT NOT NULL,
  project TEXT NOT NULL, created_at TEXT NOT NULL, created_at_epoch INTEGER NOT NULL,
[...]

PR #4024, opened 2026-09-11
fix(sync): register a memory_session_id, do not re-register it

vorrawut:
## What

`ensureMemorySessionIdRegistered` writes whenever the offered id **differs** from the stored one, and its busiest caller offers a different one every turn.

`memory_session_id` is not an ordinary column. It is the FK parent key of `observations` and `session_summaries` (`ON UPDATE CASCADE`) and the join field `requeuePromptSync` pushes to replicas — so overwriting it rewrites every memory the session owns and re-enqueues every prompt it has.

`ClaudeProvider` starts a fresh SDK process per `query()`, which mints a new `session_id`, and `resetCarriedMemorySessionId` clears the in-memory copy before each one. Every turn therefore arrives here looking like a new identity, and `requeuePromptSync` — whose own docstring describes a one-time repair, *"Once the mapping lands"* — runs once per turn per prompt.

## Measured

On one real store, before the change:

| | |
|---|---|
| `sync_outbox` rows | **1,100,783** |
| distinct prompts behind them | **6,930** |
| amplification | **158.8x** |
| share of the database | 393 MB of 854 MB |
| worst single prompt | 3,464 rows carrying **3,464 distinct** `memory_session_id`s |
| prompts over 1,000 rows each | 226, totalling 344,377 rows (31% of the table) |

Per-prompt distribution: 2,882 under 10 ops · 1,769 at 10–99 · 2,053 at 100–999 · 226 over 1,000.

Every op was `set_prompt_session`. Nothing drains the table on a device that has never paired with a sync hub (`sync_state` empty, `sync_entity_heads` 0), so it only grows.

## Why nothing needs a later turn's id

`shouldResume` in `ClaudeProvider` is a hardcoded `false`, and every read of it (the log line, `resume_parameter`, the `[ALIGNMENT]` debug line, `hasStaleMemoryId`, and the spread at the `query()` call) is either logging or a ternary that can only take the false branch. `resume` never receives the id. The remaining readers are log lines and `buildSummaryPrompt` metadata.

## The fix

Write only when the stored id is `NULL`. *Ensure registered* means make sure one exists.

A deliberate change of identity is still available through `updateMemorySessionId`, which is what a caller that genuinely means to re-identify a session should call. A refused re-registration is logged at `debug` rather than silently dropped.

## Relationship to #3628

That PR stopped the same caller writing `NULL` over the id, which cascaded `NULL` into a `NOT NULL` child and took the whole generator run down with it:

```
[ERROR] [SESSION] [session-3529] Generator failed
  {provider=claude, error=NOT NULL constraint failed: session_summaries.memory_session_id}
```

5,407 occurrences across one machine's log history (2,904 `session_summaries`, 2,503 `observations`); each one an observation or summary that was never written. This PR is the other half — the write that **succeeds**, and costs.

## Tests

`tests/worker/sync/session-identity-is-stable.test.ts`, 5 cases against a real database:

- a missing id is registered and each prompt repaired once
- re-registering the same id is free
- ten different ids across ten turns change nothing — the regression
- `updateMemorySessionId` still works, and the test pins what it costs
- a session with no prompts emits nothing

**Falsified:** with the change reverted, *"refuses to re-register a session that already has an identity"* fails and the other four pass.

## Verification

- `tsc --noEmit` — clean
- `bun test tests/worker/sync/` — **150 pass / 1 fail / 1 error** on pristine `main`, **155 pass / 1 fail / 1 error** with this branch. The `adoptMergedWorktrees` timeout and the unhandled error reproduce identically on an untouched checkout; the +5 are the new tests.

### DIFF
--- a/src/services/sqlite/SessionStore.ts
+++ b/src/services/sqlite/SessionStore.ts
@@ -2161,7 +2161,27 @@ export class SessionStore {
       throw new Error(`Session ${sessionDbId} not found in sdk_sessions`);
     }
 
-    if (session.memory_session_id !== memorySessionId) {
+    // REGISTER, DO NOT RE-REGISTER. `memory_session_id` is the FK parent key of
+    // `observations` and `session_summaries` (ON UPDATE CASCADE) and the join
+    // field `requeuePromptSync` pushes to replicas, so overwriting it is not a
+    // field update — it rewrites every memory the session owns and re-enqueues
+    // every prompt it has.
+    //
+    // The caller that made this matter is ClaudeProvider: a fresh SDK process
+    // mints a new session_id every turn, `resetCarriedMemorySessionId` clears the
+    // in-memory copy before each one, and nothing consumes a later turn's id
+    // (`shouldResume` is a hardcoded false, so `resume` never receives it). The
+    // condition below used to be `!==`, so every turn looked like a new identity.
+    //
+    // MEASURED on one store: sync_outbox held 1,100,783 rows for 6,930 distinct
+    // prompts — 158.8x, 393 MB of an 854 MB database — with its worst single
+    // prompt carrying 3,464 rows and 3,464 DISTINCT memory_session_ids. That is
+    // `requeuePromptSync`, whose own docstring describes a one-time repair
+    // ("Once the mapping lands"), running once per turn per prompt instead.
+    //
+    // A deliberate change of identity is still available through
+    // `updateMemorySessionId`. "Ensure registered" means make sure one exists.
+    if (session.memory_session_id === null) {
       this.db.prepare(`
         UPDATE sdk_sessions SET memory_session_id = ? WHERE id = ?
       `).run(memorySessionId, sessionDbId);
@@ -2169,8 +2189,13 @@ export class SessionStore {
 
       logger.info('DB', 'Registered memory_session_id before storage (FK fix)', {
         sessionDbId,
-        oldId: session.memory_session_id,
         newId: memorySessionId
+      });
+    } else if (session.memory_session_id !== memorySessionId) {
+      logger.debug('DB', 'Keeping the registered memory_session_id', {
+        sessionDbId,
+        registered: session.memory_session_id,
+        offered: memorySessionId
       });
     }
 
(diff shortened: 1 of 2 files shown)


thedotmack:
NIGHT SHIP after Batch H — issue-patch-ship: Gate **PASS**.

RCA is in this PR (1.1M `sync_outbox` rows / 158× prompt amplification from per-turn re-registration). Narrow two-file patch; the other half of shipped #3629/#3628.

Fork workflows are `action_required` (first-time contributor), so CI cannot go green here. Same-repo rehost for required checks: #4025. Will squash-merge that one if CI is green, then close this PR pointing at the merge SHA.

thedotmack:
NIGHT SHIP after Batch H — issue-patch-ship: Gate **FAIL** after CI.

`typecheck · build · test · bundle-size` failed: existing tests treat `ensureMemorySessionIdRegistered` as the **supported re-key path** (fk-constraint-fix, session_id_usage_validation, v7 CASCADE rebuild, synced-at). ClaudeProvider even documents that: “Legitimate re-keying flows through ensureMemorySessionIdRegistered.”

The outbox-amplification RCA is real, but refusing every non-NULL overwrite here is too coarse — it breaks worker-restart re-key and CASCADE rotation. The narrow fix belongs at the ClaudeProvider call site (don’t offer a fresh id every turn), not by changing this function’s contract.

Leaving #4025 draft / unmerged. No version bump.

thedotmack:
Follow-up with the 8 Linux CI test rewrites (NULL-only register kept; deliberate re-key now goes through `updateMemorySessionId`): same-repo branch `cursor/session-identity-stable-3ab4`. Will squash-merge that PR if CI is green, then close this one as superseded with the merge SHA.

thedotmack:
Superseded by #4027, squash-merged to main as `7d3f0047c5648841daf785a2f40c0b57f6069831`.

Your NULL-only register fix landed as-is. The 8 Linux CI failures were a polarity flip (keep the first registered id; drive deliberate re-key through `updateMemorySessionId`), not a restore of `!==` overwrite. Credit stands.

PR #4064, opened 2026-09-12
fix(logger): recompute the log filename when the UTC date changes

### LINKED ISSUE #4063: Logger latches its log filename, so long-lived processes write to a stale file
# Logger latches its log filename, so long-lived processes write to a stale file

## What happens

`Logger.ensureLogFileInitialized()` (`src/utils/logger.ts:90`) computes the log filename once and
guards it behind a boolean latch:

```ts
private ensureLogFileInitialized(): void {
  if (this.logFileInitialized) return;
  this.logFileInitialized = true;
  try {
    const logsDir = paths.logsDir();
    if (!existsSync(logsDir)) mkdirSync(logsDir, { recursive: true });
    const date = new Date().toISOString().split('T')[0];
    this.logFilePath = join(logsDir, `claude-mem-${date}.log`);
  } catch ...
}
```

The date is read inside the latch, so `logFilePath` is frozen at the day the process started. The
worker daemon runs for days. It therefore keeps appending to the file named for whichever UTC day it
happened to start on, while every line it writes carries the current timestamp.

## Why it is easy to miss

Short-lived hook processes recompute the filename on each spawn, so a correctly named file for today
also exists and grows. The log directory looks right. The tell is **two files with recent mtimes**.

Measured on one machine, `~/.claude-mem/logs/`:

```
claude-mem-2026-09-05.log   10,479,949 bytes   last written Sep 7 21:37   <- 2 days of drift
claude-mem-2026-09-06.log    2,848,654 bytes   last written Sep 7 14:06
...
claude-mem-2026-09-11.log      609,167 bytes   last written Sep 12 01:19
claude-mem-2026-09-12.log      319,215 bytes   last written Sep 12 12:33
```

The 2026-09-05 file is the largest in the directory because the daemon that opened it never let go.

## A caveat for anyone verifying this

`toISOString()` is UTC while the log *lines* are stamped in local time. On a UTC+2 machine the
filename rolls at 02:00 local, so between midnight and 02:00 a file correctly named with yesterday's
UTC date legitimately holds lines dated today. Comparing a filename against the local date reports a
false positive for two hours every night. Compare against `date -u +%F`.

## Fix

Compute the date before consulting the latch and compare it, so the path is recomputed on the first
write after UTC midnight. Patch attached / PR to follow. It is 8 lines including the comment.

## Scope

The compiled form of this method appears in 7 bundles per release (5 under `scripts/`, plus
`sqlite/SessionStore.js` and `sqlite/observations/files.js`), so a local workaround has to patch all
seven. Present in v13.15.3 and still present in v13.24.23.

## Related: the log directory is never pruned

`rotateLog`, `pruneLogs`, `maxLogFiles` and `retentionDays` return no matches anywhere in the source.
The same directory above reached 73 MB before being cleared by hand. Fixing rotation makes retention
matter more, not less. Happy to open that as a separate issue if it is wanted.

trencho:
Fixes #4063.

`Logger.ensureLogFileInitialized()` computed the date **inside** the latch, so `logFilePath` was
frozen at the day the process started. The worker daemon runs for days, so it kept appending to the
file named for whichever UTC day it happened to start on, while every line it wrote carried the
current timestamp.

Short-lived hook processes recompute per spawn, so a correctly named file for today also exists and
grows, and the log directory looks right. **The tell is two files with recent mtimes.**

Measured on one machine: `claude-mem-2026-09-05.log` was still being written on Sep 7 and was by far
the largest file in the directory at 10.5 MB, because the daemon that opened it never let go.

## The change

8 lines including the comment. The date is computed before the latch is consulted and compared
against it, so the path is recomputed on the first write after UTC midnight.

```ts
const date = new Date().toISOString().split('T')[0];
if (this.logFileInitialized && this.logFileDate === date) return;
this.logFileInitialized = true;
this.logFileDate = date;
```

The early return is kept rather than restructured, so the diff is as small as the fix.

## A caveat for verifying it

`toISOString()` is UTC while the log *lines* are stamped in local time. On a UTC+2 machine the
filename rolls at 02:00 local, so between midnight and 02:00 a file correctly named with yesterday's
UTC date legitimately holds lines dated today. **Comparing a filename against the local date reports a
false positive for two hours every night.** Compare against `date -u +%F`.

## How this was checked

Not by looking at the log directory once after a restart: a fixed latch and a freshly latched broken
one both produce a file named with today's date. Both methods were extracted from the built bundles
and run against an advanced clock, asserting that the **original still latches** and the patched one
rolls. Asserting only the second half leaves you unable to tell a working fix from a harness that
never exercised the guard.

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -81,6 +81,7 @@ class Logger {
   private useColor: boolean;
   private logFilePath: string | null = null;
   private logFileInitialized: boolean = false;
+  private logFileDate: string | null = null;
 
   constructor() {
     this.useColor = process.stdout.isTTY ?? false;
@@ -88,8 +89,14 @@ class Logger {
   }
 
   private ensureLogFileInitialized(): void {
-    if (this.logFileInitialized) return;
+    // The date is computed BEFORE the latch is consulted, so a long-lived process rolls onto a new
+    // log file at UTC midnight. Latching on the boolean alone freezes logFilePath at the day the
+    // process started, and the daemon then writes entries stamped with today's date into a file
+    // named for a previous day.
+    const date = new Date().toISOString().split('T')[0];
+    if (this.logFileInitialized && this.logFileDate === date) return;
     this.logFileInitialized = true;
+    this.logFileDate = date;
 
     try {
       const logsDir = paths.logsDir();
@@ -98,7 +105,6 @@ class Logger {
         mkdirSync(logsDir, { recursive: true });
       }
 
-      const date = new Date().toISOString().split('T')[0];
       this.logFilePath = join(logsDir, `claude-mem-${date}.log`);
     } catch (error: unknown) {
       console.error('[LOGGER] Failed to initialize log file:', error instanceof Error ? error.message : String(error));


thedotmack:
Routing this to **#3603 (plan-15, Worker Port & Liveness Authority)** — the log-filename latch and rotation is called out explicitly in that master. The bug is real and still on `main`: `ensureLogFileInitialized()` computes the date after the `logFileInitialized` guard, so a long-lived worker daemon pins `logFilePath` to whatever UTC day it booted on while every line it writes carries today's timestamp.

Not a duplicate of anything open and nothing on `main` has fixed it, so it stays open. 8 lines against a daemon that runs for days is the right shape for this cluster.

PR #4101, opened 2026-09-17
fix(worker): use latest user_prompts row on session rehydration

### LINKED ISSUE #4047: [Bug] Observer <user_request> falls back to the session's FIRST prompt on rehydration (sdk_sessions.user_prompt never updated) → stale, duplicated decision observations
## Summary

When a session is **rehydrated from the database** (worker restart, or the session evicted from memory), the observer's `<user_request>` falls back to `sdk_sessions.user_prompt`. That column holds the session's **first** prompt and is never updated. `lastPromptNumber` is loaded from `user_prompts` and is current, so the observer gets the right prompt number with the wrong request text. With a small local observer model, this produced a feedback loop of stale, near-duplicate `decision` observations restating prompt #1 long after the user had changed course.

## Environment

- macOS, Claude Code, marketplace install (running worker 13.24.5; the bug is confirmed still present on `main` @ `ae7053b` / 13.24.19)
- Provider: `openrouter` settings pointed at a local OpenAI-compatible Ollama endpoint (`qwen2.5:14b-instruct`, 32k context). The code path is provider-independent.

## Where

In the worker's session initialization, when the session is not in memory:

```js
let c = r || a.user_prompt;   // r = currentUserPrompt (often absent for observation-triggered starts)
// ...
userPrompt: c,
lastPromptNumber: n || this.dbManager.getSessionStore().getPromptNumberFromUserPrompts(a.content_session_id, e),
```

These log lines anchor it:
- `Discarding stale memory_session_id from previous worker instance (Issue #817)`
- `No currentUserPrompt provided for new session, using database`

Both the init prompt and the continuation prompt then embed `e.userPrompt` as `<observed_from_primary_session><user_request>…</user_request>`.

## Impact observed (one day, one session, 11 user prompts)

- 116 × `Generator auto-starting … historyLength=0` and 118 × `Briefed the observer generation with session-start context`. This was amplified by the #3940-style version-mismatch kill loop: 169 × `Worker version mismatch — killing stale worker {pluginVersion=13.24.5, workerVersion=13.24.1}`.
- On every restart the observer got `<user_request>` = prompt #1, plus a briefing that already contained the previous stale observations. It re-recorded prompt #1's intent ("use tool X with model A") as a fresh `decision` through prompt #11, even though the user switched to model B at prompt #4.
- **47 of ~70 observations** that day were stale restatements (titles such as "Leveraging X and model A for Efficiency" ×20+). One was a false `bugfix` ("duplicate decision records fixed") written before anything was fixed. 3 session summaries also carried the stale `request`.
- Every observation still carried the correct `prompt_number`, which made the stale content look legitimate.

## Expected

On rehydration, `<user_request>` should be the latest prompt for the session, i.e. the same source that `lastPromptNumber` comes from.

## Suggested fix

Either:
1. when `currentUserPrompt` is absent, load the latest `user_prompts.prompt_text` for the session instead of `sdk_sessions.user_prompt`; or
2. update `sdk_sessions.user_prompt` on every UserPromptSubmit (or rename the column to make its "first prompt" meaning explicit and stop using it as the current request).

## Workaround (verified red/green on a DB copy, then applied)

```sql
CREATE TRIGGER IF NOT EXISTS local_fix_session_latest_user_prompt
AFTER INSERT ON user_prompts
WHEN NEW.prompt_text IS NOT NULL AND length(trim(NEW.prompt_text)) > 0
BEGIN
  UPDATE sdk_sessions SET user_prompt = NEW.prompt_text WHERE id = NEW.session_db_id;
END;
```

After the workaround and a one-off backfill of the affected session, new observations immediately tracked the current prompt, and session-start context contained none of the stale entries. (Caveat for maintainers: a user-added trigger on `user_prompts` that references `sdk_sessions` could interfere with the table-rebuild migrations, `DROP TABLE` + `RENAME`, so a real fix belongs in the worker.)

## Related

- #817: same class of bug; DB-loaded session state (`lastPromptNumber`) bypassed the resume guard. `userPrompt` has the same problem.
- #3940 / #3961: the version-mismatch recycle storm multiplied the rehydrations that trigger this.
- #3606: observer response pipeline plan. The init-reply change in 13.24.9 reduces, but does not remove, the stale `<user_request>` input.

nero-harsh:
## Problem

When a session is rehydrated in `SessionManager.initializeSession()` (e.g. after a worker restart, when the session is no longer in the in-memory map) and no `currentUserPrompt` is passed in, the code falls back to `dbSession.user_prompt`, which comes from `sdk_sessions.user_prompt`. That column is only ever written once, at session creation, so it always holds the first prompt of the conversation.

Every subsequent prompt is recorded in the `user_prompts` table instead. On rehydration the session ends up labeled with the very first prompt of the conversation instead of the most recent one (issue #4047).

## Fix

On rehydration, look up the latest row from `user_prompts` via the existing `SessionStore.getLatestUserPrompt()` and prefer it over the stale `sdk_sessions.user_prompt`, only falling back to the old column when `user_prompts` has no rows yet (e.g. a session with only its first prompt).

`lastPromptNumber` computation is unchanged.

## Testing

Added `tests/session-manager-rehydration-prompt.test.ts` covering:
- rehydration with multiple prompts recorded picks the latest one, not the first
- rehydration with only the first prompt recorded still falls back correctly

Also updated the mock `getSessionStore()` in `tests/worker/session-manager-null-prompt.test.ts` to include the new `getLatestUserPrompt` call so the existing null-prompt tests keep passing.

Fixes #4047


harshith-Cognitbotz:
Closing, this was already fixed on main in 7bfc401 (fix(session): use latest user_prompts text on rehydration, closes #4047), which landed before I opened this PR. Same root cause and same fix direction (load the latest user_prompts row on rehydration instead of falling back to sdk_sessions.user_prompt), just a different implementation. No action needed here.

nero-harsh:
Closing, this was already fixed on main in 7bfc401 (fix(session): use latest user_prompts text on rehydration, closes #4047), which landed before I opened this PR. Same root cause and same fix direction (load the latest user_prompts row on rehydration instead of falling back to sdk_sessions.user_prompt), just a different implementation. No action needed here.

PR #4103, opened 2026-09-17
fix(context): stop hiding the last summary when it ties the newest observation

nero-harsh:
## Problem

`shouldShowSummary` in `src/services/context/sections/SummaryRenderer.ts` hides the last session summary whenever:

```ts
mostRecentSummary.created_at_epoch <= mostRecentObservation.created_at_epoch
```

`SessionStore.storeObservations()` computes one `timestampEpoch` per call and stamps every observation row and the session's `session_summaries` row with that same value in a single insert transaction (`src/services/sqlite/SessionStore.ts`, around line 3137). That means a session's final observations and its own summary routinely share the exact same `created_at_epoch`.

`ContextBuilder.buildContextOutput()` calls `shouldShowSummary` with `summaries[0]` and `observations[0]`, both newest-first, so on the very next session's context injection, the tied case is the common case, not an edge case. The `<=` treats a tie as "newer work exists after the summary" and suppresses it, even though nothing happened after the summary was written, it was written in the same batch as the observation next to it.

## Fix

Change the comparison to `<`, so the summary is only hidden when an observation is strictly newer than it (i.e. real work actually happened after the summary).

## Testing

Added `tests/context/sections/summary-renderer.test.ts` covering the tied-epoch case (must show), the strictly-newer-observation case (must hide), the strictly-newer-summary case (must show), no-observation case, `showLastSummary: false`, and an empty-content summary. Confirmed the tied-epoch test fails against the original `<=` and passes with the fix. Full existing suite passes (one unrelated pre-existing flaky test in `tests/worker/sync/mutation-sites.test.ts`, a temp-dir SQLite race under parallel execution).


nero-harsh:
On the bundle point: the maintainer's own fix for the same class of bug (7bfc401, session rehydration prompt selection, #4047) also only touched src/ and tests/, no plugin/ bundle changes in that commit. Looks like bundle regeneration is handled separately at release time rather than per source-fix PR in this repo, so I left plugin/ untouched here too, consistent with that precedent.

Issue #4068, opened 2026-09-12
Quota guard aborts on a seven_day snapshot past its resetsAt (13.24.23), and the only workaround, restart, discards the in-memory queue


mashuroff:
## Before submitting

- [x] I searched existing issues and confirmed this is not a duplicate

## Related

- **Defect 1 is a known bug, not a new one.** The stale-snapshot mechanism is #3870 (closed 2026-09-11, consolidated into #3606). The `seven_day` variant with an already-reset window is in the #3606 thread from @edgar-eaca (13.24.1), and @yhi264 reproduced the guard on 13.24.23 with `/api/health` snapshots. It is missing only from the #3606 plan text. The Defect 1 section below is one more 13.24.23 repro; please read it as corroboration of that thread.
- **Defect 2 is the part that is not tracked.** Queue loss on restart: #3813 was consolidated into #3603 (port/liveness), and #3609 describes a DB-backed `PendingMessageStore`. Neither matches the shipped build, where the observer queue lives only in memory. @yhi264 mentioned a discarded `pendingCount=160` in passing; the scale and the cause are below. Summary also posted in #3606: https://github.com/thedotmack/claude-mem/issues/3606#issuecomment-5651781960

Filing separately because Defect 2 is not covered by any open issue, and the only recovery for Defect 1 is what triggers it.

## Bug description

Two defects that compound each other:

1. **The quota guard keeps aborting on a `seven_day` snapshot whose `resetsAt` is already in the past.** Memory capture stays off after the weekly window has reset, until the worker restarts.
2. **The only workaround, a graceful worker restart, silently discards every queued observation.** The queue lives only in process memory and the shutdown path neither drains nor persists it.

## Environment

- claude-mem **13.24.23** (bun daemon `scripts/worker-service.cjs`)
- Claude Code CLI 2.1.268, Claude desktop app 1.52386.3
- macOS 14.5 (Darwin 23.5.0), Bun 1.3.11
- Provider `claude`, auth: Claude Code OAuth token from keychain, subscription plan
- Observer model `claude-haiku-4-5-20251001`
- All times below are local, UTC+5

## Defect 1: stale `seven_day` snapshot outlives `resetsAt`

`GET /api/health` at 23:07, same worker process (up since the previous day):

```
seven_day  utilization=0.93  status=allowed_warning  observedAt=18:24:14  resetsAt=19:00:00
five_hour                    status=allowed          observedAt=22:53:24  resetsAt=03:20:00 (next day)
[...]

PR #4148, opened 2026-09-20
fix(search): match mixed-script queries term by term

goodluckstrategy-hash:
## Problem

`SessionSearch` cannot find mixed-script (CJK + Latin) content in two related ways.

**1. Substring fallback treats the whole query as one contiguous string.**
When FTS5 returns nothing, the fallback runs a single `LIKE %query%`. A query
like `payload 优先` only matches if those characters are adjacent in that order
in the stored text — so any query mixing scripts, or with words in a different
order than the source, silently returns zero rows.

**2. The tokenizer glues a Latin run to adjacent ideographs.**
`payload优先使用LLM` tokenizes as a *single* token. An exact phrase match for
`payload` can never reach it, because `payload` is only a prefix of that token,
not the token itself (follow-up to #3801 / #4068).

Net effect: searching Chinese-language sessions that contain English technical
terms — the common case for bilingual users — returns nothing, with no error.

## Fix

- `buildSubstringClause` now splits the query into terms and ANDs together a
  `LIKE` group per term, so term order and intervening text no longer matter.
- New `buildFtsMatch` wraps the query as a quoted prefix match (`"query"*`),
  with embedded quotes doubled, so a Latin run that *starts* a glued token is
  reachable.
- Added `UNSEGMENTED_RUN` to detect the glued-token shape.

## Tests

`tests/services/sqlite/search-cjk-fallback.test.ts` (new) covers the mixed-script
fallback and the prefix-match path.

**Disclosure: I was not able to run the suite locally.** `bun install` and
`bun test` are both blocked in the environment this was prepared in, so the
tests here are written but unexecuted on my side. Please let CI run them before
merging; if anything fails I'll fix it promptly.

Everything else was verified by reading: the diff is self-consistent, and the
third commit is a pure comment move (byte count unchanged) fixing a docblock
that commit 2 had stranded above the wrong function.


goodluckstrategy-hash:
Correction to the disclosure in the description: I wrote "please let CI run them," but
this branch reports no checks on this repo — so nothing is going to run these tests
automatically.

To be precise about what is and isn't verified here:

- **The tests are written but have never been executed**, by me or by CI. `bun install`
  and `bun test` were both unavailable in the environment this was prepared in.
- Everything else was verified by reading only: the diff is self-consistent, and the
  third commit is a pure comment move (byte count identical before and after).

So please run the suite locally before merging rather than taking the green-looking PR
at face value. If anything fails I'll fix it promptly — and if you'd rather I drop the
test file and ship the `SessionSearch.ts` change alone, say the word.

PR #4155, opened 2026-09-21
fix(worker): stop the quota guard aborting on a stale rate-limit snapshot (#4068)

### LINKED ISSUE #4068: Quota guard aborts on a seven_day snapshot past its resetsAt (13.24.23), and the only workaround, restart, discards the in-memory queue

pengqun:
Fixes #4068.

## What happens today

`RateLimitStore` keeps one bucket per `rateLimitType` and only ever writes the bucket of the window that is binding for that event. Once `seven_day` is the binding window at ≥93%, the guard aborts (correct). But every event after that is `five_hour`-typed, so the `seven_day` bucket is never overwritten. When the week resets, the bucket still says 98% and `shouldAbortForQuota` keeps aborting every observer session on it. The 30-minute cooldown probe goes through the same store and is aborted the same way, so the worker never recovers on its own.

Log from a 13.25.2 worker this morning, one full window after the number was true:

```
[2026-09-21 08:18:34.636] [WARN ] [SDK   ] Aborting session for quota guard: quota:seven_day utilization 98.0% >= 93% {…, window=seven_day, authMethod=Claude Code OAuth token (read from system keychain at spawn)}
```

while the live `rate_limit_event` on the same account said `seven_day: { utilization: 0.08 }`.

## Fix

- `shouldAbortForQuota` skips any bucket whose `resetsAt` is already in the past. A snapshot of a window that has reset says nothing about the current window; wait for a fresh event instead. This alone breaks the self-lock and lets the cooldown probe succeed after a reset.
- `set()` also refreshes the other buckets from the `unifiedWindows` map Claude Code attaches to every `rate_limit_event` (`{ five_hour: { utilization, resetsAt }, seven_day: {…} }`). It is not in the SDK type, so it is optional and only used when present. With it, a window that stopped being the binding one gets fresh numbers on every event rather than only at reset. The primary bucket is written last so explicit top-level fields still win, and `unifiedWindows` itself is not persisted on entries.
- `resetsAt` is normalized (epoch seconds or ms) in one helper, `toEpochMs`, and used everywhere. The `five_hour` reset-grace check compared raw seconds against `Date.now()` and could never fire; it now does.

No behavior change for API-key auth (still exempt) or for snapshots without `resetsAt` (still abort, since staleness cannot be determined).

## Tests

`tests/worker/rate-limit-store.test.ts`: 41 → 55 cases. New cases cover stale ms/seconds `resetsAt`, stale `rejected` snapshots, live snapshots still aborting, the #4068 sequence (98% seven_day → five_hour event with `unifiedWindows.seven_day` at 8% → no abort), `unifiedWindows` merge order and filtering, and the grace buffer with seconds.

```
bun test tests/worker/rate-limit-store.test.ts   # 55 pass
bun test tests/worker                            # 754 pass, 1 flake (mutation-sites 5s timeout, passes in isolation on main too)
npx tsc --noEmit -p .                            # clean
```

Bundle not included; `node scripts/build-hooks.js` on this branch reproduces `main`'s `worker-service.cjs` byte-for-byte before the change, so the rebuilt artifact differs only by this diff.


bromiao:
Corroborating data point: this reproduces unchanged on **13.25.3**, which is newer than the branch point of every open PR in this cluster. Full payload, log repro and source audit here: https://github.com/thedotmack/claude-mem/issues/4068#issuecomment-5773921273

The live `/api/health` in that comment happens to exercise all three things this PR changes, so it may be useful as a real-world fixture:

- **`seven_day` is not in `rateLimits` at all** — only `five_hour` is keyed, while `unifiedWindows.seven_day` reports the account's real 9%. Since every event this worker sees is `five_hour`-typed, the sibling refresh added here is the only mechanism that can ever populate or unstick that bucket.
- **The binding `five_hour` entry has no top-level `utilization`** — the 0.64 lives only in `unifiedWindows.five_hour`, so `entry.utilization` is `undefined` for the window that is actually current. Worth noting for the `primarySnapshot` merge in `set()`, which covers this.
- **`resetsAt` is epoch seconds while `observedAt` in the same object is epoch ms** (`1790071200` vs `1790067882309`). That makes `toEpochMs` load-bearing rather than cosmetic: an un-normalised `resetsAt <= now` would read every seconds-valued snapshot as 1970, mark every bucket stale and quietly switch the guard off entirely — a worse and much less visible failure than the one being fixed. The two expired-bucket tests here (epoch-ms and epoch-seconds) pin exactly that.

One observation while comparing the open PRs, purely as a reviewer note: this one also normalises the reset-grace comparison (`(toEpochMs(entry.resetsAt) as number) - now`), which is the same defect #4102 fixes on its own. So #4102 looks subsumed by this branch rather than complementary.

For what it is worth, the only workaround that restored capture on 13.25.3 was stopping the worker, deleting `quota-cooldown.json`, then starting it — the cooldown file is persisted specifically to survive a restart, so restarting alone is not enough, and the SessionStart banner advises against restarting at all.

thedotmack:
PlanDo: reopening to re-run CI after fork workflow approve (prior runs stuck action_required / unre-runnable).

PR #4160, opened 2026-09-21
fix(sync-hub): index canonical_ops range scans and stop counting the log per status

necatisozer:
## Problem

Every `canonical_ops` range query filters on `LENGTH(seq)`:

```sql
WHERE LENGTH(seq) > LENGTH(?) OR (LENGTH(seq) = LENGTH(?) AND seq > ?)
ORDER BY LENGTH(seq), seq
```

`seq` is the `TEXT PRIMARY KEY`, but wrapping it in `LENGTH()` stops SQLite from seeking that index, so each call scans the whole log. `getStatus` additionally runs an unfiltered `COUNT(*) FROM canonical_ops`. The affected paths are hot: every `getChanges` poll and its `more` probe, `fanOutCommitted` on every push with a socket attached, and every `getProjectionPage`.

Physical compaction is disabled for launch, so the table only grows — and rows read per request grows with it.

```
before: SCAN canonical_ops USING COVERING INDEX sqlite_autoindex_canonical_ops_1
        USE TEMP B-TREE FOR ORDER BY
after:  SEARCH canonical_ops USING INDEX canonical_ops_seq_order (<expr>>?)
```

## Change

- **An expression index**, `canonical_ops_seq_order`, on `seq` zero-padded to the uint64 digit width. Equal-width padded decimals sort lexicographically across the whole uint64 range, so the five range queries become index seeks. SQLite only uses an expression index when the expression appears verbatim, so one `SEQ_ORDER` constant is spliced into both the `CREATE INDEX` and every query.
- **`op_count` moves to a `meta` counter**, seeded once and maintained by `pushOps`, replacing the per-probe `COUNT(*)`.

Because it indexes an expression rather than adding a column, there is no schema migration and no backfill. The ordering key is storage-internal: the wire format stays the unpadded canonical decimal, so cursors already persisted by clients remain valid.

`op_count` is kept as its own counter rather than derived from `head_seq`. The two are numerically equal today, but `head_seq` is a uint64 decimal that `METADATA-CONTRACT.md` says must never pass through a JavaScript `number`, and a row count is not a sequence.

## Measured

Same deployment (3 devices, ~14k ops), same day, switching between current `main` and this branch. Both builds carried the same deployment config, which does not touch `SyncHub.ts`, so the sync-hub code is the only variable. Figures from the Durable Objects analytics dataset:

| build | window | requests | rows read | rows / request |
|---|---|---|---|---|
| current `main` | 15 min | 445 | 4,460,524 | ~10,000 |
| this branch | ~5.5 h | ~4,635 | ~347,000 | ~75 |

~10,000 rows per request is roughly one full scan of the ~14k-row log per call.

How the branch row is counted: `rowsRead` is only reported per UTC day, so the branch window was cut off by that day's quota running out, after which ~1,260 requests failed without reading any rows. Those are excluded from the request count — including them would drag the average down to ~61 and overstate the improvement. The ~14k rows read by the one-time index build are excluded too.

## Relation to recent work

Complementary to the recent change that cut per-page projection-drain RPCs. That reduced how many Durable Object calls happen; this reduces the rows each call reads. The `main` row above already includes that change, and still scans on every poll.

## Trade-offs

- **One extra index entry written per inserted op**, and measurably slower inserts — roughly 15–30% in local benchmarks, depending on how much fixed overhead the measurement includes. Reads dominate on this workload, but it does add to rows written.
- **A one-time index build** on the first cold start after deploy, before any request is served. Its cost scales with log size.
- **`CREATE INDEX IF NOT EXISTS` matches on name, not definition.** Changing the expression later must also rename the index, or deployed objects keep the stale one and silently fall back to full scans. Noted in a comment beside the constant.
- **`op_count` only increments.** That is correct while nothing deletes from `canonical_ops`; re-enabling compaction must decrement it at the deletion site. Also noted in a comment.

## Testing

- `test/seq-order-index.test.ts` asserts the **query plan** — an index seek, with no scan of `canonical_ops` and no temp B-tree — for all five range-query shapes. It imports `SEQ_ORDER` rather than restating it, so it tests the expression objects are actually built with.
- It also runs the real `getChanges` and `getStatus` on a 300-op log and **counts the rows they read**. On current `main` those two calls read 611 and 306 rows; with this change, fewer than 40. A drifted expression would still return correct results, so only a plan or rows-read assertion can catch it.
- It also covers numeric ordering across the decimal width change, `op_count` seeding, and that an idempotent re-push is not counted.
- Full suite passes: 124 tests, plus 9 WebSocket tests.


necatisozer:
Closing as superseded by #4218, which fixes the same full-scan predicates with rowid range reads and needs no migration writes.

For reference, measurements from a self-hosted hub (~14k ops, 3 devices, Workers Free plan):

- Before: ~10,000 rows read per request, i.e. a full scan of `canonical_ops`. The 5M/day rows-read cap was exhausted within ~40 minutes of each daily reset.
- With this branch's index deployed: 7–8 rows read per request, steady over the following three days.

One thing that may still bite on the Free plan is rows written: `touchDevice` writes on nearly every request, and after the read fix we still saw up to ~60k of the 100k/day write cap on a busy day.

PR #4192, opened 2026-09-22
fix(transcripts): checkpoint only processed JSONL records

AayushBehera:
## Summary

- Keep the transcript tailer's live read cursor separate from its durable restart checkpoint.
- Preserve incomplete trailing JSONL records across worker restarts without rereading buffered prefixes during normal live appends.
- Calculate the durable checkpoint as a byte offset and cover multibyte Unicode content.

## Bug

A transcript record can be written across multiple filesystem updates. The watcher must retain the incomplete record in memory while still advancing its live read position to EOF.

The previous fix used the durable checkpoint as the next live read offset. When the remainder of a partial record arrived, the watcher reread the already-buffered prefix, duplicated the JSON payload, and silently dropped the completed event on parse failure.

## Fix

`FileTailer` now tracks two positions:

- `readOffset`: the next byte to read during the current process lifetime; advances to the observed EOF after each read.
- `offset`: the durable restart checkpoint; stops before an incomplete trailing record.

The checkpoint is calculated as:

`checkpointOffset = fileSize - Buffer.byteLength(partial, 'utf8')`

A restarted worker therefore resumes at the beginning of the incomplete record, while a running worker reads only newly appended bytes and combines them with its in-memory partial buffer.

File truncation resets both positions and the retained partial buffer.

## Regression Coverage

The existing transcript watcher tests now cover:

1. complete records followed by an incomplete record and restart recovery;
2. byte-accurate offsets with multibyte Unicode content;
3. a live partial record completed by a later append without duplicating its prefix;
4. existing truncation behavior.

## Scope

- 2 existing files changed
- **78 changed lines**
- No new dependencies
- No database migration
- No state-format change


AayushBehera:
c

AayushBehera:
Addressed the P1 live-tail regression from the automated review. The tailer now maintains separate live-read and durable checkpoint offsets, so a partial JSONL record is read once in-memory during live appends and is reread only after a restart. Added regression coverage for live completion plus restart recovery, including multibyte Unicode byte offsets.

PR #4193, opened 2026-09-22
Contrib/atomic transcript state

AayushBehera:
Problem

`FileTailer.readNewData()` persisted the transcript file offset at EOF **before** processing the newly read data.

For JSONL transcripts, the final record can be only partially written when the watcher reads the file:

```text
complete record\n
partial record...
```

The watcher stores the EOF offset while the partial record is still only held in memory. If the worker restarts before the record is completed, the persisted offset points past those bytes, so the remaining portion is never read and the record is permanently lost.

The previous ordering also allowed buffered records to be checkpointed before their asynchronous handlers had finished processing them.

Fix

The watcher now:

1. Reads the newly appended bytes.
2. Separates complete lines from the trailing partial line.
3. Processes all complete records.
4. Persists the offset only through the last complete line.

The persisted checkpoint is calculated as:

```ts
const processedOffset =
  size - Buffer.byteLength(this.tailState.partial, 'utf8');
```

This is intentionally byte-based because `createReadStream()` resumes using byte offsets.

 Regression Coverage

The existing transcript watcher test now covers the exact restart scenario:

* complete record is processed;
* incomplete trailing record is retained;
* persisted offset stops before the incomplete record;
* the remaining bytes are appended later;
* a fresh watcher resumes from the persisted offset and successfully processes the recovered record;
* multibyte Unicode content verifies that byte offsets remain correct.

 Scope

* 2 existing files changed
* 47 changed lines
* No new dependencies
* No database migration
* No state-format change
* Existing truncation handling remains unchanged

This is a targeted correctness fix for transcript ingestion data loss during partial JSONL writes and process restarts.


AayushBehera:
Addressed both automated review findings: restored `dirname`/`mkdirSync` so checkpoint saves create missing parent directories, and corrected the focused test to expect the numeric JSON value `99`. The final branch now contains the atomic watch-state persistence change and its regression coverage.

AayushBehera:
c

